PR #100, opened 2018-12-07
cmd/pmem-ns-init/main.go: code and comment cleanup

okartau:
golint points out zero value assignement is redundant,
also delete unused const (leftover) and old comment
about phys device detection which has no real value.


okartau:
if we want to streamline more, it appears that initNVdimms() contains just one line of actual code (apart from comments and log line) and that is calling next function, so we can remove extra function and put code in one function. I think original idea was to have HW presence check there, but we dont have this now.

avalluri:
> if we want to streamline more, it appears that initNVdimms() contains just one line of actual code (apart from comments and log line) and that is calling next function, so we can remove extra function and put code in one function. I think original idea was to have HW presence check there, but we dont have this now.

I agree with this

okartau:
based on agreement, I made that change and merged

PR #102, opened 2018-12-11
Namespacemode: Added dual Capacity support

okartau:
To get two Namespace modes properly working, capacity
has to be maintained for both fsdax and sector mode,
which are separate measures in LVM2 mode,
and can be one in Direct NVDIMM mode.
pmem-registry.proto was changed and interface recompiled,
the version came up as 3 which does not build, so this
line was commented out in generated pmem-registry.pb.go


okartau:
verified on a v1.12.3 cluster created using test/ system, can create pvc and run _pod/my-csi-app_

pkg/pmem-csi-driver/registryserver.go
```diff
@@ -16,8 +16,9 @@ type NodeInfo struct {
 	NodeID string
 	//Endpoint node controller endpoint
 	Endpoint string
-	//Capacity node capacity(unused)
-	Capacity uint64
+	//Capacity node capacity
+	CapacityFsdax  uint64
+	CapacitySector uint64
```
avalluri: @olve, i would prefer capacity as `map{string}uint64`, so that we can expand this api easily and even at controller side we can check something like `if capacity[vol.NsMode] > asked`

pkg/pmem-registry/pmem-registry.proto
```diff
@@ -42,5 +43,6 @@ message GetCapacityRequest {
 }
 
 message GetCapacityReply {
-    uint64 capacity = 1;
-}
\ No newline at end of file
+    uint64 capacityFsdax  = 1;
```
avalluri: same as above, what about considering capacity as `map[string]uint64`

okartau:
all use now via maps, tested with unified scripts and also verified deployment cycle (v1.12) on test/ cluster.

Issue #104, opened 2018-12-12
NodeID on master is set to value of one of other nodes


okartau:
In a default-config cluster of test/ env:
* Master:host-0
* Nodes: host-1, host-2, host-3:

Master node pmem-driver log of volume deployment shows that master knows it's own nodeID as host-3:
```
ControllerPublishVolume: cs.Node: host-3 req.volume_id: f1d7cf84-fddc-11e8-8ac7-9aaa10ac8d47, req.node_id: host-2
```

Note the `cs.Node: host-3` reported. In code this comes from `cs.Driver.nodeID`.
On nodes other-than-master, this field correctly has ID of local node, but on master this contains host-3 i.e. the last node registered to master.
We should clean this up, master should have correct local nodeID, or something special noting master role, otherwise we risk with potential issues. Right now master does not use this for anything else than this log msg, as code does not reach checking the value in mode=Controller.

okartau:
false alarm, as I expected csi controller to run on cluster master, but actually it does not, as cluster master is NoSchedule, means one of the nodes runs csi controller master in addition to Node controller, and that is host-3 in current case, so the nodeID is correct, it just looks bit misleading in this log. However energy was not completely wasted as this case produced a log improvement PR #105

PR #105, opened 2018-12-12
change logging in registryserver.go

okartau:
I don't see fmt.printf messages in log with default setting
so I replace thise with glog.Info. Also, correcting
spelling and wording.


okartau:
the issue #104 turned out as false alarm, but lets have the log improvement in this PR

Issue #107, opened 2018-12-13
question: Cannot create pmem-csi volume 


obedmr:
When I tried to create a pmem-csi PVC, it seems like it's not having the capacity to provide it.

My k8s cluster is running with 1 master and 3 workers, only 1 worker has the pmem-csi labeled.

**This is the error I get in the pvc creation:**
```
failed to provision volume with StorageClass "pmem-csi-sc": rpc error: code = Unavailable desc = No node found with 8589934592 capacaity
```

**This is the relevant log from the pod that is running the node with persistence memory enabled:**

```
.
.
.
I1213 17:46:31.993469       1 glog.go:79] total used: 0
I1213 17:46:33.070068       1 glog.go:79] CheckVG: Bus: ndbus0
I1213 17:46:33.070457       1 glog.go:79] Region: region1
I1213 17:46:33.070465       1 glog.go:79] NsMode: fsdax
I1213 17:46:33.071043       1 glog.go:79] No active namespaces in region region1
I1213 17:46:33.071058       1 glog.go:79] NsMode: sector
I1213 17:46:33.071121       1 glog.go:79] No active namespaces in region region1
I1213 17:46:33.071128       1 glog.go:79] Region: region0
I1213 17:46:33.071133       1 glog.go:79] NsMode: fsdax
I1213 17:46:33.071724       1 glog.go:79] No active namespaces in region region0
I1213 17:46:33.071732       1 glog.go:79] NsMode: sector
I1213 17:46:33.071793       1 glog.go:79] No active namespaces in region region0
I1213 17:46:37.382378       1 main.go:138] Version: v0.4.1-0-gb3ef1f69a
```

**Here's the command how my VM is started:**
```
qemu-system-x86_64 -drive if=virtio,file=/data/cow_image.qcow2,format=qcow2 \
 -drive file=/data/seed.iso,if=virtio,format=raw -nodefaults \
-device virtio-balloon-pci,id=balloon0 -realtime mlock=off -msg timestamp=on \
-chardev pty,id=charserial0 -device isa-serial,chardev=charserial0,id=serial0 \
-serial stdio -enable-kvm -machine accel=kvm,usb=off -vga qxl \
-display none -m 4G,slots=4,maxmem=32G -smp 4 -machine pc,accel=kvm,nvdimm=on \
-device virtio-net-pci,netdev=net0,mac=02:42:ac:11:00:09 -netdev tap,id=net0,vhost=on,fd=3 \
[...]

Issue #58, opened 2018-11-05
pre-allocated namespace size considerations in LVM2 mode


okartau:
Currently implemented method is: we pre-allocate fixed-size namespaces which will become a Phys.Volumes for LVM2. The fixed size for pre-allocation can be given as argument and defaults to 32GB. 
This idea of fixed-size pre-alloc was implemented based on example on openstack side, I think.
What is advantage to pre-allocate multiple namespaces per Region overall?
In theory, the advantage is increased flexibility, for example one can reduce number of Phys.Volumes without need to resize existing. But our driver will not do/need that.
If we leave manual re-config options off the table, driver would work equally if it had just one namespace per region configured (or two if we support fsdax and sector modes with separate Namespaces).
Single-namespace mode has advantage that we don't waste smaller-than-block space that would remain as leftover, also we consume minimum amount of labels and namespace-size overhead.
But we dont want to hard-code single-namespace mode either, right.
To keep maximum flexibility, I propose we add special value to pre-allocated size,   expressed with some impossible option value  like 0 or -1, meaning "single space" instead of fixed-size blocks.

okartau:
shortcomings of current approach are addressed in #108

okartau:
merge of #108 addressed most of concerns raised here.
The two remaining ones are:
* do we need special mode creating single namespace instead of fixed blocks (asked by blocksize=-1)
* any gain from using striped mode of LVM2 in this driver

As this issue contains multiple solved aspects and longish discussion, what about closing this and creating new more focused issues (if the two listed ideas are worth it) ?

okartau:
I propose, no additional modes/options, let's keep what we have, and close this issue

PR #108, opened 2018-12-14
pmem-ns-init/main.go: simplify namespace creation, use leftover

okartau:
Namespace creation is simplified and instead of creating fixed
amount of namespaces we calculate, we advance in while-style loop
until all allowed space is used. The last,smaller namespace will use
all remaining space, thus solving unused-leftover problem we had,
also it solves "cant fit 4G into 4G region" risk.
In addition we try tp do it all without hard-coding namespace
overhead. But we start taking account alignement rounding.
Also, this commit adds check for "own" namespace name
if adding up already use space, this check remained missing
in an earlier change when we introduced own vs foreign checks.

### DIFF
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -187,8 +187,9 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
 	if opts.Size != 0 {
 		ways := uint32(C.ndctl_region_get_interleave_ways(ndr))
 		if opts.Size%uint64(opts.Align*ways) != 0 {
-			return nil, fmt.Errorf("size(%v) must align to interleave-width(%v) and alignment: %v",
-				opts.Size, ways, opts.Align)
+			opts.Size &= ^uint64(opts.Align*ways - 1)
+			glog.Warningf("%s: NS size must align to interleave-width(%v) and alignment: %v, force-align to %v",
+				regionName, ways, opts.Align, opts.Size)
 		}
 	}
 
--- a/cmd/pmem-ns-init/main.go
+++ b/cmd/pmem-ns-init/main.go
@@ -11,12 +11,6 @@ import (
 	"github.com/intel/pmem-csi/pkg/ndctl"
 )
 
-const (
-	// TODO: try to get rid of hard-coded overhead
-	// overhead likely comes from alignment (default 2M) and mapping_mode="dev"
-	namespaceOverhead = 4 * 1024 * 1024
-)
-
 var (
 	/* generic options */
 	//TODO: reading name configuration not yet supported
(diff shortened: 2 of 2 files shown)


okartau:
This addresses shortcomings discussed in #58, also some which were stepped to in #107, as it creates namespaces using all available space. 
The scenario with configured 4G namespace in 4G region should also work now.
I have tested changed namespaces creation in single-host unified mode, test/ cluster (emulated), and by running driver binaries stand-alone on real HW with few 128 GB devices, using default 32GB namespace size.

okartau:
code-wise, this should do what we need. Some aspects may be fine-tuned, like definition of alignment: default mib2 is already defined in ndctl, side, but it's not cleanly usable from outside, so what about creating common sub-dir constants/ and putting such constant declaration there?

okartau:
even better idea by @avalluri: avoid use of alignment at that level and leave it to be done in lower level, ndctl. I think I saw an error there caused by arbitrary size I used (result of percentage calc). We need to clarify ans ensure forced alignment works the way we try in this PR

okartau:
It seems we can not currently call r.CreateNamespace() with unaligned value as we have
this code in pkg/ndctl/region.go:
```
	if opts.Size != 0 {
		ways := uint32(C.ndctl_region_get_interleave_ways(ndr))
		if opts.Size%uint64(opts.Align*ways) != 0 {
			return nil, fmt.Errorf("size(%v) must align to interleave-width(%v) and alignment: %v",
				opts.Size, ways, opts.Align)
		}
	}
```

means, we have to align the Size not just to opts.Align (which gets default value as 2MB), but opts.Align multiplied by `ways` (and what is value of that, should we log it to get more visibility).
If was want to implement plan to avoid aligning size in upper level, we have to modify r.CreateNamespace to force-align the size instead of returning error. I think force-align is good idea there, returning error does not help in driver use model (where there is no human to read the message and retry with different parameter, as is the idea in command-line ndctl case)

okartau:
I force-pushed another commit changing alignment error-return to forced-down alignment, and removing all alignment-related code from upper level

pkg/ndctl/region.go
```diff
@@ -187,8 +187,9 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
 	if opts.Size != 0 {
 		ways := uint32(C.ndctl_region_get_interleave_ways(ndr))
 		if opts.Size%uint64(opts.Align*ways) != 0 {
-			return nil, fmt.Errorf("size(%v) must align to interleave-width(%v) and alignment: %v",
-				opts.Size, ways, opts.Align)
+			opts.Size &= ^uint64(opts.Align*ways - 1)
+			glog.Infof("%s: NS size must align to interleave-width(%v) and alignment: %v, force-align to %v",
```
avalluri: @okartau  would it make sense to make this warning?

okartau:
I changed that Info to Warning, force-pushed.
ok to merge now?

avalluri:
yes

> I changed that Info to Warning, force-pushed.
> ok to merge now?

yes

PR #114, opened 2019-01-04
Fix issues in namespace nullify

### LINKED ISSUE #113: issues in namespace.nullify
I started to experiment with ndctl-mode of device manager, trying first simple unified-mode lifecycle as in util/lifecycle-unified.sh. It goes well until deletion, but there are few issues in namespace.nullify:
- blockdevice name needs "/dev/" prepended to work as device file name in `func (ns *Namespace) nullify() error {`
With that one changed like this:
```
file, err = os.OpenFile("/dev/" + ns.BlockDeviceName(), os.O_RDWR|syscall.O_DIRECT|os.O_EXCL, os.ModeDevice)
```

there will be 
```
panic: runtime error: slice bounds out of range

goroutine 10 [running]:
github.com/intel/pmem-csi/pkg/ndctl.(*Namespace).nullify(0x18dbae0, 0x9351b2, 0x3)
        go/src/github.com/intel/pmem-csi/pkg/ndctl/namespace.go:474 +0x4c1
```
Seems something is wrong with copying code.
The idea is to copy 4k  of zeroes to block device start, wiping volume format context, but not data itself?
This kind of overlaps with "erase" functionality, but erase is optional (and wipes entire device)

avalluri:
Simplified the logic by avoiding creating new byte array while writing to pmem
block file. Also fixed memory leak.

Fixes #113

### DIFF
--- a/pkg/ndctl/namespace.go
+++ b/pkg/ndctl/namespace.go
@@ -453,32 +453,27 @@ func (ns *Namespace) nullify() error {
 	var file *os.File
 
 	C.ndctl_namespace_set_raw_mode(ndns, 1)
+	defer C.ndctl_namespace_set_raw_mode(ndns, 0)
 	if err = ns.Enable(); err != nil {
-		C.ndctl_namespace_set_raw_mode(ndns, 0)
 		return err
 	}
 
-	if rc = C.posix_memalign(&buf, C.size_t(kib), C.size_t(kib)); rc < 0 {
-		C.ndctl_namespace_set_raw_mode(ndns, 0)
+	if rc = C.posix_memalign(&buf, C.size_t(kib4), C.size_t(kib4)); rc < 0 {
 		return fmt.Errorf("memory error: %s", cErrorString(rc))
 	}
+	defer C.free(buf)
+	C.memset(buf, 0, C.size_t(kib4))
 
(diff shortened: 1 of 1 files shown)


pkg/ndctl/namespace.go
```diff
@@ -453,32 +453,27 @@ func (ns *Namespace) nullify() error {
...
+	defer C.free(buf)
+	C.memset(buf, 0, C.size_t(kib4))
 
-	file, err = os.OpenFile(ns.BlockDeviceName(), os.O_RDWR|syscall.O_DIRECT|os.O_EXCL, os.ModeDevice)
+	file, err = os.OpenFile("/dev/"+ns.BlockDeviceName(), os.O_RDWR|syscall.O_DIRECT|os.O_EXCL, os.ModeDevice)
 	if err != nil {
-		C.ndctl_namespace_set_raw_mode(ndns, 0)
 		return err
 	}
+	defer file.Close()
```
okartau: Note, there is only one return point after defer file.Close, but yes we can have it for uniform style

avalluri: ya, you are write, we can avoid this defer.

avalluri:
@olev, can we merge this?

Issue #115, opened 2019-01-04
Context.CreateNamespace hides error from lower level calls


okartau:
I spotted this while running util/csi-sanity.sh against unified mode stand-alone pmem-csi in drivermode=ndctl (i.e. bypassing LVM2).
Volume creation sometimes fails in that test like this:
```
GRPC request: name:"sanity-node-full-836087B2-33695F1A" volume_capabilities:<mount:<> access_mode:<mode:SINGLE_NODE_WRITER > > 
CreateVolume: Name: sanity-node-full-836087B2-33695F1A, req.Required: 4194304 req.Limit; 0
CreateVolume: Special create volume in Unified mode
setting namespace sector size ...: 0
setting pfn
rpc error: code = Internal desc = CreateVolume: failed to create volume: Failed to create namespace
```
Here the only information given is "Failed to create namespace" without pointing out which phase of creation actually failed. In Region.CreateNamespace() there is some trace ("setting..." prints) but real error remains hidden, although it is available in `err` that returned from lower calls.

pohly:
Olev Kartau <notifications@github.com> writes:

> Seems we lose original lower-layer error in ndctl.go in Context.CreateNamespace
> where return stmt creates a new error message, losing the err that was
> returned from failed call.

https://github.com/pkg/errors is a solution for that: always add
additional information to existing errors with errors.Wrap() when
returning a new error.

okartau:
resolved via fixes coming from merge of direct-nvdimm mode in #133

PR #116, opened 2019-01-04
CreateNamespace: do not hide error from lower call

okartau:
This simple change fixes error hiding issue described in #115.
But note that it also changes semantics on failure:
now we return after 1st attempt failure, but original code would try creation in other remaining regions.
If we want to keep original semantics, we have to choose which error to report/return,
in case we try multiple times and there are multiple errors.
Perhaps should add log stmt showing error, in addition to returning error?

### DIFF
--- a/pkg/ndctl/ndctl.go
+++ b/pkg/ndctl/ndctl.go
@@ -6,7 +6,12 @@ package ndctl
 //#include <ndctl/libndctl.h>
 //#include <ndctl/ndctl.h>
 import "C"
-import "fmt"
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"k8s.io/klog/glog"
+)
 
 const (
 	kib  uint64 = 1024
@@ -62,14 +67,19 @@ func (ctx *Context) GetBuses() []*Bus {
 
 //CreateNamespace create new namespace with given opts
 func (ctx *Context) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
+	var err error
+	var ns *Namespace
 	for _, bus := range ctx.GetBuses() {
 		for _, r := range bus.ActiveRegions() {
-			if ns, err := r.CreateNamespace(opts); err == nil {
+			if ns, err = r.CreateNamespace(opts); err == nil {
+				glog.Infof("Namespace %s created in %s", ns.Name(), r.DeviceName())
 				return ns, nil
+			} else {
+				glog.Errorf("Namespace creation failure in %s: %s", r.DeviceName(), err.Error())
 			}
 		}
 	}
-	return nil, fmt.Errorf("Failed to create namespace")
+	return nil, errors.Wrap(err, "failed to create namespace")
 }
 
 //DestroyNamespaceByName deletes namespace with given name


pkg/ndctl/ndctl.go
```diff
@@ -66,6 +66,8 @@ func (ctx *Context) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error
 		for _, r := range bus.ActiveRegions() {
 			if ns, err := r.CreateNamespace(opts); err == nil {
 				return ns, nil
+			} else {
+				return nil, err
```
avalluri: As of i remember, it was intentional, to try with all available regions to serve the creation request. If upper layers are interested in error would call Region.CreateNamespace().

pohly: As @avalluri pointed out, this changes the semantic.

Otherwise a shorter way to write this if/else would be
`return r.CreateNamespace(opts)`.

But even better is to wrap errors with `errors.Wrap` from https://github.com/pkg/errors because then additional information and a stack trace gets added at each level.

gRPC logging however currently doesn't show the stack trace, so that part isn't useful at the moment.

okartau:
Thanks for Wrap idea, it really seems good solution for multiple/nested errors case. I force-pushed change to use Wrap, fixes the issue without changing semantics. 
If there are multiple Regions involved, the output may get crowded, but the most important part is not to miss any information, I think.
There are few wrapping  choices here, we can either:
- use empty addition in the loop and add one at the end (current code),
- wrap with added string in the loop, and not string add at the end (can use plain _wrappederr_ var there then)
- add some string in both places (makes output even more messy so I think we should not).

The current code produces this message in case of one Region:
```
CreateNamespace returned error:Failed to create namespace: : pfn: failed to set namespace
```

extra colon mark(s) seems result of Wrap() with empty string, but this is just cosmetic problem

okartau:
another minor thing, should we use import() block instead of multiple import statements

okartau:
I tried this on a system with 6 Regions were same error happens, and contrary to my expectation (to see 6 times same message repeated) there is just single one:
```
GRPC error: rpc error: code = Internal desc = CreateVolume: failed to create volume: Failed to create namespace: : pfn: failed to set namespace
```

Does error.Wrap avoid repeating same msg? or is there something wrong with how Wrap is used here.

pkg/ndctl/ndctl.go
```diff
@@ -62,14 +66,18 @@ func (ctx *Context) GetBuses() []*Bus {
 
 //CreateNamespace create new namespace with given opts
 func (ctx *Context) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
+	var err error
+	var ns *Namespace
 	for _, bus := range ctx.GetBuses() {
 		for _, r := range bus.ActiveRegions() {
-			if ns, err := r.CreateNamespace(opts); err == nil {
+			if ns, err = r.CreateNamespace(opts); err == nil {
```
pohly: Here you get a fresh error, straight from `r.CreateNamespace`. Whatever was in err previously, like an error from a previous loop iteration, is lost here.

pkg/ndctl/ndctl.go
```diff
@@ -62,14 +66,18 @@ func (ctx *Context) GetBuses() []*Bus {
...
 				return ns, nil
+			} else {
+				err = errors.Wrap(err, "")
```
pohly: And here you add an empty prefix. I don't know exactly how `errors.Wrap` handles that. Probably it simply ignores the string and just adds a stack entry.

pkg/ndctl/ndctl.go
```diff
@@ -62,14 +66,18 @@ func (ctx *Context) GetBuses() []*Bus {
...
 			}
 		}
 	}
-	return nil, fmt.Errorf("Failed to create namespace")
+	return nil, errors.Wrap(err, "Failed to create namespace")
```
pohly: It's expected that you only get the last error, with "Failed to create namespace" as prefix. 

If you want to collect all errors, then you need to collect them in a list and then somehow concatenate. If it fails the same way for all regions, then that's probably overkill and you can simply clarify that you return the latest error with "failed to create namespace in all regions, last error is".

BTW, error strings should always start lowercase. That's because another prefix might be added higher up, and then mixed case looks weird.

okartau: ah yes, silly me, I tried to "optimize" num of vars there... we need more variables. 
Same from all regions: I was thinking same way, but it feels bit overkill to start analyzing error msg contents in that function

okartau:
what about instead creating more complex error msg manipulation and analyzing there, we just log errors after we get them from lower/layer call, and we return single (new) error, as it was originally. Thats actually much simpler than wrap/analyze. The potential issue with that will be, list of per-Region errors is not returned via RPC but written in log instead. But it's still better than original state where error was not shown at all.

pohly:
> what about instead creating more complex error msg manipulation and
> analyzing there, we just log errors after we get them from lower/layer
> call, and we return single (new) error, as it was originally. Thats
> actually much simpler than wrap/analyze. The potential issue with that
> will be, list of per-Region errors is not returned via RPC but written
> in log instead. But it's still better than original state where error
> was not shown at all.

How easily can the error be discovered in the log?

This probably needs to go into an "admin cookbook" section in the
README.md.

okartau:
can we merge this now? At least I am happy with how error messages are shown

okartau:
> How easily can the error be discovered in the log? This probably needs to go into an "admin cookbook" section in the README.md.

we don't have "admin cookbook" section in README. Should we create one?

pohly:

>> How easily can the error be discovered in the log? This probably needs to go into an "admin cookbook" section in the README.md.
>
> we don't have "admin cookbook" section in README. Should we create one?

Yes.

okartau:
I change the style towards "less PRs" so that I dont create separate PR for small changes. Thus moving this change into new #133 , closing this PR

PR #118, opened 2019-01-08
Improve log messages

okartau:
In DeleteVolume, if volume was not created, log about deletion
was still emitted.


okartau:
lets have this open for some time, more improvements can be added in here

PR #121, opened 2019-01-08
ndctl/region.go: fix fallback alignement in Region.CreateNamespace

okartau:
This code seems to mimic it's counterpart in ndctl namespace.c
However the check for lower bits being non-zero is other way.


okartau:
cherry-picked that commit into #122 to reduce num of PRs, closing this

PR #123, opened 2019-01-09
pmd-ndctl.go: implement FlushDeviceData

okartau:
call flushDevice on block device, this improves pass rate
of csi-sanity in direct-nvdimm mode.


okartau:
verified in direct-nvdimm mode, unified-mode on single host, in qemu emulation so far, running util/sanity and util/lifecycle

okartau:
yes, true, I think while I was testing, nullify was not working properly. 
csi-sanity test rate should get improved by nullify as well (I need to verify that separately)
But nullify is not replacement of Flush, because nullify zeros 4k of device start, while Flush wipes all data, and thats needed to fulfill privacy requirement.

okartau:
change from this PR was moved into #133, closing

PR #124, opened 2019-01-09
nodeserver.go: simplify NodePublish and NodeUnpublish

okartau:
Removing, creating directory is not duty of this driver.
Stop checking for existing bind-mount, should not be needed.
Stop retrieve of Fstype and log of it's value, this has no meaning in bind-mounting


okartau:
tested  and works in:
- unified mode single node, using util/lifecycle and util/sanity
- deployment in cluster using test/, k8s 1.13

okartau:
I force-pushed small change, we still check for existing mount. 
Note in current version, we will not check,store returned err at all, is that potential problem?

PR #126, opened 2019-01-10
Dockerfiles: advance to ndctl version 63, use single version definition

okartau:
This change required addition of bash-completion to deps,
and --without-systemd in config options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ IMAGE_VERSION_pmem-csi-driver=canary
 IMAGE_VERSION_pmem-ns-init=canary
 IMAGE_VERSION_pmem-vgm=canary
 IMAGE_TAG=$(REGISTRY_NAME)/$*:$(IMAGE_VERSION_$*)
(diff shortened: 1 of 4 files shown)


okartau:
Docker images are buildable with those changes.
I also tried --without-bash to get rid of bash-completion dep requirement, but this did not work.
However --without-systemd is required, otehrwise build fails on systemd (which we dont need anyway)
Should we try to remove with_systemd_unit_dir=no option now also?

okartau:
while changing ndctl version, I notice we have the value "63" now defined in 9 places (3x3 Dockerfiles).
I would like reduce that repetition, so I tried to use ARG=, which could be sourced from single vaue in Makefile, but there is slight problem with that: WORKDIR value seems not substituting variable name to actual path name, and after untarring ndctl tarball, the directory name is hardcoded to ndctl-63 so it has to carry version info. I wonder, could we just rename it after  unpacking to path without version name part, like ndctl

PR #127, opened 2019-01-10
pmem-ns-init: Avoid zero-size namespace creation request

### LINKED ISSUE #125: initial namespace creation in LVM2 mode attempts zero-size creation
Seeing that in "kubectl logs pmem-csi-8swq2 pmem-ns-init"
```
I0110 Create fsdax-namespaces in region0, allowed 100 %:
total       :      34357641216
avail       :                0
can use     :      34357641216
I0110 createNS: Exists: Size       9510584320 Mode:fsdax Device:namespace0.2 Name:pmem-csi
I0110 createNS: Exists: Size       9510584320 Mode:fsdax Device:namespace0.1 Name:pmem-csi
I0110 createNS: Exists: Size       5280628736 Mode:fsdax Device:namespace0.3 Name:pmem-csi
I0110 createNS: Exists: Size       9510584320 Mode:fsdax Device:namespace0.0 Name:pmem-csi
I0110 545259520 bytes calculated available after scan for existing fsdax-mode namespaces
I0110 Less than configured namespacesize remaining, change desired size to 545259520
I0110 Calculated canUse:545259520, available by Region info:0
I0110 Available in Region:0 is less than desired nsSize, limit nsSize to that
I0110 Create next 0-bytes fsdax-namespace
W0110 Failed to create namespace:Failed to set namespace size: Invalid argument
```
It's not really damaging anything, but calling ndctl layer wit zero size is not correct, we should avoid it.

okartau:
At some point in creation loop, we typically reach the state
where calculated size remains available but actual available
is zero, so detect this and break, instead of calling
zero-size creation request and breaking loop
based on returned error.
Fixes #125

### DIFF
--- a/cmd/pmem-ns-init/main.go
+++ b/cmd/pmem-ns-init/main.go
@@ -100,6 +100,10 @@ func createNS(r *ndctl.Region, nsSize uint64, uselimit int, nsmode ndctl.Namespa
 			glog.Infof("MaxAvailableExtent in Region:%v is less than desired nsSize, limit nsSize to that", r.MaxAvailableExtent())
 			nsSize = r.MaxAvailableExtent()
 		}
+		// In typical case, NSize drops to zero at some point, avoid zero-size create request
+		if nsSize == 0 {
+			break
+		}
 		glog.Infof("Create next %v-bytes %s-namespace", nsSize, nsmode)
 		_, err := r.CreateNamespace(ndctl.CreateNamespaceOpts{
 			Name: "pmem-csi",


okartau:
tested on cluster with test/ system, namespaces init, and deployment after manual deletion of namespaces

cmd/pmem-ns-init/main.go
```diff
@@ -100,6 +100,10 @@ func createNS(r *ndctl.Region, nsSize uint64, uselimit int, nsmode ndctl.Namespa
 			glog.Infof("MaxAvailableExtent in Region:%v is less than desired nsSize, limit nsSize to that", r.MaxAvailableExtent())
 			nsSize = r.MaxAvailableExtent()
 		}
+		// In typical case, NSize drops to zero at some point, avoid zero-size create request
+		if nsSize == 0 {
```
avalluri: @okartau we can move this check to bit ahead of AvailableSize checks.

okartau: My idea here is, nsSize can be set to zero in both checks above, so checking it against zero covers both. And if nsSize is set to zero in upper check, the lower check will do nothing so it is safe

avalluri: yes true.

PR #128, opened 2019-01-10
Topology awareness to pmem-csi provisioned volume

avalluri:
The changes:
- Offload capacity bookkeeping from RegistryServer
- Fill Topology information to newly provisioned volume.


deploy/kubernetes-1.12/pmem-csi.yaml
```diff
@@ -94,7 +94,7 @@ spec:
       - name: external-provisioner
         imagePullPolicy: Always
         image: quay.io/k8scsi/csi-provisioner:v0.4.1
-        args: [ "--v=5", "--provisioner=pmem-csi", "--csi-address=$(ADDRESS)" ]
+        args: [ "--v=5", "--provisioner=pmem-csi", "--csi-address=$(ADDRESS)", "--feature-gates=Topology=true"]
```
okartau: should there be space near end of line? (likely just style issue, not functional)

okartau:
needs rebase as there were some commits recently.
By reading, seems OK,
but I also want to run this through existing test cases

deploy/kubernetes-1.12/pmem-csi.yaml
avalluri: Yes, i agree. Done

avalluri:
> needs rebase as there were some commits recently.
> By reading, seems OK,
> but I also want to run this through existing test cases

Now i rebased my changes.

okartau:
This comes, which is new compared to previous state
```
------------------------------
• Failure [0.002 seconds]
Node Service
go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  NodeGetInfo
  go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:142
    should return approproate values [It]
    go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:153

    Expected
        <*csi.Topology | 0x0>: nil
    not to be nil

    go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:164
------------------------------
```

avalluri:
> This comes when unified-mode, standalone driver is tested with util/sanity-unified.sh, which is new compared to previous state
> 
> ```
> ------------------------------
> • Failure [0.002 seconds]
> Node Service
> go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
>   NodeGetInfo
>   go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:142
>     should return approproate values [It]
>     go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:153
> 
>     Expected
>         <*csi.Topology | 0x0>: nil
>     not to be nil
> 
>     go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:164
> ------------------------------
> ```

Ya, it breaks Unified Mode :(, I will fix this.

avalluri:
@okartau i fixed Unified mode failure, can you please have a look.

okartau:
I still get same error. I started to suspect my tree did not pick up recent change, but git tells it is same as remote, also commit IDs match the ones in this PR, so I believe the code is same as here.
Bit inconvenient that force-pushing does not leave trace of changes.

```
• Failure [0.002 seconds]
Node Service
go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  NodeGetInfo
  go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:142
    should return approproate values [It]
    go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:153

    Expected
        <*csi.Topology | 0x0>: nil
    not to be nil
```

pohly:
> Bit inconvenient that force-pushing does not leave trace of changes.

If you pulled the previous revision earlier, then you can still find it
locally with "git reflog".

okartau:
thanks for idea. indeed, `git diff c09dd925 f9859e23` gives me idea the recent diff was added block `topology = ...` at controllerserver.go:168

okartau:
I think what happens is recent fix attempt has no effect because it affects topology upon volume creation, but test in question fails before any volume was created. Failing location is in tester code at go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:142 which tests NodeGetInfo before volumes creation. And that part was not running before, since test code is wrapped in check:
```
    if accessibilityConstraintSupported {
            Expect(ninfo.GetAccessibleTopology()).NotTo(BeNil())
    }
```

and that part activates now as this PR enables
`PluginCapability_Service_ACCESSIBILITY_CONSTRAINTS`

okartau:
util/sanity-unified.sh passes now

avalluri:
@pohly Is it ok to merge this change?

avalluri:
> Can you update the README.md and the architecture diagram accordingly? That'll make it easier to understand what the code change is intended to do and how it changes user-visible behavior and/or the architecture.

Ok, I will update the sequence diagram.
> 
> Regarding topology support: have you tested this? As far as I know, Kubernetes 1.13 does not support it out of the box for CSI drivers, because it depends on the `CSIDriverInfo` CRD and that is still alpha. But I haven't used it, so I might be mistaken.

I tested these changes on, 1.13.2. The dynamically provisioned volume now filled with NodeAffinity:

```
apiVersion: v1
items:
- apiVersion: v1
  kind: PersistentVolume
  metadata:
    annotations:
      pv.kubernetes.io/provisioned-by: pmem-csi
    creationTimestamp: 2019-01-21T13:53:08Z
    finalizers:
    - kubernetes.io/pv-protection
    - external-attacher/pmem-csi
    name: pvc-c25d0a37-1d83-11e9-9ee2-deadbeef0100
    namespace: ""
    resourceVersion: "430991"
    selfLink: /api/v1/persistentvolumes/pvc-c25d0a37-1d83-11e9-9ee2-deadbeef0100
    uid: e2277794-1d83-11e9-9ee2-deadbeef0100
  spec:
    accessModes:
    - ReadWriteOnce
    capacity:
      storage: 8Gi
    claimRef:
      apiVersion: v1
      kind: PersistentVolumeClaim
      name: pmem-csi-pvc
      namespace: default
      resourceVersion: "430977"
      uid: c25d0a37-1d83-11e9-9ee2-deadbeef0100
    csi:
      driver: pmem-csi
      fsType: ext4
      volumeAttributes:
        name: pvc-c25d0a37-1d83-11e9-9ee2-deadbeef0100
        nsmode: fsdax
        size: "8589934592"
        storage.kubernetes.io/csiProvisionerIdentity: 1548077214178-8081-pmem-csi
      volumeHandle: e215d5fa-1d83-11e9-ac8c-decdb10014b0
    nodeAffinity:
      required:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/hostname
            operator: In
            values:
            - host-3
        - matchExpressions:
          - key: kubernetes.io/hostname
            operator: In
            values:
            - host-2
        - matchExpressions:
          - key: kubernetes.io/hostname
            operator: In
            values:
            - host-1

```

avalluri:
> > Regarding topology support: have you tested this? As far as I know, Kubernetes 1.13 does not support it out of the box for CSI drivers, because it depends on the `CSIDriverInfo` CRD and that is still alpha. But I haven't used it, so I might be mistaken. I tested these changes on, 1.13.2. The dynamically provisioned volume now filled with NodeAffinity:
> But does that have the desired effect? Actually, what *is* the desired effect?

The desired effect is obvious, that the PV's Topology is respected by Kubernetes and considers this information while scheduling the pod which uses this PV. i.e, schedule only on nodes where this PV can be attachable. I could dig code traces to PV NodeAffinity: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/index.go
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/persistentvolume/pv_controller.go

> nodeAffinity: required: nodeSelectorTerms: - matchExpressions: - key: kubernetes.io/hostname operator: In values: - host-3 - matchExpressions: - key: kubernetes.io/hostname operator: In values: - host-2 - matchExpressions: - key: kubernetes.io/hostname operator: In values: - host-1
> If Kubernetes honors this information (not sure at the moment, I'll ask), then it only helps the first pod which uses a volume. A second pod (regardless whether it runs at the same time as the first one or later) may still land on the wrong node.

Atleast this PR is not targeted to fix the Second Pod's scheduling issue, its my bad if my connects sounded like this :( 

> IMHO we shouldn't merge in topology support if it isn't clear how we'll address all the known issues.

Currently Topology support is in Kubernetes beta(https://kubernetes.io/blog/2018/10/11/topology-aware-volume-provisioning-in-kubernetes/) and in alfa in CSI. In my opinion If something really not working as it should be, still its fix is not in our driver context.  This change shouldn't break the current driver's functionality, the worst it could do is it might have no effect as you are suspecting.

pohly:

>> Amarnath Valluri <notifications@github.com> writes:
>> > Regarding topology support: have you tested this? As far as I know, Kubernetes 1.13 does not support it out of the box for CSI drivers, because it depends on the `CSIDriverInfo` CRD and that is still alpha. But I haven't used it, so I might be mistaken. I tested these changes on, 1.13.2. The dynamically provisioned volume now filled with NodeAffinity:
>> But does that have the desired effect? Actually, what *is* the desired effect?
>
> The desired effect is obvious,

From "allows the kubernetes scheduler to take *right* decision while
scheduling the pod which claim to use PV." (emphasis mine) it wasn't
obvious which aspects of the problem that we are facing it is meant to
solve. It wasn't until you quoted the result that it became clearer that
it doesn't lock the PV to a single node, for example.

We need a definition of what the "right" decision is, then we can check
whether the code achieves that.

BTW, why is "Offload GetCapacity() from RegistryServer to Node
ControllerServer" better than the previous approach? Your commit message
explains the "what" of the change, but not the "why".

avalluri:
> We need a definition of what the "right" decision is, then we can check whether the code achieves that.


Now i updated the commit message to describe what is that *right* decision. Can you have a look if it's sounds right what this topology information is used for.

avalluri:
> BTW, why is "Offload GetCapacity() from RegistryServer to Node ControllerServer" better than the previous approach? Your commit message explains the "what" of the change, but not the "why".

I added more description to commit message why this change is required.

avalluri:
> > Can you update the README.md and the architecture diagram accordingly? That'll make it easier to understand what the code change is intended to do and how it changes user-visible behavior and/or the architecture.
> 
> Ok, I will update the sequence diagram.

I updated the sequence diagram with additional step to show when this GetCapacity() is used. And added notes to explain what are the steps involved in csi.Controller.CreateVolume() call.

pkg/pmem-csi-driver/controllerserver.go
```diff
@@ -161,20 +167,44 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-		} else /*if cs.mode == Unified */ {
-			foundNode := false
-			for _, nodeInfo := range cs.rs.nodeClients {
-				if nodeInfo.Capacity[nsmode] >= asked {
-					foundNode = true
-					break
+			topology = append(topology, &csi.Topology{
+				Segments: map[string]string{
+					"kubernetes.io/hostname": cs.Driver.nodeID,
+				},
```
pohly: This is where it becomes interesting. A CSI driver shouldn't need to know about Kubernetes host names. It's the other way around: the CSI driver has its own local host name, and the sidecar containers take care of mapping that to the Kubernetes host name.

According to Michelle, external-provisioner does that based on the information in `CSIDriverInfo`. Without `CSIDriverInfo` there should be no `NodeAffinity` in the `PV`.

@avalluri Were you testing with the `CSINodeInfo` CRD installed and the corresponding feature gate enabled?

My current theory is that `external-provisioner` just passed through the CSI driver values. Whether it should do that is open for debate...

pohly:
>> We need a definition of what the "right" decision is, then we can check whether the code achieves that.
>
>
> Now i updated the commit message to describe what is that *right*
> decision.

"This allows the kubernetes scheduler to consider this NodeAffinity while
scheduling the pod which claims this PV, and schedule only on nodes where
this PV can be attachable." - if that is the goal, then the
implementation doesn't meet that goal because Kubernetes may still
schedule pods onto the wrong node (second pod case).

To me it still looks like we must have further discussions around the
expected scheduling behavior for apps using persistent local storage
before we merge topology related code changes, because the outcome of
that discussion determines whether the desired NodeAffinity is "all
nodes with pmem" or "the single node where the volume was created".

pkg/pmem-csi-driver/controllerserver.go
avalluri: The change tested with external-provisioner v4.0.1, I don't think that CSINodeInfo is not implemented in this version.

avalluri:
> "This allows the kubernetes scheduler to consider this NodeAffinity while scheduling the pod which claims this PV, and schedule only on nodes where this PV can be attachable." - if that is the goal, then the implementation doesn't meet that goal because Kubernetes may still schedule pods onto the wrong node (second pod case).

@pohly  This PR alteast makes sure the first pod schedules on right node which is bigger issue currently the driver has. Sharing PMEM namespace/volume between applications is wider than the scope of this change, which needs some brainstorming at least for me. Lets deal that separately.

avalluri:
> > "This allows the kubernetes scheduler to consider this NodeAffinity while scheduling the pod which claims this PV, and schedule only on nodes where this PV can be attachable." - if that is the goal, then the implementation doesn't meet that goal because Kubernetes may still schedule pods onto the wrong node (second pod case). @pohly This PR alteast makes sure the first pod schedules on right node which is bigger issue currently the driver has.
> Then say so in the commit message and you are setting the expectations correctly.

Added a note with this to commit message, hope it solves.

pohly:
Now it is clear what this PR does and what it doesn't. I still think that eventually we'll need a different solution and the current use of of `external-provisioner` only happens to work by chance (it's not designed to work without `CSINodeInfo`, so we got lucky that it seems to work also without), but as an incremental improvement over the current situation it is worth merging.

avalluri:
> I still think that eventually we'll need a different solution and the current use of of `external-provisioner` only happens to work by chance (it's not designed to work without `CSINodeInfo`, so we got lucky that it seems to work also without),

@pohly As per the [CSI Volume Plugins in Kubernetes Design Doc](https://github.com/verult/community/blob/7302f0a73403d493126f517cb10638f2faffc164/contributors/design-proposals/storage/container-storage-interface.md)

> Upon initialization of the external “CSI volume driver”, kubelet must call the CSI method NodeGetInfo to get the mapping from Kubernetes Node names to CSI driver NodeID and the associated accessible_topology. It must:
> 
> Create/update a CSINodeInfo object instance for the node with the NodeID and topology keys from accessible_topology.

As of my understanding, CSINodeIfo is created by kubelet by calling "NodeGetInfo", part of this PR, we are filling the topology information in NodeGetInfoResponse. So i feel its working as expected only not may be by chance. Or am i still missing something :(

pohly:
Yes, let's merge it.

Issue #117, opened 2019-01-07
direct mode: CreateVolume fails with: pfn: failed to set namespace


okartau:
in direct-NVDIMM mode, util/sanity-unified fails with pfn: failed to set namespace, both in qemu-emulated and HW-based systems.
In qemu-emulated system with 1 region:
```
CreateVolume: Special create volume in Unified mode
setting namespace sector size ...: 0
setting pfn
GRPC error: rpc error: code = Internal desc = CreateVolume: failed to create volume: Failed to create namespace: : pfn: failed to set namespace
```

```
CreateVolume: Special create volume in Unified mode
setting namespace sector size ...: 0
setting pfn
setting namespace sector size ...: 0
setting pfn
setting namespace sector size ...: 0
setting pfn
setting namespace sector size ...: 0
setting pfn
setting namespace sector size ...: 0
setting pfn
setting namespace sector size ...: 0
setting pfn
GRPC error: rpc error: code = Internal desc = CreateVolume: failed to create volume: Failed to create namespace: : pfn: failed to set namespace
```

note that this output comes ftom code where #116 is applied, to make the inner error string visible.
As the error is same on qemu and HW-based system, we can not blame emulation this time.
There is some difference in direct-nvdimm mode namespace creation which causes this,
as LVM2-mode creation has worked on same HW

okartau:
What makes this more interesting is that creating a single namespace in simple lifecycle script using csc, works, also some creation events in csi-sanity test have worked. Failure is  triggered as result of certain sequence of csi-sanity.

okartau:
Played with driver bit more, appears nvdimm does not like to be allocated/freed dynamically by small amounts. There is some discussion: https://github.com/pmem/ndctl/issues/76 , problems I saw are similar.
Taking advice from mentioned thread, I tried to use _Align_ option to namespace creation.
I can't set it to 128MB as current ndctl code does not support that value. It supports 1GB, but when I try this, I hit different problem(s) in kernel (thats on qemu system I have used so far in trials):

`nd_pmem pfn0.0: namespace0.0 unable to satisfy requested alignment.`

So the theory now is, we can nicely use NVDIMM mem in LVM2 mode, because there we
pre-allocate namespaces in n*1GB pieces, which automatically guarantees 1GB alignment,
and we never try dynamic deletion and re-creation of namespaces (as dynamics work on top of LVM2 in this mode).
[...]

Issue #9, opened 2018-09-25
qemu: in a 2-nvdimm system, 2nd namespace creation on region1 fails with "invalid altmap"


okartau:
Qemu based system with 2 x 1GB NVDIMM which will show as 2 regions in VM.
VM is configured with 16GB of mem, 32 GB of maxmem, maxmemory slots=16

I can create multiple (for example 4) namespaces on region0:
(last one is smaller intentionally, to fit in 1GB region)
```
ndctl create-namespace -r region0 -s 268435456 -n nspace1
ndctl create-namespace -r region0 -s 268435456 -n nspace2
ndctl create-namespace -r region0 -s 268435456 -n nspace3
ndctl create-namespace -r region0 -s 266338304 -n nspace4
```

Same does not work for region1: 
(NVDIMMs are zeroed in between and labeling init re-done in VM)
1st nspace creation works:
```
ndctl create-namespace -r region1 -s 268435456 -n nspace1
```
2nd creation fails:
```
ndctl create-namespace -r region1 -s 268435456 -n nspace2
libndctl: ndctl_pfn_enable: pfn1.1: failed to enable
```

in kernel messages, there is:
```
[   95.112129] memory add fail, invalid altmap
...
[   95.112226] Call Trace:
[   95.112233]  arch_add_memory+0x42/0x50
[   95.112238]  devm_memremap_pages+0x351/0x4a0
[   95.112243]  pmem_attach_disk+0x42e/0x560 [nd_pmem]
[   95.112246]  nd_pmem_probe+0x7e/0xa0 [nd_pmem]
[   95.112250]  nvdimm_bus_probe+0x69/0x110
[   95.112255]  driver_probe_device+0x31e/0x490
[   95.112257]  bind_store+0xed/0x170
[   95.112260]  drv_attr_store+0x27/0x40
[   95.112266]  sysfs_kf_write+0x3c/0x50
[   95.112268]  kernfs_fop_write+0x125/0x1a0
[   95.112271]  __vfs_write+0x1b/0x40
[   95.112272]  vfs_write+0xb1/0x1a0
[   95.112274]  SyS_write+0x55/0xc0
[   95.112280]  do_syscall_64+0x73/0x130
```
Why does this work differently in regions?

okartau:
[...]

PR #131, opened 2019-01-15
pmd-ndctl.go: align namespace size up to 1GB boundary

okartau:
This avoids alignement-related problems in direct-nvdimm mode.


okartau:
this will fix #117 and likely also older #9 (to be checked).
Note that we redefine 1024 -based integers few times here and there, should the constants defined somewhere, be made global?

okartau:
no it will not fix #9

PR #140, opened 2019-01-21
VGroups: create one group per NS-mode

### LINKED ISSUE #137: LVM mode: current volume groups creation scheme limits scalability
Volume Groups are currently created so that Phys Volumes from one Region form one region-specific Volume Group. This was initial prototype-level design while we typically worked with one Region only.
But this approach will pose a serious limitation to possible max. Logical Volume size in multi/many Regions cases, as one Logical Volume can not span more than one Volume Group.
What we should have instead, is one Volume Group per namespace mode, totaling exactly 2 groups:
- fsdax-mode Volume group containing fsdax mode phys volumes from all Regions
- sector Volume group containing sector mode phys volumes from all Regions

This should be doable by slightly re-arranging groups naming and loop over phys. volumes in vgm, plus according changes in code which walks existing volume groups

okartau:
Previously we were creating one NS-mode group per region,
which resulted in twice Vgroups as there were regions,
thus limiting max LVolume size to be region size.
There is no reason to maintain multiple groups that
logically serve same role, which goes by NS mode,
so let's create one VGroup per NS mode instead.

Fixes #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The PMEM-CSI driver follows the [CSI specification](https://github.com/container
 
 ### Architecture and Operation
 
-The PMEM-CSI driver uses LVM for Logical Volumes Management to avoid the risk of fragmentation that would become an issue if Namespaces were served directly. The logical volumes of LVM are served to satisfy API requests. This also ensures the region-affinity of served volumes, because physical volumes on each region belong to a region-specific LVM volume group.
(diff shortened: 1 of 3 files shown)


okartau:
it is better not to combine groups this way right now, closing this PR, code change will remain in forked repo branch just in case if needed later.

PR #144, opened 2019-01-25
Fix to test failure

avalluri:
In reset change 911d0b97fb4308fabf8d076012b9e3a95a40b9b6, where Capacity field in
RegisterControllerRequest is removed. This needs fixes to registry tests.


okartau:
that change helps me to achieve clean 'make test' result.
I need #150 additionally on openSUSE 15.0 where tail does not accept old-style +2 option

PR #152, opened 2019-01-29
update Kubernetes to master

pohly:
The previous vendor update was incomplete and broke building the E2E
test suite. To avoid that from happening again, `make build` now
ensures that all tests at least build. However, it isn't the default
target so developers have to remember running it.

This update replaces the Kubernetes fork (which was necessary to avoid
pulling in unneeded drivers and their dependencies) with current
master. 1.13 unfortunately isn't suitable because it still has that
problem.

"make test_e2e" now runs again, but still has test failures.


pohly:
FWIW, here are the test failures that I get:
```
Summarizing 9 Failures:

[Fail] sanity Controller Service CreateVolume [It] should not fail when creating volume with maximum-length name 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:479

[Fail] sanity Controller Service CreateVolume [It] should return appropriate values SingleNodeWriter NoCapacity Type:Mount 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:237

[Fail] sanity Node Service [It] should work 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:386

[Fail] sanity Controller Service CreateVolume [It] should fail when requesting to create a volume with already exisiting name and different capacity. 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:403

[Fail] sanity Controller Service CreateVolume [It] should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:285

[Fail] sanity Controller Service CreateVolume [It] should not fail when requesting to create a volume with already exisiting name and same capacity. 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:330

[Fail] sanity Controller Service DeleteVolume [It] should return appropriate values (no optional values added) 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:557

[Fail] sanity Controller Service ValidateVolumeCapabilities [It] should return appropriate values (no optional values added) 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:628

[Fail] PMEM Volumes [Driver: pmem-csi] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision storage with defaults 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/provisioning.go:228

Ran 34 of 61 Specs in 1904.872 seconds
FAIL! -- 25 Passed | 9 Failed | 0 Pending | 27 Skipped
```

Gopkg.toml
```diff
@@ -59,11 +75,36 @@
   name = "k8s.io/apiserver"
   branch = "master"
 [[override]]
-  name = "k8s.io/kube-aggregator"
```
avalluri: @pohly we are removing here and adding back in line 96, to make diff cleaner can you please keep this here and add  "k8s.io/apiextensions-apiserver" as a new [override]

pohly: The entries are sorted now. I think that makes more sense than sticking to some historic ordering just to keep diffs smaller.

pohly:
Some of my test failures may have come from not having the latest container images in my local Docker registry.

pohly:
And not having the CSI driver deployed. I thought "make start" did that, but it doesn't.

okartau:
I was also surprised by that number of failures. I saw 2 failures with some code omissions to get e2e to compile

pohly:
I ended up with:
- clarifying the deployment files (explicit Kubernetes version)
- installing the driver as part of "make start"
- using the installed driver for all tests

See updated PR.

In that mode I got the provisioning test to pass. Or at least it passed once. When rerunning everything, I then got:
```
Summarizing 4 Failures:

[Fail] PMEM Volumes [Driver: pmem-csi] [Testpattern: Dynamic PV (default fs)] provisioning [BeforeEach] should provision storage with defaults 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/k8s.io/kubernetes/test/e2e/framework/create.go:81

[Fail] sanity Node Service [It] should work 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:447

[Fail] PMEM Volumes [Driver: pmem-csi] [Testpattern: Dynamic PV (default fs)] provisioning [BeforeEach] should provision storage with mount options 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/k8s.io/kubernetes/test/e2e/framework/create.go:81

[Fail] PMEM Volumes [Driver: pmem-csi] [Testpattern: Dynamic PV (default fs)] provisioning [BeforeEach] should create and delete block persistent volumes 
/nvme/gopath/src/github.com/intel/pmem-csi/vendor/k8s.io/kubernetes/test/e2e/framework/create.go:81

Ran 37 of 61 Specs in 1580.456 seconds
```

okartau:
changes here are independent of fixes in #151 ? when I try on this branch on top of devel, I hit same issues as without #151, so perhaps we should merge #151 first, or we need manual combine

pohly:
These changes here don't conflict with #151, but for a successful test cluster setup it is needed.

okartau:
I can now run test_e2e and see 4 failures matching your result, that's good.
But I think we need to recheck what is really going on, as driver seems to do most of operations as role=Controller, which in intended deployment plan should be controller part talking to cluster Nodes.
This seems not happening, as Controller attempts volume operations, thinking it is running in Unified mode i.e. alone.
Seems the real clustered deployment which runs underneath, is not actually taking part of  volumes creation, and that may be reason behind some of failures.
Is this expected as being current state of things in e2e, or do we have some config still off target?
i.s. should we merge this, or keep studying more in scope this PR.

okartau:
perhaps we should merge this, to get also #156 merged and other activities moving.
We can continue with e2e improvement in new PRs

Gopkg.toml
avalluri: agreeed

pohly:

> I can now run test_e2e and see 4 failures matching your result, that's good.
> But I think we need to recheck what is really going on, as driver seems to do most of operations as role=Controller, which in intended deployment plan should be controller part talking to cluster Nodes.
> This seems not happening, as Controller attempts volume operations, thinking it is running in Unified mode i.e. alone.
> Seems the real clustered deployment which runs underneath, is not actually taking part of  volumes creation, and that may be reason behind some of failures.
> Is this expected as being current state of things in e2e, or do we
> have some config still off target?

There's nothing special about deployment for E2E anymore. It's literally
a "kubectl create -f deploy/kubernetes-1.13/pmem-csi.yaml", just as
users are told to use it. So no, this is not an expected failure and
needs to be analysed.

PR #156, opened 2019-01-30
test: used fixed sort order

okartau:
The ordering of text files is locale-specific. When comparing against
the reference files we have to ensure that the same ordering is used
regardless of the environment and/or the actual "sort" implementation.

LC_ALL=C ensures that for GNU sort and glibc. LANG=C is added just as
an extra precaution for other implementations.

### DIFF
--- a/test/test.make
+++ b/test/test.make
@@ -27,7 +27,7 @@ test: test_vendor_bom
 test_vendor_bom:
 	@ if ! diff -c \
 		<(tail -n +2 vendor-bom.csv | sed -e 's/;.*//') \
-		<((grep '^  name =' Gopkg.lock  | sed -e 's/.*"\(.*\)"/\1/') | sort); then \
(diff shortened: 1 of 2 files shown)


pohly:
> The ordering of text files is locale-specific. When comparing against
> the reference files we have to ensure that the same ordering is used
> regardless of the environment and/or the actual "sort" implementation.
>
> LC_ALL=C ensures that for GNU sort and glibc. LANG=C is added just as
> an extra precaution for other implementations.

Please let's merge my other PRs first. It'll be easier to redo this
change here on top of those (just sort the vendor-bom.csv) than to
manually sort out merge conflicts.

okartau:
I created PR just to get it closer to merge.
Perhaps commit msg should be cleaned as well here

avalluri:
@pohly your other PR is merged, can you please rebase this

avalluri:
Thanks @pohly , looks good. I will merge this change.

PR #157, opened 2019-01-30
Add initial Jenkinsfile

obedmr:
Signed-off-by: Obed N Munoz <obed.n.munoz@intel.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,7 @@ pipeline {
 
             options {
 
-                timeout(time: 15, unit: "MINUTES")
+                timeout(time: 60, unit: "MINUTES")
                 retry(3)
 
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,6 +35,7 @@ pipeline {
 
             steps {
 
+                sh 'docker build --target build -t $BUILD_IMAGE .'
                 sh 'docker run --rm \
                 -v `pwd`:$PMEM_PATH $BUILD_IMAGE \
                 bash -c "ldconfig; make test"'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,133 @@
+pipeline {
+
+    agent {
+
+        label "pmem-csi"
+    }
+
+    options {
+
+        timeout(time: 2, unit: "HOURS")
+
+    }
+
+    environment {
+
+        PMEM_PATH = "/go/src/github.com/intel/pmem-csi"
+        IMAGE_VERSION = "ci-${env.BUILD_ID}"
+        BUILD_IMAGE = "clearlinux-builder"
+        CLUSTER= "ci-$IMAGE_VERSION"
+        GOVM_YAML="_work/$CLUSTER/deployment.yaml"
+        TEST_BOOTSTRAP_IMAGES="localhost:5000/pmem-csi-driver:$IMAGE_VERSION  localhost:5000/pmem-csi-driver-test:$IMAGE_VERSION"
+
+    }
+
+    stages {
+
+        stage('make test') {
+
+            options {
+
+                timeout(time: 15, unit: "MINUTES")
+                retry(3)
+
+            }
+
+            steps {
+
+                sh 'docker run --rm \
+                -v `pwd`:$PMEM_PATH $BUILD_IMAGE \
+                bash -c "ldconfig; make test"'
+
+            }
+
+        }
+
+        stage('Build test image') {
+
+            options {
+
+                timeout(time: 60, unit: "MINUTES")
+                retry(2)
+
+            }
+
+            steps {
+
+                sh 'docker run --rm \
+                    -e BUILD_IMAGE_ID=1 \
+                    -e IMAGE_VERSION=$IMAGE_VERSION \
+                    -v /var/run/docker.sock:/var/run/docker.sock \
+                    -v /usr/bin/docker:/usr/bin/docker \
+                    -v `pwd`:$PMEM_PATH \
+                    $BUILD_IMAGE \
+                    bash -c "make build-images"'
+
+            }
+
+        }
+
+        stage('E2E') {
+
+            options {
+
+                timeout(time: 90, unit: "MINUTES")
+                retry(2)
+
+            }
+
+            steps {
+
+                    sh 'cd deploy; grep -rl ":canary"  | xargs -L1 sed -i "s|canary|$IMAGE_VERSION|" | echo "Change image tag in deployment files"'
+                    sh 'docker run --rm \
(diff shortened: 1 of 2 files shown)


Jenkinsfile.e2e.vms
```diff
@@ -0,0 +1,51 @@
...
+  stages {
+    stage('Build Images') {
+      steps {
+        sh "sed -i 's|canary|$CI_ID|' Makefile"
+        sh 'docker run --rm -v $PWD:/go/src/github.com/intel/pmem-csi -v /var/run/docker.sock:/var/run/docker.sock pmem-csi-ci:go-alpine make build-images'
+      }
+    }
+    stage('Setup K8s') {
+      steps {
+	sh '$HOME/dev/k8s/ci-scripts/deploy-k8s.sh $CI_ID $HOME/vms/images/clear-latest-cloud.img  $BUILD_USER'
```
pohly: Please don't use deploy-k8s.sh.

We need to verify that the cluster start method which is used by everyone else (`make start`) works.

Jenkinsfile.e2e.vms
```diff
@@ -0,0 +1,51 @@
...
+  sh 'docker run --rm -v $KUBECONFIG:/tmp/kube/config -e KUBECONFIG=/tmp/kube/config pmem-csi-ci:go-alpine kubectl get pods --all-namespaces'
+      }
+    }
+    stage('Setup pmem-CSI driver and Quick E2E') {
+      steps {
+	sh '$HOME/dev/k8s/ci-scripts/e2e-pmem.sh $CI_ID $BUILD_USER'
```
pohly: What does this script do?

Whatever it is, it needs to be in this repo, not in the CI.

Jenkinsfile
```diff
@@ -0,0 +1,27 @@
...
+  }
+  stages {
+    stage('Build') {
+      steps {
+        sh 'docker run --rm -v `pwd`:/go/src/github.com/intel/pmem-csi pmem-csi-ci:go-alpine bash -c "make"'
+      }
+    }
+    stage('Test') {
+      steps {
+        sh 'docker run --rm -v `pwd`:/go/src/github.com/intel/pmem-csi pmem-csi-ci:go-alpine bash -c "make test"'
```
pohly: How do we control which version of Go is used? This has to be configured in the pmem-csi repo, because it may depend on the branch and has to be changed together with the code.

chuyd:
Adding Jenkinsfile with make_test

pohly:
@chuyd: Jenkins doesn't seem to run this job. It's in "pending" for three days now.

chuyd:
@pohly I moved the Jenkins, now it's up and running

Jenkinsfile
```diff
@@ -0,0 +1,134 @@
...
+
+                timeout(time: 60, unit: "MINUTES")
+                retry(2)
+
+            }
+
+            steps {
+
+                sh 'docker run --rm \
+                    -e BUILD_IMAGE_ID=1 \
```
pohly: Why fixed to 1?

Jenkinsfile
```diff
@@ -0,0 +1,134 @@
...
+                    -e IMAGE_VERSION=$IMAGE_VERSION \
+                    -v /var/run/docker.sock:/var/run/docker.sock \
+                    -v /usr/bin/docker:/usr/bin/docker \
+                    -v `pwd`:$PMEM_PATH \
+                    $BUILD_IMAGE \
+                    bash -c "make build-images"'
```
pohly: This `make build-images` ends up rebuilding the build image because `CACHEBUST=1` here while it was the default value when building `BUILD_IMAGE` initially during the `make test` step.

If we set `BUILD_IMAGE_ID` consistently, ideally based on the current build number (`${env.BUILD_ID}`?), do we get Docker to reuse the layers?

Jenkinsfile
```diff
@@ -0,0 +1,134 @@
...
+                retry(3)
+
+            }
+
+            steps {
+
+                sh 'docker build --target build -t $BUILD_IMAGE .'
+                sh 'docker run --rm \
+                -v `pwd`:$PMEM_PATH $BUILD_IMAGE \
+                bash -c "ldconfig; make test"'
```
pohly: Is the `ldconfig` workaround still necessary? We now have it in the file:
https://github.com/intel/pmem-CSI/blob/4d781f15560cc512f067bf0684f03c68cee05aaa/Dockerfile#L13

Jenkinsfile
```diff
@@ -0,0 +1,134 @@
...
+                        -e TEST_CHECK_SIGNED_FILES=false \
+                        -e TEST_LOCAL_REGISTRY=localhost:5000 -e IMAGE_TAG="$IMAGE_VERSION" \
+                        -v /var/run/docker.sock:/var/run/docker.sock \
+                        -v `pwd`:$PMEM_PATH \
+                        -v /usr/bin/docker:/usr/bin/docker \
+                        -v `pwd`:`pwd` \
+                        -w `pwd` \
+                        $BUILD_IMAGE \
+                        bash -c "swupd bundle-add openssh-server &&  \
+                            make start && cd $PMEM_PATH && \
```
pohly: How can `make start` work if we are not already in the source code directory? `cd $PMEM_PATH` looks redundant here.

Hmm, is the source code available in two places (pwd and `$PMEM_PATH`)? That's confusing and I am worried that it'll break when testing at some point. Can't we have it just in one place?

Can you remove the separate `make start`? `make test_e2e` will do that automatically. As it stands now, we end up doing some operations like deploying the PMEM-CSI driver twice.

Jenkinsfile
pohly: This gets resolved by doing everything inside the build image specified by our own Dockerfile.

Jenkinsfile
pohly: Yes, it's necessary. Without it, libndctl isn't found when running tests inside the build container. This is different from the pkg-config error.

Jenkinsfile
pohly: Setting `CACHEBUST=1` consistently was not enough to increase cache usage. Even during `make push-images` is the "swupd update" command run twice:
http://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-CSI/branches/PR-305/runs/2/nodes/39/log/?start=0

That doesn't happen on my development machine (Docker 18.06.0-ce). The CI is using 18.09.6. Not sure where this different behavior is coming from.

Jenkinsfile
pohly: Also, why is openssh-server needed?

pohly: `openssh-server` provides the ssh client.

pohly:
Merged via https://github.com/intel/pmem-CSI/pull/305.

PR #158, opened 2019-01-31
test/e2e: use kubernetes-1.13 path for manifest

okartau:
This reduces e2e test failure count 4->3
We may consider using symlinks in deploy/ directory instead of explicit path
which we have to maintain then in sources, but as elsewhere already kubernetes-1.13 was used,
I continue same style here.

### DIFF
--- a/test/e2e/storage/csi_volumes.go
+++ b/test/e2e/storage/csi_volumes.go
@@ -102,7 +102,7 @@ var _ = Describe("PMEM Volumes", func() {
 						ClientNodeName: "host-1",
 					},
 				},
-				scManifest: "deploy/kubernetes/pmem-storageclass.yaml",
+				scManifest: "deploy/kubernetes-1.13/pmem-storageclass.yaml",
 				// Renaming of the driver *not* enabled. It doesn't support
 				// that because there is only one instance of the registry
 				// and on each node the driver assumes that it has exclusive


test/e2e/storage/csi_volumes.go
```diff
@@ -102,7 +102,7 @@ var _ = Describe("PMEM Volumes", func() {
 						ClientNodeName: "host-1",
 					},
 				},
-				scManifest: "deploy/kubernetes/pmem-storageclass.yaml",
+				scManifest: "deploy/kubernetes-1.13/pmem-storageclass.yaml",
```
pohly: A better solution would be to retrieve the Kubernetes version at runtime and then generating the path dynamically. As it stands now, if we ever have to introduce a "deploy/kubernetes-1.14" with a different storage class, we need to remember to update this code.

okartau: yes, as I stated, another approach would be symlinked dir kubernetes -> kubernetes-1.13 which we have to maintain, but idea there is that we only maintain that single symlink and keep (multiple) occurrences in source as "kubernetes". But as you already started putting exact kubernetes-1.13 in source, I continue that style in this PR. Overall, this multiple directories style is messy of course.

pohly:
I did that earlier with the intention ("TODO") that the hard-coded deployment file would get removed eventually. But that can't work for the storage class because it's always going to be needed, even when moving to a model where the driver is deployed only during cluster bring-up.

But  I'm fine with merging this as-is, we'll figure something out eventually.

PR #159, opened 2019-02-01
e2e sanity: lock unified driver onto host-1

pohly:
This is necessary because we need to know on which host we have to
create staging directories before calling the driver.


pohly:
It seems to have the desired effect (driver runs on host-1), but I still got one sanity test failure:
```
KUBECONFIG=`pwd`/_work/clear-kvm-kube.config REPO_ROOT=`pwd` go test -timeout 0 -v ./test/e2e -args -ginkgo.focus=sanity.Node.Service.*should.work
...
• Failure [27.351 seconds]
sanity
/nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:39
  Node Service
  /nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
    should work [It]
    /nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:364

    Expected error:
        <*status.statusError | 0xc000b34190>: {
            Code: 3,
            Message: "mount filesystem failedexit status 32",
            Details: nil,
            XXX_NoUnkeyedLiteral: {},
            XXX_unrecognized: nil,
            XXX_sizecache: 0,
        }
        rpc error: code = InvalidArgument desc = mount filesystem failedexit status 32
    not to have occurred

    /nvme/gopath/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:447
```

okartau:
with this change, error is same. I think the fix does not work because the actually serving running controller pod was started before e2e code runs (on node-2), and that remains running and serving through the tests. There is another controller pod deployed by e2e test (where fix of current PR applies), but that pod exits approx. 30s after test start
We seem to have bigger overall confusion about who starts controller and how long it runs,

okartau:
[commenting on merged PR]
asking just in case, @pohly when you renamed some pods to have "unified" in name,
you did not rename that one (test/e2e/storage/sanity.go, line 97)
```
		// Wait for pod before proceeding. StatefulSet gives
		// us a well-known name that we can wait for.
		f.WaitForPodRunning("pmem-csi-controller-0")
```

is this intentional?
I understand we create a pod there in BeforeEach section, but why we always wait for
that single pod that runs as pre-started deployment?
Would it make more sense to wait for pmem-csi-unified-0 ?
But if yes, what about namespace as it runs in dedicated pmem-XXXX/ namespace.
I tried a test run with rename pod name ( to pmem-csi-unified-0).
Success rate and testing time remains same, but output is somewhat (6%) shorter, so there was some effect. I will try to make sense from the diff to see what was it...
at least *error* was mentioned 1166 times in original output and 996 times in changed cases,
so statistically there was improvement. More deeper analysis will follow

pohly:
I haven't looked at the context, but probably that name should indeed have been renamed.

pohly:
The wait is there to ensure predictable runtimes for the following tests. Here we know that starting up a new pod can take a long time, so `f.WaitForPodRunning` can use a long timeout. Then the tests themselves can have shorter timeouts to fail faster when something else is wrong.

okartau:
closer look to output confirms, the diff in msgs containing word *error*, is caused by different wait pattern.
Original version: wait returns immediately (as pre-created pod runs already), followed by few extra `connection refused". Reconnecting...` messages per case, totaling in about 100 more such messages per test cycle. So it seems we indeed should replace the name there (I will prepare a PR).
But its also notable that total time for test cycle remained same.

PR #160, opened 2019-02-01
RFC: README.md: document persistency modes

pohly:
This introduces the problem, potential solutions, and what PMEM-CSI
supports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
 
 * **_Unified_** mode is intended to run the driver in a single host, mostly for testing the driver in a non-clustered environment.
 
+### Volume Persistency
+
+Normally, volumes are provided by a storage backend that is independent
+of a particular node. When a node goes offline, the volume can be
+mounted elsewhere. With PMEM-CSI, volumes are *local* and thus can
+only be used on the node where they were created.
+
+This limitation needs to be considered when designing and deploying
+applications that are to use local storage:
+
+* Persistent Volumes  
+A volume gets created independently of the application, on some node
+where there is enough free space. Applications using such a volume are
+then forced to run on that node and cannot run when the node is
+down. Data is retained until the volume gets deleted.
+* Ephemeral Volumes  
+An application is assigned to a node and when it starts to run, a new
+volume is created for it on that node. When the application stops, the
+volume is deleted.
+* Cache Volumes  
+Volumes are created on a certain set of nodes, each with its own local
+data. Applications are started on those nodes and then get to use the
+volume on their node. Data persists across application restarts. This
+is useful when the data is only cached information that can be
+discarded and reconstructed at any time *and* the application can
+reuse existing local data when restarting.
+* Delayed Persistent Volumes  
+This is a variant of persistent volumes. The only difference is that
+creating the volume is delayed until the first application using the
+volume is started, and then the volume is created on the node chosen
+for that first application run. From then on, applications will always
+have to run on that node.
+
+Not all of these modes are supported in CSI, Kubernetes and PMEM-CSI. This table shows the current status and known limitations:
+
+Volume | Kubernetes | PMEM-CSI | Limitations
+--- | --- | --- | ---
+Persistent | supported | supported | topology aware scheduling
+Ephemeral | [in design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#proposal) | supported via a fake volume | topology aware scheduling, resource constraints
+Cache | supported | not implemented | topology aware scheduling
+Delayed Persistent | supported, [CSI support might be faulty](https://github.com/kubernetes-csi/external-provisioner/issues/221) | not implemented | topology aware scheduling, topology aware volume provisioning
+
+[Topology aware scheduling](https://github.com/kubernetes/enhancements/issues/490)
+ensures that an application runs on a node where the volume was
+created. This is obviously needed for persistent volumes, but also for
+delayed persistent volumes because otherwise a restarted application
+might end up on the wrong node. Usage of cache volumes is more flexible,
+but an application still can only run on one of the nodes where the
+a cache volume is available. For CSI-based drivers like PMEM-CSI,
+Kubernetes >= 1.13 is needed. On older Kubernetes releases, pods must
+be scheduled manually onto the right node(s).
+
+[Topology aware volume provisioning](https://kubernetes.io/blog/2018/10/11/topology-aware-volume-provisioning-in-kubernetes/)
+is needed for delayed persistent volumes.
+
+The upstream design for ephemeral volumes currently does not take
+[resource constraints](https://github.com/kubernetes/enhancements/pull/716#discussion_r250536632)
+into account. If an application gets scheduled onto a node and then
+creating the ephemeral volume on that node fails, the application on
+the node cannot start until resources become available. The same
+applies to how this is currently implemented in PMEM-CSI: a fake
+volume gets created initially, without actually allocating space. Then
+when an application starts and it is time to mount that "existing"
+volume, an attempt is made to allocate the space required for the
+volume locally. Only if there is enough space can application startup
+continue.
+
 ### Driver Components
 
 #### Identity Server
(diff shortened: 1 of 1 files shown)


pohly:
The README.md is written as if the code had been changed already. It needs to be merged together with the actual code change.

README.md
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
 
 * **_Unified_** mode is intended to run the driver in a single host, mostly for testing the driver in a non-clustered environment.
 
+### Volume Persistency
```
pohly: The intend behind this section is to describe the problem without using Kubernetes-specific terminology.

Only when it comes to comments about the actual implementation does it refer to Kubernetes.

README.md
```diff
@@ -117,6 +217,8 @@ A production deployment can improve upon that by using some other key delivery m
 
 The following diagram illustrates how the PMEM-CSI driver performs dynamic volume provisioning in Kubernetes:
 ![sequence diagram](/docs/images/sequence/pmem-csi-sequence-diagram.png)
+<!-- TODO: update the diagram to show how ephemeral volumes get created and deleted -->
```
pohly: Strictly speaking, each persistency mode would need its own diagram. My proposal is to show it only for the ephemeral case, because that is a) the most important one and b) a bit unusual in how it handles PVC/PV.

README.md
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+have to run on that node.
+
+Not all of these modes are supported in CSI, Kubernetes and PMEM-CSI. This table shows the current status and known limitations:
+
+Volume | Kubernetes | PMEM-CSI | Limitations
+--- | --- | --- | ---
+Persistent | supported | supported | topology aware scheduling
+Ephemeral | [in design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#proposal) | supported via a fake volume | topology aware scheduling, resource constraints
+Cache | supported | not implemented | topology aware scheduling
+Delayed Persistent | supported, [CSI support might be faulty](https://github.com/kubernetes-csi/external-provisioner/issues/221) | not implemented | topology aware scheduling, topology aware volume provisioning
```
avalluri: The issue on "external-provisioner" does not solve the "Delayed Persistent", most close to Ephemeral.  In my opinion the "Delayed Persistent" is the result of lack of  topology awareness in CSI that time. It was a work around solution which i don't think should carry longer, once we implement the complete toplogy awareness while creating volume, this work around can be dropped.

README.md
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+* Persistent Volumes  
+A volume gets created independently of the application, on some node
+where there is enough free space. Applications using such a volume are
+then forced to run on that node and cannot run when the node is
+down. Data is retained until the volume gets deleted.
+* Ephemeral Volumes  
+An application is assigned to a node and when it starts to run, a new
+volume is created for it on that node. When the application stops, the
+volume is deleted.
+* Cache Volumes
```
avalluri: @pohly can you please help me in depict this usage on Kuberentes. Applications in the sense stateful/replica sets or might be two different applications? How does PV -> PmemVolume relation 1-1 or 1-many.

README.md
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+volume is created for it on that node. When the application stops, the
+volume is deleted.
+* Cache Volumes  
+Volumes are created on a certain set of nodes, each with its own local
+data. Applications are started on those nodes and then get to use the
+volume on their node. Data persists across application restarts. This
+is useful when the data is only cached information that can be
+discarded and reconstructed at any time *and* the application can
+reuse existing local data when restarting.
+* Delayed Persistent Volumes
```
avalluri: We might deprecate or merge this with *Persistent Volumes*, In implementation side, when there is enough information(csi.TopologyRequirement) we respect that and provision the volume accordingly, else we pick random node to provision.

README.md
avalluri: So different Applications share the same PV.

Just to make sure if we are on same lines on how does this node set are requested by Pod, does my https://github.com/kubernetes-csi/external-provisioner/issues/221#issuecomment-460552787 address this.

README.md
avalluri: What is the correct way in your opinion. Choosing node before hand without knowing Pod's other constraints might lead to more problems.

pohly: IMHO choosing the location of data is the most important choice, because it is final and cannot be changed afterwards. I think someone who sets up a cluster where some apps use persistent local data must have a plan for where those apps are meant to run. Communicating that plan to the pmem-csi driver via pod scheduling would work technically, but it seems rather indirect and non-obvious. I would prefer to have some explicit annotation on the PVC itself.

But I might be wrong - we would have to ask users.

README.md
pohly: No, I don't think that comment captures the idea. The choice between "cache volume" and "normal persistent volume" is something that the creator of the PVC must explicitly make by setting a driver-specific parameter. It is not something that pmem-csi deduces from the topology fields, and it is independent of the pods that then end up using the PV.

pohly: To phrase it differently, persistent and cache volumes are not "requested by a pod". They are requested by an admin by creating a PVC.

The only volumes that can be said to be "requested by a pod" are the ephemeral volumes, because only those are tied to a specific pod.

avalluri: > No, I don't think that comment captures the idea. The choice between "cache volume" and "normal persistent volume" is something that the creator of the PVC must explicitly make by setting a driver-specific parameter

driver-specific parameters you mean StorageClass or PVC? PVC cannot have parameters i guess(i might be wrong) other than Selector. Can't we us SotrageClass.AllowedTopologies to limit the nodes?

avalluri: PVC is more close to User/Pod, StorageClass is for cluster admins i feel.

pohly: I think the persistency type ("persistent" vs. "ephemeral" vs. "cache") belongs into the StorageClass as a parameter. That's something that the admin needs to make available, normal users usually don't have the rights to create cluster-level, non-namespaced entities like a StorageClass.

We could use use StorageClass.AllowedTopologies. That is converted for the CSI driver by external-provisioner, isn't it?

But my gut feeling is that the choice of nodes needs to be closer to the app deployer, i.e. the non-privileged user, which means offering a parameter in the PVC. But as you said, the ability to influence dynamic provisioning via a PVC are limited. There is PVC.Resources.Requests, but while resource names are strings, the values have to be a number.

I thought we could use the same concept as ceph (let users create secrets where the content of the secret includes parameters for provisioning), but looking closer at https://kubernetes-csi.github.io/docs/Usage.html#csi-provisioner-parameters I only find `${pv.name}` listed as valid template for provisioning. I'm unsure how to use that. I'll do some more research on this.


. This works by having templates in the storage class (:
```

Comment (avalluri) [reply README.md]: > I think the persistency type ("persistent" vs. "ephemeral" vs. "cache") belongs into the StorageClass as a parameter. That's something that the admin needs to make available, normal users usually don't have the rights to create cluster-level, non-namespaced entities like a StorageClass.
> 
I agree, though myself not clear how does persistency type really effect the driver behavior. Atleast i could think of knowing that this volume is used as "cache" at creation time effect the driver, but not sure about "persistent" vs. "ephemeral". My feeling is those are differentiated by the way the volume get deleted. But you can ignore me as long as your are clear with this :)

> We could use use StorageClass.AllowedTopologies. That is converted for the CSI driver by external-provisioner, isn't it?
> 
Yes, right. Thats the reason it attracted me.

> But my gut feeling is that the choice of nodes needs to be closer to the app deployer, i.e. the non-privileged user, which means offering a parameter in the PVC. But as you said, the ability to influence dynamic provisioning via a PVC are limited. There is PVC.Resources.Requests, but while resource names are strings, the values have to be a number.
> 
There is [PVC.Selector](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#selector) can be used by Pod/PVC to select the volumes only match with volume labels.

But currently these selection is not handled by external-provisioner, even if it does in near feature, it would be via CreateVolumeRequest.AccessibleTopology i feel. That is one reason i am negotiating in my issue on external-provisioner.

> I thought we could use the same concept as ceph (let users create secrets where the content of the secret includes parameters for provisioning), but looking closer at https://kubernetes-csi.github.io/docs/Usage.html#csi-provisioner-parameters I only find `${pv.name}` listed as valid template for provisioning. I'm unsure how to use that. I'll do some more research on this.

Thanks for sharing, i will check.

File: README.md [hunk 6]
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+
+This limitation needs to be considered when designing and deploying
+applications that are to use local storage:
+
+* Persistent Volumes  
+A volume gets created independently of the application, on some node
+where there is enough free space. Applications using such a volume are
+then forced to run on that node and cannot run when the node is
+down. Data is retained until the volume gets deleted.
+* Ephemeral Volumes
```
avalluri: @pohly What is the expectation from driver to support this volume type.

As of my understanding there is no way driver knows when the pod existed. To support this usecase CO must initiate DevelVolume request then driver aromatically delete the volume . Unless external component call this upon Pod exit, driver itself can do nothing.

README.md
```diff
@@ -75,6 +75,73 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+
+Normally, volumes are provided by a storage backend that is independent
+of a particular node. When a node goes offline, the volume can be
+mounted elsewhere. With PMEM-CSI, volumes are *local* and thus can
+only be used on the node where they were created.
+
+This limitation needs to be considered when designing and deploying
+applications that are to use local storage:
+
+* Persistent Volumes
```
avalluri: Keeping in mind of local storage limitations, we better support only delayed binding `WaitForFirstConsumer`.

pohly: You are looking at this from the perspective of an implementor ("let's use `WaitForFirstConsumer`"). Can we first discuss it in more abstract terms? Are you saying that "Persistent Volumes" aren't useful and that we should instead implement "Delayed Persistent Volumes"?

README.md
pohly: One possible implementation of an ephemeral volume is this:
- `CreateVolume`: return a topology that maps to all nodes where PMEM is available (in the sense of "it might be available when a pod runs", not as in "available right now")
- `NodeStageVolume`: create the volume, format and mount it
- `NodePublishVolume`: bind-mount into pod
- `NodeUnpublishVolume`: remove bind-mound
- `NodeUnstageVolume`: delete the volume

I am aware that this may lead to scheduling pods onto a node where currently no local storage is available. There is no good solution for that, we need to work with upstream to enhance the Kubernetes scheduler or write our own.

okartau: What if we, at CreateVolume, using pmem-csi internal comms, actually reserve the space on those nodes having enough free space? Then the later-following StageVolume can't fail on any of those nodes as there is namespace waiting. After one node starts using that Volume, controller will need to order freeing reservations on other nodes. That would help with "oops there was no space when we really started to create one". Downside is, nodes will perform extra allocations and have to deal with removing those, with added complexity with more error-paths. And we should add some garbage-orphan-checking pass (at plugin startup?) to detect and clean unused left-over namespaces, which is actually good idea to have anyway (we dont have such now).

pohly: The conceptual model is that there is one PVC that describes a volume of a certain size. Then each pod that uses that PVC gets its own local instance of that size.

That doesn't work when we reserve space once for one pod when creating the PVC and then free the reservation once that pod runs.

avalluri: As of my knowledge "ephemeral" volumes concept is for Pods requesting "in-line" volume claims( i know its not yet supported for csi volumes), and  are owed by single Pod, and cannot be shared by two applications, that means sharing of PVC/PV  concept does not apply here.

Ephemeral requests should reach us only after once Scheduler choose the node for the requested Pod, and we create the volume on that node, so no need of guess work here by driver.
And we strictly *deny* reattachment requests for these volumes and clean the volume on `ControllerUnpublish` unlike normal volumes where we clear on `DeleteVolume`.

If i want to say in Kuberntes terms we only support "ephemeral" volumes that are created by `WaitForFirstConsumer` StorageClass

@pohly adn @okartau can we agree to this?

pohly: @avalluri yes, ideally "ephemeral volumes" should be implemented as inline volumes. But inline volumes are in design at the moment upstream and thus not available to us. What I described and what @okartau commented on is my proposal how we can approximate the concept with a fake PV. `WaitForFirstConsumer` doesn't help, because we will have multiple consumers of that PV.

As we agreed yesterday to focus on "cache volumes", we don't need to think too hard about how to implement the other persistency models - at least not until it becomes urgent.

okartau:
changes of this PR were covered by recent csi-1.0 update, so this can be closed without merging, right?

pohly:
Was included in #177 .

PR #161, opened 2019-02-04
test/e2e: wait for pmem-csi-unified-0 in BeforeEach

okartau:
Due to recent rename of dynamically created pods,
this Wait should be checking pmem-csi-unified-0 instead
of pmem-csi-controller-0. This change reduces
amount of "connection refused, reconnecting" errors.


pohly:
Yep. Let's merge.

PR #164, opened 2019-02-06
split controller code to master and node controllers

avalluri:
Motivation behind this change is:
- we can make node controller as self contained, i.e, each node can create and
  delete volumes independent of controller.
- clearly divides what parts are handled by master and node, say for example
  nothing related to DeviceManager should be in master code.
- this is an enabler change to implement fully topology aware provisioning by
  dropping workaround solution of creating pmem device on attachment.

Chose reattachment strategy(this only valid till we implement full topology aware):
  - reattaching to same node without data lose. The assumption is that, it
    helps for "cahche volume" type usecases.
  - reattaching to different node, with no data.


pohly:
Overall this sounds like the direction, but it is not clear to me what the result or next steps will be after merging this.

Does it fully implement one of the persistency models described in PR #160? Which one?

If it changes user-visible behavior, then the README.md must be updated in the same PR, otherwise we end up describing one thing while providing another.

avalluri:
> Overall this sounds like the direction, but it is not clear to me what the result or next steps will be after merging this.
> 
The next step in my plan(which i am working on) is to make use of CSINodeInfo support in external-provier, in other words implement `CreateVolumeRequest.AccessibilityRequirement` in our driver.

> Does it fully implement one of the persistency models described in PR #160? Which one?
> 
No, i cannot say at the moment  that this change address _fully_ any of three models you proposed. I could see this as enabler towards supporting them.

> If it changes user-visible behavior, then the README.md must be updated in the same PR, otherwise we end up describing one thing while providing another.

The only change with existing behavior is,  addressing at some degree of reattaching issue. Two reasons why i did not touch README is:
 * the reattachment policy we chose is a temporary solution, it might change(atlest needs to revisit) after implementing   `CreateVolumeRequest.AccessibilityRequirement`.
 * the current README does not mention anything related to reattachment.

pohly:
>> Does it fully implement one of the persistency models described in PR #160? Which one?
>> 
> No, i cannot say at the moment  that this change address _fully_ any
> of three models you proposed. I could see this as enabler towards
> supporting them.

Is it worth merging the PR while further work is still needed?

>> If it changes user-visible behavior, then the README.md must be updated in the same PR, otherwise we end up describing one thing while providing another.
>
> The only change with existing behavior is,  addressing at some degree
> of reattaching issue.

How? Or phrased differently, what works now that didn't work before?

> Two reasons why i did not touch README is:
>  * the reattachment policy we chose is a temporary solution, it might change(atlest needs to revisit) after implementing   `CreateVolumeRequest.AccessibilityRequirement`.
>  * the current README does not mention anything related to
>  reattachment.

This leaves people currently trying to use pmem-csi in the dark about
this crucial aspect. I think it is worthwhile adding a section now, even
if it just mentions the problem and says that work is in progress.

pohly:

>> Overall this sounds like the direction, but it is not clear to me what the result or next steps will be after merging this.
>> 
> The next step in my plan(which i am working on) is to make use of
> CSINodeInfo support in external-provier, in other words implement
> `CreateVolumeRequest.AccessibilityRequirement` in our driver.

Hmm, why work on that first? The most important persistency mode is
"ephemeral volumes", and CreateVolumeRequest.AccessibilityRequirement is
not needed for that, is it?

Which of the four modes are you trying to implement?

avalluri:
> > Does it fully implement one of the persistency models described in PR #160? Which one? > No, i cannot say at the moment that this change address _fully_ any of three models you proposed. I could see this as enabler towards supporting them.
> Is it worth merging the PR while further work is still needed?

The main goal of this change is to split the controller code. that way it is self contained. The further work I mentioned was about "reliable scheduling issue", which i would like to address in next PR. So lets consider this for merging.

> > If it changes user-visible behavior, then the README.md must be updated in the same PR, otherwise we end up describing one thing while providing another. The only change with existing behavior is, addressing at some degree of reattaching issue.
> How? Or phrased differently, what works now that didn't work before?

Currently pmem-csi is deleting the volume on Unpublish call, which means we don't support volume "persistency". This change address this by keeping the volume till we get Delete call. While reattaching we check if its moving to different node or not. if its same node we keep the data(cache usecase), else clean the device on unpblished node and create new one where its being attached.

pohly:
I would prefer to merge this once it fully implements one of the described persistency modes together with a user-visible description. As it stands now, at least I don't see the full picture of how you intend to make those modes work and thus can't review.

But if you feel that it's worthwhile to merge some intermediate step already now, then I won't object.

pohly:
> Currently pmem-csi is deleting the volume on Unpublish call, which means we don't support volume "persistency".

`ControllerUnpublishVolume` or `NodeUnpublishVolume`? I assume you mean the former, because reusing a volume on the same node works fine in the E2E testing.

Speaking of the E2E testing, does that still pass with this change here?

avalluri:
closing this PR infavour of #177  as it is containing these changes

PR #166, opened 2019-02-06
advance to ndctl version 64

### LINKED ISSUE #163: verify with ndctl v64 and step up to use v64
ndctl v64, v64.1 have been released recently.
We should advance our NDCTL version and verify:
- building in Docker
- use (via make) with libs on some distros, after v64 becomes available on those

okartau:
Also, some streamlining in ndctl build deps definition, so that we can define
those once instead of three replicas that we had previously

Fixes #163

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ IMAGE_VERSION_pmem-csi-driver=canary
 IMAGE_VERSION_pmem-ns-init=canary
 IMAGE_VERSION_pmem-vgm=canary
 IMAGE_TAG=$(REGISTRY_NAME)/$*:$(IMAGE_VERSION_$*)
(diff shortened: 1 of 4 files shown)


okartau:
don't merge yet, I will run some e2e tests

okartau:
In LVM2 mode ndctl v64 does work.
In devicemode:direct this is not testable now because of #167, and as ndctl testing makes more sense in direct mode, we should wait until we can test this properly.

okartau:
with deployment config mods, and added workaround to get around read-only /sys,  I can get e2e testing passing with this PR applied, i.e. using v64 of ndctl, so from that POV this PR is good for merging.

PR #169, opened 2019-02-08
e2e fixes

pohly:


okartau:
this affects pmem-csi tester even if we fix runtime to docker mode?
I will try with this change on my system

okartau:
I can create test cluster (make start), and run e2e tests (make test_e2e):
`SUCCESS! -- 35 Passed | 0 Failed | 0 Pending | 26 Skipped`
I verified on VM that /opt/cni/bin contains symlinks to /usr/libexec/cni

k8s node versions are 1.13.3
Clear version is 27700

pohly:
> this affects pmem-csi tester even if we fix runtime to docker mode?

That's a good question. I'm not sure whether it makes any difference in that mode.

okartau:
That change is needed to get deployment running on top of csi-1.0 branch by @avalluri  (and that is, even if docker mode is selected). So I will merge this.

Issue #167, opened 2019-02-07
Cluster deployment fails using Devicemode:Direct


okartau:
Cluster deployment used to work when we merged the devicemode:direct code into devel, just short while ago. But something has changed after that. Now on a recent qemu-based test cluster, with Clear=27660 and Kubernetes=1.13.3, a pmem-csi-driver node pod will not start if /dev hostpath is passed into it (and that volumemount addition was the key turn point which made devicemode:direct deployment working).
Error msg is:
```
Error: failed to start container "pmem-driver": Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"rootfs_linux.go:58: mounting \\\"/var/lib/kubelet/pods/3fd0ec14-2ac5-11e9-8362-deadbeef0100/containers/pmem-driver/fd9b54f9\\\" to rootfs \\\"/var/lib/docker/overlay2/17c0636357f4c679d588d8a448c752ec5496271a74d682aee58382d513891a8e/merged\\\" at \\\"/var/lib/docker/overlay2/17c0636357f4c679d588d8a448c752ec5496271a74d682aee58382d513891a8e/merged/dev/termination-log\\\" caused \\\"no such file or directory\\\"\"": unknown
```

okartau:
I wonder is this related to terminationMessagePath ?
terminationMessagePath: /dev/termination-log
but why started to fail just recently?

okartau:
indeed, the problem does not happen if I add this to pod spec,
mapping termination message path under /tmp instead of /dev:
`terminationMessagePath: /tmp/termination-log`

next question, is this safe thing to do, or does it create some additional trouble.
At least mount entries appear for /tmp/termination-log, seems k8s handles this changed path.

pohly:
This looks like a regression in Clear Linux. Current master of pmem-csi works fine on Clear Linux 27500. Changing the `terminationMessagePath` is just a workaround, the default needs to work.

Where did you find this error message with "failed to start container"? I tried with Clear Linux 
27670, and there the controller fails to start because of "Pod sandbox changed, it will be killed and re-created":
```
$ kubectl get pods
NAME                    READY   STATUS              RESTARTS   AGE
pmem-csi-cdxz6          1/2     CrashLoopBackOff    22         91m
pmem-csi-controller-0   0/4     ContainerCreating   0          91m
pmem-csi-npgwp          1/2     CrashLoopBackOff    22         91m
[...]

Issue #112, opened 2018-12-21
cri-o: read-only /sys


pohly:
When bringing up the simulator with TEST_CRI=crio in a `test/test-config.d/cri.sh` file, the `/sys` filesystem is read-only in the VM guests. That breaks initialization of the simulated PMEM. It works fine when using Docker.

The same is likely to happen when using cri-o on real hardware, therefore this issue needs to be investigated.

okartau:
/sys is read-only in pmem-csi pod in recent Clear-based test cluster even with TEST_CRI=docker.
Clear version is 27700.
I found this while running e2e test in devicemode:Direct, but it means failure should also happen
in devicemode:LVM, as initial stage of namespace creation should fail.

okartau:
in LVM mode, deployment and e2e test works in same cluster, means that in pmem-ns-init container /sys must be writable. What is the diff? One diff is that pmem-ns-init is classified as initContainer, but both containers are marked as privileged.

okartau:
We repeatedly seem to stumble to read-only /sys mount where we critically need read-write access.
The problem is exposed in some indirect way and giving potentially fuzzy error messages.
Situation may get changed by external-to-pmem-csi changes (k8s, docker, crio, clear) and
is therefore difficult to connect to known changes.
What about performing check for read-write mount of /sys during start of plugin, and refuse to continue (with clearly stated error msg) if there is no write access, as continuing would lead to errors anyway.
We need to run such check at start of:
- pmem-ns-init (initContainer) in devicemode:lvm
- pmem-driver in devicemode:ndctl

okartau:
Developing this idea one step further:
it would be very useful to detect isolated /dev in container, which leads to error path like in #168 in devicemode:direct. 
Detection of isolated /dev is not as easy as read-only /sys, I guess.
But if we can detect it, we don't need to develop not-so-clean solutions to #168, like #170 is,
as then we can avoid bad-state namespace creations altogether.

okartau:
With added explicit /sys mount in manifest, e2e test becomes runnable also with TEST_CRI=crio setting.
I made a test pass using devicemode:direct
However, run-time in crio mode  (2400 sec) is longer than typical (1600 sec) in docker mode, and there was one sanity-test  error. To be investigated more.

okartau:
We have now added:
- explicit mount of /sys for containers deployment (in direct mode) -- allows use of /sys by libndctl
[...]

PR #170, opened 2019-02-08
CreateDevice: avoid space-exhaustion loop on partial volume creation failure

### LINKED ISSUE #168: space on device exhausted if /dev from host is not imported in container
Recently discovered deployment error in #167 exposes nasty error-loop scenario where all space on device gets quickly consumed, leading to denial-of-service state.
Although the bad state is a side-effect of another error, we should try to make pmem-csi plugin more robust against such scenario.
What happens is this in ControllerPublishVolume:
if VolumeID not published:
  - status = (create namespace AND zero begin of it)
  - return if error
  - mark VolumeID published

Failure happens in "zero begin of it" part, making the 1st half succeed, but 2nd half failed,
Zeroing fails because /dev is not imported into pod container and container does not see the device.
After "return if error" part, Kubernetes repeats the call with same VolumeID instantly
and we reserve next piece of same size. This repeats until device runs out of space.
The (many) allocated namespaces have same VolumeID recorded in their Name field, making one of those  useful in some next stages, but all the others will remain orphaned.
We should improve by trying one of: 
- next allocation should re-use allocated namespace for that VolumeID, this requires that we start recording created namespaces immediately after creation succeeds, not just at "mark" part.
(needs changes as both "create" and "zero" ops are isolated in the dev-manager layer now).
- at "return if error" part we restore previous state, i.e. destroy orphaned namespace

okartau:
There was failure scenario in devicemode:direct where namespace
creation succeeded but zeroing of device did not, i.e. partial
success, but returned FAIL, which caused k8s to repeat,
and as created namespace remained there, all device space
was quickly consumed. We address this by trying to clean up
on partial failure.

Fixes #168

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -78,6 +78,16 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error 
 	if nsmode != string(ndctl.FsdaxMode) && nsmode != string(ndctl.SectorMode) {
 		return fmt.Errorf("Unknown nsmode(%v)", nsmode)
 	}
+	// Check that such name does not exist. In certain error states, for example when
+	// namespace creation works but device zeroing fails (missing /dev/pmemX.Y in container),
+	// this function is asked to create new devices repeatedly, forcing running out of space.
+	// Avoid device filling with garbage entries by returning error.
+	// Overall, no point having more than one namespace with same name.
+	_, err := lvm.GetDevice(name)
+	if err == nil {
+		glog.Infof("Device with name: %s already exists, refuse to create another", name)
+		return fmt.Errorf("CreateDevice: Failed: namespace with that name exists")
+	}
 	// pick a region, few possible strategies:
 	// 1. pick first with enough available space: simplest, regions get filled in order;
 	// 2. pick first with largest available space: regions get used round-robin, i.e. load-balanced, but does not leave large unused;
--- a/pkg/pmem-device-manager/pmd-ndctl.go
+++ b/pkg/pmem-device-manager/pmd-ndctl.go
@@ -65,6 +65,16 @@ func (pmem *pmemNdctl) GetCapacity() (map[string]uint64, error) {
 }
 
 func (pmem *pmemNdctl) CreateDevice(name string, size uint64, nsmode string) error {
+	// Check that such name does not exist. In certain error states, for example when
+	// namespace creation works but device zeroing fails (missing /dev/pmemX.Y in container),
+	// this function is asked to create new devices repeatedly, forcing running out of space.
+	// Avoid device filling with garbage entries by returning error.
+	// Overall, no point having more than one namespace with same name.
+	_, err := pmem.GetDevice(name)
+	if err == nil {
+		glog.Infof("Device with name: %s already exists, refuse to create another", name)
+		return fmt.Errorf("CreateDevice: Failed: namespace with that name exists")
+	}
 	// align up by 1 GB, also compensate for libndctl giving us 1 GB less than we ask
 	var align uint64 = 1024 * 1024 * 1024
 	size /= align


okartau:
interesting... current PR code (ndctl part) fails to destroy namespace (if we add checking for error):
`Destroy returned: failed to disable namespace: No such file or directory`
This is pretty bad, as it blocks us from removing, even if the force arg is set to TRUE.
More investigation needed.
Deletion must be possible as I can delete such namespaces from command line using ndctl.
But the difference from CLI is, this code runs inside container, without /dev/ mounted.

okartau:
Another approach. As it seems we can not clean up by deleting, lets avoid creating next namespace for request with same name (i.e. VolumeID). As such namespace exists, we return OK. This leads to failure in next step: StageVolume as file system creation will fail (as there is no device). But as whole plugin is in non-functional state at that point, it does not matter much in which stage we fail.
At least we avoid repeated creation of same-name object which instantly eats up all the space in host.
Overall, refusal to create same-name object makes sense as this name (VolumeID) is used to look up namespace later, so having more than one with same name makes no sense.
Although the error state so far has been seen in devicemode:direct, I add same check on LVM2 side as well, for uniformity.

pohly:
> Another approach. As it seems we can not clean up by deleting, lets
> avoid creating next namespace for request with same name
> (i.e. VolumeID). As such namespace exists, we return OK. This leads to
> failure in next step: StageVolume as file system creation will fail
> (as there is no device). But as whole plugin is in non-functional
> state at that point, it does not matter much in which stage we fail.

Calls are supposed to be idempotent. That means that calls have to
succeed eventually when called multiple times.

When you say "plugin is in non-functional state", what exactly does that
mean?  How does that happen and do users recover from that state?

okartau:
> When you say "plugin is in non-functional state", what exactly does that mean? How does that happen and do users recover from that state?

Specifically here I describe related state I saw when debugging state of #167. It is possible to get pmem-csi plugin running so that /dev/ is not passed into container. Such problematic state is not detected/flagged (not at least by current code, worth trying perhaps, not even sure how can it be done...).
The operating result in such state will be mixed and lead to space exhaustion, because namespace creation will work, but /dev/pmemX.Y device will not appear inside container, i.e. plugin can not provision volume and file system on top of that namespace.
That's what I mean with "non-functional":plugin will not provision anything anyway.

It's true this is result of mis-configuration, invalid manifest, changes in containering layer, etc. such issues, but we can not rule out bad things can happen, and I think we should minimize bad impact. The impact would be device filled with garbage entries which need manual cleanup.

pohly:

>> When you say "plugin is in non-functional state", what exactly does that mean? How does that happen and do users recover from that state?
>
> Specifically here I describe related state I saw when debugging state
> of #167. It is possible to get pmem-csi plugin running so that /dev/
> is not passed into container. Such problematic state is not
> detected/flagged (not at least by current code, worth trying perhaps,
> not even sure how can it be done...).

In that case it makes sense to fail permanently and ideally without
causing more breakage, so this PR is an improvement.

okartau:
one more thing to think here, is it good idea to return OK if same-name namespace already detected,
or should we return error instead. I will try both ways.

okartau:
I tried with return of error "namespace with same name exists", end result is same, plugin is called number of times with same request (18 times), then calling stops(pauses?).
I think I will change code in PR to return error instead of OK, to be more clear.

okartau:
code changed now to return error, force pushed.
OK to merge?

avalluri:
@okartau sorry for late comments, in my opinion CreateDevice calling with existing volume name/id shouldn't return error, instead it should return the same device. If i understood correct, the failure occurrence was in "ClearDevice", in that case we should delete the newly created device and return err. Instead of returning error

okartau:
I start to feel more and more that current PR tries to put rubberband fix around already badly broken state-
Instead, we should try to implement detection enabling to avoid entering such state, like I expressed in recent comment under #112.
If we can detect /dev in container which is isolated i.e. not mounted from host, and we are in devicemode:direct, then we should not even start running, thus avoiding creation bad namespace entries. There is no point running if any provision attempt is known to become failure.

okartau:
Change in this PR seems the best we can do right now, so I propose to merge this, allowing to close currently open issues about this topic

okartau:
ok to merge? any comments?

PR #175, opened 2019-02-19
WIP: clean and reduce logging

okartau:
Some log messages were relevant in early prototyping days
and do not add value any more, like which Bus and Region
we walk in loop.

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -28,17 +28,14 @@ func NewPmemDeviceManagerLVM() (PmemDeviceManager, error) {
 	}
 	volumeGroups := []string{}
 	for _, bus := range ctx.GetBuses() {
-		glog.Infof("NewPmemDeviceManagerLVM: Bus: %v", bus.DeviceName())
 		for _, r := range bus.ActiveRegions() {
-			glog.Infof("NewPmemDeviceManagerLVM: Region: %v", r.DeviceName())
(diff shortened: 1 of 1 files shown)


okartau:
I noticed those debug-like Bus and Region logs lines repeated a lot in e2e test logs, so I started to look critically what can be reduced. (Thats the removed lines)
But there are also few log msgs think we can change to V(X) level dynamic logs, like V(5)
what is used here and there.
But do we have agreement what levels should we use for what?
In e2e log, I don't want to see things that are repeated there now, like what CAPABILITY we add.
But then again, I dont want to completely remove it, but assign some real-verbose level.
But level 5 is already activated for e2e testing,
should we put real-verbose stuff for some other level (>5), or should we start running e2e test
at some other. like level 3 or 4.

pohly:
> But do we have agreement what levels should we use for what?

There are guidelines regarding that in
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

One of my objections against glog as a logging library is that it
leaves the choice of logging levels to the individual developer or
project.

> But level 5 is already activated for e2e testing,
> should we put real-verbose stuff for some other level (>5), or should we start running e2e test
> at some other. like level 3 or 4.

We could make the log level configurable in the Makefile. The default
for manual runs then can be something lower, while a CI system with good
per-test output capturing and browsing could use something higher.

pohly:
The change itself is fine. We can merge and iterate further, like replacing glog with klog.

Issue #172, opened 2019-02-13
Add detection of blocking pre-conditions


okartau:
As expressed in those comments:
https://github.com/intel/pmem-CSI/issues/112#issuecomment-463167336
https://github.com/intel/pmem-CSI/issues/112#issuecomment-463172044
https://github.com/intel/pmem-CSI/pull/170#issuecomment-463175718

It would make sense to detect blocking pre-conditions during startup of pmem-csi where possible, (followed by exit with clear error message) instead of implementing error handling, workarounds and "fixes" to states which are broken beyond repair, and often detectable already at plugin start-time.

At least 2 such conditions are currently on table:
1. /sys mounted read-only in container
This is easy to detect and we should exit with error msg right away in all cases where libndctl write-operations would be needed later (pmem-ns-init in lvm mode, pmem-driver in direct mode)

2. /dev not passed from host
This seems to apply in direct mode only
(as LVM block devices appear inside container even if /dev/ is not passed from host)
Such detection may not appear simple to do, but if we can do, we dont need workaround like #170

okartau:
Detection of 2 is not simple. How can container detect was /dev passed from host as volume or not?
If there is method to detect that, we should use that.
A real-life-scenario test would be:
1. create a small namespace for trial
2. (if 1. works) check is there corresponding block device in /dev
If we don't see device in step 2. that means /dev was not passed properly to the container.
We need to admit such test code adds it's own layer of complexity and error-handling needs,
like what to do if step 1 fails for some reason (no free space), then we can not
rely on this test, what should we decide then about runnable state.
We can try to claim "if no free space at plugin startup, plugin can't operate anyway", as CSI volume lifecycle typically does start with cleaning up older volumes (or does it? Does any authority track existing CSI volumes so that some could be cleaned after start of node, to free up space)

okartau:
More ideas to solve 2. condition: 
As creating an artificial namespace feels bit clumsy, so what about middle ground method: After a namespace creation, check for existence of corresponding /dev/ block device entry. If there is no entry, exit with error. 
That method is not by original idea in this issue (i.e. start-time failure), so it's not ideal in the sense that driver may remain in operation for some (possibly long) time until first real creation causes it's death. But it is better than remain running and try to recover (which likely will never work out).

okartau:
what I proposed in previous comment, does not avoid problem of filling the media with garbage entries, just the loop will probably happen through restarts of pmem-csi driver. 
[...]

PR #176, opened 2019-02-20
pmd-ndctl.go: Add start-time check for writable /sys

okartau:
In Devicemode_direct (alias ndctl), the driver can operate
only if /sys is mounted in read-write mode.
It has been seen however that sometimes container gets
deployed in such a way that /sys is mounted read-only.
It makes sense to detect writable /sys at start-time
and refuse to run, instead of stumbling to that later in operation.

### DIFF
--- a/pkg/pmem-device-manager/pmd-ndctl.go
+++ b/pkg/pmem-device-manager/pmd-ndctl.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/intel/pmem-csi/pkg/ndctl"
 	"k8s.io/klog/glog"
+	"k8s.io/kubernetes/pkg/util/mount"
 )
 
 type pmemNdctl struct {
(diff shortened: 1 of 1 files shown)


okartau:
This implements one of two checks proposed in #172

okartau:
For Devicemode:LVM, we could add similar check in startup phase of pmem-ns-init which needs writable /sys in LVM mode. But, two points suggest we can perhaps skip adding:
- pmem-ns-init runs as initContainer, and I haven't yet seen case when /sys is not writable there
- if we still hit read-only /sys, initContainer running before actual driver, either:
   - gets "read-only system" error at namespace creation attempt and exits (clear enough indication), 
   - or, it does not need to create namespaces (set of namespaces exists) and plugin is able to operate in LVM mode on top of existing set of namespaces.

so I think in LVM mode we can live without such start-up stage check

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -19,6 +20,24 @@ func NewPmemDeviceManagerNdctl() (PmemDeviceManager, error) {
...
+	// Check is /sys writable. If not then there is no point starting
+	mounter := mount.New("")
+	mounts, err := mounter.List()
+	for i := range mounts {
+		glog.Infof("NewPmemDeviceManagerNdctl: Check mounts: device=%s path=%s opts=%s",
+			mounts[i].Device, mounts[i].Path, mounts[i].Opts)
+		if mounts[i].Device == "sysfs" && mounts[i].Path == "/sys" {
+			for _, opt := range mounts[i].Opts {
+				if opt == "rw" {
+					glog.Infof("NewPmemDeviceManagerNdctl: /sys mounted read-write, good")
```
avalluri: can't we return from here.

okartau:
the log in "rw" case is bit of in extra-for-debug-only style, as is also the "check mounts:" trace log above. In final version those two logs could both go away. But do we delete those right here or later, our decision.

okartau:
Having few log lines once during startup is not too noisy, could provide valuable info about mounts from inside container, so lets keep them and merge.

PR #177, opened 2019-02-20
Persistency model changes

avalluri:
Currently supported persistency types are:
 - cache volume
 - ephemeral volume

These changes are over csi v1.0, so might not work on kuberntes < v 1.13. The idea is we should maintain pre csi-1.0 changes in a branch(csi-0.0.3). I will review supported persistency models within the limitations of csi v0.0.3 and submit PR to csi-0-0.3 branch.


README.md
```diff
@@ -87,6 +87,50 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+The upstream design for ephemeral volumes currently does not take
+[resource constraints](https://github.com/kubernetes/enhancements/pull/716#discussion_r250536632)
+into account. If an application gets scheduled onto a node and then
+creating the ephemeral volume on that node fails, the application on
+the node cannot start until resources become available. The same
+applies to how this is currently implemented in PMEM-CSI: a fake
+volume gets created initially, without actually allocating space. Then
+when an application starts and it is time to mount that "existing"
+volume, an attempt is made to allocate the space required for the
+volume locally. Only if there is enough space can application startup
```
pohly: According to https://github.com/intel/pmem-CSI/blob/046cadf8461d429203db204e46289140d295534f/docs/images/sequence/pmem-csi-ephemeral-sequence-diagram.png, ephemeral volumes are implemented such that they can only be used once: after unattaching it, the volume is gone. Trying to use it again for another app or restarting the app will fail (volume is not going to be re-created because that happens during `CreateVolume`).

What I described in https://github.com/intel/pmem-CSI/pull/160/files#r254731929 is different. It would allow the PVC to be created once and then be used multiple times.

This is a choice between a rock and a hard place: the current implementation requires more work from the user and probably can't work when pods get restarted by a controller (because the controller doesn't know that it needs to re-created a PVC), while reusing the same PVC has the drawback that it ignores local volume availability during pod scheduling (as documented above).

@avalluri is it correct that with the current implementation pods will always land on a node which had enough space? I'm still a bit fuzzy about that aspect of pod scheduling when WaitForFirstConsumer is active, and I haven't checked whether your implementation actually looks for a node with free space. We need to check what happens when Kubernetes picks one node for the pod and then the volume gets created on another node because the one picked by Kubernetes didn't have enough space.

I'm open for debate which one we should implement. @okartau, any opinion?

avalluri: > According to https://github.com/intel/pmem-CSI/blob/046cadf8461d429203db204e46289140d295534f/docs/images/sequence/pmem-csi-ephemeral-sequence-diagram.png, ephemeral volumes are implemented such that they can only be used once: after unattaching it, the volume is gone. Trying to use it again for another app or restarting the app will fail (volume is not going to be re-created because that happens during `CreateVolume`).
> 
> What I described in https://github.com/intel/pmem-CSI/pull/160/files#r254731929 is different. It would allow the PVC to be created once and then be used multiple times.
> 
Expecting re-attachable volumes in case of "ephemeral" is not a right ask in my opinion, The current implementation supports application restarts by kubelet, as the VolumeAttachement object wont get deleted in this case.

But if we really wants to implement reattachable ephemeral volumes, what we can do best is to clean data on NodeUnpulbish/Unstage instead of deleting the device, in that case the space is not freed even the application has stopped, till the PVC gets deleted.

> @avalluri is it correct that with the current implementation pods will always land on a node which had enough space? I'm still a bit fuzzy about that aspect of pod scheduling when WaitForFirstConsumer is active

The current implementation always  try to create volume(s) on nodes which have free space in the order of requested *AccessibilityRequirement*. We have no information if that node is really picked by kubernetes(delayed binding) or not.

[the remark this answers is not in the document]
avalluri: Hmm, may be Yes. I was thinking to make it use for illustrating the normal persistent volume type, But may not be needed as Kubernetes users will have that knowledge of how normal  dynamic PVs work.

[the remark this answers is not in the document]
avalluri: I thought this section is explaining the volume types in general nothing related to Kubernetes as in your original PR #160. Hence, i provided links to diagrams to illustrate the usage.

I will add "Volumes are created on a certain set of nodes which have free capacity" in case of cache if its not clear for user where those volume are created.

@pohly what additional user documentation you feel useful and which is appropriate section for that.

README.md
```diff
@@ -87,6 +87,50 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+Cache | supported | supported | topology aware scheduling
+
+[Topology aware scheduling](https://github.com/kubernetes/enhancements/issues/490)
+ensures that an application runs on a node where the volume was created. This is obviously needed for persistent volumes, but also for
+delayed persistent volumes because otherwise a restarted application
+might end up on the wrong node. Usage of cache volumes is more flexible,
+but an application still can only run on one of the nodes where the
+a cache volume is available. For CSI-based drivers like PMEM-CSI,
+Kubernetes >= 1.13 is needed. On older Kubernetes releases, pods must
+be scheduled manually onto the right node(s).
```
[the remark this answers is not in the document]
avalluri: > This is where documentation becomes tricky. Do we also fork the documentation (describe pmem-csi for CSI 1.0 on the master branch and pmem-csi for CSI 0.3 on the older branch) or do we describe both on the same branch?
> 
> Forking documentation would be very confusing for users, we should avoid that.

Motivation behind mainlining specific branches:
-  Currently, I am not clear what parts of this new persistence types we can *really* implement using CSI 0.0.3
- frankly, i felt the code changes required to maintain both versions in one branch are more are less same as mainlining two branches, i.e, mainlining two implementation for every csi service interface and in our splitted controller case 4 branches of controller code.
- Avoid the user to use wrong combinations like csi.0.0.3 + Kuberntes v1.13(+)

README.md
```diff
@@ -87,6 +87,50 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+
+Volume | Kubernetes | PMEM-CSI | Limitations
+--- | --- | --- | ---
+Persistent | supported | supported | topology aware scheduling
+Ephemeral | [in design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#proposal) | supported | topology aware scheduling, resource constraints
+Cache | supported | supported | topology aware scheduling
+
+[Topology aware scheduling](https://github.com/kubernetes/enhancements/issues/490)
+ensures that an application runs on a node where the volume was created. This is obviously needed for persistent volumes, but also for
+delayed persistent volumes because otherwise a restarted application
```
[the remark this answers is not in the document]
avalluri: Sorry, its my bad, copy paste error :(, I will removed this.

README.md
```diff
@@ -87,6 +87,50 @@ The PMEM-CSI driver supports running in different modes, which can be controlled
...
+
+* Persistent Volumes
+A volume gets created independently of the application, on some node where there is enough free space. Applications using such a volume are then forced to run on that node and cannot run when the node is
+down. Data is retained until the volume gets deleted.
+* Ephemeral Volumes
+An application is assigned to a node and when it starts to run, a new volume is created for it on that node. When the application stops, the volume is deleted. This [diagram](/docs/images/sequence/pmem-csi-ephemeral-sequence-diagram.png) illustrates how an ephemeral volume gets provisioned in Kubernetes using PMEM-CSI driver.
+* Cache Volumes
+Volumes are created on a certain set of nodes, each with its own local data. Applications are started on those nodes and then get to use the volume on their node. Data persists across application restarts. This
+is useful when the data is only cached information that can be discarded and reconstructed at any time *and* the application can reuse existing local data when restarting. This [diagram](/docs/images/sequence/pmem-csi-cache-sequence-diagram.png) illustrates how a cache volume gets provisioned in Kubernetes using PMEM-CSI driver.
+
```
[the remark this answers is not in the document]
avalluri: my intention was to document only volume types supported by the driver, not all possible types.

README.md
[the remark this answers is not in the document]
pohly: I don't agree that maintaining two branches will be easier or less confusing for users. But luckily we don't need to discuss this further, because yesterday we agreed that supporting only CSI 1.0 and Kubernetes 1.13 is enough. I had brought up the argument that these different persistency models cannot be implemented well on older Kubernetes due to the lack of topology support and Jim added that we already expect a recent kernel, so also expecting a recent Kubernetes isn't asking too much.

README.md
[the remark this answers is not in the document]
pohly: This comment shouldn't have been attached to this particular section. As you said, it is a general introduction to the topic.

What we need is an entirely new "Usage" section that explains what users can put into their `StorageClass` and how the driver will then behave. For a cache, this behavior description must include on how many nodes the cache will be allocated. We probably need some kind of control over that. Imaging a cluster with 100 nodes that have PMEM and an app which will run with a few instances. We don't want to allocate 100 volumes.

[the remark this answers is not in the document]
pohly: But the current diagram isn't the "normal" PVC, is it? It's the fake PVC with delayed allocation on pod startup.

Do you intend to keep this special mode? I thought we agreed to implement the normal PVC mode (allocate space in `CreateVolume`) with low priority, because it isn't really needed.

pohly: "The current implementation supports application restarts by kubelet, as the VolumeAttachement object wont get deleted in this case." - I doubt that app restarts work as well you seem to think it does.

Consider the following case:
- An app with an ephemeral volume starts on one node.
- The node is marked as unschedulable in preparation for taking it down.
- The app is restarted.
- The expectation is that it starts up on another node.

Do you agree that this doesn't work with what you have implemented?

Let me phrase it differently: the definition of "ephemeral volume" is that an app gets a new volume each time it starts and (let me add that for clarity) regardless where it starts.

pohly: It seems that you understood "ephemeral" more in the sense of "one-time use", right?

Re-reading my original description ("An application is assigned to a node and when it starts to run, a new volume is created for it on that node. When the application stops, the volume is deleted.") I see that two words are missing. It should have said "*Each time* an application...". Sorry if that wasn't clear.

pohly: Here's another use-case for ephemeral volumes: running multiple instances of the same app concurrently, each with its own local PMEM storage.

The way to specify that for Kubernetes is with a ReplicaSet. An inline volume would be the natural way to specify the volume, but it's not currently supported, so we have to use a PVC. The end result then looks something like this:
```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: frontend
  labels:
    app: guestbook
    tier: frontend
spec:
  # modify replicas according to your case
  replicas: 3
  selector:
    matchLabels:
      tier: frontend
  template:
    metadata:
      labels:
        tier: frontend
    spec:
      containers:
      - name: php-redis
        image: gcr.io/google_samples/gb-frontend:v3
      volumeMounts:
      - mountPath: "/scratch"
        name: my-csi-volume
      volumes:
      - name: my-csi-volume
        persistentVolumeClaim:
          claimName: php-redis-volume
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: php-redis-volume
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 8Gi
storageClassName: pmem-csi-sc-ephemeral
```

After considering this example, I need to revise my earlier proposal from https://github.com/intel/pmem-CSI/pull/160/files#r254731929: we don't want two pods running on the same node to share data. That would be unexpected. That means that volume creation has to be done in `NodePublishVolume` and deletion in `NodeUnpublishVolume`.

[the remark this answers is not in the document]
avalluri: My intention was that reusing the old diagram by making needed changes according to normal PVC flow. There is no fake/delayed volume creation anymore.

I would request you please review the PR *code* to get how really the cache and ephemeral are implemented.

avalluri: > It seems that you understood "ephemeral" more in the sense of "one-time use", right?
> 
If i am not wrong, the ephemeral volume is for *one-time-use", creation/deletion of volume(PV) is not in driver scope, its COs job to request at app startup and shutdown. We know that currently its not supported in Kubernetes, So what i did is "in case of ephemeral delete the pmem volume on *ControllerUnpublish*, instead of waiting for *DeleteVolume* request.

avalluri: > "The current implementation supports application restarts by kubelet, as the VolumeAttachement object wont get deleted in this case." - I doubt that app restarts work as well you seem to think it does.
> 
> Consider the following case:
> 
> * An app with an ephemeral volume starts on one node.
> * The node is marked as unschedulable in preparation for taking it down.
> * The app is restarted.
> * The expectation is that it starts up on another node.
> 
> Do you agree that this doesn't work with what you have implemented?
In my humble opinion this is not in scope of our driver, CO/user has to request new PV on app startup.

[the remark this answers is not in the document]
avalluri: > We probably need some kind of control over that. Imaging a cluster with 100 nodes that have PMEM and an app which will run with a few instances. We don't want to allocate 100 volumes.

This is not how we operate for cache volumes. Currently the "volume replica/cache count" shall be coming from StorageClass(https://github.com/avalluri/pmem-CSI/blob/046cadf8461d429203db204e46289140d295534f/deploy/kubernetes-1.13/pmem-storageclass-cache.yaml#L10).

@pohly  I would request to review the code.

avalluri: > After considering this example, I need to revise my earlier proposal from https://github.com/intel/pmem-CSI/pull/160/files#r254731929: we don't want two pods running on the same node to share data.

@pohly This case has been already considered by current implementation by rejecting multiple *NodePublishVolume* requests for ephemeral volumes. Please check with the implementation if its needed any changes.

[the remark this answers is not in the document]
pohly: Of course I can reverse-engineer how this new feature works by reading the source. But user-visible documentation needs to be written for it either way, so why not do it now? Then the code review can confirm that the code works as described to the user.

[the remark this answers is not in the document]
pohly: As I said in my other comment, it is easier to review how this is meant to work by reading the user-visible documentation for it.

It makes sense to update the existing diagram. Right now it still shows that `CreateVolume` doesn't actually create a volume. Do we agree after merging this PR, `CreateVolume` will end up creating the pmem device on some node?

pohly: > In my humble opinion this is not in scope of our driver, CO/user has to request new PV on app startup.
> ...
> It rejects multiple NodePublishVolume requests for ephemeral volumes.

This is the crux of the matter: this implementation of a one-time volume isn't very useful for users. The goal was to implement something that is as close as possible to an inline ephemeral volume, because "give me local scratch space for my app" is used by apps today.

What you are saying is that users need to rewrite their app deployment logic and can't use controllers like `ReplicaSet`. Why not change the implementation so that this all works? Is there some advantage of this implementation over what I was suggesting?

avalluri: > What you are saying is that users need to rewrite their app deployment logic and can't use controllers like `ReplicaSet`. 

"cache" volume are best suits for `ReplicaSet/Deployment` . Pre-create the cached PVC with no of replicas and ask to use that PVC by ReplicaSet.

and "ephemeral" volumes are supposed to use by `StatefulSet`, because StatefulSets  supports `volumeClaimTemplate` where each volume tied to a PVC, the volumes are not shared.

pohly:
>> What you are saying is that users need to rewrite their app deployment logic and can't use controllers like `ReplicaSet`. 
>
> "cache" volume are best suits for `ReplicaSet/Deployment` . Pre-create
> the cached PVC with no of replicas and ask to use that PVC by
> ReplicaSet.

Whether such a ReplicaSet should use "cache" or "ephemeral" volumes
depends on the application. The app might expect empty volumes on
startup and might get confused when it finds old data when using "cache"
volumes. It's also a waste of disk space to keep the data around after
the app has terminated if the data is not going to be used again.

> and "ephemeral" volumes are supposed to use by `StatefulSet`, because
> StatefulSets supports `volumeClaimTemplate` where each volume tied to
> a PVC, the volumes are not shared.

StatefulSet is meant for "stateful" applications, i.e. those which need
to retain data across restarts. I can imagine that also those apps may
need local, ephemeral storage, but the way
https://akomljen.com/kubernetes-persistent-volumes-with-deployment-and-statefulset/
describes it, volumeClaimTemplate was meant to provide each pod with its
own persistent volume.

avalluri:
@pohly, updated the document with Usage details, can you check if it helps.

README.md
```diff
@@ -125,16 +125,48 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
-![sequence diagram](/docs/images/sequence/pmem-csi-sequence-diagram.png)
 <!-- FILL TEMPLATE:
 * Target users and use cases
 * Design decisions & tradeoffs that were made
 * What is in scope and outside of scope
 -->
 
+### Volume Persistency
+
+Normally, volumes are provided by a storage backend that is independent of a particular node. When a node goes offline, the volume can be mounted elsewhere. But PMEM volumes are *local* to node and thus can only be used on the node where they were created. This limits the applications using PMEM volume cannot freely move between nodes. This limitation needs to be considered when designing and deploying applications that are to use *local storage*.
```
okartau: can we express the point more precisely? Just "Normally" does not describe a definitive state. What about saying "In a typical CSI deployment" (or something like that)

okartau: > This limits the applications using PMEM volume cannot freely move between nodes.

this sentence is not entirely syntactically clear... what about "This means the applications using PMEM volume cannot freely move between nodes."

README.md
```diff
@@ -125,16 +125,48 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+Below are the volume persistency types supported by PMEM-CSI to serve different application usecases:
```
okartau: "use cases" better as two words

README.md
```diff
@@ -125,16 +125,48 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+* Persistent Volumes</br>
+A volume gets created independently of the application, on some node where there is enough free space. Applications using such a volume are then forced to run on that node and cannot run when the node is down. Data is retained until the volume gets deleted.
+
+* Ephemeral Volumes</br>
+An application is assigned to a node and when it starts to run, a new volume is created for it on that node. When the application stops, the volume is deleted. The volume cannot be shared with other applications. Data on this volume is retained only until the application runs.
+
+* Cache Volumes</br>
+Volumes are pred-created on a certain set of nodes, each with its own local data. Applications are started on those nodes and then get to use the volume on their node. Data persists across application restarts. This is useful when the data is only cached information that can be discarded and reconstructed at any time *and* the application can reuse existing local data when restarting.
```
okartau: > Volumes are pred-created

pre-created

README.md
```diff
@@ -125,16 +125,48 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+<sup>1 </sup>[Topology aware scheduling](https://github.com/kubernetes/enhancements/issues/490)
+ensures that an application runs on a node where the volume was created. For CSI-based drivers like PMEM-CSI, Kubernetes >= 1.13 is needed. On older Kubernetes releases, pods must be scheduled manually onto the right node(s).
+
+<sup>2 </sup>The upstream design for ephemeral volumes currently does not take [resource constraints](https://github.com/kubernetes/enhancements/pull/716#discussion_r250536632) into account. If an application gets scheduled onto a node and then creating the ephemeral volume on that node fails, the application on the node cannot start until resources become available.
+
+#### Usage on Kubernetes
+
+Kuberntes cluster administrators can expose above mentioned [volume persistency types](#volume-persistency) to applications using [`StorageClass Parameters`](https://kubernetes.io/docs/concepts/storage/storage-classes/#parameters). An optional <code>persistencyType</code> parameter differentiates how the provisioned volume can be used, if no <code>persistencyType</code> specified in StorageClass then it is treated as normal persistant volume.
```
okartau: > Kuberntes cluster

Kubernetes cluster

avalluri: Done

README.md
avalluri: Done

README.md
avalluri: Done

README.md
avalluri: s/limits/means Done

avalluri: Agreed, Done

avalluri:
@okartau and @pohly  I updated the PR with:
 -  reverting ephemeral support
 - merged default {identity,node}servers code with actual implementation
 - updated the documentation with persistent models and usage.

deploy/kubernetes-1.13/pmem-app.yaml
```diff
@@ -0,0 +1 @@
+../kubernetes-1.11/pmem-app.yaml
```
pohly: Can you squash this commit with the corresponding code change?

My goal is to look at the PR commit by commit (it's too large otherwise), and without this change here the code change is incomplete.

You also need change the documentation in that same commit and explain that Kubernetes >= 1.13 is needed.

Remove the `deploy/kubernetes-1.11` and `deploy/kubernetes-1.12` directories.

deploy/kubernetes-1.13/pmem-csi.yaml
```diff
@@ -0,0 +1,256 @@
...
+kind: ServiceAccount
+metadata:
+  name: csi-service-account
+  namespace: default
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: csi-cluster-role
+rules:
```
pohly: I suspect that some of these RBAC entries are out-dated, for example the 

Did you copy-and-paste from previous deployments or from the canonical upstream definitions of the sidecar RBACs? Those are:
- https://github.com/kubernetes-csi/external-attacher/blob/v1.0.1/deploy/kubernetes/rbac.yaml
- https://github.com/kubernetes-csi/external-provisioner/blob/v1.0.1/deploy/kubernetes/rbac.yaml
- https://github.com/kubernetes-csi/cluster-driver-registrar/blob/v1.0.1/deploy/kubernetes/rbac.yaml

Please double-check the RBAC rules and explain in a comment from what and how they were assembled.

test/test-config.sh
```diff
@@ -48,6 +48,10 @@ TEST_PMEM_MEM_SIZE=32768 # 32GB
 TEST_PMEM_SHARE=on
 TEST_PMEM_LABEL_SIZE=2097152
 
+# Kubernetes feature gates to enable/disable
+# featurename=true,feature=false
+TEST_FEATURE_GATES="CSINodeInfo=true,CSIDriverRegistry=true"
```
pohly: The installation of the corresponding CRDs also belongs into the cluster setup. Currently the PR has that as part of the driver installation (like in some older CSI documentation), but that's conceptually the wrong place and was just a workaround for cluster setup tools not knowing about those CRDs yet.

test/start-stop.make
```diff
@@ -40,6 +40,11 @@ start: _work/clear-kvm.img _work/kube-clear-kvm _work/start-clear-kvm _work/ssh-
 		touch _work/clear-kvm.secretsdone; \
 	fi
 	_work/ssh-clear-kvm kubectl version --short | grep 'Server Version' | sed -e 's/.*: v\([0-9]*\)\.\([0-9]*\)\..*/\1.\2/' >_work/clear-kvm-kubernetes.version
+	if [ -e deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/csinodeinfo-crd.yaml ]; then \
+	   	if ! _work/ssh-clear-kvm kubectl get crd/csinodeinfos.csi.storage.k8s.io  >/dev/null 2>&1; then \
+			_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/csinodeinfo-crd.yaml; \
+		fi \
```
pohly: This is more suitable for `test/setup-clear-kvm.sh`. Also squash it into the previous commit.

test/e2e/storage/csi_volumes.go
```diff
@@ -95,11 +95,9 @@ var _ = Describe("PMEM Volumes", func() {
...
 
 					Config: testsuites.TestConfig{
-						Framework: f,
-						Prefix:    "pmem",
-						// Ensure that all pods land on the same node. Works around
-						// https://github.com/intel/pmem-csi/issues/132.
-						ClientNodeName: "host-1",
+						Framework:       f,
+						Prefix:          "pmem",
+						TopologyEnabled: true,
```
pohly: Which commit is getting fixed here? Better squash it into that one.

Overall I'm getting the impression that the following commits are all related. Squash them?
- Implemented cache persistency model
- fix e2e tests
- ReadME.md: updated documentation with different persistency models.

What I don't find is the commit that removes (or updates?) persistent volume support. Are you going to do that later?

deploy/kubernetes-1.11/pmem-unified-csi.yaml
```diff
@@ -2,52 +2,6 @@
 # That CSI drivers gets exposed with a NodePort service and the fixed port
 # 30123 for use with csi-test sanity testing.
```
pohly: Commit message typo: "unifined".

deploy/kubernetes-1.13/pmem-csi.yaml
avalluri: > I suspect that some of these RBAC entries are out-dated, for example the CRD entries.

csi-cluster-driver-registar v1.0 is using CRDs to create CSIDriverCRD,  here is the code trace:

https://github.com/kubernetes-csi/cluster-driver-registrar/blob/ff1f9663191e085ae6201052e5ca9c85133664e7/cmd/cluster-driver-registrar/k8s_register.go#L64

pohly: It shouldn't be doing that, and in master it was removed: https://github.com/kubernetes-csi/cluster-driver-registrar/commit/02b2bfe6bb0eea82bd1a2ab41544c1540c2b1637

I didn't know that v1.0.1 went out in that broken state (installing CRDs and not having the matching RBAC).

This is something that needs to go into our copy of the RBAC rules as a comment. We have to ensure that we remove it when updating to cluster-driver-registrar v1.1.0 or more generally, that we always keep our RBAC rules in sync with the sidecar releases that we use. That's why I was asking for a description that some other developer can use to create our copy from scratch.

deploy/kubernetes-1.13/pmem-app.yaml
avalluri: Done

deploy/kubernetes-1.13/pmem-csi.yaml
avalluri: Added the comment in deployment files related to CRD rule

test/test-config.sh
avalluri: agree. Done

test/start-stop.make
avalluri: Added cds installation to this script.

test/e2e/storage/csi_volumes.go
avalluri: Squashed the commits. Done

deploy/kubernetes-1.11/pmem-unified-csi.yaml
avalluri: Done

test/e2e/storage/csi_volumes.go
avalluri: > What I don't find is the commit that removes (or updates?) persistent volume support. Are you going to do that later?

Persistent Volume model is fixed as part of "Cache" change. And updated the diagram.

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+* `persistencyModel: cache`</br>
+Volumes of this type shall be used in combination with `volumeBindingMode: Immediate`. In this case PMEM-CSI creates set of PMEM volumes each volume on different node. The number of PMEM volumes to create can be specified by `cacheSize` StorageClass parameter. Applications which claim a `cache` volume can use `ReadWriteMany` in its `accessModes` list. Check with provided [cache StorageClass](deploy/kubernetes-1.13/pmem-storageclass-cache.yaml) example. This [diagram](/docs/images/sequence/pmem-csi-cache-sequence-diagram.png) illustrates how a cache volume gets provisioned in Kubernetes using PMEM-CSI driver.
```
pohly: See other comment about `<br>`.

Commas and "a": "In this case, PMEM-CSI creates a set of PMEM volumes, each volume on a different node."

"can" or "must" applications use `ReadWriteMany`?

Actually, is it the application or the storage class which chooses the access mode? Or both? The example app doesn't specify it. It would be good to have example apps+pvc for both models in `deploy` and also adapt the "make start" such that it deploys both storage classes and shows the commands for using both.


Can you add a note about what happens when multiple pods run on the same node ("they share the same volume") and suggest ways to spread out the pods if that isn't desired (not sure, but daemon set might work, or a replicaset with inter-pod anti-affinity).

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+* if no `persistencyModel` parameter specified in `StorageClass` then it is treated as normal Kubernetes persistent volume. In this case PMEM-CSI creates PMEM volume on a node and the application that claims to use this volume is supposed to schedule on this node by Kubernetes. Choosing of node is depend on StorageClass `volumeBindingMode`. In case of `volumeBindingMode: Immediate` PMEM-CSI chooses a node randomly, and in case of `volumeBindingMode: WaitForFirstConsumer` Kubernetes first chooses a node for scheduling the application, and PMEM-CSI creates the volume on that node. Applications which claim a normal persistent volume has to use `ReadOnlyOnce` access mode in its `accessModes` list. This [diagram](/docs/images/sequence/pmem-csi-persistent-sequence-diagram.png) illustrates how a normal persistent volume gets provisioned in Kubernetes using PMEM-CSI driver.
```
pohly: s/on this node/be scheduled onto this node/

avalluri:
> That doesn't look right. I did `make push-images` before rebuilding the cluster, so I should be running the latest driver. However, this is hard to verify because the driver does not identify itself in the log output and has no unique version number. I've filed #183 to track this.
> 
> From the logs it looks like the driver treated this volume as a cache volume?
> 
> ```
> $ kubectl log pmem-csi-controller-0 pmem-driver
> ...
> I0301 10:35:10.049651       1 glog.go:58] GRPC call: /csi.v1.Controller/CreateVolume
> I0301 10:35:10.049673       1 glog.go:58] GRPC request: name:"pvc-b0354387-3c0d-11e9-895f-deadbeef0100" capacity_range:<required_bytes:8589934592 > volume_capabilities:<mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > > accessibility_requirements:<requisite:<segments:<key:"kubernetes.io/hostname" value:"host-0" > > requisite:<segments:<key:"kubernetes.io/hostname" value:"host-1" > > requisite:<segments:<key:"kubernetes.io/hostname" value:"host-2" > > requisite:<segments:<key:"kubernetes.io/hostname" value:"host-3" > > preferred:<segments:<key:"kubernetes.io/hostname" value:"host-0" > > preferred:<segments:<key:"kubernetes.io/hostname" value:"host-1" > > preferred:<segments:<key:"kubernetes.io/hostname" value:"host-2" > > preferred:<segments:<key:"kubernetes.io/hostname" value:"host-3" > > > 
> I0301 10:35:10.049861       1 glog.go:79] CreateVolume: Name: pvc-b0354387-3c0d-11e9-895f-deadbeef0100 req.Required: 8589934592 req.Limit: 0
> I0301 10:35:10.049965       1 glog.go:79] Connecting to 192.168.8.8:10001
> I0301 10:35:10.114505       1 glog.go:79] node host-3 has requested capacity
> I0301 10:35:10.114536       1 glog.go:79] Connecting to 192.168.8.4:10001
> I0301 10:35:10.181366       1 glog.go:79] node host-1 has requested capacity
> I0301 10:35:10.181404       1 glog.go:79] Connecting to 192.168.8.6:10001
> I0301 10:35:10.249569       1 glog.go:79] node host-2 has requested capacity
> I0301 10:35:10.249621       1 glog.go:79] Found nodes: [host-3 host-1 host-2]
> I0301 10:35:10.249667       1 glog.go:79] CreateVolume: Record new volume as {b03c6245-3c0d-11e9-ba0d-26c6b0ca2305 pvc-b0354387-3c0d-11e9-895f-deadbeef0100 8589934592 1  true fsdax}
> ```

I doubt if your cluster is built on https://github.com/intel/pmem-CSI/pull/177/commits/e81ec79a014f79487060584559a6cfec6233aaab . The logs you shared
 > node host-3 has requested capacity

 are not present in this commit. Can you please verify this.

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
-![sequence diagram](/docs/images/sequence/pmem-csi-sequence-diagram.png)
 <!-- FILL TEMPLATE:
 * Target users and use cases
 * Design decisions & tradeoffs that were made
 * What is in scope and outside of scope
 -->
 
+### Volume Persistency
+
+In a typical CSI deployment, volumes are provided by a storage backend that is independent of a particular node. When a node goes offline, the volume can be mounted elsewhere. But PMEM volumes are *local* to node and thus can only be used on the node where they were created. This means the applications using PMEM volume cannot freely move between nodes. This limitation needs to be considered when designing and deploying applications that are to use *local storage*.
```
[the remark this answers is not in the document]
avalluri: I changed intentionally as other parts of the README are not wrapped.

README.md
avalluri: > "can" or "must" applications use `ReadWriteMany`?
> 
> Actually, is it the application or the storage class which chooses the access mode? Or both?

Application(PVC) requests the volume access modes. Application can request `ReadWriteMany` volume in case of cache volume, but its not must, if app wants to use only on a single node still it can request `ReadOnlyOnce`

pohly:
Indeed, I was one commit before the persistency changes. We need this embedded version number to catch such mistakes faster :-/

I'll continue testing.

avalluri:
> Commit message [e81ec79](https://github.com/intel/pmem-CSI/commit/e81ec79a014f79487060584559a6cfec6233aaab):
> 
> > the newly created volume to that node by using 'Volume.Topology'. OC has to make
> 
> s/OC/The container orchestrator/
Done

> You've spelled "Cache" and further down also "Persistent" with capital letter when talking about the model. I think "cache volume" and "persistent volume" are normal English words and not specific names, so normal capitalization seems more appropriate.

s/Cache/cache, s/Persistent/persistent. Done.
> 
> `deploy/kubernetes-1.13/pmem-pvc.yaml` says `storageClassName: pmem-csi-sc # defined in pmem-csi.yaml`. That should say `deploy/kubernetes-1.13/pmem-storageclass.yaml`.

Fixed the comment.

README.md
```diff
@@ -158,13 +158,11 @@ The driver does not create persistent memory Regions, but expects Regions to exi
...
 
-The driver deployment in Kubernetes cluster has been verified on:
+PMEM-CSI driver implements CSI specification version 1.0.0, which only supported by Kubernetes versions >= v1.13. The driver deployment in Kubernetes cluster has been verified on:
 
-| Branch            | Kubernetes branch/version      |
-|-------------------|--------------------------------|
-| devel             | Kubernetes 1.11 branch v1.11.3 |
-| devel             | Kubernetes 1.12                |
-| devel             | Kubernetes 1.13                |
+| Branch            | Kubernetes branch/version      | Required Alfa feature-gates |
```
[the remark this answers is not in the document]
avalluri: Done

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
 * Target users and use cases
 * Design decisions & tradeoffs that were made
 * What is in scope and outside of scope
 -->
 
+### Volume Persistency
+
+In a typical CSI deployment, volumes are provided by a storage backend that is independent of a particular node. When a node goes offline, the volume can be mounted elsewhere. But PMEM volumes are *local* to node and thus can only be used on the node where they were created. This means the applications using PMEM volume cannot freely move between nodes. This limitation needs to be considered when designing and deploying applications that are to use *local storage*.
+
+Below are the volume persistency models considered for implementing by PMEM-CSI to serve different application use cases:
```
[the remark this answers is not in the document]
avalluri: > s/for implementing by PMEM-CSI/for implementation in PMEM-CSI/

Done

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+
+* Persistent Volumes</br>
```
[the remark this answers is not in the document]
avalluri: Thanks for the tip. I fixed all </br>s

README.md
```diff
@@ -125,16 +125,47 @@ The [`test/setup-ca-kubernetes.sh`](test/setup-ca-kubernetes.sh) script shows ho
...
+A volume gets created independently of the application, on some node where there is enough free space. Applications using such a volume are then forced to run on that node and cannot run when the node is down. Data is retained until the volume gets deleted.
+
+* Ephemeral Volumes</br>
+An application is assigned to a node and when it starts to run, a new volume is created for it on that node. When the application stops, the volume is deleted. The volume cannot be shared with other applications. Data on this volume is retained only until the application runs.
```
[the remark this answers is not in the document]
avalluri: Done

README.md
avalluri: > Commas and "a": "In this case, PMEM-CSI creates a set of PMEM volumes, each volume on a different node."

Done

README.md
avalluri: Done

README.md
avalluri: > Can you add a note about what happens when multiple pods run on the same node ("they share the same volume") and suggest ways to spread out the pods if that isn't desired (not sure, but daemon set might work, or a replicaset with inter-pod anti-affinity).

Added a note, and added  replicaset example that uses the cache volume with nodeAntiAffinity.

avalluri: > It would be good to have example apps+pvc for both models in `deploy` and also adapt the "make start" such that it deploys both storage classes and shows the commands for using both.

Done.

README.md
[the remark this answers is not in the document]
pohly: Okay, then let's do a separate PR once all doc changes are merged (including the one from @okartau).

avalluri:
Offline, @okartau agreed to merge this change.

Issue #182, opened 2019-03-01
bring up Kubernetes deployment without restarts


pohly:
When I tested the CSI 1.0 support in https://github.com/intel/pmem-CSI/pull/177 (https://github.com/intel/pmem-CSI/pull/177/commits/eab77d77a74a91d3b94552c2c9bfe980e82c1fa7), the pods came up eventually, but only after some intermittent errors:
```
$ kubectl get pods
NAME                    READY   STATUS    RESTARTS   AGE
my-csi-app              1/1     Running   0          6m30s
pmem-csi-bqdwp          2/2     Running   11         10m
pmem-csi-controller-0   4/4     Running   15         10m
pmem-csi-h7dfn          2/2     Running   11         10m
pmem-csi-w4p2f          2/2     Running   9          10m
```

I've not caught it in the logs, but I think the node pmem-driver was restarting because the registry wasn't up yet.

This is confusing. In the sidecar containers, we chose the approach of waiting forever for a peer to show up, with regular logging while in that wait loop. The pmem driver should do the same. See https://github.com/kubernetes-csi/csi-lib-utils/pull/11.

okartau:
Do the startup phase errors/retries happen differently on 1st vs next times, i.e. what's the impact caused by 1st fetch of docker images.
Also, is csi-1.0 branch behavior different from latest devel-branch, i.e. is there more related issues created by 1.0 work. Because I don't recall seen that high restart counts in devel-branch recently.
I will run some trials as well.

avalluri:
If i am not wrong, this exists in driver for long time, not specific to 1.0 changes. I can take this task.

avalluri:
with change #184, driver deployment looks like:

```
$ kubectl get po
NAME                    READY   STATUS    RESTARTS   AGE
pmem-csi-4d6xj          2/2     Running   0          107s
pmem-csi-controller-0   4/4     Running   0          107s
pmem-csi-g5v9h          2/2     Running   0          107s
pmem-csi-tszbz          2/2     Running   0          107s
```

PR #184, opened 2019-03-04
Keep retry registering node controller with RegistryServer

avalluri:
Instead of exiting with error, driver has to wait and retry 'node controller'
registration till the registry server up and registration get succeed.

FIXES: #182

### DIFF
--- a/pkg/pmem-csi-driver/pmem-csi-driver.go
+++ b/pkg/pmem-csi-driver/pmem-csi-driver.go
@@ -17,6 +17,8 @@ import (
 	pmemgrpc "github.com/intel/pmem-csi/pkg/pmem-grpc"
 	registry "github.com/intel/pmem-csi/pkg/pmem-registry"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/klog/glog"
 )
 
@@ -166,19 +168,22 @@ func (pmemd *pmemDriver) Run() error {
 
 		if pmemd.cfg.Mode == Node {
 			if pmemd.cfg.Endpoint != pmemd.cfg.ControllerEndpoint {
-				if err := s.Start(pmemd.cfg.Endpoint, nil, ids, ns); err != nil {
+				if err := s.Start(pmemd.cfg.ControllerEndpoint, pmemd.serverTLSConfig, cs); err != nil {
 					return err
 				}
-				if err := s.Start(pmemd.cfg.ControllerEndpoint, pmemd.serverTLSConfig, cs); err != nil {
+				if err := pmemd.registerNodeController(); err != nil {
+					return err
+				}
+				if err := s.Start(pmemd.cfg.Endpoint, nil, ids, ns); err != nil {
 					return err
 				}
 			} else {
 				if err := s.Start(pmemd.cfg.Endpoint, nil, ids, cs, ns); err != nil {
 					return err
 				}
-			}
-			if err := pmemd.registerNodeController(); err != nil {
-				return err
+				if err := pmemd.registerNodeController(); err != nil {
+					return err
+				}
 			}
 		} else /* if pmemd.cfg.Mode == Unified */ {
 			if err := s.Start(pmemd.cfg.Endpoint, pmemd.serverTLSConfig, ids, cs, ns); err != nil {
@@ -194,31 +199,40 @@ func (pmemd *pmemDriver) Run() error {
 }
 
 func (pmemd *pmemDriver) registerNodeController() error {
-	fmt.Printf("Connecting to Registry at : %s\n", pmemd.cfg.RegistryEndpoint)
 	var err error
 	var conn *grpc.ClientConn
 
-	for {
-		conn, err = pmemgrpc.Connect(pmemd.cfg.RegistryEndpoint, pmemd.clientTLSConfig, connectionTimeout)
-		if err == nil {
-			glog.Info("Connected to RegistryServer!!!")
-			break
-		}
-		/* TODO: Retry loop */
-		glog.Infof("Failed to connect RegistryServer : %s, retrying after 10 seconds...", err.Error())
-		time.Sleep(10 * time.Second)
-	}
-	client := registry.NewRegistryClient(conn)
 	req := registry.RegisterControllerRequest{
 		NodeId:   pmemd.cfg.NodeID,
 		Endpoint: pmemd.cfg.ControllerEndpoint,
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
-	defer cancel()
 
-	if _, err = client.RegisterController(ctx, &req); err != nil {
-		return fmt.Errorf("Fail to register with server at '%s' : %s", pmemd.cfg.RegistryEndpoint, err.Error())
+	for {
+		glog.Infof("Connecting to registry server at: %s\n", pmemd.cfg.RegistryEndpoint)
+		conn, err = pmemgrpc.Connect(pmemd.cfg.RegistryEndpoint, pmemd.clientTLSConfig, connectionTimeout)
+		if err == nil {
+			break
+		}
+		glog.Infof("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
+		time.Sleep(retryTimeout)
 	}
+
+	client := registry.NewRegistryClient(conn)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	for {
+		glog.Info("Registering controller...")
+		if _, err = client.RegisterController(ctx, &req); err != nil {
+			if s, ok := status.FromError(err); ok && s.Code() == codes.InvalidArgument {
+				return fmt.Errorf("Registration failed: %s", s.Message())
+			}
+			glog.Infof("Failed to register: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
+			time.Sleep(retryTimeout)
+		} else {
+			break
+		}
+	}
+	glog.Info("Registration success!!!")
 
 	return nil
 }
--- a/pkg/pmem-csi-driver/registryserver.go
+++ b/pkg/pmem-csi-driver/registryserver.go
@@ -9,6 +9,8 @@ import (
 	registry "github.com/intel/pmem-csi/pkg/pmem-registry"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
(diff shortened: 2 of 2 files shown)


okartau:
good improvement!
retrytimeout was just not used and we set again to 10 sec it seems.. good fix as well.

okartau:
Confirmed, works fine in my "make test" cluster

avalluri:
@pohly  is it good enough to merge?

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -193,34 +192,39 @@ func (pmemd *pmemDriver) Run() error {
...
+		glog.Infof("Connecting to registry server at: %s\n", pmemd.cfg.RegistryEndpoint)
 		conn, err = pmemgrpc.Connect(pmemd.cfg.RegistryEndpoint, pmemd.clientTLSConfig, connectionTimeout)
 		if err == nil {
-			glog.Info("Connected to RegistryServer!!!")
 			break
 		}
 		/* TODO: Retry loop */
-		glog.Infof("Failed to connect RegistryServer : %s, retrying after 10 seconds...", err.Error())
-		time.Sleep(10 * time.Second)
+		glog.Infof("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
```
pohly: What is this TODO about now? Still valid or has this been solved?

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -177,9 +177,8 @@ func (pmemd *pmemDriver) Run() error {
 					return err
 				}
 			}
-			if err := pmemd.registerNodeController(); err != nil {
-				return err
-			}
+			go pmemd.registerNodeController()
```
pohly: This allows the driver to go into a "ready" state (= accepting CSI calls) before it has successfully registered. Is that what we want? We don't have readiness probes at the moment, but if we had something based on the driver responding to CSI requests, then this would incorrectly report "ready" when in reality the driver is still starting up.

It's just a gut feeling, but I think I prefer calling `pmemd.registerNodeController()` in blocking mode.

pohly: Also, what happens if registration fails? Does the driver now keep running despite the failure?

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: yes, we can remove. I will do that

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: > Also, what happens if registration fails? Does the driver now keep running despite the failure?

Yes, i would keep retry. In reality, the only valid path is missing arguments, both are validated at client side.

But, i agree, it should fail on real error case. Now i fixed this.

avalluri: > This allows the driver to go into a "ready" state (= accepting CSI calls) before it has successfully registered. Is that what we want? We don't have readiness probes at the moment, but if we had something based on the driver responding to CSI requests, then this would incorrectly report "ready" when in reality the driver is still starting up.
> 
> It's just a gut feeling, but I think I prefer calling `pmemd.registerNodeController()` in blocking mode.

Moved this to blocking call.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -194,31 +196,40 @@ func (pmemd *pmemDriver) Run() error {
...
-		return fmt.Errorf("Fail to register with server at '%s' : %s", pmemd.cfg.RegistryEndpoint, err.Error())
+	for {
+		client := registry.NewRegistryClient(conn)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+		defer cancel()
+
+		glog.Info("Registering controller...")
+		if _, err = client.RegisterController(ctx, &req); err != nil {
+			if s, ok := status.FromError(err); ok && s.Code() == codes.InvalidArgument {
+				return fmt.Errorf("Registration filed: %s", s.Message())
```
pohly: s/filed/failed/

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -194,31 +196,40 @@ func (pmemd *pmemDriver) Run() error {
...
-	}
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
-	defer cancel()
-
-	if _, err = client.RegisterController(ctx, &req); err != nil {
-		return fmt.Errorf("Fail to register with server at '%s' : %s", pmemd.cfg.RegistryEndpoint, err.Error())
+	for {
+		client := registry.NewRegistryClient(conn)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+		defer cancel()
```
pohly: `cancel` is only called when when the function returns. That means each loop iteration ends up allocating more resources. You need to call `cancel()` after `client.RegisterController`, regardless whether the call succeeded or failed.

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: I had a closer look at how `pmemd.registerNodeController` is called. Making it a blocking call is not enough, it also needs to be moved above the `s.Start` calls, otherwise the gRPC server is already accepting incoming requests while the registration is still pending.

avalluri: @pohly I feel, registering is not a hard requirement from driver at node, and it might not effect the 'readiness' state of node driver. It still can server valid node operations coming from kubelet and  driver registrar. I couldn't find a failure case here.

Can you help me understand if i miss something here.

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: Yes, using `defer` inside loop is bad idea. I think we can start loop after calling `defer cancel() `as we can reuse `client` and `ctx`.

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: Done

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: The failure is that the system as a whole isn't operational when nodes don't register themselves. Consider how an admin currently checks whether the pmem-csi driver deployment is working. Most likely they'll just check that all pods (including the sidecars) are running and "kubectl describe" doesn't show failures.

But if the node currently fails to register, all pods are running and yet provisioning will fail or (worse) silently ignore nodes it doesn't know about.

An admin would have to grep the logs to determine whether everything is alright. That's so unusual, I doubt people will do it, even if we put it in the documentation.

avalluri: Sorry if my earlier comment sounded like "node registration failure can be ignored". That was not my intention. It was in the context of your suggestion to move `pmemd.registerNodeController()`  before starting gRPC {identity, controller, node}servers.  If we start controller after registration, Any requests to this node controller made by Master from its registration handler might fail. So, we should start ControllerServer by the time we call `registry.RegisterController()`

Starting identity and node servers prior to registration, which allows to register the driver at node with kubelet, prior to registration. Does this leads to issues? I couldn't see any.

And after reading your comments, i felt like we should not keep retry to connect forever, fail after may be n-tries?

pohly: > If we start controller after registration, Any requests to this node controller made by Master from its registration handler might fail.

No, they don't. The listening socket is open, so connections from master get accepted and queued. The response will be generated as soon as the driver continues from "registering myself" to "serving requests".

> Starting identity and node servers prior to registration, which allows to register the driver at node with kubelet, prior to registration. Does this leads to issues? I couldn't see any.

My idea was to not create the Unix domain `csi.sock` and thus indicating to sidecars that the driver is not ready yet. However, that alone doesn't help an admin, because the sidecars will be running. What we need is a "readiness probe" sidecar. I'll ask in the CSI standup meeting why they implemented a "liveness probe" sidecar, but not a "readiness probe". IMHO "readiness" is more important.

> And after reading your comments, i felt like we should not keep retry to connect forever, fail after may be n-tries?

I still think trying to connect forever is the right approach, in particular if an admin can tell that the driver is still initializing (see "readiness" above).

avalluri: > No, they don't. The listening socket is open,

Listening Endpoint socket is only open in `Start()` call, without calling that no listening socket and if Master controller tries to connect to "Endpoint" we asked to register from its registration handler might fail?

pohly: Forget what I said then. I was thinking of "Start" as "start gRPC server" because that is how it is done elsewhere (if I remember correctly), but here it is indeed different.

We probably need to revisit this entire "bring up pmem-csi" topic sometime on the future when considering what it means for pmem-csi to be "ready" and what different failure scenarios could be.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -194,31 +196,40 @@ func (pmemd *pmemDriver) Run() error {
...
-	req := registry.RegisterControllerRequest{
-		NodeId:   pmemd.cfg.NodeID,
-		Endpoint: pmemd.cfg.ControllerEndpoint,
-	}
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
 	defer cancel()
-
-	if _, err = client.RegisterController(ctx, &req); err != nil {
-		return fmt.Errorf("Fail to register with server at '%s' : %s", pmemd.cfg.RegistryEndpoint, err.Error())
+	for {
```
pohly: Still problematic. Consider what happens after 10 minutes: all gRPC calls will immediately time out and the for loop will keep running forever.

avalluri: Yes, i agree. What about using no timeout context - `WithCancel()`.

avalluri: Fixed using` WithCancel() `context

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: Ok, i did what the best we can do with current code:
- Start ControllerServer
- Initiate Register, exit if fail.
- Start identity and node servers.

PR #186, opened 2019-03-05
Implement release versioning

### LINKED ISSUE #183: add unique version number to each build
When looking at a running cluster it is impossible to tell which revision of the pmem-csi driver is running. Let's build a unique version string into the binary with a `-version` flag (as in https://github.com/intel/oim/commit/e38285d4729b80c56c8107aa4e9f730658ade13b) and in addition, always print that when the driver runs (currently missing in OIM).

avalluri:
Setting driver version from $(git describe). Currently this version is not
generated from "tags", which should be fixed once we have release tagging.

All driver commands have "-version" flag which shows the current version and
exits.

FIXES #183

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@
 IMPORT_PATH=github.com/intel/pmem-csi
 SHELL=bash
 
+ifeq ($(VERSION), )
+VERSION=$(shell git describe --long --dirty --tags --match='v*')
+endif
+
 REGISTRY_NAME=localhost:5000
 IMAGE_VERSION_pmem-csi-driver=canary
 IMAGE_VERSION_pmem-ns-init=canary
@@ -35,6 +39,8 @@ ifneq ($(no_proxy),)
 	BUILD_ARGS:=${BUILD_ARGS} --build-arg no_proxy=${no_proxy}
 endif
 
+BUILD_ARGS:=${BUILD_ARGS} --build-arg VERSION=${VERSION}
+
 # Build main set of components.
 all: pmem-csi-driver pmem-ns-init pmem-vgm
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -16,6 +16,11 @@
 IMPORT_PATH=github.com/intel/pmem-csi
 SHELL=bash
 
+#FIXME: replace --all with --tags once we started release tagging
+ifeq ($(VERSION), )
+VERSION=$(shell git describe --long --dirty --all)
```
okartau: oim example used --tags --match=v* to tie versioning to tag lines, we don't want/plan that?
It would require to give first v-form tag in our repo to become functional

okartau:
overall looks good!
Definitely good improvement to tie node-info version to same value instead of hard-coded value.
I still remain skeptical about real use cases for command-line version flag. 

What about that "vendor name should be in node-info" requirement, we will handle separately from this PR?

Makefile
avalluri: > oim example used --tags --match=v* to tie versioning to tag lines, we don't want/plan that?
> It would require to give first v-form tag in our repo to become functional

I  mentioned this in #FIMXE comment. Once we are ready with release tagging we can move to  `-tags --match=v*` in place of `--all`.

avalluri:
> What about that "vendor name should be in node-info" requirement, we will handle separately from this PR?

I will submit new PR for that

Makefile
okartau: ah yes, I was too quick and mislooked FIXME. What about tagging now with some baseline like v0.0.0 so that tag option will always work (I think OIM did similar?). Then we don't need to maintain that FIXME place in code

okartau:
> `$ docker run 192.168.8.1:5000/pmem-csi-driver:canary -version`
> `heads/devel-0-g9cd0dc90`

yes, good point, helps with some of my skepticism. 
BTW is there way to tag docker image so that it indicates CID in its tag, but remains still usable via static name (like we use "canary" now) by k8s manifest? That would help to visualize what is in image.

avalluri:
> > What about that "vendor name should be in node-info" requirement, we will handle separately from this PR?
> 
> I will submit new PR for that

PR #189  fixes the driver name issue.

pohly:

>> `$ docker run 192.168.8.1:5000/pmem-csi-driver:canary -version`
>> `heads/devel-0-g9cd0dc90`
>
> yes, good point, helps with some of my skepticism. 
> BTW is there way to tag docker image so that it indicates CID in its
> tag, but remains still usable via static name (like we use "canary"
> now) by k8s manifest? That would help to visualize what is in image.

It's possible:
https://github.com/kubernetes-csi/csi-release-tools/blob/master/build.make#L68
Look for --label revision=$(REV).

Probably worth adding, because it covers another situation where one
might want to check the version (get image without running it).

avalluri:
Added initial tag: v0.0.0 and now version is generated from that tag, and same tag is set as label to docker images.

Issue #185, opened 2019-03-05
NodeStageVolume does not check for VolumeCapability in NodeStageVolumeRequest


okartau:
This causes one test in csi-sanity to fail, as VolumeCapability is required field:
```
Summarizing 1 Failure:
[Fail] Node Service NodeStageVolume [It] should fail when no volume capability is provided 
```
Somehow, this failure does not show up in `make test_e2e` in 4*VM cluster, but is showing up in unified-mode csi-sanity test that I tried for additional verification after recent csi-1.0 change.

okartau:
Seems the difference between `make test_e2e` and `util/sanity-unified.sh` results is explainable by recent changes in csi-test. `make test_e2e` uses fixed version 1.0.2, while in manual run of csi-sanity I used current head of csi-test repo which now is v1.1.0-51-ged673d8, and there has been update of NodeStageVolume test in between, looks like commit:
"Use real volumes when testing no capabilities" from Feb-4

PR #187, opened 2019-03-05
NodeStageVolume: check for VolumeCapability in request

okartau:
Lack of this check causes failure with newest csi-sanity

Fixes: #185


okartau:
I re-checked this change does not affect csi-sanity test with currently fixed version, runs ok, let's merge...

PR #188, opened 2019-03-05
fix deployment RBAC rules

avalluri:
cluster-driver-registrar sidecar needs to "create/delete" csidrivers hence need
these RBAC rules.

### DIFF
--- a/deploy/kubernetes-1.13/pmem-csi-direct.yaml
+++ b/deploy/kubernetes-1.13/pmem-csi-direct.yaml
@@ -35,6 +35,9 @@ rules:
 - apiGroups: [ "csi.storage.k8s.io" ]
   resources: [ "csinodeinfos" ] # attacher - {get, list, watch}
   verbs: [ "get", "list", "watch" ]
+- apiGroups: [ "csi.storage.k8s.io" ]
+  resources: [ "csidrivers" ] # driver-registrar
+  verbs: [ "create", "delete" ]
 # NOTE: cluster-driver-registrar v1.0.1 RBAC rules are broken and does not list
 # customresourcedefinitions in its RBAC rules:
 # https://github.com/kubernetes-csi/cluster-driver-registrar/blob/v1.0.1/deploy/kubernetes/rbac.yaml
--- a/deploy/kubernetes-1.13/pmem-csi-lvm.yaml
+++ b/deploy/kubernetes-1.13/pmem-csi-lvm.yaml
@@ -35,6 +35,10 @@ rules:
 - apiGroups: [ "csi.storage.k8s.io" ]
   resources: [ "csinodeinfos" ] # attacher
   verbs: [ "get", "list", "watch" ]
+- apiGroups: [ "csi.storage.k8s.io" ]
+  resources: [ "csidrivers" ] # driver-registrar
+  verbs: [ "create", "delete" ]
+
 # NOTE: cluster-driver-registrar v1.0.1 RBAC rules are broken and does not list
 # customresourcedefinitions in its RBAC rules:
 # https://github.com/kubernetes-csi/cluster-driver-registrar/blob/v1.0.1/deploy/kubernetes/rbac.yaml


pohly:
Short post-mortem:
* why did we miss this when reviewing and testing PR #177 ?
* should we use the opportunity to improve the process of keeping these RBAC rules up-to-date, for example by scripting the process? It doesn't have to be complete, just some script which gets the upstream files and cobbles something together on stdout would already be an improvement.

avalluri:
> * why did we miss this when reviewing and testing PR #177 ?

I am not sure, but my wild guess is: This rule was part of my initial commits, but removed while addressing  you comments about syncing with original RBAC rules. So out tests couldn't catch this as CSIDrivers crd already installed with old code.

> * should we use the opportunity to improve the process of keeping these RBAC rules up-to-date, for example by scripting the process? It doesn't have to be complete, just some script which gets the upstream files and cobbles something together on stdout would already be an improvement.

The issue here is this was broken in upstream cluster-driver-registrar v1.0 itself: https://github.com/kubernetes-csi/cluster-driver-registrar/blob/release-1.0/deploy/kubernetes/rbac.yaml

pohly:

>> * why did we miss this when reviewing and testing PR #177 ?
>
> I am not sure, but my wild guess is: This rule was part of my initial
> commits, but removed while addressing you comments about syncing with
> original RBAC rules. So out tests couldn't catch this as CSIDrivers
> crd already installed with old code.

So where is this breaking at the moment, i.e. do we have a test that
fails with the RBAC rules in master?

>> * should we use the opportunity to improve the process of keeping
>> these RBAC rules up-to-date, for example by scripting the process? It
>> doesn't have to be complete, just some script which gets the upstream
>> files and cobbles something together on stdout would already be an
>> improvement.
>
> The issue here is this was broken in upstream cluster-driver-registrar
> v1.0 itself:
> https://github.com/kubernetes-csi/cluster-driver-registrar/blob/release-1.0/deploy/kubernetes/rbac.yaml

That's a special case. I still think we should attempt to script the
creation of our RBAC section.

avalluri:
> So where is this breaking at the moment, i.e. do we have a test that fails with the RBAC rules in 
master?
This is fixed in cluster-driver-registrar sidecar master with this commit: https://github.com/kubernetes-csi/cluster-driver-registrar/commit/ff95ba0d90240784ef8b6ceb392533174d467fec#diff-59021f01acbcbd85d054c17acc3e26c8
```
I couldn't find any tests in this repo.

**Comment by pohly** [comment]
My question was about tests in pmem-csi. Does any of our tests fail because the RBAC rules are wrong? If not, then our test coverage is not good enough and we need to add a test.

**Comment by okartau** [comment]
I think the `make test_e2e` runnable in 4xVM cluster is our only "in-repo" proper test set involving k8s deployment, and I haven't seen RBAC-related errors during this test recently.

**Comment by pohly** [comment]
@avalluri if the tests aren't failing, how did you notice the missing RBAC entries?

**Comment by avalluri** [comment]
@okartau Do you see below error messages in driver-reigstar logs on pmem-csi-controller-0 Pod :

`E0308 07:20:13.842555       1 k8s_register.go:118] Failed to create CSIDriver object: csidrivers.csi.storage.k8s.io is forbidden: User "system:serviceaccount:default:csi-service-account-controller" cannot create resource "csidrivers" in API group "csi.storage.k8s.io" at the cluster scope`

Can you check the output of : `kubectl get csidrivers`

**Comment by avalluri** [comment]
One reason our e2e tests could not find this could be that we use unified mode for tests, where this is not visible. Does our tests uses driver deployment by pmem-csi-{lvm,direct}.yaml ?

**Comment by okartau** [comment]
that explains it then for e2e test cases, which use unified mode because of https://github.com/kubernetes-csi/csi-test/issues/142

What about or "deploy example" case, i.e. steps from README ending up in pmem-app mounting one directory? Does this fail currently?
I think we should create and put in repo a script performing those steps, which is then usable as one-command test case, I used to have it for trials already in my workdir, I could publish to repo, for example undel util/, or perhaps better under deploy/

**Comment by okartau** [comment]
> @okartau Do you see below error messages in driver-reigstar logs on pmem-csi-controller-0 Pod :
> 
> `E0308 07:20:13.842555 1 k8s_register.go:118] Failed to create CSIDriver object: csidrivers.csi.storage.k8s.io is forbidden: User "system:serviceaccount:default:csi-service-account-controller" cannot create resource "csidrivers" in API group "csi.storage.k8s.io" at the cluster scope`
> 

confirmed, I can see same error

> Can you check the output of : `kubectl get csidrivers`
```
$ kubectl get csidrivers
No resources found.
```

So yes, roles are missing, But if we consider steps shown in Makefile, these all work without errors
up to end (create pmem-app, get it to Running, run `df` in it  and see /data mounted.
So what is impact of this error in relation to deployment example?
Should we modify deployment example so that such problem would make it fail?

**Comment by avalluri** [comment]
> What about or "deploy example" case, i.e. steps from README ending up in pmem-app mounting one directory? Does this fail currently?

Volume provisioning itself is failing.

```
$ kubectl create -f - < deploy/kubernetes-1.13/pmem-pvc.yaml
 persistentvolumeclaim/pmem-csi-pvc created
$ kubectl get pvc 
NAME           STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
pmem-csi-pvc   Pending                                      pmem-csi-sc    54s
$ kubectl describe pvc pmem-csi-pvc
Events:
  Type       Reason                Age               From                         Message
  ----       ------                ----              ----                         -------
  Normal     ExternalProvisioning  2s (x7 over 75s)  persistentvolume-controller  waiting for a volume to be created, either by external provisioner "pmem-csi.intel.com" or manually created by system administrator
```

**Comment by okartau** [comment]
> I think we should create and put in repo a script performing those steps, which is then usable as one-command test case, I used to have it for trials already in my workdir, I could publish to repo, for example undel util/, or perhaps better under deploy/

actually, first half of this is now performed by `make start`,
and 2 remaining commands are printed at the end of `make start` so this case is already simple to run.
But still, one-liner activation would make verification even simpler.
Something similar to what hostpath has as one-script deploy example

**Comment by okartau** [comment]
> Volume provisioning itself is failing.

works here:
```
$ kubectl get pvc 
NAME           STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
pmem-csi-pvc   Bound    pvc-ad327c50-4182-11e9-b352-deadbeef0100   8Gi        RWO            pmem-csi-sc    19m
```

**Comment by avalluri** [comment]
@okartau, its my bad. At my end wrong storage class(with changed driver name) install, that was the issue for failed provisioning. Yes driver is working as its expected. Now i am not clear why this cluster wide driver registration really required ;-), who uses this driver info.

**Comment by avalluri** [comment]
I tried by removing "driver-registrar" sidecar itself from driver deployment, to my surprise everything work:

```
$ kubectl get po pmem-csi-controller-0 -o json | jq '.spec.containers[].name'
"external-provisioner"
"external-attacher"
"pmem-driver"
$ kubectl get po 
NAME                    READY   STATUS    RESTARTS   AGE
pmem-csi-68zdv          2/2     Running   2          109s
pmem-csi-controller-0   3/3     Running   0          109s
pmem-csi-tlt9b          2/2     Running   0          109s
pmem-csi-zjr89          2/2     Running   2          109s

$ kubectl create -f - <  deploy/kubernetes-1.13/pmem-pvc.yaml 
persistentvolumeclaim/pmem-csi-pvc created
$ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                  STORAGECLASS   REASON   AGE
pvc-64cdcbca-4189-11e9-a367-deadbeef0100   8Gi        RWO            Delete           Bound    default/pmem-csi-pvc   pmem-csi-sc             5s
$ kubectl create -f - <  deploy/kubernetes-1.13/pmem-app.yaml 
pod/my-csi-app created
$ kubectl exec my-csi-app df /data
Filesystem           1K-blocks      Used Available Use% Mounted on
/dev/ndbus0region0fsdax/64e4edd3-4189-11e9-bc9f-1e8ce08dafc3
                       8191416     36852   7718752   0% /data

```

Does Kubernetes really uses this driverinfo ? @pohly  you have any idea.

**Comment by pohly** [comment]
> that explains it then for e2e test cases, which use unified mode because of kubernetes-csi/csi-test#142

Wait, we have to distinguish between actual E2E testing (use a driver through Kubernetes) and csi-sanity testing (invoke driver directly). Both is part of the test/e2e suite because both need a cluster where the driver can run in a container. But only csi-sanity testing is using unified mode. E2E testing uses the driver that was deployed by "make start", i.e. pmem-csi.yaml = pmem-csi-lvm.yaml.

> Does Kubernetes really uses this driverinfo ? @pohly you have any idea.

It does use it to determine whether a driver wants to skip the `ControllerPublishVolume` call. If a driver doesn't support that call, it can omit installing the external-attacher.

So we can do one of these two things:
1. stop using cluster-driver-registrar
2. stop using external-attacher

I bet that with broken RBAC rules and no external-attacher, tests will start to fail with pods waiting for volumes to become attached. That's then the test case that we want for verifying that cluster-driver-registrar really works, because it should start to work once the RBAC rules are correct.

**Comment by avalluri** [comment]
So, what is the take on this PR. shall we merge this or live with driver registry failure.

**Comment by pohly** [comment]
Let's use this PR to do one of the two things I suggested, with RBAC rules updated accordingly.

I'm leaning towards removing the external-attacher.

**Comment by avalluri** [comment]
> Let's use this PR to do one of the two things I suggested, with RBAC rules updated accordingly.
> 
> I'm leaning towards removing the external-attacher.

Either case we need this change, so i would propose to take this PR and I would create new PR to remove external-attacher.

**Comment by avalluri** [comment]
my quick trail without using external-attacher failed:

```
Events:
  Type     Reason              Age                From                     Message
  ----     ------              ----               ----                     -------
  Normal   Scheduled           41s                default-scheduler        Successfully assigned default/my-csi-app to host-2
  Warning  FailedAttachVolume  10s (x2 over 26s)  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-83faeaad-41a5-11e9-a367-deadbeef0100" : attachment timeout for volume 841d3082-41a5-11e9-b889-2a0501134154

$ kubectl get csidrivers pmem-csi -o yaml
spec:
  attachRequired: false
  podInfoOnMountVersion: v1

$ kubectl get po pmem-csi-controller-0 -o json | jq '.spec.containers[].name'
"driver-registrar"
"external-provisioner"
"pmem-driver"

```

Somehow Kuberentes still trying to attach this volume,  irrespective of `attachRequired: false`

**Comment by pohly** [comment]
I've had a look at this myself. Key points:
- cluster-driver-registrar keeps running despite the error; that part wasn't clear to me
- csinodeinfo.csi.storage.k8s.io gets created by node-driver-registrar, so topology works

The incorrect RBAC rules can be detected by checking for csidrivers.csi.storage.k8s.io. Here's what it needs to look like:
```
$ kubectl describe csidrivers.csi.storage.k8s.io
Name:         pmem-csi
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  csi.storage.k8s.io/v1alpha1
Kind:         CSIDriver
Metadata:
  Creation Timestamp:  2019-03-11T07:13:29Z
  Generation:          1
  Resource Version:    2577
  Self Link:           /apis/csi.storage.k8s.io/v1alpha1/csidrivers/pmem-csi
  UID:                 2bbff8c4-43cd-11e9-ae9e-deadbeef0100
Spec:
  Attach Required:            true
  Pod Info On Mount Version:  v1
Events:                       <none>
```

However, just checking for this doesn't guarantee that it was created by the currently running cluster-driver-registrar. Adding a check is still useful when combined with testing in a CI environment where the cluster is brought up in a clean state for each test run. I'll look into that.

In the meantime, let's merge this PR.

PR #189, opened 2019-03-06
Set default driver name to pmem-csi.intel.com

avalluri:
As per CSI specification plugin name should include plugin's host company name
and the plugin name. And it must follow domain name notation format.

Hence set default plugin name to 'pmem-csi.intel.com', which can be altered
using -driverName command-line argument.

### DIFF
--- a/cmd/pmem-csi-driver/main.go
+++ b/cmd/pmem-csi-driver/main.go
@@ -20,7 +20,7 @@ import (
 
 var (
 	/* generic options */
-	driverName       = flag.String("drivername", "pmem-csi", "name of the driver")
+	driverName       = flag.String("drivername", "pmem-csi.intel.com", "name of the driver")
 	nodeID           = flag.String("nodeid", "nodeid", "node id")
 	endpoint         = flag.String("endpoint", "unix:///tmp/pmem-csi.sock", "PMEM CSI endpoint")
 	mode             = flag.String("mode", "unified", "driver run mode : controller, node or unified")
--- a/deploy/kubernetes-1.13/pmem-csi-direct.yaml
+++ b/deploy/kubernetes-1.13/pmem-csi-direct.yaml
@@ -130,7 +130,7 @@ spec:
       - name: external-provisioner
         imagePullPolicy: Always
         image: quay.io/k8scsi/csi-provisioner:v1.0.1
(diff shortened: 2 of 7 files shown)


okartau:
If default name is already set, why do we set it again to same value in activation in deployment file?
Is it for showing example how to change it?
But then, if CSI spec requires it be in certain form, why do we even provide method of overriding it via CLI, as changed value will potentially lead to violating CSI spec?

avalluri:
> If default name is already set, why do we set it again to same value in activation in deployment file?
> Is it for showing example how to change it?
> But then, if CSI spec requires it be in certain form, why do we even provide method of overriding it via CLI, as changed value will potentially lead to violating CSI spec?

Yes, i agree. These questions are valid.
One quick thing in my mind is when we support 'ephemeral' volumes, one might need to deploy the diver with different driver name(pmem-csi-ephemeral.intel.com) , as [kubrenetes inline volume design](https://github.com/vladimirvivien/kubernetes-enhancements/blob/3fc30a1e84acb9adf1fe9a768976cbc673b4b99a/keps/sig-storage/0035-20190122-inline-csi-volumes-support.md#persistent-vs-ephemeral-lifecycle-setting) expects driver can implement either "persistent" or "ephemeral" volumes.

pohly:
> If default name is already set, why do we set it again to same value in activation in deployment file?
> Is it for showing example how to change it?

That, and to make it more visible. The same string also occurs in the
storage class .yaml. It's redundant, but I'm in favor of keeping it.

> But then, if CSI spec requires it be in certain form, why do we even
> provide method of overriding it via CLI, as changed value will
> potentially lead to violating CSI spec?

The E2E testing could use this to install the driver multiple times in
parallel. It no longer does that, but perhaps we want it back for some
test.

Having the driver running multiple times with different behavior is a
more realistic use-case. As @avalluri said, ephemeral volumes will have
to do that.

okartau:
we do not change the drivername in deploy example of unified mode? That's the one we use in e2e test now, I plan to run this PR through e2e

avalluri:
> we do not change the drivername in deploy example of unified mode?

Ya, i missed this, shall I change this too.

avalluri:
I fixed the driver name in unified mode deployment.

avalluri:
Is it good enough for merging?

avalluri:
@okartau Ping

PR #191, opened 2019-03-08
Make DeviceMode:Direct e2e testing easier

okartau:
Add more direct-mode deployment support.
Test cluster start and e2e testing can use either
lvm- or direct-mode deployment file,
based on config.variable TEST_DEVICEMODE
Mention use of this variable in README.


README.md
```diff
@@ -447,6 +447,16 @@ $ REPO_ROOT=`pwd` ginkgo '-focus=sanity' -failFast ./test/e2e/
 ...
 ```
 
+#### Running E2E tests in DeviceMode:Direct
+
+Modify two test scripts to use deployment in Direct mode:
+
+```sh
+    $ sed -i 's/csi.yaml/csi-direct.yaml/' test/e2e/storage/sanity.go test/start-stop.make
```
avalluri: how about changing the link, instead?

deploy/kubernetes-1.13/pmem-unified-csi-direct.yaml
```diff
@@ -0,0 +1,77 @@
...
+  serviceName: pmem-csi-unified
+  template:
+    metadata:
+      labels:
+        app: pmem-csi-unified
+    spec:
+      initContainers:
+      containers:
+      - name: pmem-driver
+        image: 10.237.72.78:5000/pmem-csi-driver:canary
```
avalluri: wrong image registry :)

deploy/kubernetes-1.13/pmem-unified-csi-direct.yaml
```diff
@@ -0,0 +1,77 @@
...
+        imagePullPolicy: Always
+        terminationMessagePath: /tmp/termination-log
+        args:  [ "-v=5",
+                 "-drivername=pmem-csi",
```
avalluri: As PR #189 merged,  can u change the driver name.

README.md
okartau: I agree, changing link is cleaner than patching testing files. For that, we need to create link for unified file as well. I will do that.

pohly: Both approaches have the problem that they need to modify content of the git repo. A better solution would be to make sanity.go and start-stop.make configurable such that they do the right thing when an env variable is set. Env variables (in contrast to command line variables) can be checked both by make and Go code.

avalluri: > Both approaches have the problem that they need to modify content of the git repo. A better solution would be to make sanity.go and start-stop.make configurable such that they do the right thing when an env variable is set. Env variables (in contrast to command line variables) can be checked both by make and Go code.

I agree.

okartau: I agree as well, modifying state of repo objects is messy way, so let's avoid it. I will try to introduce env.vars, with default leading to LVM mode.

okartau:
pushed the method using env.vars. Its quite brutal i.e. uses the env var value directly in file name, but it should work for normal use, and it saves some code with checking and setting more variables. Sure, setting env.var to illegal value will cause tester failure but isn't that true for other parameters as well.

okartau:
updated commit msg to reflect use of env.var instead of sed patch. Also, made README chapter bit more clear about the use.
Does it look usable now?

README.md
```diff
@@ -447,6 +447,21 @@ $ REPO_ROOT=`pwd` ginkgo '-focus=sanity' -failFast ./test/e2e/
 ...
 ```
 
+#### Starting test cluster and running E2E tests in DeviceMode:Direct
+
+Export environment variable that guides testing scripts to select deployment files for Direct mode:
+
+```sh
+    $ export PMEM_CSI_DEVICEMODE_DIRECT="-direct"
```
avalluri: I would prefer more generic than making it specific to "direct". Like choosing "device mode" : `PMEM_CSI_DEVICEMODE="lvm"  # for lvm`
     or
`PMEM_CSI_DEVICEMODE="direct" # for direct`

And would prefer to prefix the test variables with TEST_,  how about TEST_DEVICEMODE?

okartau:
Pushed new version. That makes Direct mode equal citizen with LVM mode,
and mode has to be selected explicitly, there is no default:LVM which was before
activated without selection.
There is check in Makefiles to refuse testing if DeviceMode is not set.
Note that symlinks in deployment directory still support default being LVM.

deploy/kubernetes-1.13/pmem-unified-csi-direct.yaml
```diff
@@ -0,0 +1,77 @@
+# Deploys the pmem-csi driver onto exactly one node, running in unified mode.
+# That CSI drivers gets exposed with a NodePort service and the fixed port
+# 30123 for use with csi-test sanity testing.
```
avalluri: Just curious to konw, where are we doing this? i mean fixed port.

okartau: I have no idea. This text is in pmem-unified-csi-direct.yaml only because this file is based on pmem-unified-csi-lvm.yaml

okartau:
@avalluri pointed out why not define MEM_CSI_TEST_DEVICEMODE into test/test-config.sh.
Yes we can do it, probably best solution. taking away forced need from user to export env variable .
The downside however will be, majority of use will be fixed to that "default" mode then which is defined in repo file, and other mode will get less coverage. But real CI will hopefully fix that, running both modes.
I will make one more iteration putting value in test/test-config.sh

okartau:
next iteration pushed, we use variable in test/test-config.sh which is picked up by test scripts.
The default value in repo is lvm

deploy/kubernetes-1.13/pmem-unified-csi-direct.yaml
```diff
@@ -0,0 +1,77 @@
...
+          privileged: true
+        volumeMounts:
+          # We create temporary staging and target directories under /var/tmp
+          # during csi-sanity testing and must have those visible inside the
+          # container (see test/e2e/storage/sanity.go).
+          - name: var-tmp
+            mountPath: /var/tmp
+          - name: dev-dir
+            mountPath: /dev
+            mountPropagation: Bidirectional
```
avalluri: `Bidirectional` Still required?

okartau: thanks, good catch, I also found it myself around same times, force pushed deletion.
Once creeping in, this line gets copied around at same speed we find and delete it,
but we get better.

README.md
```diff
@@ -1,36 +1,36 @@
...
-        - [QEMU + Kubernetes](#qemu--kubernetes)
-        - [Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster)
-        - [Running commands on test cluster nodes over ssh](#running-commands-on-test-cluster-nodes-over-ssh)
-        - [Running E2E tests](#running-e2e-tests)
-    - [Communication and contribution](#communication-and-contribution)
+PMEM-CSI for Kubernetes
+----------------
+
+Table of Contents
+----------------
```
avalluri: I think, we can removed this ToC header.

test/test.make
```diff
@@ -97,6 +97,8 @@ RUNTIME_DEPS += LC_ALL=C LANG=C sort -u
 # Use count=1 to avoid test results caching, does not make sense for e2e test.
 .PHONY: test_e2e
 test_e2e: start
+	. test/test-config.sh && \
+	export TEST_DEVICEMODE && \
```
pohly: I prefer this:
```
	. test/test-config.sh && \
	TEST_DEVICEMODE=$$TEST_DEVICEMODE \
        KUBECONFIG=`pwd`/_work/clear-kvm-kube.config \
        REPO_ROOT=`pwd` \
        go test -count=1 -timeout 0 -v ./test/e2e
```
Then it is consistent with how the other env vars are passed.

README.md
```diff
@@ -611,6 +611,8 @@ and use `kubectl` binary on the host running VMs.
 Use `make stop` to stop the virtual machines. The cluster state
 remains preserved and will be restored after next `make start`.
 
+The DeviceMode (lvm or direct) used in testing is selected using variable TEST_DEVICEMODE in [test-config.sh](test/test-config.sh).
+
```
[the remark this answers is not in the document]
okartau: DeviceMode:LVM is my creation, I wanted to introduce new mode without using just "mode" which already was overloaded with 3 meanings, so I called it devicemode, one word to make it sound/feel same way as "mode", i.e. one-word. CamelCase is just style, to stress even more that just meant to add one mode... Colon and LVM I used to make it, again, hm, one word, I meant to express DeviceMode=LVM as in "devicemode is LVM" with short expression. If that is all to fuzzy, fine, we can spell it out to different words if you like that more.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
```diff
@@ -150,9 +150,11 @@ spec:
       - name: pmem-driver
         image: 192.168.8.1:5000/pmem-csi-driver:canary
         imagePullPolicy: Always
+        terminationMessagePath: /tmp/termination-log
```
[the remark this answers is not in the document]
okartau: We need to pass /dev to container, and if we do this, we (sometimes?) hit container creation error because termination-log is located on /dev/ by default. Adding this clause works around failure.

deploy/kubernetes-1.13/pmem-unified-csi-lvm.yaml
```diff
@@ -0,0 +1,74 @@
+# Deploys the pmem-csi driver onto exactly one node, running in unified mode.
+# That CSI drivers gets exposed with a NodePort service and the fixed port
+# 30123 for use with csi-test sanity testing.
```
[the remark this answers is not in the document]
okartau: you mean we leave the text "That CSI drivers gets exposed with a NodePort service" and delete the rest of that sentence?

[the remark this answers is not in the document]
pohly: Yes.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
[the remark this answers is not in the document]
pohly: Can you add a comment explaining that?

As it stands, `/tmp/termination-log` is useless because nothing knows that it needs to write at that location, so someone seeing this will wonder about it.

okartau:
force-pushed with recent comments addressed. marking all resolved (except DeviceMode:LVM thread).
Needs another round of tests as there were changes in makefile and deploy files, so WIP,
will do soon...

test/start-stop.make
```diff
@@ -40,8 +40,9 @@ start: _work/clear-kvm.img _work/kube-clear-kvm _work/start-clear-kvm _work/ssh-
 		touch _work/clear-kvm.secretsdone; \
 	fi
 	_work/ssh-clear-kvm kubectl version --short | grep 'Server Version' | sed -e 's/.*: v\([0-9]*\)\.\([0-9]*\)\..*/\1.\2/' >_work/clear-kvm-kubernetes.version
+	. test/test-config.sh && \
```
okartau: This line was added as one last change just now, @pohly you agree it's ok? (it somehow had escaped my earlier testing, but we need to get TEST_DEVICEMODE to env here as well)

README.md
```diff
@@ -1,38 +1,34 @@
...
+        -   [Software required](#software-required)
+        -   [Hardware required](#hardware-required)
+    -   [Supported Kubernetes versions](#supported-kubernetes-versions)
+    -   [Setup](#setup)
+        -   [Get source code](#get-source-code)
+        -   [Build PMEM-CSI](#build-pmem-csi)
+        -   [Run PMEM-CSI on Kubernetes](#run-pmem-csi-on-kubernetes)
+    -   [Automated testing](#automated-testing)
+        -   [Unit testing and code quality](#unit-testing-and-code-quality)
+        -   [QEMU + Kubernetes](#qemu-kubernetes)
```
pohly: The tool (pandoc?) that was used to generate this version of the TOC is slightly buggy: the correct link here is `#qemu--kubernetes`, not `#qemu--kubernetes`. That wasn't obvious from the diff because each line had changed whitespace.

I'll submit a fix, but long term we'll need to figure out how we want to generate this TOC.

PR #192, opened 2019-03-08
docs

pohly:
This reformats the docs (first commit) and then makes it more consistent (second commit).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 <!-- based on template now, remaining parts marked as FILL TEMPLATE:  -->
 
-# Intel PMEM-CSI plugin for Kubernetes
+# Intel PMEM-CSI for Kubernetes
 
 ## Table of Contents
 
-- [Intel PMEM-CSI plugin for Kubernetes](#intel-pmem-csi-plugin-for-kubernetes)
+- [Intel PMEM-CSI for Kubernetes](#intel-pmem-csi-for-kubernetes)
     - [Table of Contents](#table-of-contents)
     - [About](#about)
     - [Design](#design)
@@ -33,8 +33,8 @@
     - [Supported Kubernetes versions](#supported-kubernetes-versions)
     - [Setup](#setup)
         - [Get source code](#get-source-code)
-        - [Build plugin](#build-plugin)
-        - [Run plugin](#run-plugin)
+        - [Build PMEM-CSI](#build-pmem-csi)
+        - [Run PMEM-CSI on Kubernetes](#run-pmem-csi-on-kubernetes)
     - [Automated testing](#automated-testing)
         - [Unit testing and code quality](#unit-testing-and-code-quality)
         - [End-to-end testing (E2E)](#end-to-end-testing-e2e)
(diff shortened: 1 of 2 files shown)


okartau:
This looks as good improvement!
Does 2-level TOC help compared to 1-level, given we have short chapters?

pohly:
I think having a complete table of contents helps with grasping at a glance what the document and each chapter is about. But I don't have a strong opinion about this.

MCamp859:
My 2 cents: I think the new 2nd level of TOC is helpful to give the readers more context, but the 3rd level makes the TOC quite long, especially when viewing in the main readme.md file. Other opinions?

pohly:
@avalluri what do you think?

README.md
```diff
@@ -4,23 +4,66 @@
...
+            - [Node Controller Server](#node-controller-server)
+            - [Node Server](#node-server)
+        - [Communication channels](#communication-channels)
+        - [Security](#security)
+        - [Volume Persistency](#volume-persistency)
+            - [Usage on Kubernetes](#usage-on-kubernetes)
+    - [Prerequisites](#prerequisites)
+        - [Software required](#software-required)
+        - [Hardware required](#hardware-required)
+        - [Persistent memory device initialization](#persistent-memory-device-initialization)
```
avalluri: I think we can merge this section with above "Hardware required"  as its talk about `impctl`, which can be used only on hardware.

README.md
```diff
@@ -4,23 +4,66 @@
...
+            - [Namespace modes in DeviceMode:Direct](#namespace-modes-in-devicemodedirect)
+            - [Using limited amount of total space in DeviceMode:Direct](#using-limited-amount-of-total-space-in-devicemodedirect)
+        - [Driver modes](#driver-modes)
+        - [Driver Components](#driver-components)
+            - [Identity Server](#identity-server)
+            - [Node Registry Server](#node-registry-server)
+            - [Master Controller Server](#master-controller-server)
+            - [Node Controller Server](#node-controller-server)
+            - [Node Server](#node-server)
+        - [Communication channels](#communication-channels)
```
avalluri: I feel like "Communication channels" sounds incorrect, shall we rename this to something like "Driver communication".

README.md
```diff
@@ -4,23 +4,66 @@
...
+            - [Running commands on test cluster nodes over ssh](#running-commands-on-test-cluster-nodes-over-ssh)
+            - [Running E2E tests](#running-e2e-tests)
+    - [Communication and contribution](#communication-and-contribution)
+
 
 ## About
 
 ---
-*Note: This is Alpha code and not production ready.*
+ *Note: This is Alpha code and not production ready.*
```
avalluri: This additional space intentional ?

README.md
```diff
@@ -1,10 +1,10 @@
 <!-- based on template now, remaining parts marked as FILL TEMPLATE:  -->
 
-# Intel PMEM-CSI plugin for Kubernetes
+# Intel PMEM-CSI for Kubernetes
```
avalluri: After removing "plugin" in this sentence, i feel like lost its meaning :)

avalluri:
> @avalluri what do you think?

I totally agree with @MCamp859.

README.md
pohly: This isn't a sentence :-)

Trying to explain what PMEM-CSI is in the title of the document is bound to fail. The title just needs to tell the reader which document he is looking at. What the thing is that the document describes is explained in the introduction.

README.md
pohly: Yes, it was treated like a section without that.

README.md
pohly: Fine with me.

README.md
pohly: Okay, will do.

README.md
```diff
@@ -4,23 +4,66 @@
...
+        - [Driver Components](#driver-components)
+            - [Identity Server](#identity-server)
+            - [Node Registry Server](#node-registry-server)
+            - [Master Controller Server](#master-controller-server)
+            - [Node Controller Server](#node-controller-server)
+            - [Node Server](#node-server)
+        - [Communication channels](#communication-channels)
+        - [Security](#security)
+        - [Volume Persistency](#volume-persistency)
+            - [Usage on Kubernetes](#usage-on-kubernetes)
```
pohly: I suspect part of our problem with deep nesting is that we simply have too much of it.

For example, this "Usage on Kubernetes" sub-section for volume persistency adds usage instructions to a section that is about design. I think it would be more useful to users to have top-level "Design + Architecture" and "Usage" sections, where the former is about background information and the "Usage" section has the actual details for someone trying to install and use the driver.

I'll leave it like this for now, though. A lot more work is needed before we have good usage instructions, and it'll change again once we have published images.

pohly:
I've added more changes.

README.md
avalluri: fair enough.

Issue #195, opened 2019-03-13
LVM mode: restarted pod does not see block device of previously created volume


okartau:
This happens when my-csi-app deployment steps (per README) are executed,
with added "delete pod which created pvc" step between pvc creation and my-csi-app creation.

After "kubectl delete pod", the pod will be restarted.
The following my-csi-app creation does not succeed, remaining in ContainerCreating state with describe error:
```
MountVolume.MountDevice failed for volume "pvc-9e15b576-457d-11e9-aaef-deadbeef0100" : rpc error: code = Unknown desc = no device information for /dev/ndbus0region0fsdax/9e1cd240-457d-11e9-8238-7a57be51a003
```

The corresponding node pmem-csi pod log has:
```
I0313 NodeStageVolume: VolumeID is 9e1cd240-457d-11e9-8238-7a57be51a003
I0313 NodeStageVolume: Staging target path is /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-9e15b576-457d-11e9-aaef-deadbeef0100/globalmount
I0313 NodeStageVolume: Requested fsType is ext4
I0313 Executing: lvs --noheadings --nosuffix -o lv_name,lv_path,lv_size --units B ndbus0region0fsdax
I0313 Output:   9e1cd240-457d-11e9-8238-7a57be51a003 /dev/ndbus0region0fsdax/9e1cd240-457d-11e9-8238-7a57be51a003 8589934592
  b900ab17-4573-11e9-a430-964763c9242d /dev/ndbus0region0fsdax/b900ab17-4573-11e9-a430-964763c9242d 8589934592
I0313 Executing: file -bsL /dev/ndbus0region0fsdax/9e1cd240-457d-11e9-8238-7a57be51a003
I0313 Output: cannot open `/dev/ndbus0region0fsdax/9e1cd240-457d-11e9-8238-7a57be51a003' (No such file or directory)
```

What seems to happen is that pmem-csi driver instance creates a new Logical Volume and there will corresponding block device. Then this container is terminated and new one starts, but the it does not get previously created block device visible under its /dev file system.
This is another instance of problems from the class "container does not see fully host /dev or /sys contents".
In DeviceMode:direct, there has been work around of such problem by explicit pass of /dev mount into container. It might possibly help here, too?

okartau:
Theory of cause gets confirmed by trying same in DeviceMode:direct. The problem is not reproducible.

okartau:
This needs re-checking.
[...]

PR #196, opened 2019-03-14
 Persisting driver state and reload state on restart.

### LINKED ISSUE #25: Clean handling of driver restart
What is the state of provisioned volumes after driver restart?
Lvolumes may exist, and may be mounted or remain unmounted.
Host maintains that state, but driver may restart.
As driver does not have permanently stored volumes state,
it does not have context after restart, and that means trouble for next operations.
All existing volumes or mounts are orphaned in context of driver.
Driver should try to reach the same, or at least some reasonable state about existing volumes and mounts.
Here, similar to #11, it is also possible that some Lvolumes exist in host context which were not created by PMEM-CSI driver, so driver needs to recognize own vs foreign volumes as well.

avalluri:
Upon restarts, driver has to recognize the provisioned volumes on last run.  For
that driver has to persist its state(provisioned volume information). We have two options:
    
 1. Per node state: Persist the driver state on each node specific to that node.
    Upon driver start, it has to load the persisted information. Whenever a node
    registers, Controller/Master has to collect the volume information stored on
    that node.
    
 2. Cluster level state: Persist the consolidated cluster level driver state, say
    to a configmap on Kubernetes. This is supposed to done at Controller side, but not
    clear how to pass node specific information to a node.
    
This change implements the fist option, i.e, per node state.
    
Defined a new interface, StateManager, that handles the state persistency irrespective of either per node or cluster. Also provided an implementation of  StateManager that persists the state to a file.
    
Made changes to RegistrySrerver, so that it informs the listeners upon registering of a node controller. The ControllerServer on master, listens on  this and collects the volume information on node, by calling csi.ControllerServer.ListVolumes.
    
FIXES #25 - Clean handling of driver restart


avalluri:
@okartau i tested without Bidirectional /dev mount and it works fine. Now i removed that from my change.

okartau:
msg of 3rd commit has some typos and extra chars, 
also better to be specific what was restarted, so I propose:
This is workaround for issue #195, i.e, upon pod restart, driver could not see the
block devices created on last run. Now we mount /dev into the driver container.

okartau:
overall when we state "restart", I think we should be more specific about what was restarted,
we can easily have at least 3 levels: pod, node, cluster.
(more levels by combining/mixing those, for example reboot of subset of nodes in cluster).
Which levels of restart this change will cover? pod for sure, but what about node restart (i.e. host reboot).
Is the state restore supposed to work over node reboot?
In that case we should pay attention to mount of store dir, i.e. should not be on a RAM-based file system.

avalluri:
> msg of 3rd commit has some typos and extra chars,
> also better to be specific what was restarted, so I propose:
> This is workaround for issue #195, i.e, upon pod restart, driver could not see the
> block devices created on last run. Now we mount /dev into the driver container.

Thanks @okartau, now i rephrased the commit message. Hope it sounds better now.

avalluri:
> overall when we state "restart", I think we should be more specific about what was restarted,
> we can easily have at least 3 levels: pod, node, cluster.
> (more levels by combining/mixing those, for example reboot of subset of nodes in cluster).
> Which levels of restart this change will cover? pod for sure, but what about node restart (i.e. host reboot).
> Is the state restore supposed to work over node reboot?

Though i didn't test, but It supposed to work on all levels of restart. As the driver pod is started by Kubernetes on both node and cluster restart cases. So our focus is 'on Pod restart'.

> In that case we should pay attention to mount of store dir, i.e. should not be on a RAM-based file system.

Where to store the state is configurable via `-driverStatePath`, the one who deploy the driver has to choose the location on host that is mounted to driver container as shown in our deployment examples, similar to socket path. In the example we used `/var/lib/kubelet/plugins/pmem-csi.intel.com/volumes`.

okartau:
> Where to store the state is configurable via `-driverStatePath`, the one who deploy the driver has to choose the hostpath on node

I noticed the default for driverStatePath is on /tmp, and some example file also contains path under /tmp.
If this such deployment will happen on host having /tmp mounted to RAM-based, file system, there will be no saved state after reboot?

okartau:
I did put this PR through regular cycle, i.e. deploy ecample by README, and make test_e2e, no problems. But lets note as these tests do not touch the meaningful change i.e. restarting, this is merely smoke-test to verify existing functionality was not broken.

avalluri:
@pohly could you get time to review this.

avalluri:
> It even overwrites the only copy of the data in-place instead of doing the usual write+rename

Why write + rename needed? StateManager.Create() is called from CreateVolume() with newly created volume id and data. There is no data overwrite expected assuming that volume id is unique.

avalluri:
> If the goal is to support "all levels all restart", then the code also must be resilient against sudden powerloss

"all levels of restart" comment was in the context of @okartau' s comment about "Pod vs node vs cluster" restarts. This PR considered only "safe restarts" not power loss cases.

pohly:

>> If the goal is to support "all levels all restart", then the code also must be resilient against sudden powerloss
>
> "all levels of restart" comment was in the context of @okartau' s
> comment about "Pod vs node vs cluster" restarts. This PR considered
> only "safe restarts" not power loss cases.

I think we should consider a solution that covers all bases instead of
implementing something that just will have to be updated again in the
future.

avalluri:
@pohly I made changes to make file writes synchronous, hoping that this will address the sudden power loss case for some extent.

But i am not sure what should the driver do the case where, pmem device creation was success but writing it to state file failed. Do we need to treat it as CreateDevice request failed?.

We could avoid this maintaining of state file, if DeviceManager itself supports saving volume metatdata. Atleast LVM support this through "tags" but, ndctl does not support such tagging.

okartau:
I agree with @pohly that we should make effort to make the driver as safe as we can with available methods. Surely we can't avoid all errors caused by sudden power loss, but we should attempt to make  malfunction window smaller using reasonable effort.

okartau:
> But i am not sure what should the driver do the case where, pmem device creation was success but writing it to state file failed. Do we need to treat it as CreateDevice request failed?

The case with partial success is always more difficult to cover, similar to one we already had with missing block device entry in /dev (by namespace got created). Ideally, we should go back to previous state or try to repeat, to reach desired state, but in case of errors we likely can't do any of those, so we have to prepare for half-succeeded state, with recovery path later (garbage collection? which we currently don't have)

pohly:
> But i am not sure what should the driver do the case where, pmem
> device creation was success but writing it to state file failed. Do we
> need to treat it as CreateDevice request failed?.

There has been a failure, so you need to undo everything and return an
error from CreateDevice. You can't just return an error, because there's
no guarantee that the caller will try again.

I is also important not just to write the file atomically, but
also to consider potential sequences of events. In this case it is better
to:
1. write the state file first
2. then allocate storage

The restore procedure then needs to check for garbage entries in the
state file (= entries which do not actually have a volume) and remove
them.

Removing a volume must do it in the inverse order.

With that approach, we handle sudden powerloss between the two steps,
because it is always guaranteed that we know about a potentially created
volume.

It gets more tricky when the state file needs information from the
second step. Then one can try to write a placeholder entry and update it
later.

avalluri:
> There has been a failure, so you need to undo everything and return an error from CreateDevice. You can't just return an error, because there's no guarantee that the caller will try again. I is also important not just to write the file atomically, but also to consider potential sequences of events. In this case it is better to:
> 1. write the state file first
> 2. then allocate storage

This sounds good soultion, I updaed the PR with this chage.

avalluri:
@pohly Is this good enough to review?

pkg/pmem-state/pmme-state.go
```diff
@@ -0,0 +1,147 @@
...
+type fileState struct {
+	location string
+}
+
+var _ StateManager = &fileState{}
+
+// NewFileState instantiates the file state manager with given directory
+// location. It ensures the provided directory exists.
+// Returns error, if fails to create the direcotry incase of not pre-existing.
+func NewFileState(loc string) (StateManager, error) {
```
pohly: If `loc` must be a directory, why not call it `dir`? Same for `ensureLocation`. Or perhaps `directory`, depending whether you prefer abbreviations or not.

pkg/pmem-state/pmme-state.go
```diff
@@ -0,0 +1,147 @@
...
+	if err := json.NewEncoder(fp).Encode(data); err != nil {
+		// cleanup file entry before returning error
+		fp.Close() //nolint: errcheck, gosec
+		if e := os.Remove(file); e != nil {
+			glog.Warningf("file-state: fail to remove file %s: %s", file, e.Error())
+		}
+		return errors.Wrap(err, "file-state: failed to encode metadata")
+	}
+
+	return fp.Close()
```
pohly: What about making these IO operations crash-resilient?

pkg/pmem-state/pmme-state.go
```diff
@@ -0,0 +1,147 @@
...
+func ensureLocation(dir string) error {
+	if _, err := os.Stat(dir); err != nil {
+		if os.IsNotExist(err) {
+			if err := os.Mkdir(dir, 0750); err != nil { //nolint: vetshadow
+				return err
+			}
+		} else {
+			return err
+		}
+	}
```
pohly: To catch the error early you could check that the dir is actually a directory.

pkg/pmem-state/pmme-state.go
```diff
@@ -0,0 +1,147 @@
+/*
+Copyright 2019  Intel Corporation.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+package pmemstate
```
pohly: Typo in the filename.

Can you add unit tests to this package?

pkg/pmem-state/pmme-state.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/pkg/errors"
+	"k8s.io/klog/glog"
+)
+
+// GetAllFunc callback function used for StateManager.GetAll().
+// This function is called with ID, for each entry found in the state.
+type GetAllFunc func(id string) bool
+
+// StateManager manages the driver persistent state, i.e, volumes information
```
pohly: How does the state manager handle concurrency? This needs to be documented to ensure that either the manager or the caller do the necessary locking.

DEVELOPMENT.md
```diff
@@ -150,7 +150,7 @@ argument name        | meaning                                           | type
 -mode string               | driver run mode                                | string | controller, node, unified | unified
 -nodeid string             | node id                                        | string |              | nodeid
 -registryEndpoint string   | endpoint to connect/listen registry server     | string |              |
-
+-statePath                 | Directory path where to persist the state of the driver running on a node | string | absolute directory path on node | /tmp/pmem-volumes
```
pohly: This default value is going to be confusing for admins. It's clearly not suitable for persisting state across machine restart. If there is no good default, leave it empty and error out if not set.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
```diff
@@ -233,6 +233,7 @@ spec:
                   "-deviceManager=ndctl", 
                   "-endpoint=$(CSI_ENDPOINT)",
                   "-nodeid=$(KUBE_NODE_NAME)",
+                  "-statePath=/csi/volumes",
```
pohly: This isn't a suitable location. The `plugin-socket-dir` should be used for interaction with kubelet and sidecars, not as private data store for the driver. The data then ends up in /var/lib/kubelet, which is a directory that might get wiped when reinstalling the cluster.

As PMEM survives a cluster reinstallation (unless wiped, too, see blow), the persistent data also must survive.

I suggest to use `/var/lib/pmem-csi.intel.com` as state directory.

Cluster administrators will need to know about this. Please add documentation. It needs to explain that a) we have persistent state outside of the PMEM itself b) where that is c) how to reset the node (re-initialize PMEM, clean the state directory.

avalluri: > I suggest to use `/var/lib/pmem-csi.intel.com` as state directory.

Who is supposed to create this directory on node host? Do we expect that the admin would create this on every node before deploying the driver, or it must be created by the driver itself?. If its driver, then our daemonset has to mount '/var/lib' to the driver container.

pkg/pmem-state/pmme-state.go
avalluri: Now file writes are synchronous so that no writes to cache. What are other things we should do to make crash-resilient?

deploy/kubernetes-1.13/pmem-csi-direct.yaml
pohly: I would say the driver. It already runs as a privileged pod, so whether it also has access to `/var/lib` won't make a big difference.

pkg/pmem-state/pmme-state.go
pohly: Here's a good article: https://lwn.net/Articles/667788/

In a nutshell, fsync the file content and fsync the parent directory.

Your crash recovery also needs to be resilient against incomplete files: validate that the file you are reading is complete, remove it if it isn't.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
pohly: Actually, we can limit access to just the dir inside `/var/lib` with:
```
      volumes:
        - hostPath:
            path: /var/lib/pmem-csi.intel.com
            type: DirectoryOrCreate
          name: pmem-state-dir
```

avalluri: > e: DirectoryOrCreate
>           name: pmem-state-dir

Thanks, I will give it a try.

pkg/pmem-state/pmme-state.go
avalluri: > Your crash recovery also needs to be resilient against incomplete files: validate that the file you are reading is complete, remove it if it isn't.

As we are using json decoder, I believe this(incomplete file content) should be handled by json package?

pohly: Probably you'll get an error because the file isn't valid JSON. But your code then has to have error handling for the broken file.

It's worth writing a unit test for this.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
avalluri: Done

DEVELOPMENT.md
avalluri: Done. Now defaults to /var/lib/<drivername>

pkg/pmem-state/pmme-state.go
avalluri: Done. Implemented concurrency. Added tests to test this.

pkg/pmem-state/pmme-state.go
avalluri: Done

pkg/pmem-state/pmme-state.go
avalluri: Done

pkg/pmem-state/pmme-state.go
avalluri: Added unit test for this. Done

pkg/pmem-state/pmme-state.go
avalluri: Done

deploy/kubernetes-1.13/pmem-csi-direct.yaml
avalluri: > Cluster administrators will need to know about this. Please add documentation. It needs to explain that a) we have persistent state outside of the PMEM itself b) where that is c) how to reset the node (re-initialize PMEM, clean the state directory.

@pohly  What is the right place to document it. Shall we take this part of "cluster admin document", which is currently missing.

pohly: I would start a new section in the README.md for it. When it gets too large, we can still split it out later.

pohly:
@avalluri I've not reviewed the code again. Instead I worked on automated testing for this PR. It's not ready for merging, but can be used already: https://github.com/intel/pmem-CSI/pull/228

You can try that out by checking out your branch, getting my repo, and the merging the branch into yours.

Then run:
```
REPO_ROOT=`pwd` ginkgo -stream -v -focus="sanity node" ./test/e2e
```

The good news: persisting the node state seems to work, at least for the simple case with just a single volume.
The bad news: trying to publish that volume after a reboot fails:
```
• Failure [63.682 seconds]
sanity
/nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:44
  node
  /nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:136
    can mount again after reboot [It]
    /nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:204

    Unexpected error:
        <*status.statusError | 0xc0006a0d70>: {
            Code: 3,
            Message: "File system with different type exists",
            Details: nil,
            XXX_NoUnkeyedLiteral: {},
            XXX_unrecognized: nil,
            XXX_sizecache: 0,
        }
        rpc error: code = InvalidArgument desc = File system with different type exists
    occurred
```

That's for the `publishVolume`. As the code stands right now, it calls `unpublishVolume` first, and that also fails although it should simply do nothing because `NodeUnpublishVolume` needs to be idempotent.

```
• Failure [63.619 seconds]
sanity
/nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:44
  node
  /nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:136
    can mount again after reboot [It]
    /nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:204

    Unexpected error:
        <*status.statusError | 0xc000dbcc30>: {
            Code: 13,
            Message: "Unmount failed: exit status 32\nUnmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.LBNldK/target\nOutput: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.LBNldK/target: no mount point specified.\n\n",
            Details: nil,
            XXX_NoUnkeyedLiteral: {},
            XXX_unrecognized: nil,
            XXX_sizecache: 0,
        }
        rpc error: code = Internal desc = Unmount failed: exit status 32
        Unmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.LBNldK/target
        Output: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.LBNldK/target: no mount point specified.
        
        
    occurred

    /nvme/gopath/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:401
```

Can you check whether the tests are valid and look into these failures?

avalluri:
```
After rebasing my change over #228, I tried running sanity tests, the failures are bit different: 
  /home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:127

• Failure in Spec Setup (BeforeEach) [180.052 seconds]
sanity
/home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:44
  node [BeforeEach]
  /home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:136
    can mount again after reboot
    /home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:204

    Timed out after 180.000s.
    pmem-node-testing service running
    Expected
        <bool>: false
    to be true

    /home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:245
------------------------------
SSSSSSSSSSSS
------------------------------
```

pohly:
> pmem-node-testing service running

Check the state of the `pmem-node-testing` service. It should have an endpoint and a cluster port.

pohly:
Remember that you need to re-deploy PMEM-CSI after switching branches.

avalluri:
Now tests are passing, there were two reasons for failure:
1) In LVM mode after a node reboot, the driver was not able to access the lvm devices created in the last run. Now fixed it by explicitly mounting the `/dev` volume under driver container.
2) The tests are not passing `FsType` in NodeServer.StageVolume() request, where the driver was not properly handling the missing of "fsType" parameter. Now it is fixed.

With those two changes, the end-end sanity tests are passing.

@pohly, In "`can mount again after reboot`" test, unpublishing of volumes after node reboot is not required, as those mounts do not exist.

pohly:
> In "can mount again after reboot" test, unpublishing of volumes after node reboot is not required, as those mounts do not exist.

True, that extra unpublish is redundant. Doing the call despite that checks that the driver correctly handles the situation, which in this case means "does nothing".

I re-ran the updated tests from https://github.com/pohly/pmem-CSI/tree/disruptive-testing but don't get them to pass cleanly.

First, persisting the file in face of a sudden power loss seems to fail again. Looks like whatever flushing we do now is no longer sufficient. I get further with this patch which makes testing less demanding:
```
diff --git a/test/e2e/storage/sanity.go b/test/e2e/storage/sanity.go
index f51a9045..460cf9b3 100644
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -521,6 +521,8 @@ func restartNode(cs clientset.Interface, nodeID string) {
        By(fmt.Sprintf("shutting down node %s", nodeID))
        shutdown := exec.Command(ssh)
        shutdown.Stdin = bytes.NewBufferString(`sudo sh -c 'echo 1 > /proc/sys/kernel/sysrq'
+sudo sh -c 'echo s > /proc/sysrq-trigger'
+sudo sync
 sudo sh -c 'echo b > /proc/sysrq-trigger'`)
        out, _ = shutdown.CombinedOutput()
 
```

That makes "stores state across reboots for single volume" pass.

But "can mount again after reboot" fails in the first (redundant) NodeUnpublishVolume:
```
    node unpublish volume
    Unexpected error:
        <*status.statusError | 0xc000563310>: {
            Code: 13,
            Message: "Unmount failed: exit status 32\nUnmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target\nOutput: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target: not mounted.\n\n",
            Details: nil,
            XXX_NoUnkeyedLiteral: {},
            XXX_unrecognized: nil,
            XXX_sizecache: 0,
        }
        rpc error: code = Internal desc = Unmount failed: exit status 32
        Unmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target
        Output: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target: not mounted.
        
        
    occurred
```

I works when commenting out that `unpublishVolume` call.

There's one more error scenario: when the node that gets rebooted has the controller pod (happens randomly), then the PMEM-CSI registry also restarts and comes up with an incomplete view of the nodes in the cluster. Only the restarted node service on that node registers itself, the others don't. That breaks the new "supports cache volumes" test and probably also other tests.

The current error handling design doesn't cover this situation, does it? This occurred to me already two weeks ago and I wanted to write an explicit test for it to demonstrate the problem before discussing it, there just wasn't time earlier.

avalluri:
> There's one more error scenario: when the node that gets rebooted has the controller pod (happens randomly), then the PMEM-CSI registry also restarts and comes up with an incomplete view of the nodes in the cluster. Only the restarted node service on that node registers itself, the others don't. That breaks the new "supports cache volumes" test and probably also other tests.

This shouldn't happen in theory,  06c0d7ab0645f24c0cbc91d33adeb0547863c4f8 is supposed to handle this case of controller restart. Nodes are supposed to re-register themselves whenever controller restart.

avalluri:
> But "can mount again after reboot" fails in the first (redundant) NodeUnpublishVolume:
> 
> ```
>     node unpublish volume
>     Unexpected error:
>         <*status.statusError | 0xc000563310>: {
>             Code: 13,
>             Message: "Unmount failed: exit status 32\nUnmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target\nOutput: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target: not mounted.\n\n",
>             Details: nil,
>             XXX_NoUnkeyedLiteral: {},
>             XXX_unrecognized: nil,
>             XXX_sizecache: 0,
>         }
>         rpc error: code = Internal desc = Unmount failed: exit status 32
>         Unmounting arguments: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target
>         Output: umount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.APJDGc/target: not mounted.
>         
>         
>     occurred
> ```
> 
> I works when commenting out that `unpublishVolume` call.

Isn't it expected failure as the mount point is not existing after reboot?

avalluri:
> True, that extra unpublish is redundant. Doing the call despite that checks that the driver correctly handles the situation, which in this case means "does nothing".

Why is it expected to "do nothing", it tries to unmount the given path as it does in normal unpublish case, Isn't it expected behavior?

pohly:
>> I works when commenting out that `unpublishVolume` call.
>
> Isn't it expected failure as the mount point is not existing after
> reboot?

NodeUnpublishVolume is supposed to be idempotent, under all
circumstances. If the CO looses track whether the volume is published,
it can call NodeUnpublishVolume to ensure that it really is unpublished,
and it must not get an error when the volume was already unpublished.

That means the implementation must check whether some real work still
needs to be done and if not, return without an error.

pohly:
> This shouldn't happen in theory,
> 06c0d7ab0645f24c0cbc91d33adeb0547863c4f8 is supposed to handle this
> case of controller restart. Nodes are supposed to re-register
> themselves whenever controller restart.

Even if gRPC detects the connection loss (which isn't as easy as it
sounds, see below), there's always going to be a time frame where the
controller is up again and doesn't yet have a complete view of the
cluster.

For the cache volume test case the result was that either all replicas
of the cache were allocated on the same node or only a single replica
was allocated despite the request to create three (haven't
checked). What does the code do when the cache size is higher than the
number of nodes? Either way, both isn't what the user wanted.

Regarding connection loss: in this case, the machine with the controller
dies, which means that the TCP connection to it doesn't get closed. The
TCP stack detects such broken connections only after a very long timeout
(>10 minutes). The solution is to use gRPC ping, see
https://github.com/kubernetes-csi/csi-test/commit/6b8830bf5959a1c51c6e98fe514b22818b51eeeb

But that just minimizes the time window where the controller has
incomplete information, it doesn't eliminate the problem entirely.  I
think we can do better than that by persisting also the list of node
endpoints in the controller.

pohly:
> This shouldn't happen in theory, 06c0d7a is supposed to handle this case of controller restart. Nodes are 
> supposed to re-register themselves whenever controller restart.

A new test in my disruptive testing branch (https://github.com/pohly/pmem-CSI/commit/2539c91e1d1701c0dc22f52edfc2ad38e19ff416) shows that they don't do that within in a minute and probably take a lot longer.

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -154,13 +218,15 @@ func (cs *nodeControllerServer) CreateVolume(ctx context.Context, req *csi.Creat
...
+	resp = &csi.CreateVolumeResponse{
 		Volume: &csi.Volume{
 			VolumeId:           vol.ID,
 			CapacityBytes:      vol.Size,
 			AccessibleTopology: topology,
 		},
-	}, nil
+	}
+
+	return resp, nil
```
pohly: nit (i.e. don't change it unless there is some other reason to touch the PR): this local `resp` variable seems unnecessary. The previous code without it should still work fine, shouldn't it?

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -230,12 +226,51 @@ func (pmemd *pmemDriver) registerNodeController() error {
...
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+
+	connectionLost := false
+
+	for {
+		s := conn.GetState()
+		if s == connectivity.Ready {
+			if connectionLost {
+				glog.Info("ReConnected.")
```
pohly: nit: ReConnected -> Reconnected

test/e2e/storage/sanity.go
```diff
@@ -504,25 +521,30 @@ func unpublishVolume(s csi.ControllerClient, c csi.NodeClient, sc *sanity.Sanity
...
 }
 
-var unreachable = v1.Taint{Key: "node.kubernetes.io/unreachable", Effect: "NoSchedule"}
+//var unreachable = v1.Taint{Key: "node.kubernetes.io/unreachable", Effect: "NoSchedule"}
 
 // restartNode works only for one of the nodes in the QEMU virtual cluster.
 // It does a hard poweroff via SysRq and relies on Docker to restart the
 // "failed" node.
-func restartNode(cs clientset.Interface, nodeID string) {
+func restartNode(cs clientset.Interface, nodeID string, sc *sanity.SanityContext) {
```
pohly: nit: if this is not needed, it can be removed entirely.

test/test.make
```diff
@@ -110,7 +110,7 @@ space:= $(empty) $(empty)
 RUN_E2E = KUBECONFIG=`pwd`/_work/$(CLUSTER)/kube.config \
 	REPO_ROOT=`pwd` \
 	CLUSTER=$(CLUSTER) \
-	go test -count=1 -timeout 0 -v ./test/e2e -ginkgo.skip='$(subst $(space),|,$(TEST_E22_SKIP))'
+	go test -count=1 -parallel 1 -timeout 0 -v ./test/e2e -ginkgo.skip='$(subst $(space),|,$(TEST_E22_SKIP))'
```
pohly: It doesn't hurt, but was it needed?

We only have a single Test* function inside the E2E binary, so it shouldn't matter, because "go test" runs those functions in parallel.

A commit message with a "why" explanation would have helped...

okartau:
I repeated 10 rounds of make test_e2e on 4xVM cluster, and I see 2 success/ 8 failures.
If other testing shows good, does it mean there is something wrong with my setup?
I test as I used to do, using 'make start' to set up the cluster and no changes to setup (i.e. LVM mode gets used)
I fix CLEAR version to 29820 per knowledge this is the working setup ATM (is this correct?)
On a positive side, there are no leftover LVolumes after 10 rounds, which is good news for #292.
Here is errors grep from output files (rounds 02 and 03 were OK):
```
$ grep '\[Fail' out??
out01:[Fail] sanity pmem csi [It] capacity is restored after controller restart 
out04:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
out05:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
out06:[Fail] sanity pmem csi cluster [It] supports cache volumes 
out06:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
out07:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
out07:[Fail] sanity pmem csi cluster [It] supports cache volumes 
out08:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
out08:[Fail] sanity pmem csi cluster [It] supports cache volumes 
out09:[Fail] sanity pmem csi [It] capacity is restored after controller restart 
out10:[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
```

pohly:

> I repeated 10 rounds of make test_e2e on 4xVM cluster, and I see 2 success/ 8 failures.
> If other testing shows good, does it mean there is something wrong
> with my setup?

My guess is that the increased 2min (?) timeout isn't enough on real
hardware, which probably takes longer to reboot.

But to verify that you need to look at the complete failure output, not
just which test failed.

> I test as I used to do, using 'make start' to set up the cluster and no changes to setup (i.e. LVM mode gets used)
> I fix CLEAR version to 29820 per knowledge this is the working setup
> ATM (is this correct?)

29880 is the last working version. The authoritative source of that
information is now our Jenkinsfile.

okartau:
> My guess is that the increased 2min (?) timeout isn't enough on real hardware, which probably takes longer to reboot.

That 10x series  was run on 4xVM cluster, no real-HW.
Log of failing run shows it really was 2min tmo issue:
```
    mcapacity is restored after controller restart [It]
    test/e2e/storage/sanity.go:138

    Timed out after 120.000s.
    total capacity after controller restart
    Expected
        <int64>: 96624181248
    to equal
        <int64>: 32208060416
```

bit puzzling is, where the value 96624181248 originates from.
All nodes show repeated GetCapacity as 32208060416, but Controller somehow uses value 96624181248. Is this something that is old value kept in Controller and not updating correctly ?
2 min should be plenty to restart in VM env.
My baremetal tests are still coming.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: It helps in defer func, to determine if the call was a success or not, if not we revert the state.

test/e2e/storage/sanity.go
avalluri: If I understood correctly your comment is about above change (line #542), yes we can remove

pohly:
> bit puzzling is, where the value 96624181248 originates from.
> All nodes show repeated GetCapacity as 32208060416, but Controller somehow uses value 96624181248. Is this something that is old value kept in Controller and not updating correctly ?

Yes, this looks like a genuine issue. The expected value is the capacity before the restart. The current value shouldn't grow beyond that.

@avalluri, any thoughts? Can you try to reproduce by running that one test repeatedly against a four-node VM cluster? I'll try the same.

test/test.make
avalluri: I myself was not clear if this really needed :) This change was just to make sure the test failures were not caused by running in parallel.

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: > I fix CLEAR version to 29820 per knowledge this is the working setup ATM (is this correct?)

I was testing it on 29880, but it shouldn't make much difference I guess.

avalluri:
> I fix CLEAR version to 29820 per knowledge this is the working setup ATM (is this correct?)

I was testing it on 29880, but it shouldn't make much difference I guess.

avalluri:
> > bit puzzling is, where the value 96624181248 originates from.
> > All nodes show repeated GetCapacity as 32208060416, but Controller somehow uses value 96624181248. Is this something that is old value kept in Controller and not updating correctly ?
> 
> Yes, this looks like a genuine issue. The expected value is the capacity before the restart. The current value shouldn't grow beyond that.
> 
> @avalluri, any thoughts? Can you try to reproduce by running that one test repeatedly against a four-node VM cluster? I'll try the same.

This looks like too less timeout, only one node(in our test cluster each node size is 32208060416)   reconnected after reboot.

avalluri:
> > > bit puzzling is, where the value 96624181248 originates from.
> > > All nodes show repeated GetCapacity as 32208060416, but Controller somehow uses value 96624181248. Is this something that is old value kept in Controller and not updating correctly ?
> > 
> > 
> > Yes, this looks like a genuine issue. The expected value is the capacity before the restart. The current value shouldn't grow beyond that.
> > @avalluri, any thoughts? Can you try to reproduce by running that one test repeatedly against a four-node VM cluster? I'll try the same.
> 
> This looks like too less timeout, only one node(in our test cluster each node size is 32208060416) reconnected after reboot.

My bad, it's reverse, something went wrong, by the time test start, only one is connected with controller :(

pohly:
> My bad, it's reverse, something went wrong, by the time test start, only one is connected with controller :(

That sounds like a good explanation. This is one of the problems we currently have with deploying our driver: there's no good way to be sure that everything is up-and-running.

I meant to add a sanity check before starting the test suite, but haven't gotten around to it yet.

test/e2e/storage/sanity.go
pohly: Yes.

test/test.make
pohly: Then drop the commit. Without an explanation it just looks odd.

pohly:
I need to stop now.

Before we can merge this, we need to:
- go back to 29820 as Clear Linux version, because in 29980 Docker sometimes fails to restart properly for me (https://github.com/clearlinux/distribution/issues/891)
- add a sanity check for the driver before the test suite starts

Such a sanity check needs to verify that each node with the "storage: pmem" label also has the CSI node info (i.e. driver was found by kubelet). At that point each node also has registered with the controller. @avalluri does the controller then also have the right overall capacity?

Can someone of you write such a check or do you need me to add it? I might not have time today.

avalluri:
> Such a sanity check needs to verify that each node with the "storage: pmem" label also has the CSI node info (i.e. driver was found by kubelet). At that point each node also has registered with the controller. @avalluri does the controller then also have the right overall capacity?
> 
If we make sure that every node registered with the controller before each test, Because this ensures that setup is correct(does not effect by previous node reboot test) before running the test case.

> Can someone of you write such a check or do you need me to add it? I might not have time today.

I can give it a try.

okartau:
Not guaranteed this happened by changes of this PR (as I havent tested on bare metal with other code at all yet), but recent tests of this PR on a bare metal host left some bad-looking state on host:
There are 4 device-mapper mounts with crazy sizes:
```
/dev/dm-1               60689887065540446456 60689887065540426792         0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.fIHglA/target
/dev/dm-2               38318223432365555488 38318223432365535824         0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.KCgDgp/target
/dev/dm-3               62145250087106600048 62145250087106580384         0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.ADbMBC/target
/dev/dm-4               24855614196206392712 24855614196206373048         0 101% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.EINODk/target
```

These sizes are so large that df -h uses Z as human-readable unit (zetta bytes):
```
/dev/dm-1                 53Z   53Z     0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.fIHglA/target
```

There are also 4 leftover LVs (sane sizes), leftovers means more bad news for #292:
```
# lvs
  LV                                   VG                 Attr       LSize    Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  074162ae-97ec-11e9-a8ec-82f9d0a5abac ndbus0region0fsdax -wi-ao----   24.00m                                                    
  7a12102b-97ec-11e9-a8ec-82f9d0a5abac ndbus0region0fsdax -wi-ao----   24.00m                                                    
  9087b42f-97ea-11e9-a8ec-82f9d0a5abac ndbus0region0fsdax -wi-ao----   24.00m                                                    
  d91237f1-97e8-11e9-a8ec-82f9d0a5abac ndbus0region0fsdax -wi-ao----   24.00m                                                    
```

dmesg shows that host kernel had troubles with those file systems:
```
[86868.970707] EXT4-fs error (device dm-3): ext4_map_blocks:592: inode #2: block 5: comm ls: lblock 0 mapped to illegal pblock 5 (length 1)
[86868.977260] EXT4-fs warning (device dm-3): htree_dirblock_to_tree:979: inode #2: lblock 0: comm ls: error -117 reading directory block
[86868.983087] EXT4-fs error (device dm-3) in ext4_reserve_inode_write:5903: Out of memory
[86868.985899] EXT4-fs error (device dm-4): ext4_map_blocks:592: inode #2: block 5: comm ls: lblock 0 mapped to illegal pblock 5 (length 1)
[86868.990762] EXT4-fs warning (device dm-4): htree_dirblock_to_tree:979: inode #2: lblock 0: comm ls: error -117 reading directory block
[86868.994299] EXT4-fs error (device dm-4): ext4_journal_check_start:61: Detected aborted journal
[86868.996070] EXT4-fs (dm-4): Remounting filesystem read-only
[86868.997864] EXT4-fs error (device dm-1): ext4_map_blocks:592: inode #2: block 5: comm ls: lblock 0 mapped to illegal pblock 5 (length 1)
[86869.000870] EXT4-fs error (device dm-1): ext4_readdir:179: inode #2: comm ls: path /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.fIHglA/target: directory contains a hole at offset 0
[86869.003224] EXT4-fs error (device dm-1): ext4_journal_check_start:61: Detected aborted journal
[86869.004386] EXT4-fs (dm-1): Remounting filesystem read-only
```

What is causing this? Is pod restarting creating invalid contents?
Just random errors would look like garbage, but here all 4 file systems show unrealistic sizes that are in similar scale, so this looks like systematic.
Difference in VM and host interpretations of same data? hard to believe in ext4 case.

pohly:
> Not guaranteed this happened by changes of this PR (as I havent tested on bare metal with other code at all yet), but recent tests of this PR on a bare metal host left some bad-looking state on host

Perhaps the abrupt poweroff is breaking LVM and/or PMEM content.

@avalluri can you work with @okartau on debugging this?

If it turns out to be caused by the disruptive testing, then IMHO we can ignore it for 1.0 because it isn't very likely to happen in real data centers.

pkg/registryserver/registryserver.go
```diff
@@ -74,11 +86,17 @@ func (rs *RegistryServer) RegisterController(ctx context.Context, req *registry.
...
+	node := &NodeInfo{
 		NodeID:   req.NodeId,
 		Endpoint: req.Endpoint,
 	}
 
+	rs.nodeClients[req.NodeId] = node
+
+	for l := range rs.listeners {
+		l.OnNodeAdded(ctx, node)
+	}
```
pohly: As discussed on chat, this code deadlocks:
`RegisterController()->OnNodeAdded()->ConnectToNodeController()->GetNodeController()`

If `rs.listeners` doesn't need to be locked because it is only set once during initialization *and* the callback code itself is resilient against race conditions, then you can do:
```
rs.mutex.Lock()
defer func() {
    rs.mutex.Unlock()
    for l := range rs.listeners {
		l.OnNodeAdded(ctx, node)
    }
} ()
```

avalluri: I would prefer to lock and unlock only where needed, instead of locking on beginning of the function:

```
rs.mutex.Lock()
rs.nodeClients[req.NodeId] = node
rs.mutex.Unlock()
```

pkg/registryserver/registryserver.go
```diff
@@ -74,28 +85,41 @@ func (rs *RegistryServer) RegisterController(ctx context.Context, req *registry.
...
 
-	if _, ok := rs.nodeClients[req.NodeId]; !ok {
+	rs.mutex.Lock()
+	node, ok := rs.nodeClients[req.NodeId]
+	rs.mutex.Unlock()
+	if !ok {
 		return nil, status.Errorf(codes.NotFound, "No entry with id '%s' found in registry", req.NodeId)
 	}
 
+	for l := range rs.listeners {
```
pohly: `UnregisterController` is now no longer atomic: if you had two competing `UnregisterController` calls for the same node, both might get pass the "ok" check and then the rest of the code gets executed twice whereas before the second call would have gotten an error.

It's a theoretic problem, I'm just pointing out the drawback of locking just subsections of the function.

pkg/registryserver/registryserver.go
pohly: I don't like that because it is a less effective protection against race conditions - see my other comment about `UnregisterController`.

But if you prefer this, then so be it.

pkg/registryserver/registryserver.go
pohly: Another race is a RegisterController that comes in after UnregisterController: if the RegisterController takes the mutex in the middle of UnregisterController, then the controller gets removed by the rest of UnregisterController.

okartau:
I ran my repeated e2e test on recent state (commit ddeb1a).
1st round succeeds (that makes CI looking OK) but all the following rounds fail with:
`stores state across reboots for single volume`

`same volumes as before node reboot`
In 2nd round output:
Expected
<DETAILED LIST OF 3 VOLUMES HERE>
to consist of
<DIFFERENT DETAILED LIST OF 3 VOLUMES HERE>
These lists seems to contain same volumes but in different order.
In 3rd round output: similar mismatch with 8 entries, and so on, amount grows.

After 11 rounds of repeated tests, there are 36 leftover Lvolumes on node-1

okartau:
In 4/11 there was also this Failure reported:
```
stress test [It]
src/github.com/intel/pmem-csi/test/e2e/storage/csi_volumes.go:147
Unexpected error:
    <*errors.errorString | 0xc0011b82b0>: {
        s: "Gave up after waiting 15m0s for pod \"pvc-volume-tester-reader-worker-1-volume-23-6zmj4\" to be \"success or failure\"",
    }
    Gave up after waiting 15m0s for pod "pvc-volume-tester-reader-worker-1-volume-23-6zmj4" to be "success or failure"
```

avalluri:
> I ran my repeated e2e test on recent state (commit ddeb1a).
> 1st round succeeds (that makes CI looking OK) but all the following rounds fail with:
> `stores state across reboots for single volume`
> 
> `same volumes as before node reboot`
> In 2nd round output:
> Expected
> <DETAILED LIST OF 3 VOLUMES HERE>
> to consist of
> <DIFFERENT DETAILED LIST OF 3 VOLUMES HERE>
> These lists seems to contain same volumes but in different order.
> In 3rd round output: similar mismatch with 8 entries, and so on, amount grows.
> 
> After 11 rounds of repeated tests, there are 36 leftover Lvolumes on node-1

These leftover volumes might be a result of failed stress test.

avalluri:
I am seeing this failure consecutively 2nd time:

```
• Failure in Spec Teardown (AfterEach) [156.460 seconds]
sanity
/home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:243
  pmem csi
  /home/avalluri/work/go/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
    capacity is restored after controller restart [AfterEach]
    /home/avalluri/work/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:138

    [rmdir /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.blnalg] in socat container, stderr:
    rmdir: '/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.blnalg': No such file or directory
    Unexpected error:
        <exec.CodeExitError>: {
            Err: {
                s: "command terminated with exit code 1",
            },
            Code: 1,
        }
        command terminated with exit code 1
    occurred
```

pkg/registryserver/registryserver.go
avalluri: @pohly  I addressed these issues https://github.com/avalluri/pmem-CSI/commit/dacc98491ffcf6fad66cc886a4a360e51c645f66 . Now the registry calls are idempotent.

pkg/pmem-csi-driver/controllerserver-master.go
```diff
@@ -75,17 +74,65 @@ func NewMasterControllerServer(rs *registryserver.RegistryServer) *masterControl
...
+	if err != nil {
+		glog.Warningf("Failed to connect to node controller at : %s on node %s: %s", node.Endpoint, node.NodeID, err.Error())
+		return
+	}
+
+	csiClient := csi.NewControllerClient(conn)
+	resp, err := csiClient.ListVolumes(ctx, &csi.ListVolumesRequest{})
+	if err != nil {
+		glog.Warningf("Failed to get volumes on node %s: %s", node.NodeID, err.Error())
+	}
```
pohly: I'm a bit concerned about the lack of error handling here. If anything goes wrong, all we get are some warnings in the controller log.

No admin is going to find them there, if they even know that they need to look and what they mean.

There's no recovery mechanism. Even if an admin had some way to fix the situation, having to do so manually is not how Kubernetes is meant to work.

Let's fix this after merging this PR. I'll file an issue.

okartau:
I improved my repeat_e2e script to list leftover LVs after each round, for each node separately, to help in working with those long test logs. In a next run, I see again many leftover LVs.
One case is this:
- NodeUnpublish : umount=OK, wipe entire/shred=OK
  but then lvremove fails:
```
Executing: lvremove -fy /dev/ndbus0region0fsdax/f40d9212-9bc2-11e9-8716-2eecce431ed5
node: Output:   Logical volume ndbus0region0fsdax/f40d9212-9bc2-11e9-8716-2eecce431ed5 contains a filesystem in use.
node: rpc error: code = Internal desc = Failed to delete volume: exit status 5
controller: Failed to delete volume f40d9212-9bc2-11e9-8716-2eecce431ed5 on pmem-csi-clear-govm-worker3: rpc error: code = Internal desc = Failed to delete volume: exit status 5
controller: DeleteVolume: volume f40d9212-9bc2-11e9-8716-2eecce431ed5 deleted
```
The last line BTW indicates another getting-out-of-sync issue we have, controller marks Volume as deleted in own books, even though deletion failed on node.

About deletion failure:
LVol deletion failed when steps get done in quick succession: umount, shred, lvremove.
But later after e2e test completed, I can delete same volumes on CLI just fine.
We seem to have "LV state update delay" class of problems. One instance is "device not found" after lvcreate, where additional check-and-wait loop was added recently in Volume creation path.
There are also other instances, the 97th volume (our of 100) issue being one.
We could probably get deletion also working by adding another delay, but all of those
are workarounds.

pohly:
> Executing: lvremove -fy /dev/ndbus0region0fsdax/f40d9212-9bc2-11e9-8716-2eecce431ed5
> node: Output:   Logical volume ndbus0region0fsdax/f40d9212-9bc2-11e9-8716-2eecce431ed5 contains a filesystem in use.
> node: rpc error: code = Internal desc = Failed to delete volume: exit status 5
> controller: Failed to delete volume f40d9212-9bc2-11e9-8716-2eecce431ed5 on pmem-csi-clear-govm-worker3: rpc error: code = Internal desc = Failed to delete volume: exit status 5
> controller: DeleteVolume: volume f40d9212-9bc2-11e9-8716-2eecce431ed5 deleted
> ```
> The last line BTW indicates another getting-out-of-sync issue we have,
> controller marks Volume as deleted in own books, even though deletion
> failed on node.

Can you file a bug for that?

> About deletion failure:
> LVol deletion failed when steps get done in quick succession: umount, shred, lvremove.
> But later after e2e test completed, I can delete same volumes on CLI just fine.
> We seem to have "LV state update delay" class of problems. One instance is "device not found" after lvcreate, where additional check-and-wait loop was added recently in Volume creation path.
> There are also other instances, the 97th volume (our of 100) issue being one.
> We could probably get deletion also working by adding another delay, but all of those
> are workarounds.

Retry-loops for non-final errors are fine in Kubernetes. "Eventually"
things work. So let's add those retries. The deadline for them is the
gRPC deadline, available in the "ctx" parameter.

In order to avoid premature timeouts, we may have to tweak the
external-provisioner gRPC timeouts.

However, the issue above also needs to be addressed, because even with a
longer timeout we may still run into it eventually.

pkg/registryserver/registryserver.go
```diff
@@ -13,12 +12,22 @@ import (
...
 	"k8s.io/klog/glog"
+	"k8s.io/utils/keymutex"
 )
 
+type RegistryListener interface {
+	// OnNodeAdded is called by RegistryServer whenever a node controller registered.
+	OnNodeAdded(ctx context.Context, node *NodeInfo)
+	// OnNodeDeleted is called by RegistryServer whenever a node controller unregistered.
+	OnNodeDeleted(ctx context.Context, node *NodeInfo)
+}
```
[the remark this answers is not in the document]
avalluri: @pohly I added a point about the state of registry's in-memory data by the time of calling `RegistryListener` API  call. And mutex is RegistryServer internal thing which shouldn't affect the `RegistryListener` API

pkg/registryserver/registryserver.go
```diff
@@ -13,12 +12,28 @@ import (
...
+// RegistryListener is an registry server listener interface
+// All the callbacks are called once after updating the in-memory
+// registry data.
+type RegistryListener interface {
+	// OnNodeAdded is called by RegistryServer whenever a new node controller is registered
+	// or node controller updated its endpoint.
+	OnNodeAdded(ctx context.Context, node *NodeInfo)
+	// OnNodeDeleted is called by RegistryServer whenever a node controller unregistered.
+	// Callback implementations has to note that by the time this method is called,
+	// the NodeInfo for that node have already removed from in-memroy registry.
```
pohly: That's redundant, the same is said above for RegistryListener interface. There's also a typo (`in-memroy`). Do you want to fix this or get it merged as-is?

pkg/registryserver/registryserver.go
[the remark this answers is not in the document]
pohly: > And mutex is RegistryServer internal thing which shouldn't affect the RegistryListener API

But it did in a previous revision of the code. Anyway, let's leave that point out.

pkg/registryserver/registryserver.go
avalluri: Fixed, typo.

pohly:
Let's merge it.

PR #197, opened 2019-03-14
Direct-mode deploy example: no Bidirectional, use /sys

okartau:
Bidirectional was accidentally copy-pasted in original
version of that file, it may potentially cause
some side effects propagating state from container to host /dev
so we should not have it there.

Pass of /sys is vital for direct mode to work. This was
tested and tuned in e2e tests which uses another deployment
file, so it was never carried over here, but it is needed.


okartau:
I added 2nd commit here.
Although these touch same file, I keep them separate, as they are actually on overlapping lines

PR #198, opened 2019-03-18
crd installation + test

pohly:
We had broken RBAC rules in one of the previous PRs and didn't notice because `cluster-driver-registrar` continued running and we didn't have tests for it's expected behavior.

This PR adds such a test and also updates the CRD installation.

### DIFF
--- a/test/e2e/storage/pmem_csi.go
+++ b/test/e2e/storage/pmem_csi.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2019 Intel Corporation.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package storage
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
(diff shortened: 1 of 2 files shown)


test/e2e/storage/pmem_csi.go
```diff
@@ -0,0 +1,31 @@
...
+	f := framework.NewDefaultFramework("pmem")
+
+	// This checks that cluster-driver-registrar added the
+	// CSIDriverInfo for pmem-csi at some point in the past. A
+	// full test must include resetting the cluster and installing
+	// pmem-csi.
+	It("has CSIDriverInfo", func() {
+		dc := f.DynamicClient
+		csiDriverGVR := schema.GroupVersionResource{Group: "csi.storage.k8s.io", Version: "v1alpha1", Resource: "csidrivers"}
+		_, err := dc.Resource(csiDriverGVR).Namespace("").Get("pmem-csi", metav1.GetOptions{})
```
avalluri: I think we should check CSIDrvier with "pmem-csi.intel.com" name ?

pohly: Yes, now that it has been renamed. I developed this test against a revision which still had "pmem-csi". I'll rebase.

This is something that a CI pre-submit check would catch, because it would run the test against the merge result with current master :-/

pohly: Done.

okartau:
deployment and e2e works on my test system

Issue #78, opened 2018-11-21
Document repository elements make steps which are not part of Makefiles


okartau:
We have diagram source(s) in docs/diagrams,
currently one file that is turned into flow diagram using `plantuml`,
should there be also generating process documented/implemented, like
via separate Makefile under docs/ perhaps, or part of top-level Makefile?
Right now, making change in diagram has to happen in off-repo style.

okartau:
also, we should add TOC generation as make step, as discussed in #199

okartau:
Not all creation steps are suitable to become make targets (like graph creation performed using a GUI program), so let's do this:
We create a chapter in DEVELOPMENT.md describing create/update steps for everything not covered by make targets.
From that list, we can later pick single items that can be turned to make steps, and handle separately.

PR #199, opened 2019-03-20
README.md: fix QEMU + Kubernetes link

pohly:
The plus sign leads to a special link target that pandoc doesn't seem
to handle correctly. To avoid that issue in the future, let's use
"QEMU and Kubernetes" as section title.

This fixes https://github.com/intel/pmem-CSI/pull/191/commits/5321c4e879a7c5f1fc5055a2fe086e03fa97cc3d#r266401518

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
         -   [Run PMEM-CSI on Kubernetes](#run-pmem-csi-on-kubernetes)
     -   [Automated testing](#automated-testing)
         -   [Unit testing and code quality](#unit-testing-and-code-quality)
-        -   [QEMU + Kubernetes](#qemu-kubernetes)
+        -   [QEMU and Kubernetes](#qemu-and-kubernetes)
         -   [Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster)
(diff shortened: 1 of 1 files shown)


okartau:
what about adding some howto-notes in DEVELOPMENT.md how we generate TOC, plus some hints, like "avoid plus sign in headers". Otherwise this becomes scattered information in our heads and in some issues/PR. Another question: is there a tool/method to verify all TOC links, or does it need manual re-clicks of everything to discover TOC generation problems. 
This topic touches same area as #78

pohly:
Do we agree one one tool for generating the TOC? If so, we can add a make target which invokes it and does the necessary mangling.

I'm fine with pandoc myself, despite the issue.

okartau:
I used pandoc because I did not see simple method to add tooling in emacs working for TOC generation. I agree we can use pandoc and have make target which documents it all, plus tells some ideas what to avoid in titles.
pandoc as standalone tool is also better suited for automated processing by make target, compared to generating in emacs buffer (although it can likely achieved with emacs side tools as well)

okartau:
I added note under #78 to also add TOC gen as make step, so we can merge this PR without idea getting lost

PR #200, opened 2019-03-22
README.md: mention need to set up certificates in Setup section

### LINKED ISSUE #194: README.md: mention certificate installation in setup section
https://github.com/intel/pmem-csi#setup must mention that certificates must be created and installed.

okartau:
If a user does not use 'make start' method, driver deployment
will fail by steps because steps don't mention certs setup.

Resolves #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,30 @@ machines, see the ["QEMU and Kubernetes"](#qemu-and-kubernetes) section below.
     $ kubectl label node <your node> storage=pmem
 ```
 
+- **Set up certificates**
+
+Certificates are required as explained in [Security](#security).
+If you are not using the test cluster described in
+[Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster)
+where certificates are created automatically, you must set up certificates manually.
+This can be done by running the `./test/setup-ca-kubernetes.sh` script for your cluster.
+This script requires "cfssl" tools which can be downloaded.
+These are the steps for manual set-up of certificates:
+
+- Download cfssl tools
+
+```sh
+   $ curl -L https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o _work/bin/cfssl --create-dirs
+   $ curl -L https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64 -o _work/bin/cfssljson --create-dirs
+   $ chmod a+x _work/bin/cfssl _work/bin/cfssljson
+```
+
+- Run certificates set-up script
+
+```sh
+   $ KUBCONFIG="<<your cluster kubeconfig path>> PATH="$PATH:_work/bin" ./test/setup-ca-kubernetes.sh
+```
+
 - **Deploy the driver to Kubernetes using DeviceMode:LVM**
 
 ```sh


README.md
```diff
@@ -428,6 +428,30 @@ machines, see the ["QEMU and Kubernetes"](#qemu-and-kubernetes) section below.
...
+This script requires "cfssl" tools which can be downloaded.
+These are steps for manual set-up of certificates:
+
+- Download cfssl tools:
+```sh
+   $ curl -L https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o _work/bin/cfssl --create-dirs
+   $ curl -L https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64 -o _work/bin/cfssljson --create-dirs
+```
+
+or,  can run above via `make _work/.setupcfssl-stamp`
```
pohly: This "or, " looks odd. How about making it a separate sentence?

> Alternatively, one can run these commands via `make _work/.setupcfssl-stamp`.

Careful with the formatting. It'll look best if indendent under the "Download bullet point".

README.md
```diff
@@ -428,6 +428,30 @@ machines, see the ["QEMU and Kubernetes"](#qemu-and-kubernetes) section below.
...
+
+- run certificates set-up script
```
pohly: For consistency, this should be "Run" (as in "Download" above).

README.md
okartau: Should we give "can run these via make target" note before the curl commands example? If user goes step-by-step, it's too late to show  (somewhat easier) alternative after user already performed those curl commands.

okartau:
should we make bit stronger point about certificates setup, as deployment will fail without certs.
Now we say "... you can set up certificates by..."
I propose we use stronger form, like:
"... you need to set up certificates by..."
or even more:
"... you must set up certificates by..."

pohly:
> should we make bit stronger point about certificates setup, as deployment will fail without certs.
>...
> "... you must set up certificates by..."

Good point. I prefer "must".

okartau:
we describe two curl cmds, but in Makefile there are two curl cmds and chmod to make the binary executable. curl leaves it non-executable, so those steps seem incomplete. We need to show chmod cmd as well in README, isn't it.

okartau:
pushed new state. I addressed latest comments, added 3rd chmod command, moved note "Alternatively" above the commands,  is that also what you meant with your comment @pohly ?

avalluri:
> > should we make bit stronger point about certificates setup, as deployment will fail without certs.
> > ...
> > "... you must set up certificates by..."
> 
> Good point. I prefer "must".

my two cents, must here is not misleading, that, one must use only this script to generate certificates?

okartau:
> 
> my two cents, must here is not misleading, that, one must use only this script to generate certificates?

It is misleading only if user does not read full sentence. Start of sentence limits the scope to cases without test cluster.

README.md
avalluri: > It is misleading only if user does not read full sentence. Start of sentence limits the scope to cases without test cluster.

`If you are not using the test cluster described in [Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster) where certificates are created automatically, you must set up certificates by running the `./test/setup-ca-kubernetes.sh` script manually in your cluster.`

This sounds me like if its my own custom cluster, i must use this scritp. Isn't it?.

README.md
```diff
@@ -428,6 +428,31 @@ machines, see the ["QEMU and Kubernetes"](#qemu-and-kubernetes) section below.
...
+If you are not using the test cluster described in
+[Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster)
+where certificates are created automatically, you must set up certificates by
+running the `./test/setup-ca-kubernetes.sh` script manually in your cluster.
+This script requires "cfssl" tools which can be downloaded.
+These are steps for manual set-up of certificates:
+
+- Download cfssl tools
+
+  Alternatively, one can run three commands below via make _work/.setupcfssl-stamp.
```
pohly: Now the "alternatively" makes no sense anymore. Let's remove the sentence completely. We should just show the recommended way of doing this, mentioning alternatives just makes the reader wonder which one they should pick.

The rationale behind showing the individual commands is:
- no need for make
- users can see what they need to do if they don't have curl or want to use something else

README.md
pohly: Let's  break it down into smaller sentences.

> If you are not using the test cluster described in [Starting and stopping a test cluster](#starting-and-stopping-a-test-cluster) where certificates are created automatically, you must set up certificates manually. This can be done by running the ./test/setup-ca-kubernetes.sh script for your cluster.

README.md
okartau: > Let's remove the sentence completely.

removed it.

okartau:
I found out I had not addressed "let's break sentence up" proposal yet. Done now.

PR #201, opened 2019-03-22
WIP: unit testing minor cleanup and improvement

okartau:
Showing tests coverage is overall useful, let's have it.
Also, use TEST_CMD in two invocations instead of just one.

### DIFF
--- a/pkg/pmem-csi-driver/registryserver_test.go
+++ b/pkg/pmem-csi-driver/registryserver_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/container-storage-interface/spec/lib/go/csi"
 	pmemcsidriver "github.com/intel/pmem-csi/pkg/pmem-csi-driver"
 	pmemgrpc "github.com/intel/pmem-csi/pkg/pmem-grpc"
 	registry "github.com/intel/pmem-csi/pkg/pmem-registry"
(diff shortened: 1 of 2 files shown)


okartau:
I think showing coverage is overall useful, it indicates how well (or sometimes, how badly) the testing code covers all code. For example, right now we get with added -cover option, unit tests:
```
ok      github.com/intel/pmem-csi/pkg/pmem-csi-driver   coverage: 5.8% of statements
```

Same option also applicable in e2e test (this PR adds it in both), where we now get:
```
ok       github.com/intel/pmem-csi/test/e2e      1590.494s       coverage: 62.5% of statements
```

With coverage analysis, it's also easy to get coverage shown line-by-line (html-formatted output) so it's easy to see the areas which remain uncovered. But that part is not in this PR.

pohly:
> Same option also applicable in e2e test (this PR adds it in both), where we now get:

This most likely is the test coverage of the code in the *test suite* itself (i.e. all code outside of `_test.go` which gets linked into `e2e.test`), which isn't useful. What we would need is some way to collect coverage information of the PMEM-CSI binaries running in the cluster and then combine that with different unit test runs. I don't know how to do that.

okartau:
> This most likely is the test coverage of the code in the _test suite_ itself

good point, means we should not add -cover option in e2e case, I will take it away

okartau:
As I found some dead-looking code and leftover OIM reference in comment, I added 2nd commit and rename this PR to be more generic unit-test improvement

okartau:
I cherry-picked two commits from here into #246, so we can close this now

PR #202, opened 2019-03-24
README: simplify text describing node labeling

okartau:
Previous text repeated almost same thing twice, and telling
about labeling need _after_ example command is misleading,
giving wrong idea that something more needs to be done.
Change node name in labeling example to be template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,13 +422,11 @@ with at least one node that has persistent memory device(s). For development or
 testing, it is also possible to use a cluster that runs on QEMU virtual
 machines, see the ["QEMU and Kubernetes"](#qemu-and-kubernetes) section below.
 
-- **Label the cluster nodes that have persistent memory support**
+- **Label the cluster nodes that provide persistent memory device(s)**
 
 ```sh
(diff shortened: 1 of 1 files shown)


okartau:
Q: should we also change the node name in command example, it is now pmem-csi-4.
It's not so obvious that this name came as copied sample command being a real node name in some running cluster. I can imagine someone copying this cmd from README and only after failure realizing that node name needs to be substituted. Usually in such case, we try to indicate more clearly that command example contains sample name, like  <YOUR_NODE>

avalluri:
> Q: should we also change the node name in command example, it is now pmem-csi-4.
> It's not so obvious that this name came as copied sample command being a real node name in some running cluster. I can imagine someone copying this cmd from README and only after failure realizing that node name needs to be substituted. Usually in such case, we try to indicate more clearly that command example contains sample name, like <YOUR_NODE>

Yes, I agree.

okartau:
force-pushed state where we use <your node> instead of pmem-csi-4

okartau:
added 2nd commit fixing another cosmetic issue with example deployment `df` output

PR #203, opened 2019-03-25
Deploy examples: Clean and streamline LVM vs Direct files

okartau:
Reduce diff between LVM and Direct variants for clarity.
Use terminationPath clause and deviceManager option in node part
only as these are relevant only in mode=node case.
Give deviceManager option in both deviceMode files for clarity.


okartau:
I plan to verify those changes in running cluster, therefore added WIP in title.

okartau:
deployment works in both LVM and Direct device modes, removing WIP

PR #206, opened 2019-03-26
WIP: Add mutex lock around device-oriented operations

### LINKED ISSUE #205: Attempt to create two volumes crashes driver
There is no mutex protecting CreateVolume in Node, leading to crash if deployment tries to create two volumes instead of one (one volume requested in current deploy example).
I experimented with 2 volumes in my-csi-app. I doubled volumes in storageclass and pvc, and mounts in app. The result is that two CreateVolume request are sent to same Node in parallel, and two threads run nodeControllerServer.CreateVolume at same time, and one of threads will crash in
```
signal SIGSEGV: segmentation violation code=0x1 addr=0x3c pc=0x7fcaca6cb770]

runtime stack:
runtime.throw(0x98a841, 0x2a)
        /usr/local/go/src/runtime/panic.go:608 +0x72
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:374 +0x2f2

goroutine 48 [syscall]:
runtime.cgocall(0x84c050, 0xc0000e1578, 0x0)
        /usr/local/go/src/runtime/cgocall.go:128 +0x5e fp=0xc0000e1548 sp=0xc0000e1510 pc=0x40816e
github.com/intel/pmem-csi/pkg/ndctl._Cfunc_ndctl_namespace_get_id(0x0, 0x0)
        _cgo_gotypes.go:718 +0x49 fp=0xc0000e1578 sp=0xc0000e1548 pc=0x811b09
github.com/intel/pmem-csi/pkg/ndctl.(*Namespace).ID.func1(0x0, 0x0)
        /go/src/github.com/intel/pmem-csi/pkg/ndctl/namespace.go:84 +0x56 fp=0xc0000e15b0 sp=0xc0000e1578 pc=0x81ace6
github.com/intel/pmem-csi/pkg/ndctl.(*Namespace).ID(0x0, 0xc000409d20)
        /go/src/github.com/intel/pmem-csi/pkg/ndctl/namespace.go:84 +0x2b fp=0xc0000e15d0 sp=0xc0000e15b0 pc=0x814b4b
github.com/intel/pmem-csi/pkg/ndctl.(*Namespace).MarshalJSON(0x0, 0xc00018a3f0, 0x24, 0x280000000, 0x0, 0x40000000)
        /go/src/github.com/intel/pmem-csi/pkg/ndctl/namespace.go:334 +0x40 fp=0xc0000e16a8 sp=0xc0000e15d0 pc=0x816050
github.com/intel/pmem-csi/pkg/pmem-device-manager.(*pmemNdctl).CreateDevice(0xc00000e0c8, 0xc00018a3f0, 0x24, 0x200000000, 0x973939, 0x5, 0x0, 0x0)
        /go/src/github.com/intel/pmem-csi/pkg/pmem-device-manager/pmd-ndctl.go:92 +0x160 fp=0xc0000e1820 sp=0xc0000e16a8 pc=0x836640
github.com/intel/pmem-csi/pkg/pmem-csi-driver.(*nodeControllerServer).CreateVolume(0xc00013cd20, 0xa00580, 0xc00019f320, 0xc0001d49a0, 0xc00013cd20, 0x1, 0x1)
        /go/src/github.com/intel/pmem-csi/pkg/pmem-csi-driver/controllerserver-node.go:134 +0x2db fp=0xc0000e19f8 sp=0xc0000e1820 pc=0x84017b
```

okartau:
When two volumes are requested at once, Creation typically
causes two parallel Creation calls to run, and without
mutex locking we will run into troubles.

Fixes #205

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -78,6 +78,8 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error 
 	if nsmode != string(ndctl.FsdaxMode) && nsmode != string(ndctl.SectorMode) {
 		return fmt.Errorf("Unknown nsmode(%v)", nsmode)
 	}
+	devicemutex.Lock()
+	defer devicemutex.Unlock()
 	// Check that such name does not exist. In certain error states, for example when
 	// namespace creation works but device zeroing fails (missing /dev/pmemX.Y in container),
 	// this function is asked to create new devices repeatedly, forcing running out of space.
@@ -129,6 +131,8 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error 
 }
 
 func (lvm *pmemLvm) DeleteDevice(name string, flush bool) error {
+	devicemutex.Lock()
+	defer devicemutex.Unlock()
 	device, err := lvm.GetDevice(name)
 	if err != nil {
 		return err
@@ -142,6 +146,8 @@ func (lvm *pmemLvm) DeleteDevice(name string, flush bool) error {
 }
 
 func (lvm *pmemLvm) FlushDeviceData(name string) error {
+	devicemutex.Lock()
+	defer devicemutex.Unlock()
 	device, err := lvm.GetDevice(name)
 	if err != nil {
 		return err
--- a/pkg/pmem-device-manager/pmd-util.go
+++ b/pkg/pmem-device-manager/pmd-util.go
@@ -2,12 +2,28 @@ package pmdmanager
 
 import (
 	"fmt"
+	"sync"
 
 	pmemexec "github.com/intel/pmem-csi/pkg/pmem-exec"
 	"k8s.io/klog/glog"
 	"os"
 	"strconv"
 )
+
+// Mutex protecting shared device access by threads running in parallel.
+// Create, Delete, Flush may operate on same phys.device from parallel threads.
+// The mutex is defined here and used in different device managers code.
+// Ndctl manager definitely will crash without this mutex protection
+// in 2-volume creation scenario on same Node.
+// For LVM manager, situation is likely not that risky,
+// but we use similar protection in LVM Manager for clarity and unified style,
+// as LVM state is also single instance for a Node.
+//
+// Note that while main idea is to protect against two parallel threads
+// accessing shared entity with different requests (like 2 creations), this
+// mutex also protects against repeated similar requests in different threads
+// which also have been seen when Kubernetes repeats operations rapidly.
+var devicemutex = &sync.Mutex{}
 
 func ClearDevice(device PmemDeviceInfo, flush bool) error {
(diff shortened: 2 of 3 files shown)


okartau:
still seeing crash... should mutex be used with single name, as 2nd run uses new name?

okartau:
Turns out, the mutex use added in first attempt here does not solve the problem, as keymutex works separately for key values, and for different volume Names the different mutex locks were used so it did not protect the region from each other as was plan. The problem is we try to re-use idea from original locking code where the idea was to protect against quickly repeating operations concerning same volume. The new problem we face here is, multiple volume request cause same-time device-side operations (which are not thread-safe). So we need to use simple mutex protecting whole device. I will change the code here. The original mutex use I added in first commit, likely is not even needed

okartau:
I changed keymutex lock to simple lock on device manager side. It's straight from Go example, feedback welcome if something else/more/better is needed.
Now it only activates on devicemode direct, this is were problem was faced.
I briefly checked LVM mode and did not hit race problem, but I think safer is to add similar locking there as well.

okartau:
added LVM side locks. lock itself is defined in utils file to make it neutrally located

okartau:
tested in VM-cluster deployment with 2 parallel volumes, both LVM and direct modes

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -129,6 +131,8 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error
 }
 
 func (lvm *pmemLvm) DeleteDevice(name string, flush bool) error {
+	devicemutex.Lock()
+	defer devicemutex.Unlock()
```
avalluri: The locking should be on device `name` I guess. Deleting two different devices parallelly shouldn't be problem.

okartau: Delete or create tries to change single system in host, also in LVM mode.
I think it adds safety if we avoid change requests from parallel threads at same time.
Introducing another set of locks would make the code more complex,
and there is no perf. requirement to achieve high speeds here, so my idea is to play safe.

pkg/pmem-device-manager/pmd-util.go
```diff
@@ -2,13 +2,16 @@ package pmdmanager
...
 	"fmt"
+	"sync"
 
 	pmemexec "github.com/intel/pmem-csi/pkg/pmem-exec"
 	"k8s.io/klog/glog"
 	"os"
 	"strconv"
 )
 
+var devicemutex = &sync.Mutex{}
```
pohly: Please add an explanation what is protected by this mutex, i.e. which operations must be done while holding the mutex.

pkg/pmem-device-manager/pmd-lvm.go
pohly: I agree that it is better to play it safe here.

pkg/pmem-device-manager/pmd-util.go
okartau: added a comment, force-pushed

PR #207, opened 2019-03-26
Fix issue in passing dax mount option.

avalluri:
Dax mount option should be used while mounting a fsdax namespace. Fsdax is the
default volume namespace mode.

Volume.VolumeContext is only filled with volume parameters that were passed by
CreateVolumeRequest.Parameters, Hence default volume parameters are not filled
in VolumeContext. Code in NodeServer.StageVolume using dax mount option only in
case of valid VolumeContext, so dax option was not used if 'nsmode' is not found
in volume context.

This change fixes this, by considering the default volume parameters if no volume
context passed.


okartau:
verified. I tried this change in my cluster where the lack of dax was noticed, and it fixes it:
```
/dev/ndbus0region0fsdax/86d03712-4fc6-11e9-b861-06ac96387a66 on /data-ext4 type ext4 (rw,relatime,dax)
/dev/ndbus0region0fsdax/86cfdf2f-4fc6-11e9-b861-06ac96387a66 on /data-xfs type xfs (rw,relatime,attr2,dax,inode64,noquota)
```

PR #208, opened 2019-03-27
Use two volumes instead of one in deploy example

okartau:
Using two volumes serves multiple purposes:
- makes use and demonstrates configuration of ext4 vs xfs formats
- stresses driver more as 2 creations on same node go in parallel
Note that in storageclass we change
volumeBindingMode from Immediate to WaitForFirstConsumer
to make deployment work for two volumes

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,19 +492,30 @@ future, currently untested releases.
 ```
 
 The application uses **storage: pmem** in its <i>nodeSelector</i>
-list to ensure that it runs on the right node.
+list to ensure that it runs on the right node, and it requests mount of two volumes,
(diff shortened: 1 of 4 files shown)


okartau:
yes, I very much agree. My trial with two volumes was initially meant to demonstrate fsType parameter use. When I saw crashes caused by two simultaneous creations, I realized that was probably first time when more than one volume was created in parallel by this driver. All of deploy examples and e2e, sanity tests so far have been single-thread, create-one-and-delete-it type, without putting any stress on concurrency, scalability etc. sides. So yes, we need to create and run tougher tests for sure.

PR #209, opened 2019-03-28
Add chapter in DEVELOPMENT.md describing generated elements, redraw comms.diagram

### LINKED ISSUE #78: Document repository elements make steps which are not part of Makefiles

okartau:
Describe creation of the diagrams and other elements
in repository which are created using external tools.

2nd commit redraws comms diag with dia and adds the source.

Resolves #78


okartau:
As first draft, I threw together some notes/ideas accumulated so far. Let's improve from here. For example, @avalluri  do you remember plantuml command, lets write it down.

pohly:
> As first draft, I threw together some notes/ideas accumulated so
> far. Let's improve from here. For example, @avalluri  do you remember
> plantuml command, lets write it down.

FWIW, plantuml can be automated:
https://github.com/intel/oim/blob/master/doc/doc.make

okartau:
I re-created the comms diagram in dia-tool, to get the source of it as manageable piece.
I think the new README can be seen rendered here, how does it look ?
https://github.com/intel/pmem-CSI/blob/7c98dc7c6adaa49d79929648902eba93c8fb0742/README.md

I tried to copy almost exactly, with few changes(slave->node, removed docker as specific container mentioned in a box (as we can have others)). Should text color  be darker to be readable better ?
If we agree we move over to this diagram, I will change the TODO part in first commit here.

okartau:
Improved diagram more, better contrast and reduced vertical span:
https://github.com/intel/pmem-CSI/blob/ed19c21bbf904b26cb0ccf725f8845de82b8c968/README.md

DEVELOPMENT.md
```diff
@@ -213,3 +219,63 @@ will not be visible inside container. The driver does not detect the
...
+The PNG file is commited as repository elements in docs/images/communication/.
+
+Diagrams describing provisioning sequence
+-----------------------------------------------------------------------
+Two diagrams are generated using [plantuml program](http://plantuml.com/).
+Source files:
+- [source file for regular sequence](/docs/diagrams/sequence.wsd)
+- [source file for cache sequence](/docs/diagrams/sequence-cache.wsd).
+
+The PNG files are commited as repository elements in docs/images/sequence/.
```
avalluri: s/commited/committed

DEVELOPMENT.md
```diff
@@ -213,3 +219,63 @@ will not be visible inside container. The driver does not detect the
...
+The [single source file](/docs/diagrams/pmem-csi.dia) has
+layers: {common, lvm, direct} so that two diagram variants can be produced from single source.
+Image files are produced by saving in PNG format with correct set of layers visible.
+The PNG files are commited as repository elements in docs/images/devicemodes/.
+
+Top-level README diagram describing communication channels
+----------------------------------------------------------------
+This diagram was created using the [_dia_ drawing program](https://en.wikipedia.org/wiki/Dia_(software)).
+Image file is produced by saving in PNG format.
+The PNG file is commited as repository elements in docs/images/communication/.
```
avalluri: s/commited/committed

DEVELOPMENT.md
```diff
@@ -213,3 +219,63 @@ will not be visible inside container. The driver does not detect the
...
+-----------------------------------------------------------------------
+Two diagrams are generated using [plantuml program](http://plantuml.com/).
+Source files:
+- [source file for regular sequence](/docs/diagrams/sequence.wsd)
+- [source file for cache sequence](/docs/diagrams/sequence-cache.wsd).
+
+The PNG files are commited as repository elements in docs/images/sequence/.
+
+protobuf-compiler
+-----------------
```
avalluri: Instead of "protobuf-compiler" section, It would be good if can have something like "RegistryServer spec" that explains how to regenerate the source from the spec.

DEVELOPMENT.md
okartau: fixed

DEVELOPMENT.md
okartau: fixed

DEVELOPMENT.md
okartau: fixed

okartau:
fixed by recent comments, plus other improvements, like corrected TOC entry which was bad link, 
added more links in comms-diagrams section, pushed

avalluri:
>  @avalluri do you remember plantuml command, lets write it down.

Oops, I missed this comment. My editor(`vscode`) has plantuml plugin, which takes care of generating imags, so I myself never run `plantuml` to generate diagrams.

okartau:
should we mention that editor plugin in DEVELOPMENT.md ?
The visibility of message sequence charts has been reduced as these are not visible any more in markup files, so how likely it is user actually sees those?

avalluri:
> should we mention that editor plugin in DEVELOPMENT.md ?

I don't think it's worth mentioning about editor plugin. If I am not wrong, one intention of this document is for developers who want to contribute to our source. So if someone feels to change sequence diagrams, should be enough to mention that they have to use `plantuml`.

> The visibility of message sequence charts has been reduced as these are not visible any more in markup files, so how likely it is user actually sees those?

There are links in [Usage on Kubernetes](https://github.com/intel/pmem-CSI#usage-on-kubernetes) section.

pohly:
Instead of documenting how to run plantuml and protoc manually it would be better to add make targets for those. They can be copied from OIM, which had both. But in the meantime this PR is a worthwhile enhancement.

Not sure whether dia can be automated.

PR #210, opened 2019-03-28
deployment: Remove node selector from application deployment

avalluri:
As we implemented csi topology, scheduler has to consider this and launch
application only on node where the volume created, so applications need not be
aware of "stoage:pmem" label anymore.

### DIFF
--- a/deploy/kubernetes-1.13/pmem-app.yaml
+++ b/deploy/kubernetes-1.13/pmem-app.yaml
@@ -10,8 +10,6 @@ spec:
       volumeMounts:
       - mountPath: "/data"
         name: my-csi-volume
-  nodeSelector:
-    storage: pmem
   volumes:
   - name: my-csi-volume
     persistentVolumeClaim:


okartau:
Can we stop using label storage:pmem overall ?
If yes then we should clean it's use up in other locations:
- other deploy files
- test setup script
- README

okartau:
change seems to work in my test cluster (where I run two-volume app)

avalluri:
> I think I got too excited about that. You mean that app can stop using label, but we keep using label to start driver on nodes with device presence, right?
> But, I still got idea to try to reduce labeling.
> Use of labeling adds a separate manual admin step, which opens up window to config errors. Software can detect presence of device, so why should human admin be forced to teach this knowledge to the nodes again?
> What if we try:
> 
> * pmem-csi driver starts on all nodes

The only possible(as of my knowledge) way to deploy to all nodes in Kubernetes is using "DaemonSet". In this case even the driver exits, Kubernetes tries to restart the driver pod on the node, unless we use restart never policy, which is not suggestable.

But I support the idea, we can add startup check to exit if not PMEM devices detected. But we should live with labeling I guess.

avalluri:
> change seems to work in my test cluster (where I run two-volume app)

Our test cluster has no node without "pmem" support, so obviously should makes no diffrerence :). Real test would be to run in a multi-node cluster where few nodes does not have pmem installed.

okartau:
I removed labels from host-1 and host-2, and after few more extra cleaning steps (nodeinfo still kept the previous knowledge) we tried on my cluster. But app deployment fails. app should get scheduled to host-3, but kubernetes tries to put it on host-1 with msg:
```
W0329 topology.go:171] error getting CSINodeInfo for selected node "host-1": csinodeinfos.csi.storage.k8s.io "host-1" not found; proceeding to provision without topology information
```
late binding involved?

okartau:
another warning popped up while looking to these logs, we need to address it soon:
```
W0329 controller.go:489] "fstype" is deprecated and will be removed in a future release, please use "csi.storage.k8s.io/fstype" instead
```

avalluri:
> I removed labels from host-1 and host-2, and after few more extra cleaning steps (nodeinfo still kept the previous knowledge) we tried on my cluster. But app deployment fails. app should get scheduled to host-3, but kubernetes tries to put it on host-1 with msg:
> 
> ```
> W0329 topology.go:171] error getting CSINodeInfo for selected node "host-1": csinodeinfos.csi.storage.k8s.io "host-1" not found; proceeding to provision without topology information
> ```
> late binding involved?

Hmm, this shows that in late binding mode, Kubernetes is picking the node irrespective of CSI driver availability on that node. Ok, let's hold this change then.

pohly:
> in late binding mode

But are we using that? We shouldn't. We know that it doesn't work reliably (https://github.com/kubernetes-csi/external-provisioner/issues/221#issuecomment-473821768) and it doesn't make sense conceptually (why schedule allocation of persistent storage based on current RAM and CPU usage of a node?).

We should add a warning about this to the documentation.

okartau:
> But are we using that? We shouldn't.

But that change was just today merged with "use two volumes" change.
I even added note in commit msg about changing binding mode.
Without late binding mode, driver can't support more than one volume on same node.
Are we then about to restrict number of volumes to one?
We have to revert two-volumes change then.

pohly:
> But that change was just today merged with "use two volumes" change.
> I even added note in commit msg about changing binding mode.

Sorry, I missed that part.

The example can be changed to start two fake apps, each with a different volume.

How to support storage allocation of an app that uses more than one volume is indeed an open question. We can use parameters in the storage class, but that's rather awkward and the wrong level of abstraction: we want to specify for a PV where it gets allocated, not for all PVs created via one storage class. I need to ask whether PVC params have been considered. Perhaps they were not, because it limits portability of an app. But an app using PMEM isn't portable anyway.

pohly:
> I need to ask whether PVC params have been considered.

We discussed this on Friday. Saad wasn't present, but Michelle explained that he is opposed to anything in Kubernetes which makes apps non-portable between clusters, which includes special parameters in a PVC. This obviously makes it difficult to write apps which by design (= depend on PMEM) aren't portable.

However, that `fstype` is in the `StorageClass` is mostly just historic, it should be in the PVC because it is seen as portable. An effort (write KEP, implementation) to move it into PVC would be welcome.

The suggested solution for choosing where to allocate PVs was to fix late binding (https://github.com/kubernetes-csi/external-provisioner/issues/221#issuecomment-474248095).

In the meantime, let's:
- revert the introduction of WaitForFirstConsumer
- switch to two apps with one volume each

okartau:
> In the meantime, let's:

I will create a PR implementing this plan

pohly:
@avalluri please rebase

There are no merge conflicts, but after merging there's still a nodeSelector in the second app in deploy/kubernetes-1.13/pmem-app.yaml.

pohly:
@avalluri I'll take over as I want to work on deployment changes for Kubernetes 1.14.

avalluri:
> @avalluri I'll take over as I want to work on deployment changes for Kubernetes 1.14.

@pohly  sure, you can take this over for any further work needed.

pohly:
Merged via PR #263

PR #222, opened 2019-04-03
deploy: use separate storage class yamls

pohly:
The E2E testing can only handle a single storage class per file. When
there are two of them, it doesn't know which one to use.

While at it, also use the official parameter for
fstype (https://github.com/kubernetes-csi/docs/blob/master/book/src/external-provisioner.md).


okartau:
looks good, I will also try it in my cluster

okartau:
while testing (still going) I noticed something that remained left out from deploy-2-volumes change: `make start` output gives instruction at the end:
```
To try out the pmem-csi driver persistent volumes:
   cat deploy/kubernetes-1.13/pmem-pvc.yaml | _work/ssh-clear-kvm kubectl create -f -
   cat deploy/kubernetes-1.13/pmem-app.yaml | _work/ssh-clear-kvm kubectl create -f -
```
but we changed pmem-app.yaml to be pair of files, pmem-app-1.yaml and pmem-app-2.yaml,
means if user copies example commands, it will fail.
We need to use one of two files in example text, or both, or create symlink if we want to use name without number part, but the latter is probably too much if just for this example.

okartau:
can we make this change here, or we need a separate PR?

pohly:
> can we make this change here, or we need a separate PR?

Let's do it here.

okartau:
another leftover are comments in deploy/kubernetes-1.13/pmem-pvc.yaml
which still use previous file name

okartau:
Still getting errors with this PR changes applied (tried twice, same result)
```
Summarizing 2 Failures:
[Fail] sanity Node Service [It] should work 
..../kubernetes-csi/csi-test/pkg/sanity/node.go:436

[Fail] PMEM Volumes [Driver: pmem-csi] [Testpattern: Dynamic PV (default fs)] provisioning [It] should provision storage with defaults 
..../k8s.io/kubernetes/test/e2e/storage/testsuites/provisioning.go:457
```

I try to dig in logs to see what happens

okartau:
first error is caused by mount error:
```
mount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-e5d08c10-562d-11e9-a6e1-deadbeef0100/globalmount: 
wrong fs type, bad option, bad superblock on 
/dev/ndbus0region0fsdax/ed38cc5c-562d-11e9-87e6-cef1fff15303, 
missing codepage or helper program, or other error.
```
This does not ring much bell what is wrong. I searched for other examples but I find multitude of different scenarios and no easy ideas. There is valid ext4 filesystem on device, as :
- mkfs.ext4 was running OK 
- following 2nd stage attempt checks device with blkid and finds ext4 filesystem there
Still, next mount attempt fails with same error.
What seems strange is that request asks for 1 MByte volume which is really small,
then LVM refuses to create less than 4 MB and rounds the size up to 4 MBytes.
I wonder does the small size contribute to problem. But these tests have passed before, right.

okartau:
I tried `make test_e2e` in devicemode=direct, and there are no errors:
```
Ran 35 of 61 Specs in 1488.940 seconds
SUCCESS! -- 35 Passed | 0 Failed | 0 Pending | 26 Skipped
```

pohly:
> but we changed pmem-app.yaml to be pair of files, pmem-app-1.yaml and pmem-app-2.yaml

What was the reason for making that two files while keeping two PVCs in pmem-pvc.yaml? Wouldn't it be easier to have two pods in pmem-app.yaml?

pohly:
> Still getting errors with this PR changes applied

Worked for me. I also updated vendor dependencies (not submitted yet) and ran E2E with those in the default LVM mode, again without issues. Looks like some intermittent problem.

okartau:
> What was the reason for making that two files

Somehow I did not even consider putting two pods into same file. I think I thought there may be cases where only one is needed. We can try having those in same file, it would simplify other parts.

okartau:
I quickly tried and app deploy works OK with two pods declared in same file, do you want to make that change here in this PR?

pohly:
I'll merge the two app files here.

pohly:
> I'll merge the two app files here.

Done. @okartau, please review.

okartau:
By inspection, looks good.
I tried to test via 'make start' and example app deployment. It works, but it is deployable only with changes of #224 applied before 'make start'.

PR #224, opened 2019-04-04
Support test on 1.14 cluster

okartau:
First change:
Kubernetes 1.14 brings output change, join command example
is printed on two lines separated by `\` which is
problematic for grep. Join all multi-lines before grep
using sed. Currently there is just this one.

### DIFF
--- a/test/setup-clear-kvm.sh
+++ b/test/setup-clear-kvm.sh
@@ -375,8 +375,9 @@ fi
 ${TEST_CONFIGURE_POST_MASTER}
 
 # Let the other machines join the cluster.
+# Join multi-lines using 'sed' so that grep finds whole line
 for i in $(seq 1 $LAST_NODE); do
-    _work/ssh-clear-kvm.$i $(grep "kubeadm join.*token" _work/clear-kvm-kubeadm.0.log) $kubeadm_args
+    _work/ssh-clear-kvm.$i $(sed -z 's/\\\n//g' _work/clear-kvm-kubeadm.0.log |grep "kubeadm join.*token" ) $kubeadm_args
 done
 
 # From https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#pod-network
--- a/test/setup-clear-kvm.sh
+++ b/test/setup-clear-kvm.sh
@@ -385,7 +385,8 @@ _work/ssh-clear-kvm $PROXY_ENV kubectl apply -f https://raw.githubusercontent.co
 
 # Install addon storage CRDs, needed if certain feature gates are enabled.
 # Only applicable to Kubernetes 1.13 and older. 1.14 will have them as builtin APIs.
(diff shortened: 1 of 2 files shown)


okartau:
We need more changes, 1.14 means also our test setup parts which rely on $(cat _work/clear-kvm-kubernetes.version), stopped working. Let's start with symlink 1.14, pointing to 1.13?

okartau:
2nd commit creates 1.14 symlink.
Now test cluster creation and example app deployment works.

okartau:
With current changes of 2 commits applied, I can run `make test_e2e` in devicemode=direct without errors.
devicemode=lvm run exposes same mount-related errors I reported under #222, I think this deserves a new issue

okartau:
> Can you add a comment about Kubernetes 1.14 to the README.md? Something along the lines of "it works if you install the CRDs, an future update will make it work also without those"?

I added mention and install example of CRDs in #227, does this cover it? Although current version does not say about future update.

pohly:
That comment is about 1.13. We need to be explicit about also needing it for 1.14, and add the explanation that it is just a short-term workaround.

okartau:
I added another statement under #227 explicitly about 1.14. It is almost copying example commands but I think it's user-friendlier (can re-use command) than showing using version template

okartau:
> Are you sure the CRDs get installed with the changes in this PR when installing Kubernetes 1.14?

good point, they are not installed based on the pointed code. But I am positive I tried both example deployment and e2e on that system and both worked. Is it for sure those CRDs are not yet installed implicitly on 1.14 ?

okartau:
I found create CRD command in my cmd history (but just one instead of two, probably cmdline-edit-and-reuse side effect in bash_history save?), means I probably verified does CRD file exist on 1.14 path, causing install as side effect, and I ran tests after that. I need to verify whole chain from empty point.

pohly:
> good point, they are not installed based on the pointed code. But I am
> positive I tried both example deployment and e2e on that system and
> both worked. Is it for sure those CRDs are not yet installed
> implicitly on 1.14 ?

Yes.

test/setup-clear-kvm.sh
```diff
@@ -385,7 +385,8 @@ _work/ssh-clear-kvm $PROXY_ENV kubectl apply -f https://raw.githubusercontent.co
 
 # Install addon storage CRDs, needed if certain feature gates are enabled.
 # Only applicable to Kubernetes 1.13 and older. 1.14 will have them as builtin APIs.
-if _work/ssh-clear-kvm $PROXY_ENV kubectl version | grep -q '^Server Version.*Major:"1", Minor:"1[0123]"'; then
+# Temporarily, install on 1.14 until they get into sidecars.
```
pohly: The explanation is wrong. "Temporarily install also on 1.14 until we migrate to sidecars which use the builtin beta API."

okartau: replaced with your sentence, pushed

PR #230, opened 2019-04-10
WIP: Add govm-based targets for testing on VMs

### LINKED ISSUE #193: bring up QEMU cluster with GoVM
https://github.com/govm-project/govm looks like a good replacement for the home-grown scripts that configures the QEMU cluster and bring up Clear Linux in it. By switching to GoVM, we make the process simpler for users (automatic configuration of network) and more flexible (can support other OSes).

okartau:
Draft of govm-based testing.
Existing testing is so well coupled with clear deps that I considerd easier to write new scripts
than try to re-use/modify old.
The good part is, new approach does not change existing one at all.
As future work, we may want to re-group using common code from Clear side.
With current approach it should be easier to add next target,
as I tried to keep common vs. Ubuntu things separated.
Some TODOs are in there, comments, ideas welcome.
Currently docker is used and CRIO was left out.
On Ubuntu, there is one warning from docker side.
Device support on cloud-based kernel is not in good shape, so we upgrade kernel and reboot.

There is no doc part yet, and this code is not integrated with make targets currently.
To get Ubuntu-based VMs cluster up installed and running, run this in repo directory:
```
$ ./test/govm-start-ubuntu.sh
```

To get CentOS-based VMs cluster up installed and running, run this in repo directory:
```
$ ./test/govm-start-centos.sh
```

The pre-requisite is installation of govm, which has one dependency: docker.
The images and VM storage are stored under ~/vms. I did not start changing that as it would lead to additional option defining images location when govm  is used on host CLI.

Some more operating tips:
to get **all VMs** stopped and wiped away:
```
$ govm rm -a
```

The restart-idea of the creation script is similar to clear tester: "done" flags under _work enable restarting to continue interrupted setup. It does not support yet starting VMs from shutdown state, but this can be achieved via govm commands for VMs separately.
VM can be rebooted inside, it will restart.
VM can be shut down, it remains stopped, and can be started using `govm start NAME`
Creation script creates (similar to clear-based tester) ssh wrappers under _work/ for logging into VM as root.
Creation script sets up all nodes in parallel, saving output to node-specific files.

Resolves #193

### DIFF
--- a/test/govm-start-common.sh
+++ b/test/govm-start-common.sh
@@ -1,0 +1,244 @@
+. $(dirname $0)/test-config.sh
+NUM_NODES=4
+# override TEST_IP_ADDR that comes from test-config.sh (as clear-specific value) with govm value
+TEST_IP_ADDR=172.17.0
+NO_PROXY="${NO_PROXY:-},${TEST_IP_ADDR}.0/24,10.96.0.0/12,10.244.0.0/16"
+PROXY_ENV="env 'HTTP_PROXY=${http_proxy:-}' 'HTTPS_PROXY=${http_proxy:-}' 'NO_PROXY=$NO_PROXY'"
+_sshopts='-oIdentitiesOnly=yes -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oLogLevel=error'
+_ssh="ssh $_sshopts"
+_scp="scp $_sshopts"
+_imgdir=_work
+
+create_vmhost_manifest () {
+    _rel=$1
+    _templ=$2
+    _img=$3
+    _file=$_wdir/${_templ}-${_rel}.yaml
+    echo "---" >  $_file
+    echo "vms:" >>  $_file
+    for i in $(seq 0 $((NUM_NODES - 1))); do
+        cat $(dirname $0)/${_templ}-template.yaml \
+	    | sed "s/__HOST__/host-$i/" \
+	    | sed "s/__IMAGE__/$_img/" >> $_file
+	# add NVDIMM emulation related part on workers
+	if [ $i -ne 0 ]; then
+	    cat >> $_file <<EOF
+      - |
+        EXTRA_QEMU_OPTS=
+        -machine nvdimm=on -object memory-backend-file,id=mem1,share,mem-path=/data/nvdimm0,size=32768M -device nvdimm,memdev=mem1,id=nv1,label-size=2097152
+EOF
+	fi
+    done
+}
+
+set_docker_proxy () {
+    a=$1
+    if [ -n "${http_proxy:-}" ]; then
+        if [ -f $_wdir/govm_docker_proxy.$a.done ]; then
+	    return 0
+        fi
+        # proxy for docker
+        $_ssh root@$a mkdir -p /etc/systemd/system/docker.service.d/
+        $_ssh root@$a "cat >/etc/systemd/system/docker.service.d/proxy.conf" <<EOF
+[Service]
+Environment="HTTP_PROXY=$http_proxy" "HTTPS_PROXY=$https_proxy" "NO_PROXY=$NO_PROXY"
+EOF
+        touch $_wdir/govm_docker_proxy.$a.done
+    fi
+}
+
+set_docker_registry () {
+    a=$1
+    if [ -f $_wdir/govm_docker_registry.$a.done ]; then
+	return 0
+    fi
+    # configure insecure registry for docker
+    $_ssh root@$a "mkdir -p /etc/docker"
+    $_ssh root@$a "cat >/etc/docker/daemon.json" <<EOF
+{ "insecure-registries": ["10.237.72.78:5000"] }
+EOF
+    touch $_wdir/govm_docker_registry.$a.done
+}
+
+# kubeadm_args is used from multiple functions so we have separate function to set its value
+setup_kubeadm_args () {
+    kubeadm_args=
+# TODO: CRIO mode not tried, not complete/supported
+case $TEST_CRI in
+    docker)
+        cri_daemon=docker
+        # [ERROR SystemVerification]: unsupported docker version: 18.06.1
+        kubeadm_args="$kubeadm_args --ignore-preflight-errors=SystemVerification"
+        ;;
+    crio)
+        cri_daemon=cri-o
+        # Needed for CRI-O (https://clearlinux.org/documentation/clear-linux/tutorials/kubernetes).
+        kubeadm_config_init="$kubeadm_config_init
+nodeRegistration:
+  criSocket: /run/crio/crio.sock"
+        ;;
+    *)
+        echo "ERROR: unsupported TEST_CRI=$TEST_CRI"
+        exit 1
+        ;;
+esac
+
+}
+
+kubeadm_init () {
+    if [ -f $_wdir/govm_kubeadm_init.done ]; then
+	return 0
+    fi
+kubeadm_args_init=
+kubeadm_config_init="apiVersion: kubeadm.k8s.io/v1beta1
+kind: InitConfiguration"
+kubeadm_config_cluster="apiVersion: kubeadm.k8s.io/v1beta1
+kind: ClusterConfiguration"
+kubeadm_config_kubelet="apiVersion: kubelet.config.k8s.io/v1beta1
+kind: KubeletConfiguration"
+
+if [ ! -z ${TEST_FEATURE_GATES} ]; then
+    kubeadm_config_kubelet="$kubeadm_config_kubelet
+featureGates:
+$(IFS=","; for f in ${TEST_FEATURE_GATES};do
+echo "  $f" | sed 's/=/: /g'
+done)"
+    kubeadm_config_cluster="$kubeadm_config_cluster
+apiServer:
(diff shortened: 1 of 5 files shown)


pohly:
> The good part is, new approach does not change existing one at all.

This is fine while we still experiment with this, but ultimately we want to replace the custom startup scripts for Clear with GoVM.

test/govm-start-common.sh
```diff
@@ -0,0 +1,211 @@
+. $(dirname $0)/test-config.sh
+NUM_NODES=3
+# override TEST_IP_ADDR that comes from test-config.sh (as clear-specific value) with govm value
+TEST_IP_ADDR=172.17.0
+NO_PROXY="$NO_PROXY,${TEST_IP_ADDR}.0/24,10.96.0.0/12,10.244.0.0/16"
```
pohly:
```
test/govm-start-common.sh: line 5: NO_PROXY: unbound variable
```

test/govm-start-ubuntu.sh
```diff
@@ -0,0 +1,148 @@
...
+    # silence timesync, fails to connect outside. Other way would be to configure ntp server
+    $_ssh root@$a "systemctl stop systemd-timesyncd"
+    touch _work/govm_final_tuning.$a.done
+}
+
+########################  START   ###################################
+# get image, checking is it cached
+mkdir -p _work
+if [ ! -e $_imgdir/$_imgfile ]; then
+    curl -Ok https://cloud-images.ubuntu.com/${_ubunturelease}/current/$_imgfile
```
pohly: Allow insecure connections?! Please don't, not even for development.

I prefer long options in scripts for non-obvious flags, it's easier to review. `--insecure` would have raised a red flag immediately while `-k` is easy to overlook.

Add `--location` (allow redirects) and `--fail` (fail with an error when the URL leads to 404). I know, the `curl` invocation in `test/clear-kvm.make` doesn't have it either. Your chance to deliver something better :grin:

test/govm-start-common.sh
okartau: uh, right. all my dev was behind proxy, but I should have been verified that its runnable without. will fix soon

test/govm-start-ubuntu.sh
okartau: that cmd I copied out from govm use example on CI side. lets improve

test/govm-start-common.sh
okartau: pushed new version, should work better with undefined proxy values

okartau:
I replaced ready-made VM templates with generated ones. That reduces repetition on 2 levels:
- Ubuntu release files 18.04 vs 18.10 were very similar, differing just by image file name
- one file contained blocks for all 4 VMs with mostly repeated lines
Now we have a single template for Ubuntu VM which we repeat substituting specific parts,
and resulting generated yaml file under _work/ is used by 'govm compose'.
The readability how that file will look gets somewhat worse in repo, (you can look at it under_work/ after it gets generated) but I think that is small price for avoided repeated blocks in source files.

okartau:
pushed new state that covers some holes that were still open: kernel is updated and nodes rebooted, wait-for-root-ssh is dynamic now, etc. minor things.

pohly:
@okartau have you seen that @chuyd pushed https://github.com/intel/pmem-CSI/pull/236? Are you coordinating with him on who's doing what?

okartau:
yes I saw #236. There the focus seems different: to keep existing test style, keep Clear, and switch to govm-based VM mgmt, if I got it right.
In my work in this PR I try to focus on how to make test code modular so that we can easily add more distro bases. We haven't been coordinating, I think  these tracks have happened independently.

okartau:
@pohly you mention in #236 that govm installation could be part of tester system. That probably means you want to see it here as well. But implemented as common test-dep target, right.

pohly:
> you mention in #236 that govm installation could be part of tester system. That probably means you want to see it here as well. 

Just do something similar to:
https://github.com/intel/pmem-CSI/blob/679d6eb2ffe8bee799a395450f4fc543fc1aafc8/Makefile#L99-L110

okartau:
adapted to target-specific subdir style, creating all files now under _work/kvm-ubuntu.
I can get cluster up and e2e tests pass.
(this is still not integrated with make so e2e test command is given separately)
BTW, was the idea that **all** created files get stored under target-specific dir?
In clear-lvm case there are still those created under _work/:
```
clear-29100-kvm.img.xz  clear-29100-kvm.img.xz-SHA512SUMS 
clear-29100-kvm.img.xz-SHA512SUMS.sig  clear-kvm-29100.img
id  id.pub  OVMF.fd  passwd
```

pohly:
> adapted to target-specific subdir style, creating all files now under _work/kvm-ubuntu.

I would prefer `ubuntu-govm`. Our naming scheme then becomes "<OS>-<cluster technology>[-<custom suffix>]". "kvm" is a special case of QEMU (= QEMU with kvm acceleration).

> BTW, was the idea that all created files get stored under target-specific dir?

No. Some files are potentially shared by different clusters, like downloaded image files and the generated id + password. As long as sharing is save (i.e. read-only) that's okay.

okartau:
I had misspelled the path in hurry.. it is ubuntu-kvm as I mimicked clear style. But I agree it's more descriptive to name it ubuntu-govm (although there is qemu-kvm running underneath of govm/docker layer). So I will rename it ti ubuntu-govm.

Downloaded image files by default govm config are stored under ~/govm and that kind of makes sense, as there is no point in losing them when I want to wipe off my _work/ so in current draft I use that govm default config.

pohly:
> Downloaded image files by default govm config are stored under ~/govm
> and that kind of makes sense, as there is no point in losing them when
> I want to wipe off my _work/ so in current draft I use that govm
> default config.

Wiping _work was the former brute-force way of re-creating a
cluster. Now we just wipe _work/<cluster dir>. Therefore I'd prefer to
use _work for all files that get downloaded and may be reused for
different cluster installations.

Then I can clean my machine by wiping out _work without having to
hunt for other location were data might have been dumped. Also helps
when $HOME and my $GOPATH are on different partitions.

pohly:
I've probably missed it, but where does kubelet come from in this Ubuntu installation? Can we make the version of Kubernetes configurable? That would be very useful.

With Clear Linux, we are using the kubelet and kubeadm that come with Clear Linux, so Clear Linux decides for us which Kubernetes release we get.

okartau:
> where does kubelet come from in this Ubuntu installation

see install_kubeadm() in govm-start-ubuntu.sh.
kubelet is not spelled out as it gets installed as in same pass if "apt install kubeadm" is made.
Probably cleaner to spell out all 3 componensts there, I will add.

> Can we make the version of Kubernetes configurable?

I doubt it, the style in those repos seem to support one/current version,
But it is worth checking can another version be selected, or is there some other repo/install scheme.
Currently I followed the 1st and most mainstream docs that come up in "how to install" search.

pohly:
>> Can we make the version of Kubernetes configurable?
>
> I doubt it, the style in those repos seem to support one/current version,
> But it is worth checking can another version be selected, or is there
> some other repo/install scheme.

We could download
https://dl.k8s.io/v1.14.0/kubernetes-node-linux-amd64.tar.gz and then
use those kubelet and kubeadm binaries.

There's no systemd service file, but perhaps we can use the one from
Ubuntu and just swap out the binary that is used by the service?

okartau:
> https://dl.k8s.io/v1.14.0/

what version do you want?
Both govm-based approaches that I develop here, Ubuntu and Centos-7 (I will add here soon)
get v1.14.1 by default, so loading 1.14.0 specifically does not change much.

okartau:
I added CentOS-7 first support draft. Thanks to isolated common part, reaching working version of another target was relatively easy, taking just few hours (not complete of course, it needs to be improved from here).
Cluster comes up and pmem-csi driver gets deployed. I haven't tried test_e2e yet.
Adding another target let's us see common vs. specific parts division better.

okartau:
e2e test passes on CentOS-7 cluster.
(activated via `go test...` as not integrated with Makefile yet)

pohly:
> e2e test passes on CentOS-7 cluster.
> (activated via `go test...` as not integrated with Makefile yet)

What kind of integration would that be?

We can already do "make test_e2e CLUSTER=centos-govm" (assuming that
this is the name of the cluster).

okartau:
> We can already do "make test_e2e CLUSTER=centos-govm"

that fails caused by currently missing make target for _work/centos-govm/created,
because there is no make rules yet for these new CLUSTER targets.
I think I need to add centos-govm.make and ubuntu-govm.make next.
Such rules may end up being quite similar, with some differences in file names,
can that be abstracted perhaps?

pohly:

>> We can already do "make test_e2e CLUSTER=centos-govm"
>
> that fails caused by currently missing make target for _work/centos-govm/created,
> because there is no make rules yet for these new CLUSTER targets.

True. I hadn't actually checked whether you had that.

> I think I need to add centos-govm.make and ubuntu-govm.make next.
> Such rules may end up being quite similar, with some differences in file names,
> can that be abstracted perhaps?

You can have _work/%-govm/created rules inside your govm.make and then
switch between different cases based on the value of $*.

PR #231, opened 2019-04-11
README: Move "Verify node labels" block after "Deploy driver" block

okartau:
We can not verify both important labels before deploying
driver, because node= label is given when driver starts.

Note, this moves text block without changing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,19 +432,6 @@ The method to configure alpha feature gates may vary, depending on the Kubernete
     $ kubectl label node <your node> storage=pmem
 ```
 
-- **Verify that the node labels have been configured correctly**
-
-```sh
-    $ kubectl get nodes --show-labels
-```
-
-The command output must indicate that every node with PMEM has these two labels:
-```
-pmem-csi.intel.com/node=<NODE-NAME>,storage=pmem
-```
-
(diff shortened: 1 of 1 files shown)


okartau:
As another idea I had, we could move "check labels" to the end of deployment, and use heading "Troubleshooting", thus taking it out of steps, and adding as separate explanation that is supposed to be checked in case of troubles. Is such method better than current "step" style?

okartau:
still not in right location, labels check has to be moved to very end, as pmem-csi.intel.com/node= labels appear only after app start. will update... And at the end this actually turns to be a "troubleshoot" chapter

okartau:
above comment was not right. pmem-csi.intel.com/node= labels appear after controller part of driver reaches Running state. But current steps still will fail because there is wait-time (few minutes, I just tried and it was 3 min) from  "create -f pmem-csi-lv.yaml"  until all pods turn to Running state. By current steps, user is instructed to give command "create" and immediately check for labels, this will not work. We should insert a statement "wait until all pods reach Running", then have labels check step.

okartau:
How does it look now?
I verified currently visible steps by creating a cluster and following steps in README.

PR #232, opened 2019-04-12
nodeserver.go: Use hard-coded 4096 block size in mkfs.ext4 and mkfs.xfs

### LINKED ISSUE #225: LVM-mode: ext4 mount with -o dax fails on Clear VM in e2e test
`make test_e2e` fails in recent test clusters, if devicemode=lvm,
Same test works if devicemode=direct.
These have been verified to fail:
k8s=1.13.5 (Clear few steps older than 28640, I did not record)
Clear=28640 k8s=1.14, kernel 5.0.6

There are mount errors:
```
Executing: mount -c -o dax /dev/ndbus0region0fsdax/107eea56-56cc-11e9-93d6-36feaaf12372 /var/tmp/tmp.bvQWXy7mlR/staging
mount: /var/tmp/tmp.bvQWXy7mlR/staging: wrong fs type, bad option, bad superblock on /dev/ndbus0region0fsdax/107eea56-56cc-11e9-93d6-36feaaf12372, missing codepage or helper program, or other error.
```

The error is reproducible on plain VM CLI as well. It turns out that failure happens when `-o dax` is added, and file system type is ext4 (That's the one used by e2e test).
I can mount xfs-formatted file system with dax option without error (but it has to be bigger than 1M or 4M which is used by e2e testing, for mkfs.xfs even considering making file system).

This used to work short time ago so seems to be caused by a recent change.
@pohly verified e2e test works on his cluster which is Clear=28070

okartau:
It appears newer kernels refuse to use dax on LVM-exposed
block device if formatted ext4 file system uses 1k block
which is selected heuristically by mkfs.ext4 for smsller volumes.
e2e test code uses specifically small volumes and tests do fail.

Fixes #225

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -205,12 +205,13 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		// Empty FsType means "unspecified" and we pick default, currently hard-codes to ext4
 		cmd := ""
 		args := []string{}
+		// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
 		if requestedFsType == "ext4" || requestedFsType == "" {
 			cmd = "mkfs.ext4"
-			args = []string{"-F", device.Path}
+			args = []string{"-b 4096", "-F", device.Path}
 		} else if requestedFsType == "xfs" {
 			cmd = "mkfs.xfs"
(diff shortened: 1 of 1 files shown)


okartau:
not tested yet, I will run some tests.

okartau:
verified, with these changes e2e testing succeeds.
I added blocksize=4096 on xfs side as well, just for uniformity.

Issue #229, opened 2019-04-05
more flexible deployment


pohly:
Deployment of the driver depends on several different choices:
- direct vs. lvm
- amount of debug logging
- with or without sanity testing support (PR #228)
- Kubernetes version
- images to be used (locally built vs. official release)

This is not a complete list, there might be more. Right now we just support choosing between direct vs. lvm, and our approach for doing so (two different .yaml files) won't scale.

Several possible solutions:
- use kustomize, either as external tool or the builtin support in kubectl
- shell scripting (as in https://github.com/kubernetes-csi/csi-driver-host-path/blob/8d8ea06caf9a9d1bb4930b996869fdb90c84196e/deploy/util/deploy-hostpath.sh#L3-L6)
- an operator

pohly:
I'm currently leaning towards kustomize as a first step, then add an operator later. But this needs some exploration.

pohly:
Examples how other CSI drivers are installed via operators:
- https://github.com/openshift/csi-operator
- https://rook.io/docs/rook/master/ceph-csi-drivers.html

pohly:
I'm leaning towards considering this issue resolved with the kustomize support that we already have. Writing an operator can be tracked separately. @okartau what do you think?

I could write some documentation for making further changes with `kubectl -k`.

pohly:
Hrmm, `kubectl apply --kustomize deploy/kustomize/kubernetes-1.14-lvm-testing` doesn't work:
```
error: AccumulateTarget: AccumulateTarget: AccumulateTarget: security; file '../../patches/controller-role-patch.yaml' is not in or below '/nvme/gopath/src/github.com/intel/pmem-csi/deploy/kustomize/kubernetes-1.14/rbac'
```

I asked about that in https://github.com/kubernetes-sigs/kustomize/issues/975.

We could of course tell people to build `kustomize` with our Makefile, but it would be nicer if it just worked with some future version of `kubectl`.

PR #234, opened 2019-04-12
create .yaml files with kustomize

pohly:
Related-to: #229

### DIFF
--- a/test/start-stop.make
+++ b/test/start-stop.make
@@ -40,19 +40,10 @@ start: _work/clear-kvm.img _work/kube-clear-kvm _work/start-clear-kvm _work/ssh-
 		touch _work/clear-kvm.secretsdone; \
 	fi
 	_work/ssh-clear-kvm kubectl version --short | grep 'Server Version' | sed -e 's/.*: v\([0-9]*\)\.\([0-9]*\)\..*/\1.\2/' >_work/clear-kvm-kubernetes.version
-	. test/test-config.sh && \
-	if ! _work/ssh-clear-kvm kubectl get statefulset.apps/pmem-csi-controller daemonset.apps/pmem-csi >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-csi-$${TEST_DEVICEMODE}.yaml; \
-	fi
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-ext4 >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-ext4.yaml; \
-	fi
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-xfs >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-xfs.yaml; \
-	fi
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-cache >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-cache.yaml; \
-	fi
+	( . test/test-config.sh && _work/ssh-clear-kvm kubectl apply -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-csi-$${TEST_DEVICEMODE}.yaml )
+	_work/ssh-clear-kvm kubectl apply -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-ext4.yaml
+	_work/ssh-clear-kvm kubectl apply -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-xfs.yaml
+	_work/ssh-clear-kvm kubectl apply -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-cache.yaml
 	@ echo
 	@ echo "The test cluster is ready. Log in with _work/ssh-clear-kvm, run kubectl once logged in."
 	@ echo "Alternatively, KUBECONFIG=$$(pwd)/_work/clear-kvm-kube.config can also be used directly."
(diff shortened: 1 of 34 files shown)


pohly:
Shall we keep the longer "pmem-csi-" prefix (this PR) or use just "pmem-" (current state)?

Does this approach make sense?

What other files should we create this way? The storage classes are potential candidates.

I have tested the manual examples, but not E2E.

okartau:
> Shall we keep the longer "pmem-csi-" prefix (this PR) or use just "pmem-" (current state)?
> 
you mean in file names? I see mixed set of file names under deploy/ in current PR set,
what about we unify all to be pmem-csi-XXX ?

> Does this approach make sense?
> 
yes, I think it's good approach allowing to reduce multiplied sources.
Downside is, it's another framework, which has it's source to track, 
adding maintenance load to keep everything up-to-date.
With generated/patched elements that gets commited in repo,
there is risk someone will try to modify those. Currently there seem no marking
"this is patched/generated", is this intentional? Who/how keeps track what gets patched?

> What other files should we create this way? The storage classes are potential candidates.
> 
yes, I agree this is good match

> I have tested the manual examples, but not E2E.
>
I can try this.

pohly:
> > Shall we keep the longer "pmem-csi-" prefix (this PR) or use just "pmem-" (current state)?
>
> you mean in file names?

No, the name of entities inside the .yaml files, for example `pmem-controller` (old name) vs. `pmem-csi-controller` (new name in this PR).

> I see mixed set of file names under deploy/ in current PR set, what about we unify all to be pmem-csi-XXX ?

That's indeed another topic. I think the names there mostly make sense. "pmem-pvc" = a PVC using PMEM. Same for "pmem-app". "pmem-csi.yaml" is about the PMEM-CSI driver and already has "pmem" and "csi".

The storage classes are bit more questionable. One could argue that they are clearly tied to PMEM-CSI and thus should be called `pmem-csi-storageclass`. I'm undecided whether it is worth renaming those.

> Downside is, it's another framework, which has it's source to track,
> adding maintenance load to keep everything up-to-date.

We don't need to update it as long as it does what we want. The makefile rules should make it rather painless for developers (no need to install it separately).

> With generated/patched elements that gets commited in repo,
> there is risk someone will try to modify those. 

True.

> Currently there seem no marking
> "this is patched/generated", is this intentional?

It's a side effect of how kustomize works: it assembles the entities in memory, then serializes as .yaml. There's no way to insert comments into the output. This is actually a drawback for those files where we intend to demonstrate something to the reader.

> Who/how keeps track what gets patched?

I could add a "test-kustomize" which verifies that all output files are unmodified.

pohly:
> I could add a "test-kustomize" which verifies that all output files are unmodified.

Done. I also modified the Makefile slightly to make it simpler to extend and added kustomization of the storageclasses. The benefits there are less obvious (not really that much shorter), but I think it's worthwhile because it enables admins to create their own kustomized storage classes. kustomize cannot just pull a file from upstream, a base always has to have a `kustomization.yaml`.

pohly:
I have force-pushed a complete version that I tested against Clear Linux 28070 with Kubernetes 1.13. E2E testing passes.

Testing against 1.13 is important because of the CSIDriverInfo difference. When trying this against 1.14, `kubectl get csidrivers` will never find any such resource, because it ends up using the beta API from 1.14 and our CRD objects only implements the alpha API. The automated test however did fail until I added the CSIDriverInfo object, so we have at least some verification when using 1.14.

pohly:
FWIW, I tried to install Kubernetes 1.13 on current Clear Linux, but that fails because kubelet is too new. "kubelet version > control plane version" is not supported.

If we want to test against 1.13, we need to find a different way of bringing up a test cluster.

test/start-stop.make
```diff
@@ -35,24 +35,15 @@ start: _work/clear-kvm.img _work/kube-clear-kvm _work/start-clear-kvm _work/ssh-
...
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-ext4 >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-ext4.yaml; \
-	fi
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-xfs >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-xfs.yaml; \
-	fi
-	if ! _work/ssh-clear-kvm kubectl get storageclass/pmem-csi-sc-cache >/dev/null 2>&1; then \
-		_work/ssh-clear-kvm kubectl create -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-storageclass-cache.yaml; \
-	fi
+	( . test/test-config.sh && _work/ssh-clear-kvm kubectl apply -f - <deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-csi-$${TEST_DEVICEMODE}.yaml )
```
okartau: Are the changes here related to kustomize change? Why is it better to stop checking and create storageclasses without checking? I noticed t his because in my govm-ubuntu change set I run shell script with -e option and it stops when storageclass already exists.

okartau:
Looks good by looking at, but I also think it helps to run more broader testing. I ran deployment example with SUCCESS on govm-ubuntu VMs, but cant run e2e there for some unrelated-to-this PR reasons. (there k8s is 1.14.1)

test/start-stop.make
pohly: Switching from `create` to `apply` is unrelated. I find it simpler to use (no need to check for existence first) and it can update existing entities when their spec changes (with some limitations).

I can split it out into a separate commit, but I won't do any separate testing of that commit.

pohly:
> I can split it out into a separate commit, but I won't do any separate testing of that commit.

Done, force-pushed.

pohly:
Compared to the initial version I switched to exposing *all* node's csi.sock. This way it becomes possible to implement additional checks for the entire state of the cluster.

okartau:
I was now able to run both deploy-example and e2e test, both on top of govm-Ubuntu cluster (uses k8s 1.14.1) as in #230 (plus some manual changes in test invocation as official testing is somewhat clear-fixed).
In e2e test I see 1 failure `[It] has CSIDriverInfo`, I think it's same that you mention few comments back.

pohly:
> In e2e test I see 1 failure `[It] has CSIDriverInfo`, I think it's
> same that you mention few comments back.

This was missing initially, but should be working now (78e411d), because
of:
https://github.com/intel/pmem-CSI/blob/78e411ddbac1272e0d5f692b372d5e40b46690e9/deploy/kustomize/kubernetes-1.13/driverinfo-alpha.yaml#L1-L7

pohly:
@okartau I removed the already merged commit from my branch.

I don't think we need to do more manual testing of this PR. If we broke something, we'll notice soon because we are actively working with deployment + testing right now anyway.

PR #236, opened 2019-04-16
Add support to deploy GoVM with Clear Linux

chuyd:
This PR adds support to deploy a Kubernetes cluster with one master and three workers using
the latest  Clear Linux image.

Signed-off-by: Jesus Ornelas Aguayo <jesus.ornelas.aguayo@intel.com>


test/setup-clear-kvm.sh
```diff
@@ -1,34 +1,145 @@
...
+# The Kubernetes control plan (apiserver, etc.) fails to run otherwise
+# ("Host networking requested, not supported by runtime").
 
-# containers-basic is needed because of the CNI plugins (https://github.com/clearlinux/distribution/issues/256#issuecomment-440998525)
-# and for Docker.
-bundles="$bundles containers-basic"
+sudo bash -c "cat >/etc/systemd/system/docker.service.d/51-runtime.conf <<EOF
+[Service]
+Environment=\"DOCKER_DEFAULT_RUNTIME=--default-runtime runc\"
+EOF
```
pohly: Is this a known issue upstream?

test/start-kubernetes.sh
```diff
@@ -0,0 +1,240 @@
...
+				curl -O ${IMAGE_URL}-SHA512SUMS.sig
+				curl -O ${ca_url}
+				if [ "$CHECK_SIGNED_FILES" = "true" ]; then
+					if openssl smime -verify \
+					-in "${CLOUD_IMAGE}.xz-SHA512SUMS.sig" \
+					-inform DER \
+					-content "${CLOUD_IMAGE}.xz-SHA512SUMS" \
+					-CAfile "ClearLinuxRoot.pem"; then
+						unxz "${CLOUD_IMAGE}.xz"
+					else
```
pohly: As expected, the check fails for me on Debian with:
```
Verification failure
140480049168576:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:../crypto/pkcs7/pk7_smime.c:285:Verify error:unsupported certificate purpose
test/start-stop.make:5: recipe for target 'start' failed
make: *** [start] Error 2
```

Anyone else would be stuck at that point. Because I happen to know the issue, I could proceed with `CHECK_SIGNED_FILES=false make start`.

In the `else` branch we need add some output that explains what the problem is with a link to the issue that I filed and then explain how to proceed.

test/start-kubernetes.sh
```diff
@@ -0,0 +1,240 @@
+#!/bin/bash
+set -o errexit
+set -o pipefail
+
+TEST_DIRECTORY=${TEST_DIRECTORY:-$(dirname $(readlink -f $0))}
+TEST_CONFIG=${TEST_CONFIG:-${TEST_DIRECTORY}/test-config.sh}
+source ${TEST_CONFIG}
+DEPLOYMENT_SUFFIX=${DEPLOYMENT_SUFFIX}
+DEPLOYMENT_ID=${DEPLOYMENT_ID:-k8s-test${DEPLOYMENT_SUFFIX}-pmem}
```
pohly: Instead of `k8s-test` let's use something more descriptive.

I suggest `pmem-csi-<OS[-suffix]>`, which then matches with a `_work/<OS[-suffix]-govm` subdirectory.

test/start-stop.make
```diff
@@ -1,68 +1,23 @@
...
-	@ echo "The test cluster is ready. Log in with _work/$(CLUSTER)/ssh, run kubectl once logged in."
-	@ echo "Alternatively, KUBECONFIG=$$(pwd)/_work/$(CLUSTER)/kube.config can also be used directly."
+	test/start-kubernetes.sh && \
+	if ! [ -e _work/$(CLUSTER)/clear-kvm.secretsdone ] || [ $$(_work/$(CLUSTER)/ssh-clear-kvm kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
+		KUBECTL="$(PWD)/_work/$(CLUSTER)/ssh-clear-kvm kubectl" PATH='$(PWD)/_work/$(CLUSTER)/bin/:$(PATH)' ./test/setup-ca-kubernetes.sh && \
+		touch _work/$(CLUSTER)/clear-kvm.secretsdone; \
+	fi \
+	&& test/setup-deployment.sh
+
+	@ echo "The test cluster is ready. Log in with _work/ssh-clear-kvm, run kubectl once logged in."
```
pohly: When you rebased, you removed the usage of `$(CLUSTER)` and brought back the hard-coded `clear-kvm`. Please keep using that variable so that we can prepare and use multiple different clusters in parallel.

test/start-stop.make
```diff
@@ -1,68 +1,23 @@
...
+	@ echo "Alternatively, KUBECONFIG=$$(pwd)/_work/kube.config can also be used directly."
 	@ echo "To try out the pmem-csi driver persistent volumes:"
-	@ echo "   cat deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-pvc.yaml | _work/$(CLUSTER)/ssh kubectl create -f -"
-	@ echo "   cat deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-app.yaml | _work/$(CLUSTER)/ssh kubectl create -f -"
+	@ echo "   cat deploy/kubernetes-$$(cat _work/clear-kvm-kubernetes.version)/pmem-pvc.yaml | _work/ssh-clear-kvm kubectl create -f -"
```
pohly: When you rebased, you removed the usage of `$(CLUSTER)` and brought back the hard-coded `clear-kvm`. Please keep using that variable so that we can prepare and use multiple different clusters in parallel.

test/start-stop.make
```diff
@@ -1,68 +1,23 @@
...
-	( . test/test-config.sh && _work/$(CLUSTER)/ssh kubectl apply -f - <deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-csi-$${TEST_DEVICEMODE}-testing.yaml )
-	_work/$(CLUSTER)/ssh kubectl apply -f - <deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-storageclass-ext4.yaml
-	_work/$(CLUSTER)/ssh kubectl apply -f - <deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-storageclass-xfs.yaml
-	_work/$(CLUSTER)/ssh kubectl apply -f - <deploy/kubernetes-$$(cat _work/$(CLUSTER)/kubernetes.version)/pmem-storageclass-cache.yaml
-	@ echo
-	@ echo "The test cluster is ready. Log in with _work/$(CLUSTER)/ssh, run kubectl once logged in."
-	@ echo "Alternatively, KUBECONFIG=$$(pwd)/_work/$(CLUSTER)/kube.config can also be used directly."
+	test/start-kubernetes.sh && \
+	if ! [ -e _work/$(CLUSTER)/clear-kvm.secretsdone ] || [ $$(_work/$(CLUSTER)/ssh-clear-kvm kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
+		KUBECTL="$(PWD)/_work/$(CLUSTER)/ssh-clear-kvm kubectl" PATH='$(PWD)/_work/$(CLUSTER)/bin/:$(PATH)' ./test/setup-ca-kubernetes.sh && \
```
pohly: The `PATH` is wrong here. It has use `$(PWD)/_work/bin`, otherwise the cfssl that gets installed by `make start` isn't found.

It fails for me because I don't have the tool in my regular path.

test/start-stop.make
```diff
@@ -1,68 +1,23 @@
...
-# - creates pmem NVDIMM files if necessary
-# - starts QEMU virtual machines with pmem NVDIMMs as described in https://github.com/qemu/qemu/blob/bd54b11062c4baa7d2e4efadcf71b8cfd55311fd/docs/nvdimm.txt
-# - starts a Kubernetes cluster
+# - starts a Kubernetes cluster with NVDIMMs as described in https://github.com/qemu/qemu/blob/bd54b11062c4baa7d2e4efadcf71b8cfd55311fd/docs/nvdimm.txt
 # - generate pmem secrets if necessary
-# - installs pmem-csi driver
-#
-# It operates on the cluster in _work/$(CLUSTER).
-start: _work/$(CLUSTER)/created _work/$(CLUSTER)/start-kubernetes _work/$(CLUSTER)/run-qemu _work/$(CLUSTER)/ssh test/setup-ca-kubernetes.sh _work/.setupcfssl-stamp
+start: test/setup-ca-kubernetes.sh _work/.setupcfssl-stamp
```
pohly: Previously it was possible to do `make stop && make push-test-image && make start` to reboot the machine with a new image, without having to go through the expensive cluster setup process.

This command combination still works, but now re-creates the cluster. That's okay because it provides a simple way to reset the cluster. We just need something else for the developer use case above.

Can you add a `make restart`? It first needs to reboot each VM and then wait for them to come up again.

Rebooting the VMs can be done with `ssh sudo systemctl reboot`. Waiting can be done by checking with ssh and kubectl.

test/start-stop.make
```diff
@@ -1,68 +1,23 @@
...
-	@ echo
-	@ echo "The test cluster is ready. Log in with _work/$(CLUSTER)/ssh, run kubectl once logged in."
-	@ echo "Alternatively, KUBECONFIG=$$(pwd)/_work/$(CLUSTER)/kube.config can also be used directly."
+	test/start-kubernetes.sh && \
+	if ! [ -e _work/$(CLUSTER)/clear-kvm.secretsdone ] || [ $$(_work/$(CLUSTER)/ssh-clear-kvm kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
+		KUBECTL="$(PWD)/_work/$(CLUSTER)/ssh-clear-kvm kubectl" PATH='$(PWD)/_work/$(CLUSTER)/bin/:$(PATH)' ./test/setup-ca-kubernetes.sh && \
+		touch _work/$(CLUSTER)/clear-kvm.secretsdone; \
+	fi \
+	&& test/setup-deployment.sh
+
```
pohly: Redundant empty line. Make seems to tolerate it, but remove it anyway.

test/start-kubernetes.sh
chuyd: Hi I added a message explaining the problem and referring to the issue

test/start-kubernetes.sh
chuyd: I updated the name

test/start-stop.make
chuyd: I added the use of $CLUSTER

test/start-stop.make
chuyd: I updated the path

test/start-stop.make
chuyd: I added `make restart` to reboot the cluster,  it waits for all the nodes to have ssh, then it verifies that all the kubernetes nodes are ready

test/start-stop.make
chuyd: I removed the empty line,  and modified the the govm yaml file to save the deployments info into the cluster directory _workdir/$CLUSTER, this way all the deployment info can be deleted by runing `make stop` and deleting the _workdir/$CLUSTER

pohly:
There's some odd extra output during "make start":
```
Your public key has been saved in /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa.pub.
The key fingerprint is:
SHA256:SfufZ8WVgdObvTHbxOhshFLF1OgClSOPgxe+afPRT5w pohly@pohly-mobl1
The key's randomart image is:
+---[RSA 2048]----+
|           ..==o |
|          + +ooo.|
|        .o B +.o*|
|       ..o* + +*=|
|        S. = * +O|
|         .= . =E*|
|         ..o o + |
|           ...o .|
|            oo   |
+----[SHA256]-----+
/work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm /work/gopath/src/github.com/intel/pmem-csi
Downloading clear-29440-cloud.img.xz image
```

Why does it print `/work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm /work/gopath/src/github.com/intel/pmem-csi`? Some debug output perhaps that can be removed?

Does it still disable auto-update on Clear Linux?

`test/setup-deployment.sh` seems to run with -x:
```
+ CLUSTER=clear-govm
++++ readlink -f test/setup-deployment.sh
...
```
I think that's to verbose and can be removed.

```
storageclass.storage.k8s.io/pmem-csi-sc-cache created
The test cluster is ready. Log in with _work/ssh-clear-govm, run kubectl once logged in.
Alternatively, KUBECONFIG=/work/gopath/src/github.com/intel/pmem-csi/_work/kube.config can also be used directly.
To try out the pmem-csi driver persistent volumes:
cat: _work/clear-govm/kubernetes.version: No such file or directory
   cat deploy/kubernetes-/pmem-pvc.yaml | _work/ssh-clear-govm kubectl create -f -
...
```

That error around kubernetes.version appears a few times.

The kube.config file needs to be inside the $CLUSTER directory.

Can you rename the ssh wrapper scripts back to `$CLUSTER/ssh.<number from 0>`? That is shorter and makes it easier to call them from a test.

test/start-kubernetes.sh
```diff
@@ -0,0 +1,298 @@
...
+	chmod +x $DELETE_VMS_SCRIPT
+
+	#Create script to restart virtual machines
+	cat <<-EOF>$RESTART_VMS_SCRIPT
+	#!/bin/bash
+	echo "Rebooting virtual machines"
+	for ip in $(echo ${IPS}); do
+		ssh $SSH_ARGS ${CLOUD_USER}@\${ip} "sudo systemctl reboot"
+	done
+	echo "Waitig for ssh connectivity"
```
pohly: Waitig -> Waiting

pohly:
I wonder whether `make stop` should remove the `_work/$CLUSTER` directory. Probably yes? That then would be consistent with the semantic of `make start` = "create cluster" and `make stop` = "delete cluster".

test/start-kubernetes.sh
chuyd: Changed

chuyd:
> I wonder whether `make stop` should remove the `_work/$CLUSTER` directory. Probably yes? That then would be consistent with the semantic of `make start` = "create cluster" and `make stop` = "delete cluster".

Yes, it makes sense to remove the _work/$CLUSTER directory, changed.

> There's some odd extra output during "make start":
> 
> ```
> Your public key has been saved in /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa.pub.
> The key fingerprint is:
> SHA256:SfufZ8WVgdObvTHbxOhshFLF1OgClSOPgxe+afPRT5w pohly@pohly-mobl1
> The key's randomart image is:
> +---[RSA 2048]----+
> |           ..==o |
> |          + +ooo.|
> |        .o B +.o*|
> |       ..o* + +*=|
> |        S. = * +O|
> |         .= . =E*|
> |         ..o o + |
> |           ...o .|
> |            oo   |
> +----[SHA256]-----+
> /work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm /work/gopath/src/github.com/intel/pmem-csi
> Downloading clear-29440-cloud.img.xz image
> ```
> 
> Why does it print `/work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm /work/gopath/src/github.com/intel/pmem-csi`? Some debug output perhaps that can be removed?

It was information prompted using pushd/popd I removed this output in the latest commit.

> Does it still disable auto-update on Clear Linux?

Yes, it disables the auto-update

> `test/setup-deployment.sh` seems to run with -x:
> 
> ```
> + CLUSTER=clear-govm
> ++++ readlink -f test/setup-deployment.sh
> ...
> ```
> 
> I think that's to verbose and can be removed.

Verbose removed 

> ```
> storageclass.storage.k8s.io/pmem-csi-sc-cache created
> The test cluster is ready. Log in with _work/ssh-clear-govm, run kubectl once logged in.
> Alternatively, KUBECONFIG=/work/gopath/src/github.com/intel/pmem-csi/_work/kube.config can also be used directly.
> To try out the pmem-csi driver persistent volumes:
> cat: _work/clear-govm/kubernetes.version: No such file or directory
>    cat deploy/kubernetes-/pmem-pvc.yaml | _work/ssh-clear-govm kubectl create -f -
> ...
> ```
> 
> That error around kubernetes.version appears a few times.

Error fixed

> The kube.config file needs to be inside the $CLUSTER directory.
Text updated
> Can you rename the ssh wrapper scripts back to `$CLUSTER/ssh.<number from 0>`? That is shorter and makes it easier to call them from a test.

Wrapper scripts renamed

pohly:
After `make restart`, should all nodes be ready again?

I thought the command waited for nodes, but in practice that wasn't the case:
```
Waiting for ssh connectivity
NAME                          READY   STATUS                  RESTARTS   AGE
pmem-csi-controller-0         0/3     Error                   2          26m
pmem-csi-node-4wtvq           0/2     Init:ImagePullBackOff   0          26m
pmem-csi-node-7qwh2           0/2     Init:ImagePullBackOff   0          25m
pmem-csi-node-8pvc7           0/2     Init:ImagePullBackOff   0          25m
pmem-csi-node-testing-7b5r4   1/1     Running                 2          25m
pmem-csi-node-testing-88fsd   1/1     Running                 2          26m
pmem-csi-node-testing-hnwsb   1/1     Running                 2          25m
Waiting for kubernetes nodes to be ready
$ _work/clear-govm/ssh.0 
ssh_exchange_identification: Connection closed by remote host
...
```

Eventually the login worked again.

pohly:
```
To try out the pmem-csi driver persistent volumes:
   cat deploy/kubernetes-1.14/pmem-pvc.yaml | _work/ssh-clear-govm kubectl create -f -
...
```

Wrong ssh script name here.

pohly:
> After make restart, should all nodes be ready again?

The problem is that `restarts.sh` ends with:
```
while [ $SECONDS -lt 60 ];do
sleep 3
pending_nodes=$(/work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm/ssh-clear-govm "kubectl get nodes  -o go-template           --template='{{range .items}}{{range .status.conditions }}{{if eq .type \"Ready\"}}              {{if ne .status \"True\"}}{{printf \"%s\n\" .reason}}{{end}}{{end}}{{end}}{{end}}'"2>/dev/null)
if [ -z "$pending_nodes" ]; then
break
fi
done
exec ssh -oIdentitiesOnly=yes -oStrictHostKeyChecking=no        -oUserKnownHostsFile=/dev/null -oLogLevel=error         -i /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa clear@172.17.0.2 "sudo systemctl reboot"
exec ssh -oIdentitiesOnly=yes -oStrictHostKeyChecking=no        -oUserKnownHostsFile=/dev/null -oLogLevel=error         -i /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa clear@172.17.0.3 "sudo systemctl reboot"
exec ssh -oIdentitiesOnly=yes -oStrictHostKeyChecking=no        -oUserKnownHostsFile=/dev/null -oLogLevel=error         -i /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa clear@172.17.0.4 "sudo systemctl reboot"
exec ssh -oIdentitiesOnly=yes -oStrictHostKeyChecking=no        -oUserKnownHostsFile=/dev/null -oLogLevel=error         -i /work/gopath/src/github.com/intel/pmem-csi/_work/resources/id_rsa clear@172.17.0.5 "sudo systemctl reboot"
```

Those four "exec ssh" calls shouldn't be there. Only the first one gets executed (because of `exec` replacing the shell interpreter), but that is enough to trigger another reboot of the first node.

pohly:
Creating a second cluster with a different name fails:
```
$ CHECK_SIGNED_FILES=false make start CLUSTER=clear-govm-xyz
. test/test-config.sh && \
test/start-kubernetes.sh && \
if ! [ -e _work/clear-govm-xyz/secretsdone ] || [ $(_work/clear-govm-xyz/ssh-clear-govm-xyz kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
	KUBECTL="/work/gopath/src/github.com/intel/pmem-csi/_work/clear-govm-xyz/ssh-clear-govm-xyz kubectl" PATH='/work/gopath/src/github.com/intel/pmem-csi/_work/bin/:/work/gopath/go/bin:/work/protexIP/bin:/work/protexIP/bin:/work/protexIP/bin:/work/protextIP/protexIP/bin:/home/pohly/bin:/home/pohly/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games' ./test/setup-ca-kubernetes.sh && \
	touch _work/clear-govm-xyz/secretsdone; \
fi \
&& test/setup-deployment.sh
Kubernetes cluster clear-govm is already deployed, to start a new
cluster first stop the current cluster by running:
CLUSTER=clear-govm make stop
```

The invocation above sets `CLUSTER` as make variable. The Makefile must ensure that it passes that variable to scripts which need it. Currently doesn't do that, but that can be added with:
```
$ git diff
diff --git a/Makefile b/Makefile
index c63e872e..e4715196 100644
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ clean:
 # All clusters called "clear-kvm[-something]" are created with
 # test/clear-kvm.make. They run inside QEMU and share the
 # same IP addresses, and thus cannot run in parallel.
-CLUSTER := clear-govm
+export CLUSTER := clear-govm
```

However, that alone wasn't enough. It looks like some code still uses a hard-coded `clear-govm` name instead of `$CLUSTER`.

pohly:
`git grep clear-kvm` finds some instances of comments and documentation that need to be updated.

chuyd:
> ```
> To try out the pmem-csi driver persistent volumes:
>    cat deploy/kubernetes-1.14/pmem-pvc.yaml | _work/ssh-clear-govm kubectl create -f -
> ...
> ```
> 
> Wrong ssh script name here.

ssh script path updated


> > After make restart, should all nodes be ready again?

Yes, `kubectl get nodes  -o go-template           --template='{{range .items}}{{range .status.conditions }}{{if eq .type \"Ready\"}}              {{if ne .status \"True\"}}{{printf \"%s\n\" .reason}}{{end}}{{end}}{{end}}{{end}}'` gets all the nodes that are not ready

> Those four "exec ssh" calls shouldn't be there. Only the first one gets executed (because of `exec` replacing the shell interpreter), but that is enough to trigger another reboot of the first node.

exec removed

> However, that alone wasn't enough. It looks like some code still uses a hard-coded `clear-govm` name instead of `$CLUSTER`.

`CLUSTER=custom-cluster make start
. test/test-config.sh && \
test/start-kubernetes.sh && \
if ! [ -e _work/custom-cluster/secretsdone ] || [ $(_work/custom-cluster/ssh-custom-cluster kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
        KUBECTL="/home/jenkins-worker-1/pmem-CSI/_work/custom-cluster/ssh-custom-cluster kubectl" PATH='/home/jenkins-worker-1/pmem-CSI/_work/bin/:/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/h
ome/jenkins-worker-1/.local/bin:/home/jenkins-worker-1/bin:/home/jenkins-worker-1/.local/bin:/home/jenkins-worker-1/bin' ./test/setup-ca-kubernetes.sh && \
        touch _work/custom-cluster/secretsdone; \
fi \
&& test/setup-deployment.sh
clear-29480-cloud.img found, skipping download
INFO[0000] GoVM Instance pmem-csi-custom-cluster-master has been successfully created
INFO[0001] GoVM Instance pmem-csi-custom-cluster-worker1 has been successfully created
INFO[0001] GoVM Instance pmem-csi-custom-cluster-worker2 has been successfully created
INFO[0002] GoVM Instance pmem-csi-custom-cluster-worker3 has been successfully created`

test/test-config.sh dind't check if the variable was set
-CLUSTER=clear-govm
+CLUSTER=${CLUSTER:-clear-govm}

pohly:
One observation (but that can be handled separately): while the four machines set up Kubernetes, the output is streamed through sed. However, buffering causes delays when printing each line, so one doesn't actually see in real time what's going one.

okartau:
> can you perhaps give it a try

Tried, seeing this:
```
Downloading clear-29540-cloud.img.xz image
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 54.5M  100 54.5M    0     0   578k      0  0:01:36  0:01:36 --:--:-- 1121k
Verification failure
140195340121920:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:crypto/pkcs7/pk7_smime.c:285:Verify error:unsupported certificate purpose
Image verification failed, does not work at the moment,
(https://github.com/clearlinux/distribution/issues/85)
to skip image verification run:
CHECK_SIGNED_FILES=false make start
make: *** [test/start-stop.make:5: start] Error 2
```
This seems to be our own check in start-kubernetes.sh, points to github issue 85.
I see that was discussed above, should this just work,
or do I need to set CHECK_SIGNED_FILES=false to make this work?
It would be friendlier if this worked without need to always alter certain default setting.

P.S. As pointed issue has been closed in August 2018, does the wording "at the moment" make sense?

test/start-kubernetes.sh
```diff
@@ -0,0 +1,301 @@
...
+		echo "$CLOUD_IMAGE found, skipping download"
+	else
+		case $CLOUD_USER in
+			clear)
+				ca_url="https://cdn.download.clearlinux.org/"
+				ca_url+="releases/${IMAGE_URL//[^0-9]}/"
+				ca_url+="clear/ClearLinuxRoot.pem"
+				echo "Downloading ${CLOUD_IMAGE}.xz image"
+				curl -O ${IMAGE_URL}
+				curl -s -O ${IMAGE_URL}-SHA512SUMS
```
okartau: downloaded SHA512 and ca-related files are only needed inside next 'if', so download cmds can be moved there?

test/start-kubernetes.sh
```diff
@@ -0,0 +1,301 @@
...
+		-machine pc,accel=kvm,nvdimm=on}"
+EXTRA_QEMU_OPTS="${EXTRA_QWEMU_OPTS:-\
+		-object memory-backend-file,id=mem1,share=${TEST_PMEM_SHARE:-on},\
+		mem-path=/data/nvdimm0,size=${TEST_PMEM_MEM_SIZE:-32768}M \
+		-device nvdimm,id=nvdimm1,memdev=mem1,label-size=${TEST_PMEM_LABEL_SIZE:-2097152} \
+		-machine pc,nvdimm}"
+CLOUD_USER=${CLOUD_USER:-clear}
+CLOUD_IMAGE="${CLOUD_IMAGE:-$(\
+		curl -s https://download.clearlinux.org/image/latest-images |\
+		awk '/cloud.img/ {gsub(".xz",""); print $0}')}"
```
okartau: here we take CLOUD_IMAGE name, strip off .gz, and then numerous time add it back when using this variable, but only once we use the CLOUD_IMAGE as it is (down when creating govm description). It would reduce changes if we maintained the name with .xz part and and remove it in that one occasion

test/start-kubernetes.sh
```diff
@@ -0,0 +1,301 @@
...
+IMAGE_URL=${IMAGE_URL:-http://download.clearlinux.org/image/${CLOUD_IMAGE}.xz}
```
okartau: The method loading image from download.clearlinux.org/image/ has changed compared to previous method (used download.clearlinux.org/releases).
This is problematic as now it's impossible to load other images than latest, as download.clearlinux.org/image/ maintains one latest image only.
I propose we revert back load images from download.clearlinux.org/releases/ where all releases are accessible on same base path

okartau:
I added some inline comments, some are minor, but the inability to load previous images is major thing. In past we have had times where we need to stick to older image than later.
Another typical case is use of mirror, when mirror does not have newest images synced yet.

test/start-kubernetes.sh
pohly: Agreed. I hadn't tested installing an older version. Good catch.

pohly:
> This seems to be our own check in start-kubernetes.sh, points to github issue 85.
> I see that was discussed above, should this just work,
> or do I need to set CHECK_SIGNED_FILES=false to make this work?

That variable must be set on some host systems.

> It would be friendlier if this worked without need to always alter certain default setting.

That would then imply that we default to "don't check", which isn't a good default.

The solution is a setup.sh.d file where developers on affected systems set that variable, without checking that into git. That, or always remembering to set the variable explicitly.

> P.S. As pointed issue has been closed in August 2018, does the wording "at the moment" make sense?

Agreed, the wording could be improved. The message also implies that the verification errors is *always* something that can be ignored, which is not the case.

I propose this:
```
Verification failure
140195340121920:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:crypto/pkcs7/pk7_smime.c:285:Verify error:unsupported certificate purpose

Image verification failed, see error above.

"unsupported certificate purpose" is a known issue caused by an incompatible openssl
version (https://github.com/clearlinux/distribution/issues/85). Other errors might indicate
a download error or man-in-the-middle attack.

To skip image verification run:
CHECK_SIGNED_FILES=false make start
```

test/start-kubernetes.sh
chuyd: Thanks, I moved the SHA512  and ca-related files there

test/start-kubernetes.sh
chuyd: I changed the code to maintain the .xz

test/start-kubernetes.sh
chuyd: I updated the IMAGE_URL variable to use download.clearlinux.org/releases/ `CLOUD_IMAGE=clear-<VERSION>-cloud.img.xz make start` can download previous versions

chuyd:
> I propose this:
> 
> ```
> Verification failure
> 140195340121920:error:21075075:PKCS7 routines:PKCS7_verify:certificate verify error:crypto/pkcs7/pk7_smime.c:285:Verify error:unsupported certificate purpose
> 
> Image verification failed, see error above.
> 
> "unsupported certificate purpose" is a known issue caused by an incompatible openssl
> version (https://github.com/clearlinux/distribution/issues/85). Other errors might indicate
> a download error or man-in-the-middle attack.
> 
> To skip image verification run:
> CHECK_SIGNED_FILES=false make start
> ```

I added the message

okartau:
Thanks, recently mentioned areas are improved, I get few steps further, but still not through to end yet.

My workflow involves loading images from external registry server, and that support is not complete.
We have TEST_INSECURE_REGISTRIES config variable, used for adding entries. This adds entry in CRI case, but not in docker case, so docker config in /etc/docker/daemon.json remains without added REGISTRY.

Note also about CRI config: resulting format is non-uniform:
```
# cat /etc/containers/registries.conf
[registries.insecure]
registries = [ \'172.17.0.2:5000\',"10.123.123.123:5000" ]
```
is it intentional there are quoted `\'` around first address? does it work?

There is more about added registry address:
Deploy uses REGISTRY_IP in replacement meaning as well, making two variables noting same thing.
I propose we change the names so that we only use one variable, as (works at least in my workflow)
I use own registry which needs to be substituted both in REGISTRY_IP and TEST_INSECURE_REGISTRIES

Note that the script that now uses REGISTRY_IP, does not take values from test/test-config.sh,
but needs value given in env when starting `make`

Long story summarized:
use single addr:port value for separate registry users so that:
- can be specified in test/test-config.sh,(or alternatively on cmd line as env value)
- overrides deploy address in deploy commands
- gets added as insecure registry addr, both in cri and docker configs,

test/setup-clear-govm.sh
```diff
@@ -7,7 +7,7 @@ set -o pipefail
 
 HOSTNAME=${HOSTNAME:-$1}
 TEST_CRI=${TEST_CRI:-docker}
-TEST_IP_ADDR=${TEST_IP_ADDR:-127.0.0.1}
+TEST_INSECURE_REGISTRIES=${TEST_INSECURE_REGISTRIES:-127.0.0.1:5000}
```
pohly: This changes the semantic of `TEST_INSECURE_REGISTRIES` from "add additional insecure registries" to "override the default insecure registry".

The Intel-internal pmem-redis BKM was meant to be used with the original semantic: the local, default registry has the PMEM-CSI image, and `TEST_INSECURE_REGISTRIES` is set to enable access to the additional registry which has the pmem-redis image.

I'm leaning towards keeping the original semantic, because then the local registry will always work correctly without extra effort.

@okartau: what do you think?

test/start-stop.make
```diff
@@ -1,68 +1,27 @@
...
 
 stop:
-	for i in $$(seq 0 $$(($(NUM_NODES) - 1))); do \
-		if [ -e _work/$(CLUSTER)/qemu.$$i.pid ]; then \
-			kill -9 $$(cat _work/$(CLUSTER)/qemu.$$i.pid) 2>/dev/null; \
-			rm -f _work/$(CLUSTER)/qemu.$$i.pid; \
-		fi; \
-	done
+	. test/test-config.sh && \
+	_work/$(CLUSTER)/stop.sh
```
pohly: I just noticed that stopping twice leads to a rather unhelpful error message:
```
$ make stop
. test/test-config.sh && \
_work/clear-govm/stop.sh
bash: line 1: _work/clear-govm/stop.sh: No such file or directory
test/start-stop.make:22: recipe for target 'stop' failed
make: *** [stop] Error 127
```

`make start` twice leads to a failure, so I think it is adequate to also fail when `make stop` is called again, but it should be with a clear error message along the lines of "cluster $(CLUSTER) has already been stopped".

test/start-stop.make
```diff
@@ -1,68 +1,27 @@
...
+
+restart:
+	. test/test-config.sh && \
+	_work/$(CLUSTER)/restart.sh
```
pohly: Same here: if `	_work/$(CLUSTER)/restart.sh` does not exist, print an explicit error message.

test/start-kubernetes.sh
```diff
@@ -0,0 +1,305 @@
...
+CLOUD_IMAGE="${CLOUD_IMAGE:-$(\
+		curl -s https://download.clearlinux.org/image/latest-images |\
+		awk '/cloud.img/ {print $0}')}"
+IMAGE_URL=${IMAGE_URL:-https://download.clearlinux.org/releases/${CLOUD_IMAGE//[!0-9]/}/clear}
+SSH_TIMEOUT=60
+SSH_ARGS="-oIdentitiesOnly=yes -oStrictHostKeyChecking=no \
+	-oUserKnownHostsFile=/dev/null -oLogLevel=error \
+	-i ${SSH_KEY}"
+CREATE_REGISTRY=${CREATE_REGISTRY:-false}
+CHECK_SIGNED_FILES=${CHECK_SIGNED_FILES:-true}
```
pohly: Another thing that I just realized: some of the config variables have `TEST_` as prefix (`TEST_INSECURE_REGISTRIES`), others don't (`CHECK_SIGNED_FILES`).

I used the `TEST_` prefix to reduce the risk of conflicts with env variables that a developer might have set in his environment. Does that make sense? Should all variables have that prefix?

Or none of them?

Either way is fine, we just should avoid the inconsistency. For example, this is what I have now in a config file:
```
$ cat test/test-config.d/my-config.sh
TEST_INSECURE_REGISTRIES="xxxxxx.intel.com"
CHECK_SIGNED_FILES=false
```

okartau: I vote for having TEST_ prefix in related variable names

test/setup-clear-govm.sh
okartau: I had not been thinking about needing original and added registry, as I use "override" in my own workflow. But sure, "add additional" is more flexible and thus makes more sense.

okartau:
I tried the latest state in my workflow.
I can get support for my own registry working by specifying env.value on cmd line like this:
`TEST_INSECURE_REGISTRIES=10.123.123.123:5000 CHECK_SIGNED_FILES=false make start`
But I can't reach working deployment using `TEST_INSECURE_REGISTRIES=10.123.123.123:5000` in test/test-config.sh. In that case deployment attempts default addresses 172.x.x.x.
The reason seems to be that test/setup-deployment.sh does not process setting from test/test-config.sh.
I believe the intention still is to support configurable values equally by both methods.

pohly:
I'm no longer sure whether we agreed that `make start` and `make stop` should or should not be idempotent.

Earlier today I said that `make stop` should provide a clear error message because `make start` also errors out.

I just noticed that this is problematic, because `make test_e2e` fails when cluster is already running:
```
$ make test_e2e
if [ ! -e '_work/bin/govm' ]; then \
	tmpdir=`mktemp -d` && \
	trap 'rm -r $tmpdir' EXIT && \
	cd $tmpdir && \
	echo "module govm" > go.mod && \
	go get -v github.com/govm-project/govm@1166148359ed9b4b83df555e528aad3cd1144ed3 && \
	go build -o /nvme/gopath/src/github.com/intel/pmem-csi/_work/bin/govm github.com/govm-project/govm;\
fi
. test/test-config.sh && \
test/start-kubernetes.sh && \
if ! [ -e _work/clear-govm/secretsdone ] || [ $(_work/clear-govm/ssh-clear-govm kubectl get secrets | grep -e pmem-csi-node-secrets -e pmem-csi-registry-secrets | wc -l) -ne 2 ]; then \
	KUBECTL="/nvme/gopath/src/github.com/intel/pmem-csi/_work/clear-govm/ssh-clear-govm kubectl" PATH='/nvme/gopath/src/github.com/intel/pmem-csi/_work/bin/:/fast/work/blackduck/protexIP/bin:/home/pohly/bin:/home/pohly/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games' ./test/setup-ca-kubernetes.sh && \
	touch _work/clear-govm/secretsdone; \
fi \
&& test/setup-deployment.sh
Kubernetes cluster clear-govm is already deployed, to start a new
cluster first stop the current cluster by running:
CLUSTER=clear-govm make stop
test/start-stop.make:5: recipe for target 'start' failed
make: *** [start] Error 1
```

I think I prefer the original behavior, i.e.:
- `make start` creates the cluster if needed
- `make start` updates the driver each time
- `make stop` does nothing if already stopped

I see some advantages in this:
- I don't need to remember whether the cluster is already running before starting or stopping it.
- If I don't remember how to use it,  `make start` will always tell me.
- Some (within the limits of `kubectl apply`) config changes can be applied on the fly.

@okartau any preference from you?

okartau:
> any preference from you?

I agree with above ideas. 'make start' should not fail if cluster already runs, my routine has been to run `make start_e2e` on a running cluster and that would never work if `start` is sub-target and keeps failing.

test/setup-clear-govm.sh
pohly: The default (pulling images from `172.17.0.2`) does not work for me. I'm not sure whether it actually ever did :man_shrugging: 

`172.17.0.2` is the IP address of the first node in the virtual cluster. That's not where I am pushing the images. This comes from `setup-deployment.sh`:
```
TEST_INSECURE_REGISTRIES=${TEST_INSECURE_REGISTRIES:-$(awk '{NF=NF-1;gsub(".*@","");print $NF}' ${WORK_DIRECTORY}/ssh-${CLUSTER}):5000}
...
sed "s|${TEMPLATE_IP}|${TEST_INSECURE_REGISTRIES// */}|g" ${DEPLOYMENT_DIRECTORY}/${deployment_file} | ${KUBECTL} apply -f -
```

Ah, this special interpretation of `TEST_INSECURE_REGISTRIES` was added in 39f05be73416563726d297b6f6a708d28ef703e2. Okay, so one could argue that this is a user error and I should have updated my configuration. @okartau is that how you got it working?

I find this special meaning of the first entry surprising. Nothing in the name suggests that this meaning exists, and it prevents configuring a secure registry to pull the PMEM images from.

My proposal is this:
- Stop using `192.168.8.1:5000` as registry in our deployment files and replace it with `PMEM_REGISTRY`; the fixed IP address used to be a working default when using the `test/qemu-ifup` script, but now it is no longer useful. An explicit variable that is easy to replace in a sed expression is better.
- `ip=$(_work/clear-govm/ssh.0 ip addr show dev ens5 | grep " inet " | sed -e 's/.* inet //' -e 's;/.*;;' -e 's;\.2;.1;')` is the IP address for the host as seen by the cluster. This with port 5000 is always added as insecure registry.
- `TEST_PMEM_REGISTRY` defaults to that same ip:5000, but can be overridden (which seems to be what @okartau needs)
- `TEST_INSECURE_REGISTRIES` adds additional registries (as before)

One more problem with the current code: I ended up with just an IP address in `/etc/docker/daemon.json`, but Docker also wanted the port number there.

@okartau, @chuyd: does this make sense? I can prepare a patch on top of the current branch.

okartau: seems I created more confusion with my request to combine two registry addresses. Let's go back to two variables if that helps.
About port number: I got cluster created in my workflow by prepending that on cmd line:
`TEST_INSECURE_REGISTRIES=10.123.123.123:5000`
so yes, I used port number there.
But I have noticed current code uses hard-coded 5000 in numerous places, is use of hard-coded port value good idea?

pohly: My proposal is to hard-code 5000 only in the default value and nowhere else.

pohly:
After looking at the files locally I noticed that indention is inconsistent. Sometimes it uses spaces, sometimes tabs, and it does not always line up well.

Let's format consistently, with four spaces for shell code as in the other shell scripts.

pohly:
> Note that the script that now uses REGISTRY_IP, does not take values from test/test-config.sh,
> but needs value given in env when starting make

That's because the `. test-config.sh && <shell script>` idiom in `start-stop.make` is broken: the variables do not get exported by `test-config.sh` and thus aren't seen by the following shell scripts. Each shell script must source the config file. I'll fix this.

pohly:
I think I have implemented all of the proposals above in these commits:
* https://github.com/pohly/pmem-CSI/commit/77df7ffd85e776f105119e3ee6099d430b547b51
* https://github.com/pohly/pmem-CSI/commit/48d0e0f947a863cf4f73b34185ee6edfc99ffae1
* https://github.com/pohly/pmem-CSI/commit/c69919130e77bdd41f05371c86d863f585e168b7

The actual branch needs to be rebased (conflicts with devel otherwise) and squashed.

pohly:
A rebased and squashed commit is here: https://github.com/pohly/pmem-CSI/commits/govm

@okartau: can you test that one?
@chuyd: do you agree that my changes above make sense? If yes, then please replace your branch's content with https://github.com/pohly/pmem-CSI/commits/govm to update this PR.

pohly:
> > Those four "exec ssh" calls shouldn't be there. Only the first one gets executed (because of exec
> >replacing the shell interpreter), but that is enough to trigger another reboot of the first node.
>
> exec removed

I meant "nothing of this" should be there. Why bother rebooting the machines, waiting for them to come up, and then reboot them *again*?

pohly:
> I meant "nothing of this" should be there.

Fixed in the rebased branch.

I also reformatted the code (no tabs) and removed clever, but hard-to-read tricks like `${KVM_CPU_OPTS//$(echo -e "\t")/}`.

okartau:
I verified I can get cluster up by specifying my own registry on cmd line as TEST_PMEM_REGISTRY and  TEST_INSECURE_REGISTRIES, that is good.
I want to repeat also with edited test-config.sh, this is ongoing.

One thing I notice, probably just cosmetic effect and not functional:
NO_PROXY setting in node docker config will have double comma.
This seems to be caused by fact that 
/usr/share/clr-k8s-examples/setup_system.sh
makes own pass over interfaces adding those to NO_PROXY:
Environment="NO_PROXY=${no_proxy},${ADD_NO_PROXY}"

This loop leaves extra comma at the end.
Our script then adds node addresses, adding comma before each entry.
As result, we will have double comma in the middle.
But fixing that will need some ugly check to be added,
otherwise we will end up with extra comma somewhere (we can move it to end).

okartau:
We don't want to support scenario where I can stop VMs and start those later without re-create?
do I recall we have such possibility in current setup?
"make start" is actually create + start
"make stop" is actually stop + destroy.
i.e. we have embedded create and destroy to be part of start and stop,
without possibility to make those detached.

One could expect that "stop" will stop VMs with possibility to start VMs again without reinstall.

okartau:
I repeated cluster creation with changed values in test-config.sh, succeeds.

pohly:
> One thing I notice, probably just cosmetic effect and not functional:
> NO_PROXY setting in node docker config will have double comma.

We could sanitize that in our Makefile with `sed -e 's/,,*/,/'`.

> We don't want to support scenario where I can stop VMs and start those later without re-create?

No, that's not possible anymore. We would need some kind of "suspend" support in GoVM for that.

> do I recall we have such possibility in current setup?

Yes, the current semantic of "start/stop" is more like "optionally create and start" and "stop running machines".

I agree that it would be nice to have it, but as long as we all run this on sufficiently capable build machines, I am not worried about having those clusters active in the background.

okartau:
ok, let's adjust our habits to this start/stop semantics.
The code now works for my workflow, I already used it in few recent runs.
Ready to merge after double-comma fix?

pohly:
> ok, let's adjust our habits to this start/stop semantics.
> The code now works for my workflow, I already used it in few recent runs.
> Ready to merge after double-comma fix?

We still need @chuyd to update his branch. We could create a separate PR and merge that instead, but that should be our last resort if @chuyd doesn't get around to it. I'd prefer to keep everything in this PR and also to have him check the latest state. The "docker registry in master" feature is something that he added for the CI and I haven't tested that.

Regarding the double-comma: as that isn't breaking anything, let's fix it later. At some point we need to stop making cosmetic changes and get it merged to unblock other next steps. For example, my disruptive testing (PR #237) also needs this to be merged first.

pohly:
@chuyd pointed out on Slack that indeed, TEST_CREATE_REGISTRY support was broken. The following changes fix this for me (tried both with and without):
```
diff --git a/test/setup-clear-govm.sh b/test/setup-clear-govm.sh
index f98045f0..cc4f47fa 100755
--- a/test/setup-clear-govm.sh
+++ b/test/setup-clear-govm.sh
@@ -71,13 +71,13 @@ EOF"
     sudo mkdir -p /etc/containers
     sudo bash -c "cat >/etc/containers/registries.conf <<EOF
 [registries.insecure]
-registries = [ $(echo $TEST_INSECURE_REGISTRIES $TEST_LOCAL_REGISTRY | sed 's|^|"|g;s| |", "|g;s|$|"|') ]
+registries = [ $(echo $INSECURE_REGISTRIES | sed 's|^|"|g;s| |", "|g;s|$|"|') ]
 EOF"
 
     # The same for Docker.
     sudo mkdir -p /etc/docker
     sudo bash -c "cat >/etc/docker/daemon.json <<EOF
-{ \"insecure-registries\": [ $(echo $TEST_INSECURE_REGISTRIES $TEST_LOCAL_REGISTRY | sed 's|^|"|g;s| |", "|g;s|$|"|') ] }
+{ \"insecure-registries\": [ $(echo $INSECURE_REGISTRIES | sed 's|^|"|g;s| |", "|g;s|$|"|') ] }
 EOF"
 
     # Proxy settings for Docker.
diff --git a/test/start-kubernetes.sh b/test/start-kubernetes.sh
index 1f7c6ca8..bc1858f0 100755
--- a/test/start-kubernetes.sh
+++ b/test/start-kubernetes.sh
@@ -239,6 +239,8 @@ EOF
             chmod +x ${WORKING_DIRECTORY}/ssh-${CLUSTER}
         fi
         ENV_VARS="env$(env_vars) HOSTNAME='$vm_name' IPADDR='$ip'"
+        # The local registry and the master node might be used as insecure registries, enabled that just in case.
+        ENV_VARS+=" INSECURE_REGISTRIES='$TEST_INSECURE_REGISTRIES $TEST_LOCAL_REGISTRY pmem-csi-$CLUSTER-master:5000'"
         scp $SSH_ARGS ${TEST_DIRECTORY}/{$setup_script,$install_k8s_script} ${CLOUD_USER}@${ip}:. >/dev/null
         ssh $SSH_ARGS ${CLOUD_USER}@${ip} "$ENV_VARS ./$setup_script && $ENV_VARS ./$install_k8s_script" &> >(sed -e "s/^/$vm_name: /" | tee -a $log_name ) &
         cat <<EOF >$ssh_script
@@ -254,16 +256,23 @@ EOF
     export KUBECONFIG=${KUBECONFIG}
     # Copy images to local registry in master vm?
     if $TEST_CREATE_REGISTRY; then
-        images=( pmem-csi-driver pmem-vgm pmem-ns-init)
-        for image in "${images[@]}" ; do
-            echo "Saving $image"
-            docker save ${TEST_PMEM_REGISTRY}/$image:$IMAGE_TAG > ${WORKING_DIRECTORY}/$image
-            echo Copying image $image to master node
-            scp $SSH_ARGS $image ${CLOUD_USER}@${master_ip}:.
+        for image in $TEST_BOOTSTRAP_IMAGES; do
+            # The image might have been build for a certain registry (like my-registry-server:5000).
+            # We need to re-tag it before pushing it to the localhost:5000 registry on the master node.
+            remoteimage="$(echo "$image" | sed -e 's;^[^/]*/;localhost:5000/;')"
+            echo "Copying $image to master node"
+            docker save "$image" | ssh $SSH_ARGS ${CLOUD_USER}@${master_ip} sudo docker load
             echo Load $image into registry
-            ssh $SSH_ARGS ${CLOUD_USER}@${master_ip} "sudo docker load < $image"
-            ssh $SSH_ARGS ${CLOUD_USER}@${master_ip} "sudo docker push localhost:5000/$image:$IMAGE_TAG"
-            rm ${WORKING_DIRECTORY}/$image
+            ssh $SSH_ARGS ${CLOUD_USER}@${master_ip} sudo docker tag "$image" "$remoteimage"
+            ssh $SSH_ARGS ${CLOUD_USER}@${master_ip} sudo docker push "$remoteimage"
+        done
+
+        # TEST_PMEM_REGISTRY in test-config.sh uses this machine name as registry,
+        # so we need to ensure that the name can be resolved.
+        for ip in ${workers_ip}; do
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} <<EOF
+sudo sh -c "echo ${master_ip} pmem-csi-${CLUSTER}-master >>/etc/hosts"
+EOF
         done
     fi
 
diff --git a/test/test-config.sh b/test/test-config.sh
index 1a07e503..f2bda74f 100644
--- a/test/test-config.sh
+++ b/test/test-config.sh
@@ -28,11 +28,21 @@ fi
 # on. Here we default to the IP address of the docker0 interface.
 : ${TEST_LOCAL_REGISTRY:=$(ip addr show dev docker0 2>/dev/null | grep " inet " | sed -e 's/.* inet //' -e 's;/.*;;'):5000}
 
+# Set up a Docker registry on the master node.
+: ${TEST_CREATE_REGISTRY:=false}
+
 # The registry used for PMEM-CSI image(s). Must be reachable from
-# inside the cluster.
-: ${TEST_PMEM_REGISTRY:=${TEST_LOCAL_REGISTRY}}
+# inside the cluster. The default is the registry on the master
+# node if that is enabled, otherwise a registry on the build
+# host.
+: ${TEST_PMEM_REGISTRY:=$(if ${TEST_CREATE_REGISTRY}; then echo pmem-csi-${CLUSTER}-master:5000; else echo ${TEST_LOCAL_REGISTRY}; fi)}
 
 # The same registry reachable from the build host.
+# This is needed for "make push-images". Pushing
+# to the registry on the master node (TEST_CREATE_REGISTRY=true)
+# is only supported when making additional changes on the
+# build host (like enabling insecure access to that registry)
+# and therefore the default is always a local registry.
 : ${TEST_BUILD_PMEM_REGISTRY:=localhost:5000}
 
 # Additional insecure registries (for example, my-registry:5000),
@@ -40,6 +50,14 @@ fi
 # marked as insecure and does not need to be listed.
 : ${TEST_INSECURE_REGISTRIES:=}
 
+# When using TEST_CREATE_REGISTRY, some images can be copied into
+# that registry to boot-strap the cluster.
+#
+# To use this, do:
+# - make build-images
+# - TEST_CREATE_REGISTRY=true make start
+: ${TEST_BOOTSTRAP_IMAGES:=${TEST_BUILD_PMEM_REGISTRY}/pmem-csi-driver:canary ${TEST_BUILD_PMEM_REGISTRY}/pmem-csi-driver-test:canary}
+
 # Additional Clear Linux bundles.
 : ${TEST_CLEAR_LINUX_BUNDLES:=storage-utils}
 
@@ -71,9 +89,6 @@ fi
 # Initialize "region0" as required by PMEM-CSI.
 : ${TEST_INIT_REGION:=true}
 
-# Set up a Docker registry on the master node.
-: ${TEST_CREATE_REGISTRY:=false}
-
 # Validate signature of downloaded image files.
 # This may have to be disabled for Clear Linux depending
 # on the version of OpenSSL on the build host
```

I've included that in my "govm" branch, which is now at 426dd201.

chuyd:
@pohly I updated the branch from your `govm branch`, `make test_e2e` is working and the tests are passing.
 

```
Ran 39 of 83 Specs in 118.541 seconds
SUCCESS! -- 39 Passed | 0 Failed | 0 Pending | 44 Skipped
--- PASS: TestE2E (118.54s)
PASS
ok      github.com/intel/pmem-csi/test/e2e      118.684s`
```

okartau:
re-checked the latest, works here as well. Setting up cluster is much slower using latest Clear version compared to older version that I used before, but that's unrelated to this PR. I will merge.

PR #228, opened 2019-04-05
sanity testing normal driver

pohly:
This enhances the driver such that it passes testing with csi-sanity from csi-test v2.0.0 and improves testing.

PR #237, opened 2019-04-16
disruptive testing

pohly:
This provides tests for PR #196.

It is a superset of PR #228, which in turn is a superset of PR #234.


pohly:
@avalluri I previously pointed you to my `sanity-testing-normal-driver` branch for testing your PR. Please use this here instead.

pkg/pmem-csi-driver/controllerserver-master.go
```diff
@@ -435,3 +435,7 @@ func (cs *masterController) getVolumeByName(Name string) *pmemVolume {
 	}
 	return nil
 }
+
+func (cs *masterController) ControllerExpandVolume(context.Context, *csi.ControllerExpandVolumeRequest) (*csi.ControllerExpandVolumeResponse, error) {
```
avalluri: @pohly, can you move this default implementation to `controllerserver-default.go`, so that both master and node inherits them.

avalluri: Oops, I realized that these changes are already merged. So Ignore my comment.

pohly:
I've rebased on top of the changes for PR #236.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -184,7 +184,7 @@ func (pmemd *pmemDriver) Run() error {
 			if err := pmemd.registerNodeController(); err != nil {
 				return err
 			}
-			if err := s.Start(pmemd.cfg.Endpoint, nil, ids, ns); err != nil {
+			if err := s.Start(pmemd.cfg.Endpoint, nil, ids, ns, cs /* cs only needed for testing */); err != nil {
```
pohly: @avalluri, @okartau: what do you think about this?

It exposes the controller implementation also on a node's csi.sock where it can be accessed for testing purposes, using the already existing mechanism (socat socket forwarding).

It's currently done unconditional. There's no harm in that (anyone who can connect to csi.sock can do all kinds of havoc, whether this particular interface is there or not), but it might be seen as exposing a debug interface also in production.

I could make this conditional and only enable it in our testing deployment. I'm just not sure whether that's worth the effort.

test/e2e/storage/sanity.go
```diff
@@ -166,3 +267,176 @@ func getDaemonSet(cs clientset.Interface, setName string) *appsv1.DaemonSet {
...
+
+var unreachable = v1.Taint{Key: "node.kubernetes.io/unreachable", Effect: "NoSchedule"}
+
+// restartNode works only for one of the nodes in the QEMU virtual cluster.
+// It does a hard poweroff via SysRq and relies on Docker to restart the
+// "failed" node.
+func restartNode(cs clientset.Interface, nodeID string) {
+	if !regexp.MustCompile(`worker\d+$`).MatchString(nodeID) {
+		framework.Skipf("node %q not one of the expected QEMU nodes (worker<number>))", nodeID)
+	}
```
pohly: Is this function striking the right balance between "don't break a real system" and "ensure that it runs"?

It's using the node ID to determine the node number because it needs to find the ssh script for it. Should we ever changing the naming convention, then this will break and the test will get skipped.

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: @pohly I personally not favor of exposing interfaces that are used just for tests. In this particular case, I would prefer testcase that checks if the existing volume is "publishable" after node/cluster reboot and avoid ListVolulems check.

okartau: I also think this (exposing of test features) should be avoided if we can.

pohly: What if it is optional? Is that acceptable?

Exposing it enables writing tests that are more focused on a particular aspect and thus easier to debug when something goes wrong. I think that's important.

Testing cache volumes will also be easier that way.

okartau: I think it's acceptable if test/debug feature can be turned on/off during build time, with default=off in production version.

avalluri: > What if it is optional? Is that acceptable?
> 
"Optional" you mean exposing "cs" only in test builds?, that's fine for me.

pohly: I was thinking of doing it via a runtime command line parameter. This is conceptually similar with having debug output built into production binaries.

pohly: Please check whether 4d8ee6f0fd12b2b6df7ec778fa3025b643b8b07a addresses this concern.

pohly:
> Exposing it enables writing tests that are more focused on a particular aspect and thus easier to debug when something goes wrong. I think that's important.
> 
>Testing cache volumes will also be easier that way.

To demonstrate what I meant with that I added a test for cache volumes: cd25587ef1e00dd34e954f303512f4cee449d958

pkg/pmem-csi-driver/pmem-csi-driver.go
avalluri: > > Exposing it enables writing tests that are more focused on a particular aspect and thus easier to debug when something goes wrong. I think that's important.
> > Testing cache volumes will also be easier that way.
> 
> To demonstrate what I meant with that I added a test for cache volumes: [cd25587](https://github.com/intel/pmem-CSI/commit/cd25587ef1e00dd34e954f303512f4cee449d958)

@pohly Why tests are focusing on volume lists on a node, instead wouldn't it be good enough if we check that the volume could be publishable on the node where the volume expected to be created.

pohly: > Why tests are focusing on volume lists on a node, instead wouldn't it be good enough if we check that the volume could be publishable on the node where the volume expected to be created.

Because that makes tests more complicated and harder to debug when they fail.

avalluri:
@pohly would you mind if I take this PR in #196

pohly:
> @pohly would you mind if I take this PR in #196

Not at all, please do. Merging them together always has been the goal.

pohly:
Merged via PR #196

avalluri:
The changes already merged as part of #196

PR #238, opened 2019-04-17
WIP: Testing: Reduce clear specific naming and hard-coded entries

okartau:
There are some file names that have `clear` in name but which are not clear-specific otherwise,
and using same functionality in another distro based testing creates situation where we
have to rename to keep the naming consistent.
By selecting distro-neutral names, we can unify these parts better and reduce diff.

Note that I created separate commits for separate names, but these could (hopefully) get combined
if we want to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -690,14 +690,14 @@ The DeviceMode (lvm or direct) used in testing is selected using variable TEST_D
 
 ### Running commands on test cluster nodes over ssh
 
-`make start` generates ssh-wrappers `_work/ssh-clear-kvm.N` for each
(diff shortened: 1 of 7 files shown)


pohly:
There's one problem with the generic `_work/ssh-kvm` name: it's desirable to bring up multiple different test clusters in parallel. We can't do that right now, but that can and should be changed. Then debugging an issue that only occurs on one cluster configuration but not another will be easier, because both can be brought up in parallel.

Let's have sub-directories of _work for different cluster instances. We can have a generic "ssh" wrapper script in those.

I can take over this PR if you want.

PR #239, opened 2019-04-18
Deployment: add explicit /sys mount in pmem-ns-init in LVM mode

okartau:
Deployment failed in CRIO case in LVM mode because of
read-only /sys. This change makes it work.


okartau:
This change is under testing...

okartau:
Verified, driver initContainer pmem-ns-init succeeds creating namespaces in CRIO mode, LVM mode, in clear-based VMs cluster.

PR #240, opened 2019-04-18
better cluster filenames

pohly:
This was originally based on PR #238, but I ended up doing it differently: instead of just renaming files in `_work`, files related to a cluster installation are now in `_work/clear-kvm[-something]` sub-directories, with generic names inside that directory.

The immediate benefit is that multiple parallel clusters can prepared in parallel (but cannot run in parallel yet).

When we add alternative ways of bringing up clusters, we can just swap out the rules which set up the cluster, while code using the cluster just needs to know the directory name and then can use the generic files inside it for whatever operation it needs to do (to be decided).

### DIFF
--- a/test/clear-kvm.make
+++ b/test/clear-kvm.make
@@ -30,7 +30,7 @@
 #
 # Kubernetes does not get started by default because it might
 # not always be needed in the image, depending on the test.
-# _work/kube-clear-kvm can be used to start it.
+# _work/clear-kvm/start-kubernetes can be used to start it.
 
 # Sanitize proxy settings (accept upper and lower case, set and export upper
 # case) and add local machine to no_proxy because some tests may use a
@@ -42,11 +42,8 @@ NO_PROXY=$(shell echo "$${NO_PROXY:-$${no_proxy}},$$(ip addr | grep inet6 | grep
 export HTTP_PROXY HTTPS_PROXY NO_PROXY
 PROXY_ENV=env 'HTTP_PROXY=$(HTTP_PROXY)' 'HTTPS_PROXY=$(HTTPS_PROXY)' 'NO_PROXY=$(NO_PROXY)'
 
-_work/clear-kvm-original.img:
-	$(DOWNLOAD_CLEAR_IMG)
-
 # This picks the latest available version. Can be overriden via make CLEAR_IMG_VERSION=
-CLEAR_IMG_VERSION = $(shell curl https://download.clearlinux.org/latest)
+CLEAR_IMG_VERSION = $(shell curl --silent https://download.clearlinux.org/latest)
 
 DOWNLOAD_CLEAR_IMG = true
 DOWNLOAD_CLEAR_IMG += && mkdir -p _work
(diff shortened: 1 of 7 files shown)


pohly:
A cluster brought up from scratch with this branch passed "make test_e2e".

test/clear-kvm.make
```diff
@@ -71,14 +68,19 @@ NUM_NODES = 4
 # kubernetes-0/1/2/.... The first image is for the Kubernetes master node,
 # but configured so that also normal apps can run on it, i.e. no additional
 # worker nodes are needed.
-_work/clear-kvm.img: test/setup-clear-kvm.sh _work/clear-kvm-original.img _work/kube-clear-kvm _work/OVMF.fd _work/start-clear-kvm _work/id _work/passwd
-	$(PROXY_ENV) test/setup-clear-kvm.sh $(NUM_NODES)
-	ln -sf clear-kvm.0.img $@
+_work/$(CLUSTER)/created: _work/clear-kvm%/created: test/setup-clear-kvm.sh _work/clear-kvm%/start-kubernetes _work/clear-kvm%/run-qemu _work/id _work/passwd
```
pohly: My original approach was just a wildcard rule:
```
_work/clear-kvm%/created: test/setup-clear-kvm.sh _work/clear-kvm%/start-kubernetes _work/clear-kvm%/run-qemu _work/id _work/passwd
```

But then `make CLUSTER=clear-kvm`, only `make CLUSTER=clear-kvm-something` did. Apparently make absolutely wanted the `%` to be non-empty.

The way it is written now currently works for clusters called "clear-kvm[-something]" but it'll fail for other cluster names:
```
$ make CLUSTER=govm-foobar start
test/clear-kvm.make:71: target '_work/govm-foobar/created' doesn't match the target pattern
test/clear-kvm.make:80: target '_work/govm-foobar/run-qemu' doesn't match the target pattern
test/clear-kvm.make:112: target '_work/govm-foobar/start-kubernetes' doesn't match the target pattern
if ! [ -f _work/clear-kvm-28920.img ]; then ( true && mkdir -p _work && cd _work && version=28920 && [ "$version" ] && curl -O https://download.clearlinux.org/releases/$version/clear/clear-$version-kvm.img.xz && curl -O https://download.clearlinux.org/releases/$version/clear/clear-$version-kvm.img.xz-SHA512SUMS && curl -O https://download.clearlinux.org/releases/$version/clear/clear-$version-kvm.img.xz-SHA512SUMS.sig && sed -e 's;/.*/;;' clear-$version-kvm.img.xz-SHA512SUMS | sha512sum -c && unxz -c <clear-$version-kvm.img.xz >clear-kvm-$version.img ); fi
```

This is a problem that'll have to be solved when adding some alternative setup mechanism.

pohly: Found a solution, see the second commit.

With, it becomes possible to add `clear-govm` like this (no implementation yet...):
```
diff --git a/Makefile b/Makefile
index 1e3e053f..276561ed 100644
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,7 @@ clean:
 # same IP addresses, and thus cannot run in parallel.
 CLUSTER := clear-kvm
 include test/clear-kvm.make
+include test/clear-govm.make
 include test/start-stop.make
 include test/test.make
 
diff --git a/test/clear-govm.make b/test/clear-govm.make
new file mode 100644
index 00000000..f2e877cd
--- /dev/null
+++ b/test/clear-govm.make
@@ -0,0 +1,8 @@
+# The following rules only apply when CLUSTER starts with clear-govm.
+ifneq (,$(filter clear-govm%,$(CLUSTER)))
+
+_work/$(CLUSTER)/created: _work/clear-govm%/created:
+       echo "not implemented yet"
+       false
+
+endif
```

okartau:
finally managed to create clear VM cluster, earlier today I used DNSSEC override, now it worked in clean tree of this PR. make test_e2e passes, took 75 seconds (amazed as it used to be 1600s ... I think thats the result of previous testing change of #228 right, I just had not tried it yet on Clear-based VMs).

pohly:
> finally managed to create clear VM cluster, earlier today I used
> DNSSEC override, now it worked in clean tree of this PR.

So shall we merge?

> make test_e2e passes, took 75 seconds (amazed as it used to be 1600s
> ... I think thats the result of previous testing change of #228 right,

Correct.

PR #241, opened 2019-04-19
(I propose not to merge) pkg/ndctl/region.go: Avoid integer overflow in align

okartau:
With interleave_ways=4 and opts.Align=1GB, code gets
divide-by-zero error because align gets calculated
as 32bit unsigned value were multiplying 1GB by ways=4
overflows 32-bit value, leaving zero value, which is then
converted to 64-bit zero value. We fix this making ways
to be 64-bit, thus multiplication happens to 64-bit value.

### DIFF
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -185,8 +185,8 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
 	}
 
 	if opts.Size != 0 {
-		ways := uint32(C.ndctl_region_get_interleave_ways(ndr))
-		align := uint64(opts.Align * ways)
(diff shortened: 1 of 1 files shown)


okartau:
note that divide-by zero is hit with interleave_ways=4, but not for example with 2 or 6 which will retain 1-bits in 32-bit value after multiply.

okartau:
The reason why this problem remained uncovered is that early trials were verified on a non-interleaved configuration where interleave_ways was always 1, never overflowing 32 bit unsigned value when multiplied by opts.Align value

okartau:
I propose to merge #245 that covers the scenario that this PR attempts to fix, but in better and broader way. Let's have this PR still open as it visualizes what is the exact trouble point in code.

PR #246, opened 2019-04-23
add args unit tests for pmem-ns-init, other minor unit testing improvements 

### LINKED ISSUE #223: re-factor arguments checking in pmem-ns-init
Currently namespace init-stage arguments checking (used in LVM-mode only) implements resetting to sane values.
It was done that way to make the driver tolerant and flexible, but it has downsides:
- arguments processing testing becomes more difficult
- will hide mistakes in deployment files

It is more transparent to fail on invalid values instead of internally amending those.
Refactoring is needed, unit testing can be added in same change set.

okartau:
Instead of setting sane values if out-of-bounds values found,
return error and fail the program with error msg.

Work in progress, dont merge yet.
Unit testing of argument values can be added here in same PR.

Resolves #223

### DIFF
--- a/cmd/pmem-ns-init/main.go
+++ b/cmd/pmem-ns-init/main.go
@@ -35,6 +35,10 @@ func main() {
 
 	glog.Info("Version: ", version)
 
+	if err := CheckArgs(*namespacesize, *useforfsdax, *useforsector); err != nil {
+		fmt.Printf("Arguments check failed: %s", err.Error())
+		os.Exit(1)
+	}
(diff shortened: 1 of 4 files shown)


okartau:
I cherry-picked 2 commits from #201 to here so that we can continue in one PR and close #201.
Here we now have added unit tests for pmem-ns-init, and change in Makefile to cover cmd/* in unit testing.

okartau:
Removed coverage commit and force-pushed

PR #252, opened 2019-04-29
Write termination message to TERMINATION_MESSAGE_PATH

### LINKED ISSUE #233: record termination message
When the PMEM-CSI driver encounters a fatal problem, it should write an explanation to the termination log file before quitting. That way, `kubectl describe pods` can show the reason why the container is no longer running.

One example is `Failed to run driver: FATAL: /sys mounted read-only, can not operate`.

okartau:
Write termination message to TERMINATION_MESSAGE_PATH
   
The path is different from default /dev/termination-log
because we need to pass /dev/ explicitly to containers,
and that is known to cause container creation failure
if combined with termination-log located on /dev.
So we relocate termination log to /tmp.
We define env.var that we pass into container so that
container knows where to create the termination message.

Resolves #233

### DIFF
--- a/cmd/pmem-csi-driver/main.go
+++ b/cmd/pmem-csi-driver/main.go
@@ -10,6 +10,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"k8s.io/klog"
@@ -66,12 +67,24 @@ func main() {
 		Version:            version,
 	})
 	if err != nil {
-		fmt.Printf("Failed to Initialized driver: %s", err.Error())
-		os.Exit(1)
+		errorexit("Failed to initialize driver:", err)
 	}
 
 	if err = driver.Run(); err != nil {
-		fmt.Printf("Failed to run driver: %s", err.Error())
-		os.Exit(1)
+		errorexit("Failed to run driver:", err)
 	}
+}
+
+func errorexit(msg string, e error) {
+	str := msg + e.Error()
+	fmt.Println(str)
+	terminationMsgPath := os.Getenv("TERMINATION_LOG_PATH")
+	if terminationMsgPath != "" {
+		err := ioutil.WriteFile(terminationMsgPath, []byte(str), os.FileMode(0644))
+		if err != nil {
+			fmt.Println("Can not create termination log file:" + terminationMsgPath)
+		}
+	}
+
+	os.Exit(1)
 }
--- a/deploy/kubernetes-1.13/pmem-csi-direct-testing.yaml
+++ b/deploy/kubernetes-1.13/pmem-csi-direct-testing.yaml
@@ -272,11 +272,14 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: spec.nodeName
+        - name: TERMINATION_LOG_PATH
+          value: /tmp/termination-log
         image: 192.168.8.1:5000/pmem-csi-driver:canary
         imagePullPolicy: Always
         name: pmem-driver
         securityContext:
           privileged: true
+        terminationMessagePath: /tmp/termination-log
         volumeMounts:
         - mountPath: /certs/
           name: registry-cert
@@ -391,6 +394,8 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: status.podIP
+        - name: TERMINATION_LOG_PATH
+          value: /tmp/termination-log
         image: 192.168.8.1:5000/pmem-csi-driver:canary
         imagePullPolicy: Always
         name: pmem-driver
--- a/deploy/kubernetes-1.13/pmem-csi-direct.yaml
+++ b/deploy/kubernetes-1.13/pmem-csi-direct.yaml
@@ -261,11 +261,14 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: spec.nodeName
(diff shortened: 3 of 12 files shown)


okartau:
This implement idea in #233. Seems terminationMessagePath can not use value from already defined env.var, so we have to spell the path out there.
In Kustomize part: with current version, I only get it added to one pod (node part), but not controller part. How to specify patch stmt to get it to both?

okartau:
@pohly  can you comment on this: 
Connected to work in this PR, I want to add terminationMessagePath to more pods via kustomize.
But I am puzzled how this works. Currently the file 
deploy/kustomize/kubernetes-1.13-direct/direct-patch.yaml
specifies overlay that adds entries in 
deploy/kubernetes-1.13/pmem-csi-direct.yaml

The direct-patch.yaml says in comment:
`The PMEM-CSI driver must be in container #0`
and it uses "add" format  path: `/spec/template/spec/containers/0/`
i.e. selects `array element 0`

But the base file that is used:
deploy/kustomize/driver/pmem-csi.yaml

contains 2 container specs: 1st for controller, 2nd for node.
The patching code adds entries in the 2nd entry only.
How is this explained by the patch `addressing entry 0 `?  
And how to apply patch to 1st container in that file?
Did I understood wrong how the patching works?

My idea was, we should add terminationMessagePath to all driver containers,
so I want to patch both container entries in that file. But how to target the 1st entry then
if 2nd entry is targeted by index 0.

pohly:
> The patching code adds entries in the 2nd entry only.
> How is this explained by the patch addressing entry 0 ?

It's explained by the `kustomization.yaml`:
```
patchesJson6902:
- target:
    group: apps
    version: v1beta2
    kind: DaemonSet
    name: pmem-csi-node
  path: direct-patch.yaml
```

That says that the patch is to be applied to the DaemonSet.

> And how to apply patch to 1st container in that file?

You need to add a second entry in the `patchesJson6902` array which references the `StatefulSet`.

> My idea was, we should add terminationMessagePath to all driver containers

This is starting to get complicated. Let's reconsider our assumptions: do we really need different `terminationMessagePath` settings for direct and LVM mode? I think both modes can put the termination message into `/tmp`, and then the change can go into `deploy/kustomize/driver/pmem-csi.yaml`.

okartau:
> Let's reconsider our assumptions: do we really need different `terminationMessagePath` settings

That was my idea as well, to unify settings over all scenarios

okartau:
If we agree to unify terminationPath in all driver variants (and that is a sane plan IMHO),
then we can stop patching that part using kustomize and write the settings to base config,
and the need to add patching rules drops away.
It makes base config bigger, but advantage is reduced patching part.
I force-pushed that solution.
It seems to work in brief test:
- Regular tests (deploy, e2e) pass.
- The termination message I tested by forcing error via incorrect -mode parameter.
I can see the error written in `kubectl describe pod`:
```
    Last State:     Terminated
      Reason:       Error
      Message:      Failed to initialize driver:Invalid driver mode: badmode
```
This was tested in Clear cluster in LVM mode.
Previously we had differences in explicit mounts (which originally caused the whole terminationPath change). Some of diffs have now been unified between different deploy scenarios,
so it makes sense to verify both modes, I will test direct mode next.

okartau:
direct mode tests pass as well in Clear VM cluster

okartau:
This is simpler/smaller than #263 so we can postpone and perform rebase after #263 gets merged

cmd/pmem-csi-driver/main.go
```diff
@@ -66,12 +66,25 @@ func main() {
...
+	fmt.Println(str)
+	terminationlogpath := os.Getenv("TERMINATION_LOG_PATH")
+	if terminationlogpath == "" {
+		terminationlogpath = os.Getenv("/dev/termination-log")
+	}
+	f, err := os.Create(terminationlogpath)
+	if err == nil {
+		f.WriteString(str)
+		f.Close()
+	}
```
pohly: Using `ioutil.WriteFile` is simpler.

pohly:
You need to rebase and re-run "make kustomize".

cmd/pmem-csi-driver/main.go
okartau: I used os because it's already imported and operates on string, where ioutil.WriteFile needs new import and operates on byte[]. No big difference, we can indeed save few lines where we actually call it.

pohly: Let's always keep the source code as simple as possible in cases where runtime overhead isn't relevant (which is almost always the case).

okartau:
uses  ioutil.WriteFile now and does not write term.msg if env.var not defined.
Is use of termination message such API that we need to document it?
It may appear not obvious that undefined env.var causes missing termination cause,
instead of writing it to default location.

okartau:
not tested yet with newest changes

pohly:
Some further improvements are possible:
- The error message is mixed case and misses a space: `Message:      Failed to initialize driver:Invalid driver mode: nodexxx` - Go recommendation is to always write error message in lowercase, to allow concatenating them. This is a problem in several of the PMEM-CSI error messages. The missing space is from the new code.
- pmem-vgm and pmem-ns-init would also benefit from recording the final error message.

As I need to touch the code anyway as part of my PR #252 due to conflicts, I'll make these changes there.

Issue #251, opened 2019-04-29
revert name change?


pohly:
https://github.com/intel/pmem-CSI/pull/242 changes the name of the project. It is not certain whether `pmem-CSI` is actually the right name. We need to clarify with @marcemq what the motivation for the PR was and potentially revert it.

pohly:
Also https://github.com/intel/pmem-CSI/pull/247

PR #254, opened 2019-04-29
revert name change

pohly:
PMEM-CSI is the correct name for this project.

Fixes: #251

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -79,7 +79,7 @@ Network ports
 
 Network ports are opened as configured in manifest files:
 
-- registry endpoint: typical port value 10000, used for pmem-CSI internal communication
+- registry endpoint: typical port value 10000, used for PMEM-CSI internal communication
 - controller endpoint: typical port value 10001, used for serving CSI API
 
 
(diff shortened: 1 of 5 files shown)


okartau:
If we just revert, we go back to mixed set, as before those two style-unifying commits we had pmem-csi, Pmem-CSI. Should we add one more commit to unify all to PMEM-CSI?
Another idea, as those different spelling cases are now (mostly?) in README-qemu-notes, what is the role of that file nowadays. Could we retire it as likely no-one sets up qemu-emulated system by hand any more. If we have useful pieces of knowledge there, we could mention in DEVELOPMENT.md

pohly:
> before those two style-unifying commits we had pmem-csi, Pmem-CSI. Should we add one more commit to unify all to PMEM-CSI?

Yes. I thought I had corrected the spelling before, but  either I missed something or it came back.

> Another idea, as those different spelling cases are now (mostly?) in README-qemu-notes, what is the role of that file nowadays.

Let's remove it.

Let's merge this PR and then I'll follow up with a second PR making those changes.

okartau:
ok. will merge.


> Let's remove it.

I will cover removal of README-qemu-notes separately tracked by #256

Issue #256, opened 2019-04-30
Retire unified mode


okartau:
We can get rid of unified mode. This is tracking/discussion issue to make sure we cover all related parts.
These can be removed/cleaned:
- "unified" as one choice of DriverMode in code, comments, command line options
- Description in documents: README.md DEVELOPMENT.md
- unified mode specific deployment files and referral from other deployment files
- Retire README-qemu-notes.md as we should not promote setting up single Qemu-based dev.env, which was in early project days used in unified mode. If there are useful bits in there worth survival, we can move those into DEVELOPMENT.md
- Remove the entire test/sanity directory, has been superseded by `test/e2e/storage/sanity.go`
- unified-mode scripts in `util/`

okartau:
test/sanity/sanity_test.go still uses mode=unified, so we cant retire it yet?

pohly:
`test/sanity/sanity_test.go` has been superseded by `test/e2e/storage/sanity.go`. We can remove the entire `test/sanity` directory.

okartau:
> We can remove the entire `test/sanity` directory.

I added that in 1st comment box to have list of change entries in one place

PR #258, opened 2019-04-30
remove README-qemu-notes.md and documentation of unified mode

pohly:
It isn't useful anymore and would have to be updated to fix the
out-dated pmem-CSI spelling. The only useful information seems to be
the information why ndctl must be used after the first boot, which was
moved into a comment in start-stop.make.

The only reference to the file was for unified mode, which is
scheduled to be removed (#256), so that entire section gets removed.


okartau:
this change leaves removed headings in TOC of DEVELOPMENT, but as I am cherry-picking this commit anyway into my PR, does not matter, I can do it

pohly:
Closing in favor of PR #260

PR #260, opened 2019-04-30
Retire unified mode

### LINKED ISSUE #256: Retire unified mode

okartau:
I started with cherry-picking commit from #258, which I amended to also update TOC and command line options section.

Resolves #256


okartau:
now there is first quick cleanup pass made, I covered all that I found via simple string grep. It's possible some remained which escaped the direct string hit, so more careful checking will be needed, and testing of course.

okartau:
Unified mode happened to be default mode if option did not specify mode.
In current draft here I changed default to **node** in doc, but code does not implement that.
I propose we make change so that mode **must** be specified using cmdline option,
and driver exits with error if mode is not set. I will add another commit with that.

okartau:
in README we now have these in description of driver modes:
**_Controller_** mode is intended to be used in a multi-node cluster...
**_Node_** mode is intended to be used in a multi-node cluster...

These sentences emphasized modes difference to Unify mode, but now that
Unified-mode is removed, are these sentences misleading?
They now leave impression that multi-node clusters are handled differently.
What about clarifying those sentences by dropping the part `used in a multi-node cluster`

Something like: (or better wording, please propose):
**_Controller_** mode should run as a single instance in cluster level....
**_Node_** mode instance should run on worker nodes...

avalluri:
> in README we now have these in description of driver modes:
> **_Controller_** mode is intended to be used in a multi-node cluster...
> **_Node_** mode is intended to be used in a multi-node cluster...
> 
> These sentences emphasized modes difference to Unify mode, but now that
> Unified-mode is removed, are these sentences misleading?
> They now leave impression that multi-node clusters are handled differently.
> What about clarifying those sentences by dropping the part `used in a multi-node cluster`
> 
> Something like: (or better wording, please propose):
> **_Controller_** mode should run as a single instance in cluster level....
> **_Node_** mode instance should run on worker nodes...

Yes, this sounds good. But,  in the near future, we should try [leader election](https://medium.com/michaelbi-22303/deep-dive-into-kubernetes-simple-leader-election-3712a8be3a99) for Controller.

okartau:
passes all tests, removing WIP

okartau:
This is simpler/smaller than #263 so we can postpone and perform rebase after #263 gets merged

PR #262, opened 2019-04-30
test: set up cluster with additional insecure registries

pohly:
This is useful when there are additional local registries that don't
have an SSL certificate that Clear Linux trusts out-of-the-box.

### DIFF
--- a/test/setup-clear-kvm.sh
+++ b/test/setup-clear-kvm.sh
@@ -88,6 +88,17 @@ kill_process_tree () {
     fi
 }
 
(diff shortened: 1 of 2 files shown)


okartau:
makes sense and looks good, I plan to test it in action as well.
This reduces the patching I use to do for my local deployments.

okartau:
tried clear-kvm driver deployment with added local registry (where images are), works.

pohly:
@okartau please merge

PR #263, opened 2019-05-02
Kubernetes 1.14

pohly:
Various deployment enhancement (including PR #210), leading up to a deployment for Kubernetes 1.14 without the CSI CRDs.

Passes `make test_e2e` for me with Clear Linux 29160.


pohly:
> two commits "merge: common files" can be combined?

Oh, absolutely. That was the intend, I just forgot before pushing. Sorry!

> misspelled `stoage` in commit msg

Okay, let me fix that for Amarnath as I took over his commit.

Force pushed...

okartau:
I also tried make_e2e which passes.
But not everything is clean here.
When I switch to this branch from devel, I get 2 untracked files:
```
deploy/kustomize/kubernetes-1.13-direct-testing/controller-socat-patch.yaml
deploy/kustomize/kubernetes-1.13-lvm-testing/controller-socat-patch.yaml
```

this relates to change in git ignore set perhaps?
When I then try `make test`, I get few issues:
```
Error: AccumulateTarget: AccumulateTarget: security; file '../../patches/controller-role-patch.yaml' is not in or below em-csi/deploy/kustomize/kubernetes-1.13/rbac'
````
followed by long show of yaml,
and then:
```
deploy/kubernetes-1.13/pmem-csi-direct.yaml was modified manually
```

something wrong with state of my working tree?

**Comment by pohly** [comment]
"controller-socat-patch.yaml" needs to be removed, either by "make clean-kustomize" (on a branch where it is still used) or manually (after switching to this PR).

`test-kustomize` indeed fails. I'll fix that.

**Comment by pohly** [comment]
> test-kustomize indeed fails. I'll fix that.

Done.

File: deploy/kubernetes-1.13/pmem-app.yaml [hunk 1]
```diff
@@ -10,8 +10,6 @@ spec:
       volumeMounts:
       - mountPath: "/data"
         name: my-csi-volume
-  nodeSelector:
-    storage: pmem
```
avalluri: @pohly I think, we still not ready for this. If I remember correctly, the issue is when the cluster has schedulable nodes without "storage: pmem" label, that case Kubernetes might schedule the application pod on those nodes, even if it requires pmem volume.

okartau:
`make test` now passes clean for me

deploy/kubernetes-1.13/pmem-app.yaml
pohly: Have you considered that we currently don't use delayed binding? If we did, what you said can indeed happen. But if we always create volumes first and Kubernetes honors the topology (both should be the case), then dropping the node selector is fine.

avalluri: yes, if we ignore delayed binding. This should work.

deploy/kustomize/kubernetes-1.14/driverinfo-beta.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: pmem-csi.intel.com
+spec:
+  attachRequired: true
```
avalluri: I think we can drop this. If I am not wrong, Our driver no more needs attachment.

pohly: Can we do that in a separate PR?

pohly: If we don't attachments, then we shouldn't be running external-attacher either. This is a separate change than adding support for Kubernetes 1.14.

avalluri: yes, We can do it as separate PR.

pohly: https://github.com/intel/pmem-CSI/issues/264

PR #265, opened 2019-05-03
coverage information

pohly:
Coverage percentages and line-by-line information is useful to determine where testing is insufficient.


pohly:
Note that this approach of building test binaries can also be trivially extended to enable https://golang.org/doc/articles/race_detector.html - it's basically one additional flag and some checking of the logs.

okartau:
I tried to run tests. I see coverage information is printed at end of log, if a container gets completed (like InitContainers), but how to get coverage info from driver itself? commit log mentions "need ssh access to cluster" but are there some instructions how to do it? a new make targets ? Should we add relevant instructions to DEVELOP, README ?

okartau:
unit test coverage: when I added -cover option to unit test cmd in an earlier PR, you asked to remove it with plan to cover in later work (I understand it's here now), but unit test output shows no coverage here.

pohly:
There are new make targets: https://github.com/intel/pmem-CSI/blob/3bb4e3a2413f0fd085827e6a06e77c7e160eb25d/test/test.make#L147-L173

I'm not sure whether it is worth documenting them. Definitely not in the README (why would a user care?), perhaps in DEVELOP.

pohly:
Rebased and updated the "shut down cleanly commit".

okartau:
I used new make targets and I can get nice-looking coverage output.

pohly:
Rebased again on top of the merged https://github.com/intel/pmem-CSI/pull/252.

In the commit where I had the merge conflict due to the modified `main()` I've made further changes to the exit error message formatting and enabled it also for the init containers.

pohly:
E2E testing passed.

pohly:
@okartau I've force-pushed a revised and enhanced version of the PR.

okartau:
I tried this in Clear-based cluster, ran `make test_e2e` `make _work/coverage.txt` `make _work/coverage.html`, I see useful coverage reports created, no issues detected. Anything else to try?

pohly:
> I tried this in Clear-based cluster, ran make test_e2e make _work/coverage.txt make _work/coverage.html, I see useful coverage reports created, no issues detected.

`make coverage` is now a newer shorthand for re-running the tests and producing the text and html output.

> Anything else to try?

No, that should be it. Please merge if you agree.

Dockerfile
```diff
@@ -22,7 +22,16 @@ ADD . /go/src/github.com/intel/pmem-csi
...
 ENV PKG_CONFIG_PATH=/usr/lib/pkgconfig/
 WORKDIR /go/src/github.com/intel/pmem-csi
-RUN make VERSION=${VERSION} OUTPUT_DIR=/go/bin
+ARG BIN_SUFFIX
+# Here we choose explicitly which binaries we want in the image and in
+# which flavor (production or testing). The actual binary name in the
+# image is going to be the same, to avoid unnecessary deployment
+# differences.
+RUN make VERSION=${VERSION} pmem-csi-driver${BIN_SUFFIX} pmem-vgm${BIN_SUFFIX} pmem-ns-init${BIN_SUFFIX} && \
+    mkdir -p /go/bin/ && \
```
avalluri: I would suggest using new make variable OUTPUT_DIR

avalluri: > I can't just copy all of OUTPUT_DIR like you did before because each binary must be renamed when building the test image (pmem-csi-driver-test -> pmem-csi-driver in the image). This is done in this stage.

Why can't `make` generate binaries without -test prefix, then we need not to do this extra step of renaming them. If someone wants to generate both(with and without -test) variants of binaries, could select with different `OUTPUT_DIR`s
> 
> I also think that is is better to explicitly list in the Dockerfile what ends up in the image, instead of relying on the Makefile and what it produces. The result is more predictable and easier to review this way.

If you want to choose what goes into the image, then I would prefer in the later stage, while preparing the final image.

pohly: > Why can't make generate binaries without -test prefix

It's useful to build both locally.

> If you want to choose what goes into the image, then I would prefer in the later stage, while preparing the final image.

Why?

okartau:
tried latest version, `make coverage` works OK.

okartau:
We almost agreed to merge this few days ago, but then discussion add few more comments. Does recent discussion mean more changes will be needed/done here, or is considered good to merge?
I plant to start namespace-size change PR, and would like to have that merged before, to reduce rebasing needs later, as some code migrates

Dockerfile
avalluri: I would prefer to define new `BUILD_VARIANT` build arg, instead of `BIN_SUFFIX`, and define below make targets:
```
build: $(CMDS)
build-test: $(addsuffix -test,$(CMDS))
build-all:
   make build
   make build-test OUTPUT_DIR=${OUTPUT_DIR}/test
```
In Dockerfile:
```
ARG BUILD_VARIANT
RUN make VERSION=${VERSION} build${BUILD_VARIANT} OUTPUT_DIR=/go/bin
```

This is so convenient to use, instead of renaming binaries.

pohly:
> I would prefer to define new `BUILD_VARIANT` build arg, instead of `BIN_SUFFIX`, and define below make targets:
> ```
> build: $(CMDS)
> build-test: $(addsuffix -test,$(CMDS))
> build-all:
>    make build
>    make build-test OUTPUT_DIR=${OUTPUT_DIR}/test
> ```

I don't like this because it makes our build targets even more
confusing. We would have "make all", "make build", "make build-all". Is
that really what we want?

My proposal is to keep "make all" as-is and extend "make build" with the
additional test variants of the binaries. I'm even tempted to go further
and remove the difference between "make all" and "make build".

Olev, it looks like you get too choose between two options here
and act as a tie-breaker - please pick something and I'll adapt the PR
accordingly.

okartau:
I don't follow well your advanced dialogue about variables, directories.
The change that this PR brings to make targets, already confuses me. I vote for reducing build targets. `make all` is not logical as we have multiple build and test-related targets, does "all" mean all of those, or all of build-related, or both local-build and containers-build, etc. There are many combinations and whatever we pick, is probably not intuitive as meaning of "all".
Container images: I almost never use containers building without pushing, so for me explicit build-image target remains unused and I run push-image (which includes build).
My typical use model goes like this:
- `make test` to verify sources are ok (but target name is bit misleading as this is very different from test_e2e)
- `make` shortened version of above, sometimes
- `make push-image` to get container images built and into registry
- `make start` to create cluster
- `make test_e2e` to test on cluster

more detailed build targets may appear confusing. (make coverage is very useful but not used often).
`make all` does not really fit anywhere in that model.

pohly:
> I don't follow well your advanced dialogue about variables, directories.
> The change that this PR brings to make targets, already confuses me. I
> vote for reducing build targets.

As we were discussing it here, I'll update this PR accordingly.

> `make all` is not logical as we have multiple build and test-related
> targets,

Exactly. I think we already fooled one user into thinking that he has to
do "make" and then do something with the resulting binaries.

> My typical use model goes like this:
> - `make test` to verify sources are ok (but target name is bit misleading as this is very different from test_e2e)
> - `make` shortened version of above, sometimes
> - `make push-image` to get container images built and into registry
> - `make start` to create cluster
> - `make test_e2e` to test on cluster

So let's treat "make" aka "make all" as an alias for "make build" and
define that as "build all binaries". That then covers the "test that
everything at least compiles" aspect of development. "make test" must
include that and then add running some actual tests.

pohly:
I added the "make all/build" changes and rebased. There was a huge number of conflicts with PR #266 and I currently cannot do more than simple build testing, so please review carefully again and try whether "make coverage" still works.

I got errors from kustomize about conflicting patches in the `-testing` variants. It looks like it doesn't simply apply them one-after-the-other. I solved this by layering the coverage changes on top of the testing ones, without actually generating all of these variants in the output directories. For `deploy/kustomize' that's fairly clean, the only potentially confusing part is the mapping between kustomize variant and output files.

avalluri:
I am not convinced fully, still it sounds to me like "coverage/test" builds are kind of `BUILD_VARIANT` than `BIN_SUFFIX` and below targets makes most sense to me:
```
build: $(CMDS)
build-test: $(addsuffix -test,$(CMDS)) # "-coverage" is even better than "-test" suffix i feel
```
and use them(build${BUILD_VARIANT} in Dockerfile, instead of asking to make individual binaries. 

But, I have no strong opinion here, @okartau can take the last call :)

pohly:
Coverage analysis is one thing that can be done with these binaries, but I hope that it won't be the only thing. I also intend to enable race detection, so "-test" is a better suffix.

Regarding "build variant": I prefer a flat hierarchy of targets because it keeps the makefile simpler, and building individual targets in the Dockerfile is intentional because it avoid accidentally building things that aren't needed.

But I think we've discussed this enough now, @okartau you have @avalluri's permission to merge, so please do.

okartau:
I don't feel comfortable as tie-breaker here as I feel confused with changing make targets, variables and change variants. I could formulate opinion after some time of using those targets, but right now I can't tell which I prefer. So I will merge this to get moving, future improvement will come later if needed/agreed.
One related idea: With multiple make targets and sub-targets, areas and sub-areas, what about maintaining a list of commands with one-line description? We could show this via 'make help', or/and have chapter in DEVELOPMENT.md. It will take some extra effort to maintain this, but one line is not long to write, and it would be valuable for those who did not invent the make targets.

PR #266, opened 2019-05-06
Re-organize log statements, separating regular vs. debug levels

okartau:
This PR started as removal of debug-only features.
Real substance later turned out to be check and re-organization of log messages, and make division to regular messages that should get logged always, and debug messages that are present but only shown on elevated log levels. Deployment examples specify verbosity level=3.
For testing deployments, we elevate verbosity to v=5.

Additionally, this PR adds commit removing left-over unified-mode files.


okartau:
One candidate for removing are these traces of namespace creation:
```
I0506 07:18:03.850214       1 glog.go:79] setting namespace sector size: 0
I0506 07:18:03.852934       1 glog.go:79] setting pfn
I0506 07:18:04.034159       1 glog.go:79] enabling namespace
```
these have meaning in case of creation failure, showing what parts had been reached.
But do we need to generate those in all test logs?
One idea: go through and adjust log levels so that more detailed logs can be enabled, but are not present in regular run or test. Could go through log statements, and re-check what do we set in deployment file (v5 mostly now).

pohly:

> One candidate for removing are these traces of namespace creation:
> ```
> I0506 07:18:03.850214       1 glog.go:79] setting namespace sector size: 0
> I0506 07:18:03.852934       1 glog.go:79] setting pfn
> I0506 07:18:04.034159       1 glog.go:79] enabling namespace
> ```
> these have meaning in case of creation failure, showing what parts had been reached.
> But do we need to generate those in all test logs?

"Remove debug features from production" as a guideline is good, but then
if there are problems, one still needs an easy way to re-enable
debugging.

> One idea: go through and adjust log levels so that more detailed logs
> can be enabled, but are not present in regular run or test. Could go
> through log statements, and re-check what do we set in deployment file
> (v5 mostly now).

This is what we should do. We have the means now to use a low log level
in deploy/kustomize/pmem-csi.yaml and can dial that up to 5 in the
testing versions of the .yaml files.

okartau:
Let's rename this PR title to express the broader scope of cleaning log/debug levels.
The first commit: I still believe that msg should go away. Originally I left it as extra debug to see the other mount points "just in case". I believe it gives no added value now

pohly:
> The first commit: I still believe that msg should go away. Originally
> I left it as extra debug to see the other mount points "just in
> case". I believe it gives no added value now

Removing obsolete output of course is fine to remove noise from logs.

okartau:
In log API we can use Info, Warning, Error, Fatal, and all V-levels.
Note there is no pre-defined level for Debug. 
I don't easily find standard or recommendation for using glog V-levels.
In practice, level 5 seems to get used for verbose debug info, and is used already now in pmem-csi.
Currently we set verbosity -v=5 in deploy examples.

I propose to:
- use glog.Info, glog.Warning, glog.Error, glog.Fatal for messages that are logged always
- use glog.V(5).Infof() for extra debug info which is not shown always; optionally, we can create 2 verbosity levels for detailed debug using V-levels 4 and 5
- change deploy examples to specify verbosity level 3

In existing log messages, we can turn some debug-purpose glog.Info() into glog.V(5).
Some glog.Info() are now indicating errors/failures, those should be turned to Warning at least.
These glog.Error causing instant exit, like initialization-time problems should be turned into glog.Fatal()
I will make a pass over messages and propose some improvements here in this PR.

pohly:

> In log API we can use Info, Warning, Error, Fatal, and all V-levels.
> Note there is no pre-defined level for Debug. 
> I don't easily find standard or recommendation for using glog
> V-levels.

Yes, that's hard to find:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

klog.V(4) - Debug level verbosity

    Logging in particularly thorny parts of code where you may want to come back later and check it

klog.V(5) - Trace level verbosity

    Context to understand the steps leading up to errors and warnings
    More information for troubleshooting reported issues

> I propose to:
> - use glog.Info, glog.Warning, glog.Error, glog.Fatal for messages that are logged always
> - use glog.V(5).Infof() for extra debug info which is not shown
> always; optionally, we can create 2 verbosity levels for detailed
> debug using V-levels 4 and 5

Let's use the two levels above (4 and 5).

> - change deploy examples to specify verbosity level 3

Yes.

> These glog.Error causing instant exit, like initialization-time
> problems should be turned into glog.Fatal()

Let's reserve this for absolutely fatal problems that really cannot be
handled any other way. Such error paths break testing (because the test
program stops) and aren't recorded as termination messages (because the
program exits directly). A better approach is to return an error.

okartau:
We have some (planned future improvement?) code under pmem-common/ with log wrappers that take context, and use  SpanFromContext, but this code remains commented out, and ctx thus remains unused.
Is this something that you @pohly planned to enable , as those lines were added by you. Should we use that wrap interface from more places?

pohly:
> We have some (planned future improvement?) code under pmem-common/
> with log wrappers that take context, and use SpanFromContext, but this
> code remains commented out, and ctx thus remains unused.  Is this
> something that you @pohly planned to enable , as those lines were
> added by you. Should we use that wrap interface from more places?

I originally did that for OIM, which uses its own logging
package. Nowadays https://github.com/go-logr/logr as abstract API and
klog as implementation are probably the right way to go, except that
there is no API for passing around a logr implementation via the
context. That's a gap that I meant to work on at some point, but never
got around to actually doing.

For PMEM-CSI, let's stick with just using klog directly.

okartau:
I have no made pass over all log statements, turned some error-related to Errof from Infof, and added V() levels to all Info calls. I tried mostly to maintain previously existing levels. but I lowered some debug-nature msgs visibility by raising those to level 4 and 5.
Example deployment files specify level 3 for all pods.
I took liberty to delete some code in pmem-common/tracing that was not used.
I have verified using e2e test that less logs are printed now, and I can add logs by using v=4 and v=5.

okartau:
There is one comment that looks questionable in pmem-common/tracing.go:
In comment we warn that log could show secrets.
In code few lines below, we use protosanitizer.StripSecrets().
Is that comment up to date or should we delete the part warning about showing secrets?
```
// Warning: at log levels >= 5 the recorded information includes all
// parameters, which potentially contains sensitive information like
// the secrets.
func LogGRPCServer(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
	glog.V(3).Infof("GRPC call: %s", info.FullMethod)
	glog.V(5).Infof("GRPC request: %+v", protosanitizer.StripSecrets(req))

```

pohly:
> There is one comment that looks questionable in pmem-common/tracing.go:
> In comment we warn that log could show secrets.
> In code few lines below, we use protosanitizer.StripSecrets().
> Is that comment up to date or should we delete the part warning about
> showing secrets?

That's out-dated.

okartau:
> That's out-dated.

deleted that comment, force pushed

avalluri:
Looks good to me, but one point is how about adjusting log levels of sidecar containers.

avalluri:
@okartau This PR needs rebasing.

okartau:
> Looks good to me, but one point is how about adjusting log levels of sidecar containers.

you mean, we should not change v=5 to v=3 in sidecar deployment examples? 
TBH, I haven't checked how sidecar log messages verbosity changes because of current change.
It's true that if sidecars have different logging policy then we should adjust verbosity level in more detailed fashion in deployment examples.

pohly:
The sidecars should follow the same Kubernetes guidance on what levels to use, so `-v=3` for them makes sense in production.

deploy/kubernetes-1.13/pmem-csi-direct-testing.yaml
```diff
@@ -237,7 +237,7 @@ spec:
     spec:
       containers:
       - args:
-        - --v=5
+        - --v=3
```
pohly: I would prefer to continue running at `-v=5` in the testing deployments.

okartau:
yes, v5 level seems to add GRPC traces in practice.
Lets keep v3 for regular, and I will change to v5 for testing.

okartau:
one detail I noticed while looking around in deployment files, part of cases uses 
`--v=5` and another part `-v=5`, both have same effect. What is the idea in using double dash in options overall? Can I unify the -v option use, and to which, single or double dash?

pohly:
> one detail I noticed while looking around in deployment files, part of cases uses 
> `--v=5` and another part `-v=5`, both have same effect. What is the
> idea in using double dash in options overall?

None, both are valid.

> Can I unify the -v option use, and to which, single or double dash?

The -help (sic!) output uses single-dash, so I prefer that.

okartau:
do we want -v=5 for socat containers ?

pohly:
> do we want -v=5 for socat containers ?

No, socat is a C application and doesn't accept -v=5.

okartau:
rebased, as after attacher removal deploy files changed.
Please check does this kustomize patching make sense, I am kustomer-beginner.
Seems we need separate patch instructions to various container sets (one vs two vs three and so on) as we cant use wildcard there?
Current set seems to achieve what we want, i.e. -v=3 regularly and elevated to -v=5 in all test-related containers, and socat containers dont get any -v= option

PR #267, opened 2019-05-06
kustomize: always rerun operations

pohly:
Dependency tracking was meant to avoid "make kustomize" from
doing unnecessary work. But because the output files are under git
control, "git rebase" can lead to an inconsistent state (output files
out-dated, but were touched last by git). It's better to always
re-generate them.

Also fixes the .PHONY declaration of the other targets.


okartau:
how can I try this? once I switch to another branch, the changes of this branch are not present. 
This can properly be used/tested only after merging?

pohly:
If you are in a situation where you can actually try it (i.e. "make kustomize" doesn't do something it should do), then you can cherry-pick my commit or simply patch the Makefile.

Otherwise just check it out and observe that "make kustomize" now always creates all files, each time it is called.

okartau:
I created kustomize-area change as in #268 and tried it on top if this branch changes,  works as expected, make kustomize created new state. Also, changing branch to-other-and-back did not create any unwanted effects.

PR #271, opened 2019-05-09
Dockerfile: Use single docker image

avalluri:
Building multiple docker images with different dependencies consumes more time
and space. And even in deployment, Node hast to download all three different
images.

Instead, this change builds single docker image that holds all three binaries.
Did also needed deployment changes.

Now {build,push}-%-image(s) make targets are replaced with {build,push}-image
target.


pohly:
Let's merge this before PR #265.

PR #273, opened 2019-05-10
README.md: Changes to build steps as per new image changes

avalluri:
With resent change #271, now we build single image and make targets
{build,push}-images are not valid, instead one should use {build,push}-image.


pohly:
Should we restore the previous make targets instead?

I am adding them back in PR #265, albeit with slightly different meaning (build and push test and production image).

pohly:
Oops, merged already.

pohly:
Okay, let's leave it like that. It's user-focused, so building the test image isn't needed.

PR #276, opened 2019-05-13
Drop external-attacher usage.

### LINKED ISSUE #264: drop external-attacher
We can stop using external-attacher altogether by setting `attachRequired: false` in our driver info.

avalluri:
Practically current PMEM-CSI driver does nothing in Controller{Publish,
Unpublish} requests. Initial driver implementation, when there was no CSI(v0.3)
topology support, used these methods to create/delete real PMEM volumes. But now
this implementation is obsolete and can be droped.

With the result of this, driver deployment can drop "external-attacher"
completely. Once we set 'CSIDriver.spec.attachRequired=false', Kubernetes does
not wait for volume attach/detach for those volumes created by this driver.

Fixes #264


okartau:
seems nice and straightforward, and takes away number of lines so it's always positive change.
Passes e2e test on my Clear VM cluster.

Issue #275, opened 2019-05-13
default namespacesize of 9GB does not make sense on HW-based deployment


okartau:
We have left namespacesize=9GB in example deployment, which has been left there
from dev.times when we used emulated storage.
That size will result in creation of large amount of namespaces on HW-based systems with large devices.
We should put bigger value there.
Note that default is 32GB, which is improvement over 9 GB,
but may also feel too small for multiple hundreds of GB devices.
Nothe that putting a larger value than device size does not cause failure, pmem-ns-init will reduce the actual size to adjust for remaining size, i.e. with large setting we will get
one device-size namespace, which is also not bad for tester systems (with small emulated device)

avalluri:
@okartau maybe its time to reevaluate our design about namespaces in lvm mode, Do we really need to split the nvdimm region and merge back with LVM?. Instead, how about creating a single namespace(per namespace mode) with the size of the region. This simplifies namespace creation logic and even speeds-up driver startup time on Nodes, i feel.

okartau:
I agree, let's go to single namespace. 
The idea behind multiple namespaces was to make the potentially failing unit smaller, also to leave possibility for manual fixes and adjustments in smaller scale (for example, possibility to detach or add device slices in smaller sets).
In production case, none of that has real value, right. We don't expect admin to configure single namespaces and PVolumes. It's actually the opposite, creation of multiple objects adds risk of inconsistency.
There is another way to partition in LVM mode by setting "use for" ratios, and creating one more level by slicing space to many pre-allocated namespaces, is unneeded complexity.

I will prepare a PR replacing many namespaces during pre-allocating with single namespace. It will simplify the code.

PR #277, opened 2019-05-13
LVM-mode improvements: drop -namespacesize, wait for device creation, improve locking

okartau:
- Maintaining multiple fixed-size namespaces does not add value, so lets create one namespace per NS-mode.
- add short wait for device appearing as it may not happen instantly and we hit error "stat: no such device"
- improve use of mutex yo avoid keeping lock over whole storage for longer times (i.e. runtime of `shred`)

Resolves: #275
Resolves: #289
Resolves: #280


okartau:
I add 3rd commit here, also related to LVM mode improvement so I moved it over from #290 (which gets closed).
This is proposed work-around for missing device issues mentioned in #280 and #289. I tried it out (not exact version pushed now, but earlier incarnation where wait() was implemented inside FlushDevice, but the idea remains), and it resolved the issue: I saw cases during 2-volume reference deployment where device path was not there immediately after lvcreate, but was there 100 ms later.

okartau:
This improves also repeated e2e test success rate compared to what #280 initially reported.
I remove WIP from title, please check and comment.

avalluri:
@okartau Can you please squash first two commits(code change + documentation change).

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -121,6 +121,10 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error
 				if err != nil {
 					return err
 				}
+				err = WaitDeviceAppears(device)
```
avalluri: Is this change really required?. If I understood correctly, the issue wasn't with missing /dev mounting in LVM mode?.

okartau:
> Can you please squash first two commits

done, pushed

pkg/pmem-device-manager/pmd-lvm.go
okartau: I think it is needed. The "wait for device" is different issue from "missing /dev mount". My observations have shown that creating of LVolume and following appearance of it's entry in /dev/ is not synchronous operation:
after lvcreate returns and driver code goes on accessing /dev/ entry,
the new device entry either exists in /dev, or appears shortly later.
As it may appear asynchronously after short delay, it makes sense to retry few times instead of failing on 1st try.
With failing we destroy chance to create volume which really was just showing up with a delay.
The combination of lvcreate() + WaitDeviceAPpears() is kind an attempt to make the creation operation synchronous by semantics.

avalluri: @okartau Thanks for the clarification.

avalluri:
Looks good to me

Issue #272, opened 2019-05-09
test/setup-ca-kubernetes.sh: warn when there are no PMEM nodes


pohly:
A potential user error is to run the script before labeling nodes. In that case an empty `pmem-csi-node-secrets` is created and the deployment fails to start in a rather obscure way (PMEM-CSI driver reports that cert is not found and exits, then sidecars fail to connect).

OTOH, that same situation can also be legitimate when the cluster really has no nodes with PMEM and the script gets run just in case.

I suggest that the script deals with "no nodes" differently:
- print a warning, indicating that labels may be missing
- don't generate the empty secret

pohly:
When the `pmem-csi-node-secrets` is missing, the deployment will fail in a more obvious way (driver doesn't even get started and "kubelet describe" will show that the secret is missing).

PR #281, opened 2019-05-14
setup-ca-kubernetes.sh: Handle no pmem nodes case

avalluri:
Add check to throw a warning and exit in case of no nodes with "storage=pmem"
label.

Took the opportunity to improve the script, such that now it generates secrets
only for newly added nodes. First, it retrieves existing node secrets and
filters out those nodes from all pmem nodes and generates certificates for
leftover nodes.

Now one can add new pmem node(node with storage=pmem label) and rerun the script
for generating secrets for that new node, ever after deploying the driver.

Fixes: #272

### DIFF
--- a/test/setup-ca-kubernetes.sh
+++ b/test/setup-ca-kubernetes.sh
@@ -8,13 +8,19 @@
 
 # Directory to use for storing intermediate files
 WORKDIR=${WORKDIR:-$(mktemp -d -u -t pmem-XXXX)}
-mkdir -p $WORKDIR
-cd $WORKDIR
 
 cfssl_found=1
 (command -v cfssl 2>&1 >/dev/null && command -v cfssljson 2>&1 >/dev/null) || cfssl_found=0
 if [ $cfssl_found -eq 0 ]; then
     echo "cfssl tools not found, Please install cfssl, cfssljson."
+    exit 1
+fi
+
+# Find the nodes for which we need to create node certificates
+NODES=$($KUBECTL get no -l storage=pmem -o name | sed  -e 's;.*/;;')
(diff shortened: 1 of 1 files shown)


okartau:
I tried both positive (regular make start) and negative case (commented out labeling, causing no-label error), and both cases work as expected.
Minor usability note: when I hit unlabaled nodes issue while running  'make start', I get message
`Rerun this script after labeling the cluster nodes that provide persistent memory.`
But as I did not start 'this script' by name, user does not get instant feedback what script to rerun.
Perhaps error message could state the script name?

avalluri:
> I tried both positive (regular make start) and negative case (commented out labeling, causing no-label error), and both cases work as expected.
> Minor usability note: when I hit unlabaled nodes issue while running 'make start', I get message
> `Rerun this script after labeling the cluster nodes that provide persistent memory.`
> But as I did not start 'this script' by name, user does not get instant feedback what script to rerun.
> Perhaps error message could state the script name?

Thanks, @okartau. I did not think of "make start" with no labels case. How it sounds adding "Suppress warnings" option to script. that is supposed to be used by make start?

okartau:
I had in mind to show script name when mentioning "this script", as one script can always be invoked by another script, and term "this script" may become fuzzy. What about
"Rerun $0 after labeling", (or do we needs some more advanced shell variable).

avalluri:
Ok, now I replace 'this' with '$0'.

okartau:
yes, now I get:
`Rerun './test/setup-ca-kubernetes.sh' script after labeling the cluster nodes that provide persistent memory.`
which is more descriptive.

avalluri:
@pohly Could you get a chance to review this.

PR #282, opened 2019-05-15
Dockerfile: Provide defaults for build args

avalluri:
Docker image building might fail if one tries to build directly without using
our defined make targets. In this case, as the mandatory NDCTL* build args have
no default values that result in unexpected build failures.

With this change, Dockerfile is self-contained and uses these defaults defined
in Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
 FROM clearlinux:base AS build
 
-ARG VERSION
-ARG NDCTL_VERSION
-ARG NDCTL_CONFIGFLAGS
-ARG NDCTL_BUILD_DEPS
+ARG VERSION="unknown"
+ARG NDCTL_VERSION="64.1"
+ARG NDCTL_CONFIGFLAGS="--libdir=/usr/lib --disable-docs --without-systemd --without-bash"
+ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
 
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN swupd update && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
 
 WORKDIR /
-RUN curl --location --remote-name https://github.com/pmem/ndctl/archive/v${NDCTL_VERSION}.tar.gz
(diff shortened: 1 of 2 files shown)


okartau:
That makes sense. Originally those variables were defined in Makefile because we had 3 Dockerfiles sharing same variables.

IMAGE_BUILD_ARGS now remains unset in Makefile, but is passed into docker build.
That may be confusing, as it is not stated who and where should use this variable and
where should it be set if used.
Is the idea that we still support this variable in Makefile as way to override predefined variables?
Then we should add a comment.
Or, if we don't plan to support it, to be removed from docker build line?

Dockerfile
```diff
@@ -1,9 +1,9 @@
 FROM clearlinux:base AS build
 
-ARG VERSION
-ARG NDCTL_VERSION
-ARG NDCTL_CONFIGFLAGS
-ARG NDCTL_BUILD_DEPS
+ARG VERSION="unknown"
+ARG NDCTL_VERSION="64.1"
+ARG NDCTL_CONFIGFLAGS="--libdir=/usr/lib --disable-docs --without-systemd --without-bash"
+ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
```
pohly: Can you also add `--fail` to the curl invocation?

The lack of that was hiding the root cause of the build failure.

pohly:
I would keep `IMAGE_BUILD_ARGS=` (empty assignment) in the Makefile with a comment that it can be used to pass additional parameters to `docker build`. That may be useful.

pohly:
This PR is going to conflict with PR #265. Rebasing will not be hard this time, but then (strictly speaking) one would have to re-test. Can we merge #265 first?

avalluri:
> IMAGE_BUILD_ARGS now remains unset in Makefile, but is passed into docker build.

@okartau  I fixed this.

Dockerfile
avalluri: Done

avalluri:
> This PR is going to conflict with PR #265. Rebasing will not be hard this time, but then (strictly speaking) one would have to re-test. Can we merge #265 first?

Ok for me.

avalluri:
I reabased on latest devel. @okartau , @pohly  Is it good enough to merge?

okartau:
yes, I think we already agreed to merge, just waited for  #265. I will merge

PR #284, opened 2019-05-16
README: add chapter describing NVDIMM pre-provisioning with ipmctl

### LINKED ISSUE #217: document support for interleaving DIMMs
When combining multiple DIMMs, interleaved (hardware memory controller striped) access may be faster. We need to document that:
* the PMEM-CSI does not need to deal with interleaving
* how an admin can enable interleaving

okartau:
Resolves #217

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
     -   [Prerequisites](#prerequisites)
         -   [Software required](#software-required)
         -   [Hardware required](#hardware-required)
+        -   [Persistent memory pre-provisioning](#persistent-memory-pre-provisioning)
     -   [Supported Kubernetes versions](#supported-kubernetes-versions)
     -   [Setup](#setup)
         -   [Get source code](#get-source-code)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -375,10 +376,30 @@ development and testing can be done using QEMU-emulated persistent
...
+The PMEM-CSI driver itself can not create such regions.
+The [ipmctl](https://github.com/intel/ipmctl) utility can be used to create regions,
+see the documentation of ipmctl for the details.
+With a non-interleaved creation goal, one region is created per each NVDIMM.
+In such configuration, a PMEM-CSI volume can not be larger than one NVDIMM.
+
+Example of creating non-interleaved goal using all NVDIMMs:
+```sh
+# ipmctl create -goal PersistentMemoryType=AppDirectNotInterleaved
+```
```
pohly: Ah, so "goal" is a special term in ipmctl. I didn't know that and other readers might have the same problem. Let's write our README so that as little prior knowledge is needed as possible.

We can (and have to) use it here of course, just not in plain text.

README.md
```diff
@@ -375,10 +376,30 @@ development and testing can be done using QEMU-emulated persistent
...
+
+Multiple NVDIMMs can be combined using interleaving creation goals.
```
pohly: "Alternatively, multiple NVDIMMs can be interleaved."

README.md
```diff
@@ -375,10 +376,30 @@ development and testing can be done using QEMU-emulated persistent
...
+This causes the data to be striped over multiple NVDIMM devices
+and allows one region to be larger than single NVDIMM, thus enabling
+creation of PMEM-CSI volumes that larger than single NVDIMM.
```
pohly: "Performance is also higher." - right?

okartau:
Pushed new version. I avoid using word "goal" in text so that we dont have to explain it explicitly. It remains visible in impctl command examples, where it is (hopefully) self-explaining (or user needs to check ipmctl help/manual). 
I added sentence stating App Direct Mode has two options (borrowed that sentence from Optane config description).I realized we actually dont mention anywhere that PMEM-CSI driver needs Optane devices to be configured in App Direct mode, so hopefully that sentence helps to clarify that as well.
I re-worded some text to be shorter and avoid repetition, and added note about perf. improvement that interleaved set brings. I tried to cover all of above comments areas.

avalluri:
> Looks good to me.
> 
> @avalluri if you agree, can you merge it?

Yes, let's merge it.

PR #288, opened 2019-05-21
Use abs.path in README cert command example, modernize example pod list output, delete leftover files

### LINKED ISSUE #287: Certificate setting example command in README does not work
There is example command in README about setting certificates in case where `make start` is not used:
```
$ KUBCONFIG="<<your cluster kubeconfig path>> PATH="$PATH:_work/bin" ./test/setup-ca-kubernetes.sh
```

This command produces error message:
```cfssl tools not found, Please install cfssl, cfssljson.```

The reason is relative PATH as _work/bin, combined with script changing working directory before checking cfssl is runnable.
That command will work with absolute path passed:
PATH="$PATH:$PWD/_work/bin"

okartau:
Relative path leads to error as the script changes working dir.

Resolves #287


okartau:
Added 2nd commit taking "get pod" output to modern level

avalluri:
Looks good to me.

Issue #289, opened 2019-05-21
Multiple issues using LVM-mode on a HW-based system


okartau:
Testing has been mostly happening on emulated devices so far.
Here is some notes resulting from trial on HW-based system.
We may have multiple/cascading issues, so I write down some notes as a start.
Some of issues seem caused by underlying unclean state of namespaces.

Cluster has 3 nodes, one node has 4x128GB of Optane DIMMs organized
as two 256GB regions, interleave_set size is 2.

I tried reference deployment in LVM mode as per README steps.
- too many small namespaces (addressed already in #275, will be resolved by #277).
- volume provisioning fails, timing/race issue?
```
I0521 10:07:57.873824       1 glog.go:58] GRPC call: /csi.v1.Controller/CreateVolume
E0521 10:07:58.205146       1 glog.go:119] FlushDevice: /dev/ndbus0region0fsdax/4ed6ff15-7bb0-11e9-b208-a6ef1e106519 does not exist
E0521 10:07:58.205204       1 glog.go:119] GRPC error: rpc error: code = Internal desc = CreateVolume: failed to create volume: stat /dev/ndbus0region0fsdax/4ed6ff15-7bb0-11e9-b208-a6ef1e106519: no such file or directory
```
After namespace creation completes, corresponding device does not appear in time inside container, but is attempted to flush immediately?
Side question: does this also happen with code before moving to Clear-based container?

This seems to be cause to e2e tests failures/regression that are documented in #280.
Note that repeated e2e hits this in emulated storage as well. But seems to be even more systematic on HW.

Then I stopped driver, tried to get rid of LV state, and destroy all namespaces.
Some namespaces (belonging to same Region) were not deleted after `ndctl destroy-namespace all --force`.
That Region remains in some bad state, as next VG operations start to fail on that:
```
I0521 10:45:10.697547       1 glog.go:58] No Vgroup with name ndbus0region0fsdax, mark for creation
I0521 10:45:10.697569       1 glog.go:58] Executing: vgcreate --force ndbus0region0fsdax /dev/pmem0.10 /dev/pmem0.20 /dev/pmem0.2 /dev/pmem0.12 /dev/pmem0.22 /dev/pmem0.4 /dev/pmem0.6 /dev/pmem0.26 /dev/pmem0.8 /dev/pmem0.1 /dev/pmem0.3 /dev/pmem0.5 /dev/pmem0.7 /dev/pmem0.9
[...]

Issue #280, opened 2019-05-14
Repeated test_e2e degrades cluster ability and leaves many leftover LV volumes


okartau:
I tried to repeat findings of https://github.com/intel/pmem-CSI/pull/278#issuecomment-492104594 but leaving out version change impact.

With code from current head of branch=devel, Clear VM cluster with `make start`,
using devicemode=LVM, repeated `make test_e2e`:

- runs 1..7 PASS
- run 8 FAIL "should fail when no volume capabilities are provided"
- run 9 FAIL "sanity Node Service [It] should work"
- run 10 FAIL "should fail when no volume capabilities are provided"
- run 11 FAIL "should return appropriate values (no optional values added)"

seems repeated e2e runs degrade ability seriously, with errors varying.

Another finding: after 11 rounds, all nodes have leftover LV volumes with 4MB sizes
(so small not affecting provision ability via space exhaustion at least)
- host-1: 63 volumes
- host-2: 42 volumes
- host-3: 7 volumes

okartau:
We can suspect recent change to clear-based containers, but we don't have documented results of repeated e2e from times before container base change. That can be retested taking older code, I think

pohly:
Olev Kartau <notifications@github.com> writes:

> I tried to repeat findings of https://github.com/intel/pmem-CSI/pull/278#issuecomment-492104594 but leaving out version change impact.
>
> With code from current head of branch=devel, Clear VM cluster with `make start`,
> using devicemode=LVM, repeated `make test_e2e`:
>
> - runs 1..7 PASS
> - run 8 FAIL "should fail when no volume capabilities are provided"
> - run 9 FAIL "sanity Node Service [It] should work"
> - run 10 FAIL "should fail when no volume capabilities are provided"
> - run 11 FAIL "should return appropriate values (no optional values
> added)"

Is that perhaps because of the volume leak that @avalluri found in
csi-sanity? The fix has been merged, but to get it, one currently has to
update to csi-test master.

> seems repeated e2e runs degrade ability seriously, with errors varying.
>
> Another finding: after 11 rounds, all nodes have leftover LV volumes with 4MB sizes
> (so small not affecting provision ability via space exhaustion at least)
> - host-1: 63 volumes
> - host-2: 42 volumes
> - host-3: 7 volumes

That looks like the leak. However, as there is space left, why are tests
failing?

It sounds like we hit a scalability issue somewhere.

okartau:
[...]

PR #290, opened 2019-05-22
Wait for device to appear after creation before Flush

okartau:
In the LVM mode, there seems to be
time window when device is not accessible yet.


okartau:
This is proposed work-around for missing device issues  mentioned in #280 and #289. I tried it out (not exact version pushed now, but earlier incarnation where wait() was implemented inside FlushDevice, but the idea remains), and it resolved the issue: I saw cases during 2-volume reference deployment where device path was not  there immediately after lvcreate, but was there 100 ms later.

okartau:
I will add the commit from this PR to #277 and rename #277 to be "LVM mode improvements", to group commits related to same area closer together. I am testing those changes together anyway now.

PR #291, opened 2019-05-29
Dockerfile: Create custom bundle - pmem-csi

avalluri:
Using 'mixin' utility one can build custom bundle and can ask swupd to migrate
to custom mix.  Now we create a custom bundle named pmem-csi that holds all
packages we depend on - lvm2, e2fsprogs, xfsprogs, and file. This way we need
not to install huge 'storage-util' bundle and reduce the image size.

Now pmem-csi-driver image size came down to : 470MB. We can still reduce the
size once this issue got resolved:
https://github.com/clearlinux/swupd-client/issues/937

That allows us to remove below unneeded bundles from our image:
- findutils
- libglib
- libstdcpp
- os-core
- os-core-update
- os-core-webproxy
- p11-kit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,23 +38,38 @@ FROM clearlinux:base
 LABEL maintainers="Intel"
 LABEL description="PMEM CSI Driver"
 
-# update and install needed bundles:
+# update and install needed bundles/packages:
 # file - driver uses file utility to determine filesystem type
 # xfsprogs - XFS filesystem utilities
-# storge-utils - for lvm2 and ext4(e2fsprogs) utilities
+# e2fsprogs- Ext filesystem utilities
+# LVM2-bin - for lvm2 and ext4(e2fsprogs) utilities
 ARG CACHEBUST
-RUN swupd update && swupd bundle-add file xfsprogs storage-utils && rm -rf /var/lib/swupd
+ENV PATH="/bin:/usr/bin"
+#NOTE: This below RUN command is intentionally combined all commands, to flatten the docker layers
+RUN swupd update && swupd bundle-add mixer && \
+# Create custom 'pmem-csi' bundle that holds all the needed packages
+    mixin package add file --bundle pmem-csi && \
+    mixin package add xfsprogs --bundle pmem-csi && \
+    mixin package add e2fsprogs-extras --bundle pmem-csi && \
+    mixin package add LVM2-bin --bundle pmem-csi && \
+    mixin build && \
+# Remove unneeded bundes. swupd is fail to remove child bundles, so we had to list all the bundles mixer pulled.
+# NOTE: Do not change the order of below remove bundle list, Otherwise swupd fails to remove them saying failed dependency.
+    swupd bundle-remove mixer sysadmin-basic os-installer linux-firmware-extras acpica-unix2 containers-basic git curl diffutils dnf glibc-locale gzip htop iperf iproute2 kbd kernel-install less man-pages minicom openssl parallel patch perl-basic powertop procps-ng pygobject  python3-basic strace sudo tmux tzdata unzip wpa_supplicant linux-firmware-wifi xz zstd libX11client lib-opengl lib-openssl which && \
+    swupd update --migrate && \
+    swupd bundle-add pmem-csi && \
+    swupd clean && \
+    rm -rf /var/lib/swupd && \
+# default lvm config uses lvmetad and throwing below warning for all lvm tools
+# WARNING: Failed to connect to lvmetad. Falling back to device scanning.
+# So, ask lvm not to use lvmetad
+    mkdir -p /etc/lvm && echo "global { use_lvmetad = 0 }" >> /etc/lvm/lvm.conf
 
 # move required binaries and libraries to clean container
 COPY --from=build /usr/lib/libndctl* /usr/lib/
 COPY --from=build /usr/lib/libdaxctl* /usr/lib/
 RUN mkdir -p /go/bin
 COPY --from=build /go/bin/ /go/bin/
-# default lvm config uses lvmetad and throwing below warning for all lvm tools
-# WARNING: Failed to connect to lvmetad. Falling back to device scanning.
-# So, ask lvm not to use lvmetad
-RUN mkdir -p /etc/lvm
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -38,23 +38,38 @@ FROM clearlinux:base
...
+RUN swupd update && swupd bundle-add mixer && \
+# Create custom 'pmem-csi' bundle that holds all the needed packages
+    mixin package add file --bundle pmem-csi && \
+    mixin package add xfsprogs --bundle pmem-csi && \
+    mixin package add e2fsprogs-extras --bundle pmem-csi && \
+    mixin package add LVM2-bin --bundle pmem-csi && \
+    mixin build && \
+# Remove unneeded bundes. swupd is fail to remove child bundles, so we had to list all the bundles mixer pulled.
+# NOTE: Do not change the order of below remove bundle list, Otherwise swupd fails to remove them saying failed dependency.
+    swupd bundle-remove mixer sysadmin-basic os-installer linux-firmware-extras acpica-unix2 containers-basic git curl diffutils dnf glibc-locale gzip htop iperf iproute2 kbd kernel-install less man-pages minicom openssl parallel patch perl-basic powertop procps-ng pygobject  python3-basic strace sudo tmux tzdata unzip wpa_supplicant linux-firmware-wifi xz zstd libX11client lib-opengl lib-openssl which && \
```
pohly: I'm worried that this long list of bundles is going to be a maintenance nightmare.

Why are these bundles installed at all? Do they get pulled in when installing mixer or are they part of the clear-linux base image?

Instead of pruning that base image, perhaps start a discussion with the Clear developers around publishing a smaller image for usage as base for containers?

Dockerfile
```diff
@@ -38,23 +38,38 @@ FROM clearlinux:base
...
-RUN swupd update && swupd bundle-add file xfsprogs storage-utils && rm -rf /var/lib/swupd
+ENV PATH="/bin:/usr/bin"
+#NOTE: This below RUN command is intentionally combined all commands, to flatten the docker layers
+RUN swupd update && swupd bundle-add mixer && \
+# Create custom 'pmem-csi' bundle that holds all the needed packages
+    mixin package add file --bundle pmem-csi && \
+    mixin package add xfsprogs --bundle pmem-csi && \
+    mixin package add e2fsprogs-extras --bundle pmem-csi && \
+    mixin package add LVM2-bin --bundle pmem-csi && \
+    mixin build && \
```
pohly: The mixin approach is an interesting idea. It's something that @mythi mentioned two weeks ago. Did you have the same idea independently? What a coincidence :smile: 

This is worth sharing with the Clear Linux stacks team. They run into similar size issues with the existing bundles.

avalluri: > The mixin approach is an interesting idea. It's something that @mythi mentioned two weeks ago. Did you have the same idea independently? What a coincidence 😄
> 
I was googling around "micro bundles" then I came across mixer and mixin. :-) But mixer is such a huge bundle, around 1.5 GB it increasing the build timing.

Dockerfile
avalluri: > I'm worried that this long list of bundles is going to be a maintenance nightmare.
> 
> Why are these bundles installed at all? Do they get pulled in when installing mixer or are they part of the clear-linux base image?

These are pulled by mixin, removing mixin itself is not removing all its dependencies, hence I had to remove each one of them.

pohly: Would it be possible to run `mixin build` in one container, then start from scratch again in another where one copies over the output of that command and does the `swupd update --migrate && swupd bundle-add pmem-csi`? Then we avoid this whole fragile cleanup procedure.

If all of our containers have "swupd update" as a separate command, then the result of that command can be cached and the command only needs to run once per build.

avalluri: > Would it be possible to run `mixin build` in one container, then start from scratch again in another where one copies over the output of that command and does the `swupd update --migrate && swupd bundle-add pmem-csi`? Then we avoid this whole fragile cleanup procedure.
> 
No, it's not possible to split "mixin build" and "swupd update --migrate" to different containers, swupd loses the context to do the needed custom update.

okartau:
ATM, fails here:
```
Starting download of remaining update content. This may take a while...
Error: Could not download all files, aborting update
Update took 5.1 seconds, 0 MB transferred
Update failed.
The command '/bin/sh -c swupd update && swupd bundle-add mixer &&     mixin package add file --bundle pmem-csi &&     mixin package add xfsprogs --bundle pmem-csi &&     mixin package add e2fsprogs-extras --bundle pmem-csi &&     mixin package add LVM2-bin --bundle pmem-csi &&     mixin build &&     swupd bundle-remove mixer sysadmin-basic os-installer linux-firmware-extras acpica-unix2 containers-basic git curl diffutils dnf glibc-locale gzip htop iperf iproute2 kbd kernel-install less man-pages minicom openssl parallel patch perl-basic powertop procps-ng pygobject  python3-basic strace sudo tmux tzdata unzip wpa_supplicant linux-firmware-wifi xz zstd libX11client lib-opengl lib-openssl which &&     swupd update --migrate &&     swupd bundle-add pmem-csi &&     swupd clean &&     rm -rf /var/lib/swupd &&     mkdir -p /etc/lvm && echo "global { use_lvmetad = 0 }" >> /etc/lvm/lvm.conf' returned a non-zero code: 17
```

pohly:
>> Would it be possible to run `mixin build` in one container, then start from scratch again in another where one copies over the output of that command and does the `swupd update --migrate && swupd bundle-add pmem-csi`? Then we avoid this whole fragile cleanup procedure.
>> 
> No, it's not possible to split "mixin build" and "swupd update
> --migrate" to different containers, swupd loses the context to do the
> needed custom update.

That context is just a set of files. We just need to know what those are
and copy them.

mcastelino:
@avalluri do you have a list of RPMs you need for your minimal docker image?

avalluri:
> @avalluri do you have a list of RPMs you need for your minimal docker image?

Here is the list: `lvm2, e2fsprogs, xfsprogs`, and `file`.

And to make image compact, the `os-core-update` bundle and its dependencies (`findutils, libglib, libstdcpp, os-core-webproxy, p11-kit`) could be removed from the final image.

Issue #293, opened 2019-05-31
make build-image fails on Ubuntu 18.04.2 LTS with go1.12.5 and docker 18.06.2-ce


mcastelino:
# make build-image failure

```
checking for pkg-config... /usr/sbin/pkg-config
checking pkg-config is at least version 0.9.0... /usr/sbin/pkg-config: error while loading shared libraries: libglib-2.0.so.0: cannot open shared object file: No such file or directory
no
checking for libkmod... no
configure: error: in `/ndctl':
configure: error: The pkg-config script could not be found or is too old.  Make sure it
is in your PATH or set the PKG_CONFIG environment variable to the full
path to pkg-config.

Alternatively, you may set the environment variables KMOD_CFLAGS
and KMOD_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

To get pkg-config, see <http://pkg-config.freedesktop.org/>.
See `config.log' for more details
The command '/bin/sh -c ./configure ${NDCTL_CONFIGFLAGS}' returned a non-zero code: 1
Makefile:80: recipe for target 'build-image' failed
make: *** [build-image] Error 1
```

# Environment

## Kernel

```
5.0.11-050011-generic
```

## OS

```
Ubuntu 18.04.2 LTS
```

## Docker

```
Server Version: 18.06.2-ce
Storage Driver: devicemapper
```

pohly:
Please try with the latest "devel" branch. We only had "build-image" as target for a while and went back to building multiple images. As an aside, even if we only build one, we should stick to "make build-images" instead of going back and forth.

Another change that you might not have is that switch to Clear Linux as base in the Dockerfile.

But having said that, I am confused that you are getting this error. The build should run entirely inside the Docker container and thus it shouldn't matter what you have on the host system. `make build` builds on the host, but that's not what's failing here.

mcastelino:
@pohly it used to work before. In fact it worked yesterday on a different machine. So it could be that clearlinux base image or bundle contents have changed. Let me see if we need to add another bundle to the Dockerfile.

pohly:
I was building today and it worked for me. Which revision of pmem-csi are you using?

pohly:
Just to be sure I tried again, with success:
```
Step 7/34 : RUN swupd update && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
 ---> Running in df1ddd1ea4a0
Update started.
[...]

PR #294, opened 2019-05-31
Dockerfile: Use ldconfig to create the necessary links

mcastelino:
Use ldconfig to create the necessary links and cache to the most recent shared libraries

Note: This should not really be needed, but resolves the issues on machine where we face this issue.

Fixes: #293

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,9 @@ ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN swupd update && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
+# Workaround for "pkg-config: error while loading shared libraries" when using older Docker
+# (see https://github.com/clearlinux/distribution/issues/831)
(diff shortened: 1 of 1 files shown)


mcastelino:
https://github.com/clearlinux/distribution/issues/831

mcastelino:
This workaround for a docker bug is needed only if we want to be able to perform docker build on Clearlinux. On other distros with latest docker and overlay2 graph driver this bug is not present.

Till Clearlinux moves to the latest release of docker we will still need this workaround).

Dockerfile
```diff
@@ -8,6 +8,7 @@ ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN swupd update && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
+RUN ldconfig
```
pohly: Interesting find :+1: 

Can you add a comment before "RUN ldconfig" along the lines of:
```
# Workaround for "pkg-config: error while loading shared libraries" when using older Docker (see https://github.com/clearlinux/distribution/issues/831).
```

Then we might remember to remove this.

PR #301, opened 2019-06-05
README.md: PMEM preparation in VMs

pohly:
This used to be part of README-qemu-notes.md, which was removed
because it seemed obsolete. But labeling is still needed in virtual
machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,8 +377,15 @@ section for the commands that create such a virtual test cluster.
 
 ### Persistent memory pre-provisioning
 
-The PMEM-CSI driver needs pre-provisioned regions on the NVDIMM device(s).
-The PMEM-CSI driver itself cannot create such regions.
+The PMEM-CSI driver needs pre-provisioned regions on the NVDIMM
+device(s). The PMEM-CSI driver itself intentionally leaves that to the
+administrator who then can decide how much and how PMEM is to be used
+for PMEM-CSI.
+
+Beware that the PMEM-CSI driver will run without errors on a node
+where PMEM was not prepared for it. It will then report zero local
+storage for that node, something that currently is only visible in the
+log files.
 
 When running the Kubernetes cluster and PMEM-CSI on bare metal,
 the [ipmctl](https://github.com/intel/ipmctl) utility can be used to create regions.
@@ -458,11 +465,6 @@ enabled 1 region
 # ls -l /dev/pmem*
 ls: cannot access '/dev/pmem*': No such file or directory
 ```
-
-Beware that the PMEM-CSI driver will run without errors on a node
-where PMEM was not prepared as described above. It will then
-report zero local storage for that node, something that currently is
-only visible in the log files.
 
 ## Supported Kubernetes versions
 
--- a/README.md
+++ b/README.md
@@ -372,15 +372,16 @@ Building of Docker images has been verified using Docker-ce: version 18.06.1
 
 Persistent memory device(s) are required for operation. However, some
 development and testing can be done using QEMU-emulated persistent
-memory devices, see [README-qemu-notes](README-qemu-notes.md) for
-technical details and the ["QEMU and Kubernetes"](#qemu-and-kubernetes)
+memory devices. See the ["QEMU and Kubernetes"](#qemu-and-kubernetes)
 section for the commands that create such a virtual test cluster.
 
 ### Persistent memory pre-provisioning
 
 The PMEM-CSI driver needs pre-provisioned regions on the NVDIMM device(s).
 The PMEM-CSI driver itself cannot create such regions.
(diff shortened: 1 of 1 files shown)


okartau:
Very good description!
You use "$ sudo ndctl" but 1st occurrence misses sudo and has double $.
Just in previous block I had used # to indicate that root account is used.
Your style with sudo is more clear. But now we have two styles next to each other.
What about other parts of doc, should we check all..

pohly:
"sudo" may be clearer, but it's also longer and not always necessary. Let's stick with `#`. I've changed my commands.

I also made one more change:
```
 the virtual machine. Instead, that region must be made available for
-use with `ndctl` because when the virtual machine comes up for the
+use with PMEM-CSI because when the virtual machine comes up for the
 first time, the entire region is already allocated for use as a single
```

I had copied the text from the older QEMU readme, but looking at it again the wording seemed a bit odd. Clearly the region is already usable with `ndctl` - it's just PMEM-CSI that cannot do anything with it...

README.md
```diff
@@ -400,6 +401,68 @@ Example of creating regions in interleaved mode, using all NVDIMMs:
...
+    "iset_id":10248187106440278,
+    "persistence_domain":"unknown"
+  }
+]
+$ ls -l /dev/pmem*
+ls: cannot access '/dev/pmem*': No such file or directory
+```
+
+Beware that the PMEM-CSI driver will run without errors on a node
+where PMEM that was not prepared as described above. It will then
```
okartau: That sentence sounds bit unclear, what about changing to either
1) ...on a node with PMEM that was not..
2) ...on a node where PMEM was not..

README.md
```diff
@@ -400,6 +401,68 @@ Example of creating regions in interleaved mode, using all NVDIMMs:
...
+report zero local storage for that node, something that currently is
+only visible in the log files.
```
okartau: thats good point to raise about current driver startup behavior in LVM mode -- should driver refuse to run if there is no available_space in region and no namespaces for driver use exist? because running driver in such state will never produce anything useful on that node, should driver instead "fail early" and show the problem in exit status. That would be more operator-friendly than keeping running, nerver serving anything. and letting operator to find out the reason through log files analysis and examining node state (using ndctl). It can be estimated situations like those will happen in the field.

okartau: actually, that specific path to trouble on a clean system is qemu specific. To reach such situation on a HW-based node (i.e. what I call "field scenario") there must be previously used regions/namespaces involved, i.e. not so common path.

README.md
pohly: Option 2 was what I wanted to write. Need to remove the extra "that"...

README.md
pohly: What is the situation on real hardware with a "fresh" PMEM NVDIMM when `ipmctl create` wasn't called? Is that resulting in a failed driver pod or will it run without memory?

Either way, yes, we need to consider how to surface these setup problems through the Kubernetes API. My plan was to make it part of the PMEM CSI operator, because that then can have a status field and events that show when something is amiss.

For now let's keep everything as it is.

okartau: out-of-box NVDIMM will be in Memory mode before ipmctl goal creation (which needs reboot to activate). With NVDIMM device(s) in memory mode, no PMEM storage is available for PMEM-CSI driver, it will start and run, but will fail to create volumes. That's similar to running on any host: driver will run but can not serve volumes. That's because there is no HW detection in driver: it can't know are there device(s) present. The driver can see Regions and Namespaces over libndctl API, and these appear after "ipmctl create goal + reboot".

pohly: So it *is* exactly as in the VM case, which is what I thought.

Now I am confused by your earlier statement that "that specific path to trouble on a clean system is qemu specific". Hmm, perhaps it wasn't clear that with "PMEM not prepared as described above" I meant both cases?

This confusion could be avoided by moving that warning into the first paragraph, directly after "The PMEM-CSI driver itself cannot create such regions."

okartau: Although the end result in driver is same, states (and path to state) are different and should not be thought as same thing:
- HW: no regions exist (ipmctl was not used), driver will not find usable space.
- qemu: kernel pmem subsystem prepares one region and fills it with namespace -> driver will not find usable space.

The paths forward, to solve it, are different as well:
- HW: use ipmctl create goal, reboot -> driver will find regions with all space free, will create namespaces
- qemu: use 3 labeling commands to redo region formatting, then driver will see one region with space

That is what I meant with "path to trouble is qemu specific". On HW, you will end up with space-taken-by-existing-namespace(s) only via previous use. On qemu, this is the state after VM creation.

pohly: Ack. I pushed another commit where I move the error message. I think it is clearer this way and less likely to be missed.

PR #302, opened 2019-06-05
Makefile: fix "make test-kustomize" dependency

pohly:
On a clean checkout or after "make clean", "make test" failed with:

make: *** No rule to make target '_work/kustomize', needed by
'test-kustomize-deploy/kubernetes-1.13/pmem-csi-direct.yaml'.  Stop.

That's the result of changing how the kustomize binary is built and
then not adapting all dependencies on it.


okartau:
I can confirm I can reproduce this problem as well, and the fix eliminates it.

Issue #215, opened 2019-04-01
choosing node for volume(s)


pohly:
When creating a persistent volume, the user may want to specify on which node the volume is to be allocated. Currently this is random (= PMEM-CSI picks among the nodes which have free space). There is resistance against extending PVC with additional parameters because it is supposed to be a portable abstraction on top of a generic storage system. The suggested solution is to make late binding via `WaitForFirstConsumer` reliable (https://github.com/kubernetes-csi/external-provisioner/issues/221#issuecomment-474248095).

We should develop and submit a patch for that to external-provisioner. Problem: new features will be against master, which only supports  topology on Kubernetes >= 1.14. Backport into the release-1.0 branch?

When implementing that we also need to consider what that means for cache volumes ~~where nodes are specified in the StorageClass~~. For cache volumes, users can only choose the number of nodes, but not the nodes themselves. No good solution for that yet.

avalluri:
Submitted [PR](https://github.com/kubernetes-csi/external-provisioner/pull/282) for comments. The idea here is to add a new flag(`-strict-topology`), that handles the case of late binding and passes only selected node to `CreateVolumeRequest.AccessibilityRequirements`. So, PMEM-CSI considers only that node to create volume and returns with failure if not enough space without trying for other nodes in the cluster.

pohly:
https://github.com/kubernetes-csi/external-provisioner/pull/282 was merged. Next steps:
- tentatively change our deployment so that it uses the "canary" image of the provisioner and the `-strict-topology` flag
- create a storage class with delayed binding
- try it out
- cherry-pick the strict topology patch into the release-1.2 (upcoming release, urgent, for Kubernetes >= 1.14) and release-1.0 (Kubernetes 1.13)

avalluri:
> * cherry-pick the strict topology patch into the release-1.2 (upcoming release, urgent, for Kubernetes >= 1.14) and release-1.0 (Kubernetes 1.13)
Submitted https://github.com/kubernetes-csi/external-provisioner/pull/288 for release-1.2

avalluri:
@pohly I think we can close this issue. Do you agree?

PR #221, opened 2019-04-03
Fix issue in parsing "eraseafter" parameter

avalluri:
Fixed typo in in error checking while parsing bool value.

Also made more verbose in case of parsing failures.

FIXES #220

Signed-off-by: Amarnath Valluri <amarnath.valluri@intel.com>

PR #304, opened 2019-06-07
WIP: Deployment changes to support late binding

avalluri:
Deployment changes comprised of:
- csi-provisioner: Using "canary" image
- Moved PVC files to deploy/common
 - Added sample deployments to demonstrate late binding 

FIXES: #215


avalluri:
> Please mark the PR with "WIP: " in the subject to indicate that it isn't ready for merging (we need the upstream images first).
> 


```
$ docker run quay.io/k8scsi/csi-provisioner:canary --help
Usage of /csi-provisioner:
pflag: help requested
      --alsologtostderr                  log to standard error as well as files
      --connection-timeout               This option is deprecated. (default )
      --csi-address string               The gRPC endpoint for Target CSI Volume. (default "/run/csi/socket")
      --enable-leader-election           Enables leader election. If leader election is enabled, additional RBAC rules are required. Please refer to the Kubernetes CSI documentation for instructions on setting up these RBAC rules.
      --feature-gates mapStringBool      A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                         AllAlpha=true|false (ALPHA - default=false)
                                         Topology=true|false (BETA - default=false)
      --kubeconfig string                Absolute path to the kubeconfig file. Either this or master needs to be set if the provisioner is being run out of cluster.
      --leader-election-type string      the type of leader election, options are 'endpoints' (default) or 'leases' (strongly recommended). The 'endpoints' option is deprecated in favor of 'leases'. (default "endpoints")
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --log_file string                  If non-empty, use this log file
      --logtostderr                      log to standard error instead of files (default true)
      --master string                    Master URL to build a client config from. Either this or kubeconfig needs to be set if the provisioner is being run out of cluster.
      --provisioner                      This option is deprecated. (default )
      --retry-interval-max duration      Maximum retry interval of failed provisioning or deletion. (default 5m0s)
      --retry-interval-start duration    Initial retry interval of failed provisioning or deletion. It doubles with each failure, up to retry-interval-max. (default 1s)
      --skip_headers                     If true, avoid header prefixes in the log messages
      --stderrthreshold severity         logs at or above this threshold go to stderr
      --strict-topology                  Passes only selected node topology to CreateVolume Request, unlike default behavior of passing aggregated cluster topologies that match with topology keys of the selected node.
```

Canary images already having the change we need.

pohly:
> Canary images already having the change we need.

Yes, but we shouldn't switch to canary images in our "devel" branch, should we? Or are you suggesting that we do that?

deploy/kubernetes-1.13/pmem-csi-direct.yaml
```diff
@@ -173,6 +173,7 @@ spec:
         - --v=3
         - --csi-address=/csi/csi-controller.sock
         - --feature-gates=Topology=true
+        - --strict-topology=true
```
pohly: Even if we were to use "canary" images for kubernetes 1.14, we would still need images for 1.13.

avalluri: > > Canary images already having the change we need.
> 
> Yes, but we shouldn't switch to canary images in our "devel" branch, should we? Or are you suggesting that we do that?

With the comments you made on #221, I felt like you suggested to move to canary till  release-1.2 is ready.

avalluri: > Even if we were to use "canary" images for kubernetes 1.14, we would still need images for 1.13.

You mean csi-provisioner:master dropped support for 1.13? I didn't try myself on 1.13, as no cluster available.

avalluri:
> Can you add documentation for this in the README.md? The documentation should document how to use the feature and warn about its pitfalls (unschedulable pod when local storage is insufficient).

Currently, we have below text about late binding under persistenetency models in our ReadME:

 `in case of volumeBindingMode: WaitForFirstConsumer Kubernetes first chooses a node for scheduling the application, and PMEM-CSI creates the volume on that node.`

Shall I extend this to add pitfall of late binding?  I think we need not repeat an example to explain how to start storageclass, pvc, and app in this case of late binding.

deploy/kubernetes-1.13/pmem-csi-direct.yaml
avalluri: > > Even if we were to use "canary" images for kubernetes 1.14, we would still need images for 1.13.
> 
> You mean csi-provisioner:master dropped support for 1.13? I didn't try myself on 1.13, as no cluster available.

Ok, Now I got it. Both Kuberntes v1.13, and v1.14 using different provisioner releases. Yes, in that case, latest canary image might not work for 1.13.

pohly: > With the comments you made on #221, I felt like you suggested to move to canary till release-1.2 is ready.

No, "tentatively change our deployment" wasn't meant as "change it on devel". I was thinking of a WIP branch.

> Both Kuberntes v1.13, and v1.14 using different provisioner releases. Yes, in that case, latest canary image might not work for 1.13.

I expects the 1.14 API for CSINodeInfo. It is useful to test whether it works on 1.13 despite that (would be great if it did, because then we don't need to backport into 1.0.x), but I am not optimistic.

> as no cluster available

You can use `TEST_CLEAR_LINUX_VERSION=<something old> CLUSTER=clear-govm-1-13 make start` to bring up one; I just don't have a sufficiently old version number at hand.

pohly:
> Shall I extend this to add pitfall of late binding?

Yes.

>  I think we need not repeat an example to explain how to start storageclass, pvc, and app in this case of late binding.

Perhaps not a full example, but mentioning that the additional .yaml files exist is worthwhile.

pohly:
Also, can you create an E2E test for this?

It's hard to determine *why* a pod lands on a node, but a test that schedules one pod onto each of the nodes explicitly where each pod references a different PVC should give a good indication that it works. Increase the number of pods to increase the likelihood that it didn't just work by chance...

pohly:
external-provisioner 1.2.1 has been released, but we also need `--strict-topology` in a 1.1.x release. I've created https://github.com/kubernetes-csi/external-provisioner/pull/307 for that.

avalluri:
Closing thins, in favor of #310

Issue #226, opened 2019-04-05
mutex locking for global data structures


pohly:
We added some locking for node operations, but that isn't enough.

For example, many gRPC calls in controllerserver-master.go operate on shared data structures (like `cs.pmemVolumes` in `ListVolumes`) without locking.

This is not an exhaustive list. Basically every single gRPC call has to be checked.

pohly:
We also need tests for this, i.e. concurrent operations. This alone is unlikely to catch missing locking reliably, though, unless something really goes horribly wrong. We have to figure out how to use the https://golang.org/doc/articles/race_detector.html both for "go test" style unit testing and testing of a deployed driver.

pohly:
Preliminary race detection in this branch: https://github.com/pohly/pmem-CSI/tree/race-detector

But someone that doesn't report races in the gRPC server methods, so it's not as useful yet as it needs to be.

PR #306, opened 2019-06-19
pmem-csi-driver: fix race condition in registry server

pohly:
When nodes register, there is a race condition around the concurrent
modifications and/or usage of the node info map by the master
controller. All accesses to that field have to be protected by a
mutex.

To enforce this, the RegistryServer implementation gets moved into its
own package with only those methods exported which do proper
locking.

Fixes: https://github.com/intel/pmem-CSI/issues/226

### DIFF
--- a/pkg/pmem-csi-driver/controllerserver-master.go
+++ b/pkg/pmem-csi-driver/controllerserver-master.go
@@ -11,16 +11,16 @@ import (
 	"math"
 	"strconv"
 
+	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/google/uuid"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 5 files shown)


pohly:
I was hoping to have a regression test for this, but the Go race detector did not actually find this concurrent access (see https://github.com/golang/go/issues/32682).

@avalluri are there other gRPC services which lack locking?

avalluri:
> @avalluri are there other gRPC services which lack locking?

Nothing specific in my mind, but we should cross check once the master and node controller services.

pohly:
> we should cross check once the master and node controller services

Can you do that? Let's keep this PR open as a reminder.

avalluri:
As part of this, I made changes to the Controller(node and master) code to support concurrency.:
https://github.com/avalluri/pmem-CSI/commit/21f6448e27b1b1a8f1ee24f7957a128c6d5417d0

PR #307, opened 2019-06-19
release: support reproducible builds

pohly:
Before tagging a release, we want to ensure that the image for that
release can always be reproduced from source. For that we need to lock
down:
- the Clear Linux base image
- the version we update to

The intended usage is:
1. create a release branch
2. run this script on the branch, commit
3. test
4. tag
5. repeat at step 2 regularly to produce an up-to-date release image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,14 @@
-FROM clearlinux:base AS build
+# CLEARLINUX_BASE and SWUPD_UPDATE_ARG can be used to make the build reproducible
+# by choosing an image by its hash and updating to a certain version with -m:
+# CLEAR_LINUX_BASE=clearlinux@sha256:b8e5d3b2576eb6d868f8d52e401f678c873264d349e469637f98ee2adf7b33d4
+# SWUPD_UPDATE_ARG=-m 29970
+#
+# This is used on release branches before tagging a stable version. The master and devel
+# branches default to using the latest Clear Linux.
+ARG CLEAR_LINUX_BASE=clearlinux:latest
+ARG SWUPD_UPDATE_ARG=
+
+FROM ${CLEAR_LINUX_BASE} AS build
 
 ARG VERSION="unknown"
 ARG NDCTL_VERSION="65"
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -7,7 +17,7 @@ ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
 
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
-RUN swupd update && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
+RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
```
mcastelino: @pohly if you have already chosen a specific CLEAR_LINUX_BASE why do we the the update. If you are choosing the version you are going to update to, then what is the need for CLEAR_LINUX_BASE.

Is the fact that you know you are running `nnn` post update not enough.

Is this just so that you know exactly what happened? i.e. you picked the latest clearlinux container image which was hash `xyz` and you updated it to `nnn`

pohly: > if you have already chosen a specific CLEAR_LINUX_BASE why do we the the update.

Because the Clear Linux base image only gets updated once a month. Releasing without "swupd update" would lead to an app with a potentially insecure OS and isn't acceptable.

>  If you are choosing the version you are going to update to, then what is the need for CLEAR_LINUX_BASE.

It's necessary to ensure reproducibility. We cannot start with "clearlinux:latest" and the do "swupd update -m <someversion>" because that breaks once the base image is on a newer version than `<someversion>`, as "swupd updated" doesn't support downgrades. The Clear Linux stacks team is doing that at the moment, and older tags in their repos cannot be built anymore because of that (I checked).

hack/update-clear-linux-base.sh
```diff
@@ -0,0 +1,44 @@
...
+# Latest server version: 29970
+# There is a new OS version available: 29970
+version=$(echo "$output" | tail -n 1 | sed -e 's/.*: *//')
+echo "Update version: $version"
+
+# Do a trial-run with these parameters.
+docker run "$base" swupd update -m $version || die "failed to update"
+
+# Now update the Dockerfile.
+sed -i -e 's;^\(ARG CLEAR_LINUX_BASE=\).*;\1'"$base"';' -e 's;^\(ARG SWUPD_UPDATE_ARG=\).*;\1"-m '"$version"'";' Dockerfile || die "failed to patch Dockerfile"
```
avalluri: Same as above.

pohly: @avalluri can you repeat what you said above? GitHub only shows me this one comment here from you.

avalluri: Yes, seems my earlier comment disappeared! The comment was about using `-m` option, which is deprecated, instead i was suggesting to use` --version/-V`.

pohly: Good hint, I changed to `--version`.

I also updated the commit message to capture what I explained to @mcastelino.

Dockerfile
pohly: I need to point out one assumption: this approach relies on those `clearlinux@sha256:....` images remaining available even when they are no longer referenced by a named tag. I can't test that at the moment because I don't have a hash for an older image.

If they don't remain available, then this approach is no worse than referencing `clearlinux:latest` because it works as long as that tag is set.

PR #309, opened 2019-06-21
Jenkins enhancements

pohly:
Various Jenkins pipeline enhancements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ FROM ${CLEAR_LINUX_BASE} AS build
 
 ARG VERSION="unknown"
 ARG NDCTL_VERSION="65"
-ARG NDCTL_CONFIGFLAGS="--libdir=/usr/lib --disable-docs --without-systemd --without-bash"
+ARG NDCTL_CONFIGFLAGS="--libdir=/usr/lib64 --disable-docs --without-systemd --without-bash"
 ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
 
 #pull dependencies required for downloading and building libndctl
@@ -67,8 +67,8 @@ RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add file xfsprogs storage-u
 RUN ldconfig
 
 # move required binaries and libraries to clean container
-COPY --from=build /usr/lib/libndctl* /usr/lib/
-COPY --from=build /usr/lib/libdaxctl* /usr/lib/
+COPY --from=build /usr/lib64/libndctl.so.* /usr/lib/
+COPY --from=build /usr/lib64/libdaxctl.so.* /usr/lib/
 RUN mkdir -p /go/bin
 COPY --from=build /go/bin/ /go/bin/
 # default lvm config uses lvmetad and throwing below warning for all lvm tools
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,9 @@
 ARG CLEAR_LINUX_BASE=clearlinux:latest
 ARG SWUPD_UPDATE_ARG=
 
+# Common base image for building PMEM-CSI:
+# - up-to-date Clear Linux
+# - ndctl installed
 FROM ${CLEAR_LINUX_BASE} AS build
 
 ARG VERSION="unknown"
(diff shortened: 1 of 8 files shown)


pohly:
The attempt to run E2E testing in parallel for LVM and direct mode failed, there were test failures (timeouts):
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-CSI/detail/PR-309/12/pipeline/60

It might be worthwhile to investigate this. The modified pipeline is in 5f0be57446ac3575ca62e8b1de1877f7aeff7181.

But in the meantime I'll switch back to serial testing.

pohly:
Rebased, added retry for E2E step. The failure on Kubernetes 1.13 seemed to be a flake, I couldn't reproduce it locally.

okartau:
needs rebase, likely because of recent merge of #314

pohly:
> needs rebase, likely because of recent merge of #314

I've rebased, but before I push let me limit testing of PRs. We just
need to discuss how much we want to test there...

As discussed on chat, we'll test 1.14 with LVM and direct mode for each
PR.

pohly:
@okartau, @avalluri: pre-submit tests are finally passing in this PR. Can we merge it?

It could still break our periodic testing (and probably will, given the experience I had with trying to get the direct mode working), but that's what we have to fix via other PRs.

pohly:
Rebased to resolve the merge conflicts.

Dockerfile
```diff
@@ -67,10 +73,10 @@ RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add file xfsprogs storage-u
 RUN ldconfig
 
 # move required binaries and libraries to clean container
-COPY --from=build /usr/lib64/libndctl.so.* /usr/lib/
-COPY --from=build /usr/lib64/libdaxctl.so.* /usr/lib/
+COPY --from=binaries /usr/lib64/libndctl.so.* /usr/lib/
```
avalluri: As we build `ndctl` in the build phase, `--from=build `is appropriate right?

pohly: Both works. I find copying from  "binaries" better because then we get everything from a single source image. Not sure whether that has any impact, though,

test/test-config.sh
```diff
@@ -86,6 +86,10 @@ fi
 # This string is used as part of deployment file name.
 : ${TEST_DEVICEMODE:=lvm}
 
+# Which deployment to use during testing.
+# Allowed values: testing (default), production
+: ${TEST_DEPLOYMENTMODE:=testing}
```
marcemq: Following the name formatting mostly used in `test/test-config.sh` file, this variable should be `TEST_DEPLOYMENT_MODE`.
 Although if we decide to change it, we should change the name of `TEST_DEVICE_MODE` too, but we can track this specific change in a different PR, what do you think?

marcemq: After reading a bit more the `README.md` file, `DEVICEMODE` is used as one word, so the name `TEST_DEVICEMODE` follows the name formatting in `test/test-config.sh`.

Is `DEPLOYMENTMODE` ging to be used in the same way? If yes, then you can ignore my comment to rename it as `TEST_DEPLOYMENT_MODE`. If no, then we need to rename it.

pohly: I was struggling with that "to underscore or not to underscore" question myself. I settled on making it the same as `DEVICEMODE` simply for the sake of consistency, even though strictly speaking with underscore would make more sense.

marcemq: Got it!

test/test.make
```diff
@@ -110,6 +110,7 @@ space:= $(empty) $(empty)
 RUN_E2E = KUBECONFIG=`pwd`/_work/$(CLUSTER)/kube.config \
 	REPO_ROOT=`pwd` \
 	CLUSTER=$(CLUSTER) \
+	TEST_DEPLOYMENTMODE=$(shell source test/test-config.sh; echo $$TEST_DEPLOYMENTMODE) \
```
marcemq: I think we have an extra `$` when echoing `TEST_DEPLOYMENTMODE` variable, aren't we?
`echo $$TEST_DEPLOYMENTMODE` -> `echo $TEST_DEPLOYMENTMODE`?

pohly:
Got okay to merge from @okartau on chat.

PR #310, opened 2019-06-21
late binding

pohly:
This is PR #304 with several fixes and all of my comments addressed:
- stable v1.2.1 external-provisioner image for Kubernetes 1.14, no `--strict-topology` for 1.13
- tests added
- example visible to user if supported

@avalluri please review my commits. If further changes are needed, we can then discuss whether I do them here or you merge back my changes into your branch and finish it.


pohly:
The new stress test got stuck once because one of 100 volumes didn't get provisioned. I've lowered the numbers a bit also to make it run faster, but this is something we need to watch out for.

To run just that stress test with some non-default values, use:
```
KUBECONFIG=`pwd`/_work/clear-29880/kube.config REPO_ROOT=`pwd` go test -count=1 -timeout 0 -v ./test/e2e -ginkgo.skip='no-such-test|should.access.volume.from.different.nodes' -ginkgo.focus="late binding.*stress test" -pmem.latebinding.workers=10 -pmem.latebinding.volumes=100
```

pohly:
@marcemq this is relevant for pmem-redis. When deploying with anti-affinity (high availability), the storage class must be the one with delayed binding. Otherwise there's the risk that volumes get provisioned on the wrong node.

Can you try out this PR and then (tentatively) adapt the redis operator?

README.md
```diff
@@ -353,7 +353,7 @@ example. This
 illustrates how a cache volume gets provisioned in Kubernetes using
 PMEM-CSI driver.
 
-**NOTE**: Cache volumes are local to node not Pod. If two Pods using
+**NOTE**: Cache volumes are local to node not pod. If two pods using
```
okartau: We use "local to node" above in slightly different context (meaning not migratable in cluster) so I propose we use here different wording, like "associated with node".
Instead of two pods, lets mention multiple as this is not specifically about two.
That sentence also uses few mixed plurals/singulars and "endup". Also, I think we should use "Operators have to consider" instead of "Applications has to consider"

So what about changing to:
Cache volumes are associated with a node, not pod. Multiple pods using the same cache volume on the same node, will not get their own instance but will end up sharing the same PMEM sharing the same PMEM
volume. Operator has to consider this and use available...

Also, the link in "use available", when clicked, will show name of yaml file as "symlink", and no contents, which is not  so user-friendly as user has to navigate to that element with other means and open it (not sure how to improve this easily)

pohly: Is it better now? I avoided the "operator" term because it is unclear whether that is a Kubernetes operator (a tool) or an administrator (a human).

okartau: yes, looks good now, thanks!

test/e2e/storage/latebinding.go
```diff
@@ -0,0 +1,122 @@
...
+
+	By(fmt.Sprintf("%s: deleting claim %q/%q", id, claim.Namespace, claim.Name))
+	framework.ExpectNoError(client.CoreV1().PersistentVolumeClaims(claim.Namespace).Delete(claim.Name, nil))
+
+	// Wait for the PV to get deleted if reclaim policy is Delete. (If it's
+	// Retain, there's no use waiting because the PV won't be auto-deleted and
+	// it's expected for the caller to do it.) Technically, the first few delete
+	// attempts may fail, as the volume is still attached to a node because
+	// kubelet is slowly cleaning up the previous pod, however it should succeed
+	// in a couple of minutes. Wait 20 minutes to recover from random cloud
```
marcemq: Why choose 20 minutes over other values?

test/e2e/storage/latebinding.go
```diff
@@ -0,0 +1,122 @@
...
+	framework.ExpectNoError(framework.WaitForPodSuccessInNamespaceSlow(client, pod.Name, pod.Namespace))
+	runningPod, err := client.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred(), "get pod %s", id)
+	actualNodeName := runningPod.Spec.NodeName
+	testsuites.StopPod(client, pod)
+	pod = nil // Don't stop twice.
+
+	By(fmt.Sprintf("%s: checking the created volume is readable and retains data on the same node %q", id, actualNodeName))
+	command = "grep 'hello world' /mnt/test/data"
+	testsuites.RunInPodWithVolume(client, claim.Namespace, claim.Name, "pvc-volume-tester-reader-"+id, command, testsuites.NodeSelection{Name: actualNodeName})
```
marcemq: Not familiar with the test frameworks used in e2e test, but shouldn't we check that the `command` was successfully executed? perhaps something like? 
* `framework.ExpectNoError()` call or
* `Expect(err).NotTo(HaveOccurred(), ,)` call

test/e2e/storage/latebinding.go
pohly: This was copied from the original provisioning.go. It's probably based on experience with real clouds. It doesn't affect us, so I would just leave that as-is instead of fiddling with those values (if we can do that at all - it might be a value in an utility function that we vendor).

test/e2e/storage/latebinding.go
pohly: There's no error to be checked here. The error handling is all inside `testsuites.RunInPodWithVolume`.

test/e2e/storage/latebinding.go
marcemq: Got it.

test/e2e/storage/latebinding.go
marcemq: Got it.

marcemq:
I've tested the changes within this PR and I successfully deployed a Redis cluster trough redis-operator using [pmem.yaml](https://github.com/spotahome/redis-operator/blob/master/example/redisfailover/pmem.yaml) as it is with `TEST_CLEAR_LINUX_VERSION=29820 make start` command for K8s cluster enablement :tada:

Some output of the pods assigned to different nodes is shown below:
```
$ kubectl get po -o wide
NAME                                      READY   STATUS    RESTARTS   AGE     IP           NODE                          NOMINATED NODE   READINESS GATES
pmem-csi-controller-0                     3/3     Running   0          36m     10.244.3.2   pmem-csi-clear-govm-worker3   <none>           <none>
pmem-csi-node-6cvh6                       2/2     Running   0          35m     172.17.0.4   pmem-csi-clear-govm-worker2   <none>           <none>
pmem-csi-node-9pbq2                       2/2     Running   0          36m     172.17.0.5   pmem-csi-clear-govm-worker3   <none>           <none>
pmem-csi-node-testing-5hv6f               1/1     Running   0          36m     172.17.0.5   pmem-csi-clear-govm-worker3   <none>           <none>
pmem-csi-node-testing-tbdqr               1/1     Running   0          35m     172.17.0.3   pmem-csi-clear-govm-worker1   <none>           <none>
pmem-csi-node-testing-vq5fd               1/1     Running   0          35m     172.17.0.4   pmem-csi-clear-govm-worker2   <none>           <none>
pmem-csi-node-vljwf                       2/2     Running   0          35m     172.17.0.3   pmem-csi-clear-govm-worker1   <none>           <none>
redisoperator-688f6f4fcc-xjwnn            1/1     Running   0          20m     10.244.1.2   pmem-csi-clear-govm-worker1   <none>           <none>
rfr-redisfailover-pmem-0                  1/1     Running   0          16m     10.244.1.4   pmem-csi-clear-govm-worker1   <none>           <none>
rfr-redisfailover-pmem-1                  1/1     Running   0          12m     10.244.2.3   pmem-csi-clear-govm-worker2   <none>           <none>
rfr-redisfailover-pmem-2                  1/1     Running   0          8m38s   10.244.3.4   pmem-csi-clear-govm-worker3   <none>           <none>
rfs-redisfailover-pmem-7595857d4c-m64rq   1/1     Running   0          16m     10.244.2.2   pmem-csi-clear-govm-worker2   <none>           <none>
rfs-redisfailover-pmem-7595857d4c-qxrsn   1/1     Running   0          16m     10.244.1.3   pmem-csi-clear-govm-worker1   <none>           <none>
rfs-redisfailover-pmem-7595857d4c-xwnj8   1/1     Running   0          16m     10.244.3.3   pmem-csi-clear-govm-worker3   <none>           <none>
```

PR #314, opened 2019-06-26
Jenkinsfile: go back to Clear Linux 29820

pohly:
29980 too often failed to restart Kubernetes because
Docker forgot about the containers after a restart.


okartau:
I remained with 29820 in own testing all recent times and had no issues. I will merge.

PR #315, opened 2019-06-26
test: wait for PMEM-CSI driver before starting the tests

pohly:
Most E2E tests indirectly wait for the driver to start, for example
when waiting for volume provisioning. However, those tests then also
pass when PMEM-CSI only started on one of the desired nodes.

The new "controller restart" test doesn't wait and failed when it ran
while only some of the nodes had not registered themselves
yet (https://github.com/intel/pmem-CSI/pull/196#issuecomment-505728337).

In both cases, checking the driver readiness as part of the test suite
startup helps.

This also would have helped detect that testing ran against a cluster
in a degraded state due to the Clear Linux networking
issue (https://github.com/clearlinux/distribution/issues/904) and only
passed randomly when picking a working node for sanity testing.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -18,12 +18,15 @@ package e2e
 
 import (
 	"context"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/pkg/version"
 	"k8s.io/kubernetes/test/e2e/framework"
 	"k8s.io/kubernetes/test/e2e/framework/ginkgowrapper"
@@ -106,6 +109,8 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 	podlogs.CopyAllLogs(ctx, c, "default", to)
 	podlogs.WatchPods(ctx, c, "default", ginkgo.GinkgoWriter)
 
+	WaitForPMEMDriver(c)
+
 	dc := c.DiscoveryClient
 
 	serverVersion, serverErr := dc.ServerVersion()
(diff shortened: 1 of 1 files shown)


pohly:
During `make test_e2e` without an already running cluster I could see that the test suite had to wait for the driver to start, as expected:
```
Jun 26 16:20:26.648: INFO: e2e test version: v0.0.0-master+$Format:%h$
Jun 26 16:20:26.657: INFO: WARNING: PMEM-CSI not ready yet on some nodes: pmem-csi-clear-govm-worker1, pmem-csi-clear-govm-worker2, pmem-csi-clear-govm-worker3
pod event: ADDED: default/pmem-csi-controller-0 Pending:  [{Initialized True 0001-01-
01 00:00:00 +0000 UTC 2019-06-26 16:20:10 +0200 CEST  } {Ready False 0001-01-01 00:00
:00 +0000 UTC 2019-06-26 16:20:10 +0200 CEST ContainersNotReady containers with unrea
dy status: [pmem-driver external-provisioner socat]} {ContainersReady False 0001-01-0
1 00:00:00 +0000 UTC 2019-06-26 16:20:10 +0200 CEST ContainersNotReady containers wit
h unready status: [pmem-driver external-provisioner socat]} {PodScheduled True 0001-0
1-01 00:00:00 +0000 UTC 2019-06-26 16:20:10 +0200 CEST  }]
   external-provisioner: WAITING: ContainerCreating - 
   pmem-driver: WAITING: ContainerCreating - 
   socat: WAITING: ContainerCreating - 
...
Jun 26 16:20:46.656: INFO: WARNING: PMEM-CSI not ready yet on some nodes: pmem-csi-clear-govm-worker2, pmem-csi-clear-govm-worker3
...
```

okartau:
I ran 10 x e2e on a VM cluster, i.e. same test set that created recent trouble that led to this change set,
and result is 10/10 OK, looks good.

okartau:
actually, that was not right state to test as #196 was not present. I need to test combined this and #196

avalluri:
Lets deal this with #196 , because that is what demanded this change.

PR #318, opened 2019-06-27
Persisting driver state and reload state on restart + deadlock fix.

pohly:
This is PR #196 with one additional change to avoid the deadlock discussed in that PR.


pohly:
Let's see whether this works in the CI...

pohly:
Doesn't even build ;-} Nice try, Patrick.

Amarnath has a similar fix in his PR now, so I'm closing this one.

PR #323, opened 2019-07-02
test: move stress test to sanity testing

pohly:
We noticed that Kubernetes on Clear Linux wasn't creating pods
reliably (https://github.com/clearlinux/distribution/issues/966),
which explained at least some test failures.

By running the stress test directly against the CSI API, we avoid that
issue. We can also execute the test more quickly.

### DIFF
--- a/test/e2e/storage/cleanup.go
+++ b/test/e2e/storage/cleanup.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2018 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"context"
+	"log"
+	"sync"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-csi/csi-test/pkg/sanity"
+
+	. "github.com/onsi/ginkgo"
+)
+
+// VolumeInfo keeps track of the information needed to delete a volume.
+type VolumeInfo struct {
+	// Node on which the volume was published, empty if none
+	// or publishing is not supported.
+	NodeID string
+
+	// Volume ID assigned by CreateVolume.
+	VolumeID string
+}
+
+// Cleanup keeps track of resources, in particular volumes, which need
+// to be freed when testing is done. All methods can be called concurrently.
+type Cleanup struct {
+	Context                    *sanity.SanityContext
+	ControllerClient           csi.ControllerClient
+	NodeClient                 csi.NodeClient
+	ControllerPublishSupported bool
+	NodeStageSupported         bool
+
+	// Maps from volume name to the node ID for which the volume
+	// is published and the volume ID.
+	volumes map[string]VolumeInfo
+	mutex   sync.Mutex
+}
+
+// RegisterVolume adds or updates an entry for the volume with the
+// given name.
+func (cl *Cleanup) RegisterVolume(name string, info VolumeInfo) {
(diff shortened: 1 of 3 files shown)


test/e2e/storage/cleanup.go
```diff
@@ -0,0 +1,143 @@
...
+		cl.volumes = make(map[string]VolumeInfo)
+	}
+	cl.volumes[name] = info
+}
+
+// MaybeRegisterVolume adds or updates an entry for the volume with
+// the given name if CreateVolume was successful.
+func (cl *Cleanup) MaybeRegisterVolume(name string, vol *csi.CreateVolumeResponse, err error) {
+	if err == nil && vol.GetVolume().GetVolumeId() != "" {
+		cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetVolumeId()})
```
marcemq: Isn't `NodeID` value missing when sending this `VolumeInfo` struct?

pohly: Good catch. Yes, it's missing. However, this only affects CSI drivers which support `ControllerPublish` (see usage in `DeleteVolumes`), therefore we can ignore it here.

@marcemq as you found this issue, do you want to do the upstreaming to csi-test? Include the mutex locking, too.

pohly:
Updated:
- cleanup.go had a deadlock issue (caused cleanup to hang if there were volumes to be deleted, which also caused volumes to leak)
- uses GinkgoRecover (now I understand better how it works, see comments in the code)

pohly:
This currently includes PR #326 because the two PRs would conflict otherwise. We need to merge that other PR first.

test/e2e/storage/cleanup.go
```diff
@@ -0,0 +1,146 @@
...
+
+// RegisterVolume adds or updates an entry for the volume with the
+// given name.
+func (cl *Cleanup) RegisterVolume(name string, info VolumeInfo) {
+	cl.mutex.Lock()
+	defer cl.mutex.Unlock()
+	if cl.volumes == nil {
+		cl.volumes = make(map[string]VolumeInfo)
+	}
+	cl.volumes[name] = info
```
marcemq: What if for some reason the given name is `""`?, this is an acceptable key.
And even more, if we want to register to new volumes with `""` key, then we'll just update the current info and loss it.

test/e2e/storage/cleanup.go
```diff
@@ -0,0 +1,146 @@
...
+// UnregisterVolume removes the entry for the volume with the
+// given name, thus preventing all cleanup operations for it.
+func (cl *Cleanup) UnregisterVolume(name string) {
+	cl.mutex.Lock()
+	defer cl.mutex.Unlock()
+	cl.unregisterVolume(name)
+}
+func (cl *Cleanup) unregisterVolume(name string) {
+	if cl.volumes != nil {
+		delete(cl.volumes, name)
```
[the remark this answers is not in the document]
pohly: Empty name isn't valid. Nothing bad happens when using it, though, so there's no extra check.

test/e2e/storage/cleanup.go
pohly: Names have to be unique.

The name shouldn't be empty, but even if it is, then that's simple that one volume with the empty name.

test/e2e/storage/cleanup.go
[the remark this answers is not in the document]
marcemq: Got it!

test/e2e/storage/cleanup.go
marcemq: Got it!

avalluri:
@pohly Can you split the PR into 2 commits - cleanup fix and stress tests. Then its easier to revert back cleanup commit once it's fixed in upstream,

pohly:
> @pohly Can you split the PR into 2 commits - cleanup fix and stress tests. Then its easier to revert back cleanup commit once it's fixed in upstream,

I tried, but if I move the cleanup code in the first commit and then add the stress test, then first commit cannot simply be reverted with "git revert" because of conflicts. Ordering the commits the other way around makes no sense, because the stress test depends on the updated cleanup code.

Let's just go back to upstream cleanup code with manually removing the copy and `s/Cleanup/sanity.Cleanup/`.

avalluri:
Lets deal this in #336

Issue #322, opened 2019-07-01
"capacity restored" test fails


pohly:
After merging PR #196, the following test of the "devel" branch failed:
```
[Fail] sanity pmem csi [AfterEach] capacity is restored after controller restart 
/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:109
```

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-CSI/detail/devel/8/pipeline

avalluri:
#324 Is supposed to fix these type of test failures, caused by delay in the re-registering of Nodes on Controller restarts.

pohly:
Amarnath Valluri <notifications@github.com> writes:
> #324 Is supposed to fix these type of test failures, caused by delay
> in the re-registering of Nodes on Controller restarts.

That PR may shorten the delay, but the race condition remains.

avalluri:
#324 Has the needed fix and it's merged.

PR #324, opened 2019-07-02
Fixes to test failures

avalluri:
- Often test failures such as #322 are caused by delay the Node driver knowing about the connection failures to Controller. This PR resolves this by adding Keepalives to grpc connections.

### DIFF
--- a/pkg/pmem-grpc/grpc.go
+++ b/pkg/pmem-grpc/grpc.go
@@ -12,6 +12,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/resolver"
 	"k8s.io/klog/glog"
 
@@ -41,6 +42,11 @@ func Connect(endpoint string, tlsConfig *tls.Config) (*grpc.ClientConn, error) {
 	} else if proto == "unix" {
 		dialOptions = append(dialOptions, grpc.WithDialer(unixDialer))
 	}
+	// This is necessary when connecting via TCP and does not hurt
+	// when using Unix domain sockets. It ensures that gRPC detects a dead connection
+	// in a timely manner.
+	// Code lifted from https://github.com/kubernetes-csi/csi-test/commit/6b8830bf5959a1c51c6e98fe514b22818b51eeeb
+	dialOptions = append(dialOptions, grpc.WithKeepaliveParams(keepalive.ClientParameters{PermitWithoutStream: true}))
 
 	return grpc.Dial(address, dialOptions...)
 
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -45,7 +45,7 @@ import (
 )
 
 var (
-	cleanup              func()
+	cleanup func()
 )
 
 // Run the csi-test sanity tests against a pmem-csi driver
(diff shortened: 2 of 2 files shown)


test/test.make
```diff
@@ -110,7 +110,7 @@ space:= $(empty) $(empty)
 RUN_E2E = KUBECONFIG=`pwd`/_work/$(CLUSTER)/kube.config \
 	REPO_ROOT=`pwd` \
 	CLUSTER=$(CLUSTER) \
-	go test -count=1 -timeout 0 -v ./test/e2e -ginkgo.skip='$(subst $(space),|,$(TEST_E22_SKIP))'
+	go test -count=1 -timeout 0 -v ./test/e2e -ginkgo.skip='$(subst $(space),|,$(TEST_E22_SKIP))' -ginkgo.failFast
```
pohly: We shouldn't add that to the repo. In the CI we then only get to see one single test failure.

pkg/pmem-grpc/grpc.go
```diff
@@ -41,6 +42,11 @@ func Connect(endpoint string, tlsConfig *tls.Config) (*grpc.ClientConn, error) {
 	} else if proto == "unix" {
 		dialOptions = append(dialOptions, grpc.WithDialer(unixDialer))
 	}
+	// This is necessary when connecting via TCP and does not hurt
+	// when using Unix domain sockets. It ensures that gRPC detects a dead connection
+	// in a timely manner.
+	// code lift from https://github.com/kubernetes-csi/csi-test/commit/6b8830bf5959a1c51c6e98fe514b22818b51eeeb
+	dialOptions = append(dialOptions, grpc.WithKeepaliveParams(keepalive.ClientParameters{PermitWithoutStream: true}))
```
pohly: s/code lift/Code lifted/

There's also a typo in your commit message (connwction).

test/test.make
avalluri: Ok, I revered the change then.

avalluri:
> As discussed on chat, the real fix is to ensure that `restartNode` blocks until the cluster is in a healthy state again, i.e. all nodes are connected to the controller again.

Currently, there is no right way to know if a Node(all nodes) connected to Controller.

pohly:
> Currently, there is no right way to know if a Node(all nodes) connected to Controller.

You can check the capacity before and after the restart. It's indirect, but it works.

test/e2e/storage/sanity.go
```diff
@@ -564,25 +566,24 @@ sudo sh -c 'echo b > /proc/sysrq-trigger'`)
...
-	By("node reboot success!")
+	By("Node reboot success! Waiting for driver restore connections")
+	Eventually(func() int64 {
+		currentCapacity, err := cc.GetCapacity(context.Background(), &csi.GetCapacityRequest{})
+		if err != nil {
+			// Probably not running again yet.
+			return 0
+		}
+		return currentCapacity.AvailableCapacity
+	}, "3m", "2s").Should(Equal(capacity.AvailableCapacity), "total capacity after node restart")
```
pohly: That looks like it should do the job.

pohly:
@avalluri can you add a comment when making changes to a PR? Pushing alone doesn't trigger notifications.

pohly:
> @avalluri can you add a comment when making changes to a PR? Pushing alone doesn't trigger notifications.

Scratch that. It did send a notification. Not sure why, though. I could swear it doesn't always do that.

PR #326, opened 2019-07-03
test: sync filesystem before starting restart tests

pohly:
The following error occurred randomly:

  rmdir: '/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.ECDAPN': No such file or directory

Most recently it was observed in the teardown after an otherwise
successful node restart test. One possible reason is that the
directory was created before the test ran, but never got flushed to
disk, so when the node came back up, the directory was lost.

We may be able to avoid that by explicitly syncing. This is done only
in those tests which need it and at the very beginning, so changes
made by the driver still need to be synced by the driver itself.


marcemq:
LGTM.

PR #327, opened 2019-07-04
test direct mode and LVM

pohly:
This changes the test setup such that the CI covers LVM and direct mode on Kubernetes 1.14.

### DIFF
--- a/test/e2e/storage/csi_volumes.go
+++ b/test/e2e/storage/csi_volumes.go
@@ -17,6 +17,7 @@ limitations under the License.
 package storage
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"strings"
@@ -30,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/podlogs"
 	"k8s.io/kubernetes/test/e2e/storage/testpatterns"
 	"k8s.io/kubernetes/test/e2e/storage/testsuites"
 	"k8s.io/kubernetes/test/e2e/storage/utils"
@@ -157,6 +159,17 @@ var _ = Describe("PMEM Volumes", func() {
 			// The load here consists of n workers which
 			// create and test volumes in parallel until
 			// we've tested m volumes.
+
+			// Because this test creates a lot of pods, it is useful to
+			// log their progress.
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			to := podlogs.LogOutput{
+				StatusWriter: GinkgoWriter,
+				LogWriter:    GinkgoWriter,
+			}
+			podlogs.CopyAllLogs(ctx, f.ClientSet, f.Namespace.Name, to)
+			podlogs.WatchPods(ctx, f.ClientSet, f.Namespace.Name, GinkgoWriter)
 
 			wg := sync.WaitGroup{}
 			volumes := int64(0)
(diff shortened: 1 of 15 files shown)


test/start-kubernetes.sh
```diff
@@ -17,15 +17,14 @@ LOCKFILE="${LOCKFILE:-${REPO_DIRECTORY}/_work/start-kubernetes.exclusivelock}"
 LOCKDELAY="${LOCKDELAY:-300}" # seconds
 NODES=( $DEPLOYMENT_ID-master
         $DEPLOYMENT_ID-worker1
-        $DEPLOYMENT_ID-worker2
-        $DEPLOYMENT_ID-worker3)
+        $DEPLOYMENT_ID-worker2)
```
avalluri: Passing this Via CI environment is more appropriate in my opinion, than hardcoding in script.

test/start-kubernetes.sh
```diff
@@ -17,15 +17,14 @@ LOCKFILE="${LOCKFILE:-${REPO_DIRECTORY}/_work/start-kubernetes.exclusivelock}"
...
 CLOUD="${CLOUD:-true}"
 FLAVOR="${FLAVOR:-medium}"
 SSH_KEY="${SSH_KEY:-${RESOURCES_DIRECTORY}/id_rsa}"
 SSH_PUBLIC_KEY="${SSH_KEY}.pub"
 EFI="${EFI:-true}"
 KVM_CPU_OPTS="${KVM_CPU_OPTS:-\
- -m 2G,slots=${TEST_MEM_SLOTS:-2},maxmem=34G -smp 4\
+ -m 2G,slots=${TEST_MEM_SLOTS:-2},maxmem=34G -smp 2\
```
avalluri: I think this can be managed in CI environment configuration.

test/e2e/storage/csi_volumes.go
```diff
@@ -158,6 +160,17 @@ var _ = Describe("PMEM Volumes", func() {
...
+			// Because this test creates a lot of pods, it is useful to
+			// log their progress.
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			to := podlogs.LogOutput{
+				StatusWriter: GinkgoWriter,
+				LogWriter:    GinkgoWriter,
+			}
+			podlogs.CopyAllLogs(ctx, f.ClientSet, f.Namespace.Name, to)
+			podlogs.WatchPods(ctx, f.ClientSet, f.Namespace.Name, GinkgoWriter)
```
avalluri: As the test itself is pending(Pit) test, does this change have any effect now.

test/start-kubernetes.sh
pohly: I was considering that, but then asked myself whether it's actually useful to then have a larger cluster on a desktop. My answer was "no", therefore I'm changing the defaults here.

test/e2e/storage/csi_volumes.go
pohly: No, but it will once someone runs the test manually. IMHO it's useful to have.

test/start-kubernetes.sh
pohly: Same as for the number of CPUs: why would we want a larger configuration on a desktop than in the CI?

pohly:
Testing in the CI showed that the reduced size of the test cluster apparently caused test failures (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-327/43/pipeline) while it worked without that change (current test result in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-327/44/pipeline).

So let's merge it without that change (current state of the PR).

pohly:
> So let's merge it without that change (current state of the PR).

I forgot to force-push the removal of the size commit. Build results were clean only by chance, the first run of testing 1.14 direct had failed and now also fails on devel.

I'm not sure whether this particular commit is really the reason for the flaky test behavior, but let's revert it anyway:
https://github.com/intel/pmem-csi/pull/352

PR #330, opened 2019-07-04
Dockerfile: Fix missing image version issue

avalluri:
Change 5c5f60601246865220eeb63ecb7eba7c63f3e15d introduced 3-phase image
building, in that VERSION argument spans to different(base & binaries) phases.
This makes VERSION passed via Makefile(build-arg) is not reached properly.

This change fixes this by moving the VERSION argument to appropriate phase.


okartau:
confirmed here as well.
I rebuilt with that change included and I see Version: reported in logs again.

PR #334, opened 2019-07-04
WIP: Use driver-global mutex lock for running any system command

okartau:
Stress test exposes multiple troubles that likely
are caused by multiple threads operating on system
(mount, LV commands and alike) at same time.
Allowing one command at a time shows better stability.

### DIFF
--- a/pkg/pmem-device-manager/pmd-util.go
+++ b/pkg/pmem-device-manager/pmd-util.go
@@ -58,8 +58,10 @@ func FlushDevice(dev PmemDeviceInfo, blocks uint64) error {
 	}
 	if blocks == 0 {
 		glog.V(5).Infof("Wiping entire device: %s", dev.Path)
-		// use one iteration instead of shred's default=3 for speed
-		if _, err := pmemexec.RunCommand("shred", "-n", "1", dev.Path); err != nil {
+		// Use one iteration instead of shred's default=3 for speed.
+		// Use "NoMutex" variant because shred takes long time on bigger volumes,
+		// and regular RunCommand uses mutex always.
+		if _, err := pmemexec.RunCommandNoMutex("shred", "-n", "1", dev.Path); err != nil {
 			return fmt.Errorf("device shred failure: %v", err.Error())
 		}
 	} else {
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -136,6 +136,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		options = append(options, "ro")
 	}
 	glog.V(5).Infof("NodePublishVolume: bind-mount %s %s", stagingtargetPath, targetPath)
+	pmemexec.Commandmutex.Lock()
+	defer pmemexec.Commandmutex.Unlock()
 	mounter := mount.New("")
 	if err := mounter.Mount(stagingtargetPath, targetPath, "", options); err != nil {
 		return nil, err
@@ -162,6 +164,8 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 
 	// Unmounting the image
 	glog.V(3).Infof("NodeUnpublishVolume: unmount %s", targetPath)
+	pmemexec.Commandmutex.Lock()
+	defer pmemexec.Commandmutex.Unlock()
 	// Check if the target path is really a mount point. If its not a mount point do nothing
 	if notMnt, err := mount.New("").IsLikelyNotMountPoint(targetPath); notMnt || err != nil && !os.IsNotExist(err) {
 		return &csi.NodeUnpublishVolumeResponse{}, nil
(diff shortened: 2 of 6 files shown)


okartau:
This has positive impact on e2e stress testing.
On a centos-govm cluster, I can run 10 x e2e without error and without leftover volumes,
amazing result that never happened before (test code as from #323)
On a clear-govm system I ran 5x e2e and there were 4 failures and 1 success, few leftovers.
The error landscape seems different, this PR definitely makes things cleaner.

Why centos vs clear systems are different is also good question.
I see one more test skipped in centos, but that's not in stress area it seems.

okartau:
Let's note that this PR uses exec.RunCommand as convenient place to wrap many system commands, but there is another path where some of mount/umount commands run, in mount package. To make command coverage more complete, those commands should be protected by same mutex. Or, to be written as direct mount/umount commands instead of mount package. 

Is mount package really useful for us? Use of it saves some code lines (and there is convenient method to detect device path), but OTOH it produces some overhead: every time we create mount object, there is systemd probe which returns negative. At least in logs this is disturbing as it produces lot of useless noise there about systemd. And in runtime this is wasted effort. Another related question, why it produces negative while target actually runs systemd as PID=1 ? If check passed, mount object could take systemd-oriented approach to mounting.

okartau:
I added same mutex around other mount,umount calls in nodeserrver

pohly:
> Is mount package really useful for us?

As you noticed, it does provide some additional features that are necessary and may not be that easy to reimplement.

> OTOH it produces some overhead: every time we create mount object, there is systemd probe which returns negative. At least in logs this is disturbing as it produces lot of useless noise there about systemd. And in runtime this is wasted effort.

True. There's an effort underway to refactor the code. Perhaps that'll lead to something that is a bit leaner.

> Another related question, why it produces negative while target actually runs systemd as PID=1 ? If check passed, mount object could take systemd-oriented approach to mounting.

The check runs inside a container where PID=1 isn't systemd. systemd probably cannot even be reached from inside the container.

okartau:
I added WIP , dont merge yet, we may discover more protections to add here.

> The only exception is shredding

yes, shred lesson I already had earlier where shred remained inside locked part
and created other trouble (lots of repeated queued requests)

okartau:
speaking of recent "shred inside mutex" lesson... Looks like I am bout to re-create that problem again with current "protect everything we run in pmexec.RunCommand()" change.
What we need is "protect everything that is reasonable short", and treat shred as exception, right. Otherwise all paths that require a command, will remain blocked. e2e does not expose this as all volumes are very small.

pohly:
> Looks like I am bout to re-create that problem again with current "protect everything we run in pmexec.RunCommand()" change.

So you are now seeing an increased rate of "context deadline exceeded" errors? That cannot be addressed by making operations faster, the timeouts are much too low for that. We need to a) make the timeouts higher (I'm working on it) and b) ensure that if they are exceeded, the system still remains stable (idempotent calls, etc.).

At the moment, we get way too many timeouts, too early, and then the system seems to deteriorate (>50 min runtime of the E2E stress test for direct mode). I'm not sure yet why it gets that slow - perhaps it's because external-provisioner starts to back off exponentially.

okartau:
> So you are now seeing an increased rate of "context deadline exceeded" errors?

I haven't counted those errors, but it makes sense to count those, this PR vs HEAD.
The shred point just popped to my head because I was thinking of earlier mutex lesson, and now I inserted shred into mutex section again. I now pushed new version taking shred out of that set (creating another RuCommand method in parallel). tests going...

deteriorating tests: I dont see any problems (one e2e takes 260 sec and I run 10x of it in sequence), but I run tester code of #323, perhaps that makes difference

pohly:
> deteriorating tests: I dont see any problems (one e2e takes 260 sec and I run 10x of it in sequence), but I run tester code of #323, perhaps that makes difference

It does make a difference, because external-provisioner is not involved anymore when using the sanity stress test from #323.

okartau:
Results of my own repeated e2e of this code (test code from #323): 
OK/Total
Adding test counts as these slightly vary (skip reasons to be explained and eliminated in the future)
clear-govm 5/10  45 Passed
centos-govm 10/10 OK  44 Passed
ubuntu-govm 1/5 44 Passed
HW (1-node cluster): 3/3 41 Passed (skips reboot tests)

Based on these, current PR is clearly improvement as without that change, the results were usually 0/10.
Still remarkable that different distro-based govm clusters are quite different by counts.

pkg/pmem-device-manager/pmd-util.go
```diff
@@ -59,7 +59,7 @@ func FlushDevice(dev PmemDeviceInfo, blocks uint64) error {
 	if blocks == 0 {
 		glog.V(5).Infof("Wiping entire device: %s", dev.Path)
 		// use one iteration instead of shred's default=3 for speed
-		if _, err := pmemexec.RunCommand("shred", "-n", "1", dev.Path); err != nil {
+		if _, err := pmemexec.RunCommandNoMutex("shred", "-n", "1", dev.Path); err != nil {
```
pohly: A comment explaining why this doesn't use `RunCommand` would be worthwhile here.

okartau: comment added, force pushed

okartau:
cherry-picked commit from #323 here to see combined effect on CI

avalluri:
Seems like #336 is more reliable, Can we close this in favor of that.

okartau:
> Seems like #336 is more reliable, Can we close this in favor of that.

I agree, closing

Issue #338, opened 2019-07-08
Volume deletion code path should block calling ClearDevice,shred for volume that is in use


okartau:
Recent cases like #317 #337 have indicated there is code path where a volume that remained mounted, gets erased by shred. This has happened for various reasons, few cases been caused by incorrect paths use by tester code. The reason does not matter so much: if the code can take the path, we should handle it in less-damaging way, avoiding destruction of live file system.

On a call path: DeleteVolume-ClearDevice we should detect that the corresponding file system is in use, and return error **before** calling ClearDevice and shred.
Current code returns error anyway just one step later in lvremove, so shred of device is unnecessary destruction.
This analysis is based on LVM mode where most of recent cases have been debugged, but likely is equally relevant in direct mode as well.

pohly:
Olev Kartau <notifications@github.com> writes:
> This has happened for various reasons, few cases been caused by
> incorrect paths use by tester code. The reason does not matter so
> much: if the code can take the path, we should handle it in
> less-damaging way, avoiding destruction of live file system.

It's good if the driver has sanity checks against a malfunctioning CO,
but ultimately it's still the COs (or in this case, the test code which
pretends to be the CO) which must behave correctly.

Issue #317, opened 2019-06-27
shred of mounted volume corrupts ext4 file system showing zetabyte sizes


pohly:
The new disruptive test in PR #196 causes LVM on real hardware to become confused and report a volume with zetabytes size (for details see https://github.com/intel/pmem-CSI/pull/196#issuecomment-505846744):

```
/dev/dm-1                 53Z   53Z     0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.fIHglA/target
```

okartau:
Change in story: rebooting of node never happened. Rebooting is supposed to happen via _work/CLUSTER/ssh helper, but that does not exist in HW-based system, so it gets skipped.
But file system gets corrupted without any reboot needed.
One HW-pmem node in cluster, running e2e; After test there will be this mount:
```
Filesystem               Size  Used Avail Use% Mounted on
/dev/dm-1                 62Z   62Z     0 100% /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.mAKbPI/target
```

In dmesg:
```
[Wed Jul  3 14:43:32 2019] attempt to access beyond end of device
[Wed Jul  3 14:43:32 2019] dm-1: rw=2049, want=6053473084835904072, limit=49152
[Wed Jul  3 14:43:37 2019] EXT4-fs error (device dm-1): ext4_map_blocks:592: inode #8: block 11: comm jbd2/dm-1-8: lblock 1 mapped to illegal pblock 11 (length 1)
[Wed Jul  3 14:43:37 2019] jbd2_journal_bmap: journal block not found at offset 1 on dm-1-8
[Wed Jul  3 14:43:37 2019] Aborting journal on device dm-1-8.
```

lvs shows that device as leftover:
```
# lvs
  LV                                   VG                 Attr       LSize    Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  root                                 fedora             -wi-ao---- <930.19g                                                    
  e916406b-9d87-11e9-8078-7a34dbfd3c91 ndbus0region0fsdax -wi-ao----   24.00m                                                    
```

In e2e output, there is error logged about lvremove failing on same device:

```
pmem-csi-node-vqwmw/pmem-driver: I0703 11:44:27.361841       1 glog.go:58] Executing: lvremove -fy /dev/ndbus0region0fsdax/e916406b-9d87-11e9-8078-7a34dbfd3c91
[...]

PR #336, opened 2019-07-05
DeviceManager/LVM: Synchronize all lvm calls

avalluri:
Often LVM stress tests are failing due to race conditions in executing LVM operations.
Hence, we added mutex locking so that only one LVM operation is in progress.
    
As part of this also made changes to cache LVM volumes by device manager so that it avoids repeated lvs calls.

This is an alternative approach to #334 to attack the lvm issues

PR #340, opened 2019-07-08
Refuse to clear mounted device

okartau:
There is code path where a volume that remained mounted, gets
erased by shred. This has happened for various reasons, like
incorrect paths use by tester code.
Detect that the device is opened, and return error before
performing shred, avoiding unnecessary destruction that
would show as garbage remaining mounted as file system.

Resolves: #317
Resolves: #338

### DIFF
--- a/pkg/pmem-device-manager/pmd-util.go
+++ b/pkg/pmem-device-manager/pmd-util.go
@@ -9,6 +9,7 @@ import (
 
 	pmemexec "github.com/intel/pmem-csi/pkg/pmem-exec"
 	"k8s.io/klog/glog"
+	"k8s.io/kubernetes/pkg/util/mount"
 	"k8s.io/utils/keymutex"
 )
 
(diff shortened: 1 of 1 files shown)


okartau:
I tested this on top of #336 and HW-based system which skips reboots, so it exposes the code path for this case: volume remains mounted but is attempted to remove

pohly:
> I tested this on top of #336 and HW-based system which skips reboots, so it exposes the code path for this case: volume remains mounted but is attempted to remove

So now the cleanup code gets the "<path> is open" error?

okartau:
> So now the cleanup code gets the " is open" error?

yes:
```
cleanup: error: DeleteVolume: rpc error: code = Internal desc = Failed to delete volume 46061a5f-a177-11e9-b1d4-32f2a830ec0e on node-3: rpc error: code = Internal desc = Failed to delete volume: /dev/ndbus0region0fsdax/46061a5f-a177-11e9-b1d4-32f2a830ec0e is open
```

okartau:
addressed 3 comments, rebased

Issue #320, opened 2019-07-01
error handling in OnNodeAdded


pohly:
PR #196 added code in `OnNodeAdded` which merely logs warnings when things go wrong (https://github.com/intel/pmem-CSI/pull/196#discussion_r298898414).

We need to expose those failures as errors that are more visible to the cluster admin, ideally such that our current "is PMEM-CSI healthy" checks detect the issue, and PMEM-CSI needs to recover from those errors automatically.

pohly:
One possible solution would be to let `OnNodeAdded` return an error. That could be returned to the node which wanted to add itself, and then the retry loop there will ensure that a) the problem is visible (because the node does not become ready) and b) the problem eventually fixes itself.

Whether `OnNodeAdded` then gets called before or after making data structure changes is to be determined.

PR #342, opened 2019-07-09
RegistryServer: Report back errors from OnNodeAdded

avalluri:
Failures happened in RegistryListener.OnNodeAdded() are supposed to send back to
appropriate node, and the node registration should fail. This allows the node
driver to make this error visible to cluster admins via driver health check.

FIXES: #320

### DIFF
--- a/pkg/registryserver/registryserver.go
+++ b/pkg/registryserver/registryserver.go
@@ -7,6 +7,7 @@ import (
 
 	pmemgrpc "github.com/intel/pmem-csi/pkg/pmem-grpc"
 	registry "github.com/intel/pmem-csi/pkg/pmem-registry"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


pkg/registryserver/registryserver.go
```diff
@@ -115,7 +117,12 @@ func (rs *RegistryServer) RegisterController(ctx context.Context, req *registry.
 
 	if !found {
 		for l := range rs.listeners {
-			l.OnNodeAdded(ctx, node)
+			if err := l.OnNodeAdded(ctx, node); err != nil {
```
pohly: This is now racy again:
- RegisterController is called twice for the same node
- first call adds the node, fails in OnNodeAdded
- second call finds that the node is registered
- first call removes the node and returns an error
- second call returns success
- node thinks that it is registered but in reality it isn't

The correct solution is to lock the entire RegisterController until it either succeeds (including the OnNodeAdded) or fails.

Injecting foreign code via this callback mechanism is what makes this all so complicated. Can registering nodes and establishing a connection to them be done entirely inside this file?

avalluri: Second `RegisterController `call from same node waits till the first call completes. There is mutex lock based on `NodeId`,

pohly: Okay, that works.

Can you add explanations of how this locking works to the source code?

avalluri: > Can you add explanations of how this locking works to the source code?

Done

PR #343, opened 2019-07-09
DeviceManager/ndctl: Fix deadlocking

avalluri:
With recent change #6a8d1ee6b9ee4ba371730a23e363ad7589c11908 FlushDevice()
itself handles the mutex(volumeMutex) locking and unlocking, so
callers(LVM,ndctl) should not lock explicitly while calling this utility method.

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/intel/pmem-csi/pkg/ndctl"
 	pmemexec "github.com/intel/pmem-csi/pkg/pmem-exec"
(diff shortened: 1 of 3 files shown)


pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -19,12 +20,14 @@ var _ PmemDeviceManager = &pmemLvm{}
 var lvsArgs = []string{"--noheadings", "--nosuffix", "-o", "lv_name,lv_path,lv_size", "--units", "B"}
 var vgsArgs = []string{"--noheadings", "--nosuffix", "-o", "vg_name,vg_size,vg_free,vg_tags", "--units", "B"}
 
+var lvmMutex = &sync.Mutex{}
```
pohly: Can you briefly explain what this mutex serializes and why? Reference the known issue with LVM here.

avalluri: I can add what it is used for. But For LVM call synchronization there is no specific open issue.

avalluri: Done.

avalluri:
> This looks okay now, but the second commit fixes a problem introduced by the first one, so it would be better to squash.
> 
> Locking would be more obviously correct if both pmd-lvm.go and pmd-ndctl.go had their own mutex which is only used in that one file. With the current global `devicemutex` it is both harder to explain what it is used for and why it is used in both cases, and one has to be careful that it isn't used elsewhere.
> 
> Let's not micro-optimize by sharing the same variable instance and instead have a lvmMutex and a ndctlMutex.

Did the changes to use device manager specific mutexes. That resulted in removing obsolete `{Clear,Flus}Device()`'s  internal mutex.

pohly:
Tested successfully in the CI in combination with PR #346: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-327/41/execution/node/72/log/

PR #345, opened 2019-07-10
RestartNode: use /run for temp.file instead of /tmp

### LINKED ISSUE #325: e2e: restartNode() relies on /tmp mounted on tmpfs which is not always true
The shutdown-reboot detection logic in test/e2e/storage/sanity.go: restartNode()
relies on /tmp being mounted on tmpfs (RAM based) but not all distros do that:
At least Ubuntu and Centos have /tmp on rootfs, means the restart 
detection via file used as a flag would fail there.
We could use a path which is always mounted to tmpfs, what about /dev/shm.

Another proposal in that code:
The comment "TCPKeepAlive is necessary because otherwise..."
refers to option given in ssh helper scripts, it would add clarity to mention that in the comment

okartau:
/tmp is not guaranteed to reside on tmpfs-type file system,
it is part of rootfs on multiple distros (CentOS, Ubuntu etc).

Resolves #325

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -672,9 +672,9 @@ func restartNode(cs clientset.Interface, nodeID string, sc *sanity.SanityContext
 		os.Getenv("CLUSTER"),
 		node)
 	// We detect a successful reboot because a temporary file in the
-	// tmpfs /tmp will be gone after the reboot.
-	out, err := exec.Command(ssh, "touch", "/tmp/delete-me").CombinedOutput()
-	framework.ExpectNoError(err, "%s touch /tmp/delete-me:\n%s", ssh, string(out))
+	// tmpfs /run will be gone after the reboot.
+	out, err := exec.Command(ssh, "sudo", "touch", "/run/delete-me").CombinedOutput()
+	framework.ExpectNoError(err, "%s touch /run/delete-me:\n%s", ssh, string(out))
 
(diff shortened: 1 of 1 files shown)


okartau:
First version of this PR is trivial replace of path in 5 locations (4 code, 1 comment).
Idea: what about using single definition and use it as variable.

pohly:
`/run` (https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) is probably a better place. It's guaranteed to exist and not hold data from a previous boot. `/dev/shm` isn't standardized.

okartau:
ok, I will change to /run. I have habit myself to use /dev/shm as it's writable by user, while /run is not. But in current case that does not matter as file is created by root, right.

pohly:
> But in current case that does not matter as file is created by root,
> right.

Correct.

okartau:
actually, not by current code as CI faithfully notices:
touch: cannot touch '/run/delete-me': Permission denied
ssh helper scripts make login as user=clear and current code does not prepend sudo, so we run those commands as user. And /tmp was easier as it is user-writable.
We need to prepend sudo if we want to use /run

pohly:

> actually, not by current code as CI faithfully notices:
> touch: cannot touch '/run/delete-me': Permission denied
> ssh helper scripts make login as user=clear and current code does not prepend sudo, so we run those commands as user. And /tmp was easier as it is user-writable.
> We need to prepend sudo if we want to use /run

Right, let's do that.

PR #346, opened 2019-07-10
failure handling

pohly:
Addresses two issues seen by @okartau when testing on real hardware:
- failure not logged
- failure not captured by GinkgoRecover

The problem itself most likely was an error during directory setup or teardown. That can still happen, but now it should be visible.

The problem itself might have been the "126 = cannot execute error" that I also got in a CI run. Workaround for that also included.

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	clientexec "k8s.io/client-go/util/exec"
 	"k8s.io/kubernetes/test/e2e/framework"
 	testutils "k8s.io/kubernetes/test/utils"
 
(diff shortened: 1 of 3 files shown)


okartau:
Tried 10 x e2e on same system with code of this PR; no errors, no leftovers, clean pass.
I will try some more.

test/e2e/wrapper.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}()
+
+	ginkgo.Fail(message, skip)
+}
+
+// ginkgo adds a lot of test running infrastructure to the stack, so
+// we filter those out
+var stackSkipPattern = regexp.MustCompile(`onsi/ginkgo`)
+
```
marcemq: Missing `pruneStack` function description.

test/e2e/wrapper.go
```diff
@@ -0,0 +1,100 @@
...
+	Message        string // The failure message passed to Fail
+	Filename       string // The filename that is the source of the failure
+	Line           int    // The line number of the filename that is the source of the failure
+	FullStackTrace string // A full stack trace starting at the source of the failure
+}
+
+// String makes FailurePanic look like the old Ginkgo panic when printed.
+func (FailurePanic) String() string { return ginkgo.GINKGO_PANIC }
+
+// Fail wraps ginkgo.Fail so that it logs failures as they occur
```
marcemq: I think the first word in description should be the function full name, isn't it? If so, would please add it?

test/e2e/wrapper.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(callerSkip) > 0 {
+		skip += callerSkip[0]
+	}
+
+	_, file, line, _ := runtime.Caller(skip)
+	fp := FailurePanic{
+		Message:        message,
+		Filename:       file,
+		Line:           line,
+		FullStackTrace: pruneStack(skip),
```
marcemq: What if `pruneStack()` returns a really long stack trace to `FullStackTrace` variable, we might face an overflow error, don't you think?

test/e2e/wrapper.go
pohly: It's an internal function. Those do not have to have a description. Doesn't hurt of cause, but this is code from Kubernetes and I'm just copying it temporarily, therefore I want to make as little local changes as possible.

test/e2e/wrapper.go
pohly: True.

test/e2e/wrapper.go
pohly: Overflow of what? Go allocates everything dynamically.

test/e2e/wrapper.go
pohly: `Fail` used to be the function name, but I had to rename it later. Fixed.

pohly:
Rebased and included another workaround for a temporary error during directory creation/deletion.

test/e2e/storage/sanity.go
```diff
@@ -561,9 +561,13 @@ func (v volume) create(sizeInBytes int64, nodeID string) (string, *csi.Volume) {
 			},
 		}
 	}
-	vol, err := v.cc.CreateVolume(
-		v.ctx, req,
-	)
+	var vol *csi.CreateVolumeResponse
+	err = v.retry(func() error {
```
avalluri: @pohly I guess this change might affect other sanity tests, not just stress test. Is that expectation?

test/e2e/wrapper.go
marcemq: Got it.

test/e2e/storage/sanity.go
pohly: That's a good point. It was intended that they also retry, however, there's one small problem with that: the default context in other tests doesn't have a timeout. That needs to be added, otherwise permanent errors cause retries until the entire test suite is killed.

pohly: I've added a default timeout of 5 minutes.

PR #349, opened 2019-07-12
Better alignment handling

### LINKED ISSUE #348: volume size rounding
The driver should:
- do rounding only in the layer (LVM or ndctl) which actually knows about size limitations
- only ever round up, not down: the request is to provide at least the specified amount of storage
- report back the actual size in `CreateVolumeResponse` to ensure that the result is visible in Kubernetes (stored in the PV)

Right now rounding done downwards (issue #119) and fails in some configurations when the size is 0. The actual size is not reported.

okartau:
Stop doing aligning on 2 layers, only do it in ndctl
and align rounding up there. In pmem-device-manager
layer make sure that size what we ask is at least 2GB.

Adding "get interleve ways" function in ndctl layer so that
namespace creation function can adjust to real alignment value.

Resolves #348

### DIFF
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -188,10 +188,9 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
 		ways := uint64(C.ndctl_region_get_interleave_ways(ndr))
 		align := opts.Align * ways
 		if opts.Size%align != 0 {
-			// force-align down to block boundary. More sensible would be to align up, but then it may fail because we ask more then there is left
-			opts.Size /= align
-			opts.Size *= align
-			glog.Warningf("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
+			// Round up size to align with next block boundary.
+			opts.Size = (opts.Size/align + 1) * align
+			glog.V(4).Infof("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
 				regionName, ways, opts.Align, opts.Size)
 		}
 	}
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -83,6 +83,11 @@ func (r *Region) Readonly() bool {
 	return C.ndctl_region_get_ro(ndr) != 0
 }
 
+func (r *Region) InterleaveWays() uint64 {
+	ndr := (*C.struct_ndctl_region)(r)
+	return uint64(C.ndctl_region_get_interleave_ways(ndr))
+}
+
 //ActiveNamespaces returns all active namespaces in the region
 func (r *Region) ActiveNamespaces() []*Namespace {
 	return r.namespaces(true)
--- a/pkg/pmem-csi-driver/controllerserver-node.go
+++ b/pkg/pmem-csi-driver/controllerserver-node.go
@@ -174,12 +174,6 @@ func (cs *nodeControllerServer) CreateVolume(ctx context.Context, req *csi.Creat
 	}
 
 	asked := req.GetCapacityRange().GetRequiredBytes()
-	// Required==zero means unspecified by CSI spec, we create a small 4 Mbyte volume
-	// as lvcreate does not allow zero size (csi-sanity creates zero-sized volumes)
-	// FIXME(avalluri): This check should be handled by LvmDeviceManager??
-	if asked <= 0 {
-		asked = 4 * 1024 * 1024
-	}
 
 	if vol = cs.getVolumeByName(req.Name); vol != nil {
 		// Check if the size of existing volume can cover the new request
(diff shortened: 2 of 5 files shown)


okartau:
The first commit handles the align part. Another commit is needed to fix reporting back part

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -215,6 +215,19 @@ func (cs *nodeControllerServer) CreateVolume(ctx context.Context, req *csi.Creat
 
 		cs.mutex.Lock()
 		defer cs.mutex.Unlock()
+		devices, err := cs.dm.ListDevices()
```
avalluri: @okartau I would suggest to modify `DeviceManager.CreateDevice()` to return newly created device. It was my fault, I should have done that while designing the DeviceManager interface.

okartau: Good point, that makes sense, added a commit doing this

pkg/ndctl/region.go
```diff
@@ -188,8 +188,9 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
 		ways := uint64(C.ndctl_region_get_interleave_ways(ndr))
 		align := opts.Align * ways
 		if opts.Size%align != 0 {
-			// force-align down to block boundary. More sensible would be to align up, but then it may fail because we ask more then there is left
+			// align up to block boundary.
 			opts.Size /= align
+			opts.Size += 1
```
pohly: I think it would be cleaner to write this as:
```
// Round up size to align with next block boundary.
opts.Size = (opts.Size / align + 1) * align
```

pkg/ndctl/region.go
```diff
@@ -188,8 +188,9 @@ func (r *Region) CreateNamespace(opts CreateNamespaceOpts) (*Namespace, error) {
...
 			opts.Size *= align
 			glog.Warningf("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
```
pohly: This shouldn't be a warning. A warning implies that something unexpected happened. But users have no knowledge about the expected alignment, therefore it is normal that this happens. Better use an info message at a lower priority.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -90,11 +90,12 @@ func (pmem *pmemNdctl) CreateDevice(name string, size uint64, nsmode string) err
 		glog.V(4).Infof("Device with name: %s already exists, refuse to create another", name)
 		return fmt.Errorf("CreateDevice: Failed: namespace with that name exists")
 	}
-	// align up by 1 GB, also compensate for libndctl giving us 1 GB less than we ask
+	// align by 1 GB
```
pohly: Aligning implies that we ensure that the value is an integer multiple of some base value. Here we just enforce a minimum size. I would describe this as:
```
// Increase to minimum size supported by libndctl.
const minsize uint64 = 2 * 1024 * 1024 * 1024
if size < minsize {
    size = minsize
}
```

pkg/pmem-ns-init/main.go
```diff
@@ -72,10 +72,12 @@ func initNVdimms(ctx *ndctl.Context, useforfsdax int, useforsector int) {
 }
 
 func createNS(r *ndctl.Region, uselimit int, nsmode ndctl.NamespaceMode) {
+	align := uint64(1024 * 1024 * 1024)
```
pohly: Same as in the other commit: `align` here isn't really about alignment, it's a minimum size, right?

pkg/pmem-device-manager/pmd-ndctl.go
okartau: you are right, I was lazy and re-used align value as minsizize value, will clean this up.

pkg/pmem-ns-init/main.go
okartau: align here is actually exactly for alignment, used down in
r.CreateNamespace call:
```
Align: align,
```

The confusion is created by my lazy approach to use same value for min.size. I will make it cleaner.

okartau:
I addressed all (I hope) points from recent review round, pushed new version

pohly:
Testing in direct mode is taking a long time: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-349/

okartau:
yes I notice  that too, but it's not easy to see where time goes/went, when I look to log text of that direct stage, it covers 18:09....18:29 i.e. approx 30 min, but stage itself has been running for 2h:30 already:
testing 1.14 direct
 - 2h 30m 54s

(whole CI cycle now 3h24min)
 Did I miss something.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -90,11 +90,13 @@ func (pmem *pmemNdctl) CreateDevice(name string, size uint64, nsmode string) err
...
-	size /= align
-	size += 2
-	size *= align
+	// Increase to minimum size supported by libndctl.
+	const minsize uint64 = 2 * 1024 * 1024 * 1024
+	if size < minsize {
+		size = minsize
+	}
+	// use align by 1 GB in creation request
+	const align uint64 = 1024 * 1024 * 1024
```
pohly: Just curious: do we have documentation about these limitations (minimum size and alignment)? As it stands here, these are magic numbers with no explanation.

PR #350, opened 2019-07-12
Bugfix setup-ca-kubernetes.sh script

gabrielbriones:
The the format for the existing_secrets variable was not right, caused
to not find the nodes with the pmem-csi-node-secrets correctly. Removed
unnecesary quotes and whitespaces.

Signed-off-by: Gabriel Briones <gabriel.briones.sayeg@intel.com>


okartau:
Indeed, the exiting code produces entries with extra double space added, and whole series is enclused in single quotes. The changed code produces cleaner list. Has this ever worked before (was never used?)?-
BTW, there is extra "the" in commit msg.

okartau:
Do I need to add that change simplifying quotes in style
```
"'"'something'"'"
```
feels always welcome.
We still have double double quotes, but seems these are needed.

PR #351, opened 2019-07-15
Add deploy/kubernetes-1.15 as symlink to kubernetes-1.14

okartau:


okartau:
improved commit msg, pushed

PR #358, opened 2019-07-16
test: reduce sanity stress test size of direct mode

pohly:
The small volume size gets rounded up considerably, which makes test
execution noticably slower than for LVM mode. To get back to acceptable
execution times we create less volumes.


okartau:
What about idea to use eraseafter=false in CI testing? Do those tests actually depend/care does the media get fully erased after use?

pohly:
The tests don't care, but I think it's better to execute the normal flow.

pohly:
This cuts down test time in the CI for direct mode on 1.14 from 34min to 26min. Direct mode is still slower than 1.14 LVM (15min), but it's a bit better.

@okartau: please merge if you agree.

PR #360, opened 2019-07-18
CI caching

pohly:
This idea originally came up when the private registry was introduced
and a search then showed that this has indeed been done before
(https://andrewlock.net/caching-docker-layers-on-serverless-build-hosts-with-multi-stage-builds---target,-and---cache-from/):
with `--cache-from` we can use existing images to warm up the Docker
build cache. This keeps the Dockerfile and thus normal developer
workflow simple (no need to manually build base images first) while
making the CI faster.
    
The we get the most benefit out of caching the build image, but
pulling the PMEM-CSI driver image is also worthwhile. When doing both
and nothing changed, the two image build stages take ~8min instead of
~19min.
    
Only the non-PR jobs push build images. PR jobs pick up those images
and use the same cachbust value. Besides faster build times this also
has the advantage that PR jobs cannot break if Clear Linux changes,
because they will always use a Clear Linux version that was used
successfully before.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,12 @@ pipeline {
 
         PMEM_PATH = "/go/src/github.com/intel/pmem-csi"
         REGISTRY_NAME = "cloud-native-image-registry.westus.cloudapp.azure.com"
-        BUILD_IMAGE = "${env.REGISTRY_NAME}/pmem-clearlinux-builder"
+        // Per-branch build environment, marked as "do not promote to public registry".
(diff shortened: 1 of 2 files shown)


pohly:
This should be ready for merging now.

pohly:
@okartau clean test results, too.

The more interesting test however will be after merging into `devel`: we then need wait for the `devel` rebuild and check that future PRs benefit from caching.

pohly:
Not quite successful.

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/devel/34 created new images from scratch, but it pushed a `pmem-clearlinux-builder:null-rejected` image - `null` because `CHANGE_TARGET` was empty.

I'll have to experiment with a `ci` branch to figure that out.

PR #363, opened 2019-07-19
CI: document and implement release strategy

pohly:
This outlines what we intend to do. Not all of it is currently
implemented (no master branch, CI not fully operational).

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,6 +4,9 @@ Table of Contents
 - [Code quality](#code-quality)
     - [Coding style](#coding-style)
     - [Input validation](#input-validation)
+- [Release management](#release-management)
+-   - [Branching](#branching)
+    - [Tagging](#tagging)
 - [APIs](#apis)
     - [CSI API](#csi-api)
     - [Network ports](#network-ports)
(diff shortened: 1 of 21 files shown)


DEVELOPMENT.md
```diff
@@ -55,6 +56,50 @@ Nonetheless, input needs to be validated to catch mistakes:
   refuses to send messages that are larger
   (https://godoc.org/google.golang.org/grpc#MaxSendMsgSize)
 
+Branching strategy
+==================
+
+The `master` branch is the main branch. It is guaranteed to have
+passed full CI testing. However, it always uses the latest Clear Linux
+for building container images, so changes in Clear Linux can break the
+building of older revisions.
```
pohly: Another approach would be to lock Clear Linux also on the `master` branch. But that means that merging new changes from `devel` becomes more complicated:
- run `hack/update-clear-linux-base.sh` on the `devel` branch
- commit locally
- run the tests
- merge master with that local commit

It actually should be fairly doable, I just don't know whether it is worth it. In practice, Clear Linux updates haven't been problematic for building PMEM-CSI (although Go 2.0 entering Clear could change that). It's Kubernetes on Clear which is problematic, and that currently doesn't get addressed by `hack/update-clear-linux-base.sh` but rather by the `TEST_CLEAR_LINUX_VERSION` values in our Jenkinsfile.

pohly:
It occurred to me that we need to think and do a bit more about the content of an actual release:
- use the real registry in our deployment files
- replace "canary" image version before tagging a release

Here's my proposal.

While testing it I also ran into a special situation: the Clear Linux base image is currently already up-to-date. This made it necessary to update our update-clear-linux-base.sh script.

I verified that the Jenkinsfile change works:
- building a release-0.5-test branch does not push images: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/release-0.5-test/10/execution/node/50/log/
- building the v0.5.0-rc2 tag does push images: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/tags/job/v0.5.0-rc2/11/execution/node/50/log/

okartau:
Naming management scripts directory hack/ is a bit disturbing. I see that is quite commonly used name in k8s related repos, (is there some special need I dont know to have it named like this?) but still, could we have a different name for release & branch mgmt scripts location? hack/ for me associates with something temporary/trial/unsure, and releases making is all opposite.

pohly:
> Naming management scripts directory hack/ is a bit disturbing. I see that is quite commonly used name in k8s related repos

That's why I named it the same way. We can name it whatever we want of course. Do you have a proposal?

okartau:
let's keep it as it is, no big deal. Following common culture is also important.

PR #313, opened 2019-06-25
WIP: enable more e2e test suites.

okartau:
This is about enabling more e2e test suites.
I started with quick survey by enabling one line at a time in block of lines marked with TODO.
Brief comments are added in the file.
There was one test suite that did not fail in my run, activated, to see the impact on CI side.

PR #365, opened 2019-07-22
WIP: not-for-merge: Enable volumemode tests

okartau:
EXPERIMENTAL: use POST instead of local kubectl in volumemode tests
    
This is experimental, NOT FOR MERGE, try local changes under vendor/
to verify these pass CI.
We avoid use of local kubectl in volumemode tests,
by using ExecCommandInContainerWithFullOutput() instead of
RunKubectl(). As there are other call paths to PodExec(),
we do not change it but create a variant PodExec2().
Framework needs to be passed down is call chain.
VerifyExecinPodFail change: the error code returned
after change is not castable by old code, so the
import of exec is changed.
The code is also simplified slightly: checking err
using Expect before further checks, means the
if-check of it is not needed.


vendor/k8s.io/kubernetes/test/e2e/storage/utils/utils.go
```diff
@@ -53,9 +54,16 @@ func PodExec(pod *v1.Pod, bashExec string) (string, error) {
 	return framework.RunKubectl("exec", fmt.Sprintf("--namespace=%s", pod.Namespace), pod.Name, "--", "/bin/sh", "-c", bashExec)
 }
 
+// PodExec2 wraps RunKubectl to execute a bash cmd in target pod
+func PodExec2(f *framework.Framework, pod *v1.Pod, bashExec string) (string, error) {
+	stdout, _, err := f.ExecCommandInContainerWithFullOutput(pod.Name, pod.Spec.Containers[0].Name, "/bin/sh", "-c", bashExec)
```
pohly: This proposed `PodExec[2]` has a conceptual problem: it assumes that the pod runs in the namespace created for `f`, which might not be the case. `f` might not even *have* a namespace.

I think it makes sense to create a `ExecCommandInContainerWithFullOutput` function in `test/e2e/framework/pods` which takes client set and namespace as parameters instead of using the ones in `f`. The functions in `f` can be kept for now, they simply became wrappers for the actual implementation. Later they can be removed, if it turns out that they don't have enough value by themselves.

Once we have those new functions, `PodExec2` just needs a new client set parameter instead of `f`.

okartau: ExecCommandInContainerWithFullOutput() is mostly just wrapper which calls ExecWithOptions(), which uses f.ClientSet.
Means that we should create variants of both `ExecCommandInContainerWithFullOutput` and `ExecWithOptions` under `test/e2e/framework/pods` ?

pohly:
> Means that we should create variants of both
> `ExecCommandInContainerWithFullOutput` and `ExecWithOptions` under
> `test/e2e/framework/pods` ?

Yes.

okartau:
pushed new version to show it, to verify did I get the idea right. At least it compiles, runs and test succeeds. It's still trial so not very clean, and I accidentally messed changes into 2nd commit (which had only test enablement) so I then squashed everything to one change, it's a trial anyway (will separate that again later).

vendor/k8s.io/kubernetes/test/e2e/framework/pods.go
```diff
@@ -292,3 +295,54 @@ func (c *PodClient) PodIsReady(name string) bool {
 	ExpectNoError(err)
 	return podutil.IsPodReady(pod)
 }
+
+// ExecWithOptions executes a command in the specified container,
+// returning stdout, stderr and error. `options` allowed for
+// additional parameters to be passed.
+func (c *PodClient) ExecWithOptions(options ExecOptions, cli clientset.Interface) (string, string, error) {
```
pohly: Nit: "cli" is a very unusual parameter name for "clientset.Interface", it occurs only once in the entire Kubernetes code base. "cs" is much more common. It's also more common to have it first in the parameter list.

okartau:
thanks, I pushed new state where I separated two changes again into separate commits, and changed cli to cs

okartau:
Added 5th commit that adds some notes close to test suites list, recording the results from enabling trials performed under PR #313 so  that we can close #313

okartau:
I created a PR to k8s based on "vendor part" of this PR.
Initial feedback in 84042 and 83913 is that local kubectl is part of plans and is not going away.
So is there point in trying to affect k8s part at all, or should we adapt and start using local kubectl,
(as de facto already happened as we did Volumes test suite enabling without any effort to reduce local kubectl dependency and added local kubectl on tester host).

pohly:
> I created a PR to k8s based on "vendor part" of this PR.
> Initial feedback in 84042 and 83913 is that local kubectl is part of plans and is not going away.
> So is there point in trying to affect k8s part at all,

Yes, I think it is a worthwhile improvement of the tests.

PR #366, opened 2019-07-22
README additions: LVM/Direct mode table, supported versions

### LINKED ISSUE #355: Add an overview of LVM, Direct modes in README.md.
I would help to have an overview of LVM, Direct modes in README.md. 
Right now there's no guidance to users which mode to use.
An overview table would be nice, with two columns (LVM and direct) and several rows:
- minimum size of volumes
- alignment requirements
- handling of fragmentation
- whatever else is relevant

okartau:
Resolves #355

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -224,32 +224,31 @@ Notes about switching device mode
 ================================
 
 If device mode is switched between LVM and direct(aka ndctl), please keep
-in mind that PMEM-CSI driver does not clean up or reclaim Namespaces,
-therefore Namespaces plus other related context (LVM state)
+in mind that PMEM-CSI driver does not clean up or reclaim namespaces,
+therefore namespaces plus other related context (LVM state)
 created in previous mode will remain stored on device and most likely
 will create trouble in another device mode.
 
 Going from LVM device mode to direct device mode
 ------------------------------------------------
 
-- examine LV Groups state on a node: `vgs`
-- examine LV Phys.Volumes state on a node: `pvs`
-- Delete LV Groups before deleting namespaces: `vgremove VGNAME`, to
-  avoid orphaned VGroups
+- examine LV groups state on a node: `vgs`
+- examine LV physical volumes state on a node: `pvs`
+- delete LV groups before deleting namespaces to avoid orphaned volume groups: `vgremove VGNAME`
 
 NOTE: The next **WILL DELETE ALL NAMESPACES** so be careful!
 
-- Delete Namespaces on a node using CLI: `ndctl destroy-namespace all --force`
+- Delete namespaces on a node using CLI: `ndctl destroy-namespace all --force`
 
 Going from direct device mode to LVM device mode
 ----------------------------------------------
 
-No special steps are needed to clean up Namespaces state.
+No special steps are needed to clean up namespaces state.
 
 If PMEM-CSI driver has been operating correctly, there should not be
-existing Namespaces as CSI Volume lifecycle should have been deleted
-those after end of life of Volume. If there are, you can either keep
-those (LVM device mode does honor "foreign" Namespaces and leaves those
+existing namespaces as CSI volume lifecycle should have been deleted
+those after end of life of volume. If there are, you can either keep
+those (LVM device mode does honor "foreign" namespaces and leaves those
 alone) if you have enough space, or you can choose to delete those
 using `ndctl` on node.
 
@@ -265,7 +264,7 @@ Read-only access to /sys
 
 In some deployment schemes /sys remains mounted read-only in the
 container running pmsm-csi-driver. This creates problem for the
-driver which needs write access to /sys for Namespaces management
+driver which needs write access to /sys for namespaces management
 operations. There is start-time check for read-write mount of /sys in
 the code. An error in pod log `pmem-driver: Failed to run driver:
 FATAL: /sys mounted read-only, can not operate` is the sign of such
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -17,18 +17,18 @@ Table of Contents
     - [Specific arguments to pmem-vgm](#specific-arguments-to-pmem-vgm)
     - [Specific arguments to pmem-csi-driver](#specific-arguments-to-pmem-csi-driver)
     - [Environment variables](#environment-variables)
-- [Notes about switching DeviceMode](#notes-about-switching-devicemode)
-    - [Going from DeviceMode:LVM to DeviceMode:Direct](#going-from-devicemodelvm-to-devicemodedirect)
-    - [Going from DeviceMode:Direct to DeviceMode:LVM](#going-from-devicemodedirect-to-devicemodelvm)
+- [Notes about switching device mode](#notes-about-switching-device-mode)
+    - [Going from LVM device mode to direct device mode](#going-from-lvm-device-mode-to-direct-device-mode)
+    - [Going from direct device mode to LVM device mode](#going-from-direct-device-mode-to-lvm-device-mode)
 - [Notes about accessing system directories in a container](#notes-about-accessing-system-directories-in-a-container)
     - [Read-only access to /sys](#read-only-access-to-sys)
     - [Access to /dev of host](#access-to-dev-of-host)
--   [Repository elements which are generated or created separately](#repository-elements-which-are-generated-or-created-separately)
-    -   [Top-level README diagrams describing LVM and Direct device modes](#top-level-readme-diagrams-describing-lvm-and-direct-device-modes)
-    -   [Top-level README diagram describing communication channels](#top-level-readme-diagram-describing-communication-channels)
-    -   [Diagrams describing provisioning sequence](#diagrams-describing-provisioning-sequence)
-    -   [RegistryServer spec](#registryserver-spec)
-    -   [Table of Contents in README and DEVELOPMENT](#table-of-contents-in-readme-and-development)
+- [Repository elements which are generated or created separately](#repository-elements-which-are-generated-or-created-separately)
+    - [Top-level README diagrams describing LVM and Direct device modes](#top-level-readme-diagrams-describing-lvm-and-direct-device-modes)
+    - [Top-level README diagram describing communication channels](#top-level-readme-diagram-describing-communication-channels)
+    - [Diagrams describing provisioning sequence](#diagrams-describing-provisioning-sequence)
+    - [RegistryServer spec](#registryserver-spec)
+    - [Table of Contents in README and DEVELOPMENT](#table-of-contents-in-readme-and-development)
 
 Code quality
 ============
@@ -160,9 +160,9 @@ Command line arguments
 ----------------------
 
 Note that different set of programs is used in different
-DeviceModes. Three stages: *pmem-ns-init*, *pmem-vgm*,
-*pmem-csi-driver* run in DeviceMode:LVM. Only *pmem-csi-driver* runs
-in DeviceMode:Direct.
+device modes. Three stages: *pmem-ns-init*, *pmem-vgm*,
+*pmem-csi-driver* run in LVM device mode. Only *pmem-csi-driver* runs
+in direct device mode.
 
 Common arguments
 ----------------
@@ -205,7 +205,7 @@ argument name        | meaning                                           | type 
 -clientCertFile string | Client SSL certificate file to use for authenticating peer connections | string | | certFile
 -clientKeyFile string | Client private key associated to client certificate | string |              | keyFile
 -controllerEndpoint string | internal node controller endpoint              | string |              |
--deviceManager string      | device manager to use to manage pmem devices   | string | lvm or ndctl | lvm
+-deviceManager string      | device mode to use. ndctl selects mode which is described as direct mode in documentation. | string | lvm or ndctl | lvm
 -drivername string         | name of the driver                             | string |              | pmem-csi
 -endpoint string           | PMEM CSI endpoint                              | string |              | unix:///tmp/pmem-csi.sock
 -keyFile string            | Private key file associated to certificate     | string |              |
@@ -220,17 +220,17 @@ Environment variables
 TEST_WORK is used by registry server unit-test code to specify path to certificates in test system. 
 Note, THIS IS NOT USED IN PRODUCTION
 
-Notes about switching DeviceMode
+Notes about switching device mode
 ================================
 
-If DeviceMode is switched between LVM and Direct(ndctl), please keep
+If device mode is switched between LVM and direct(aka ndctl), please keep
 in mind that PMEM-CSI driver does not clean up or reclaim Namespaces,
-therefore Namespaces plus other related context (possibly LVM state)
+therefore Namespaces plus other related context (LVM state)
 created in previous mode will remain stored on device and most likely
-will create trouble in another DeviceMode.
+will create trouble in another device mode.
 
-Going from DeviceMode:LVM to DeviceMode:Direct
-----------------------------------------------
+Going from LVM device mode to direct device mode
+------------------------------------------------
 
 - examine LV Groups state on a node: `vgs`
 - examine LV Phys.Volumes state on a node: `pvs`
@@ -241,7 +241,7 @@ NOTE: The next **WILL DELETE ALL NAMESPACES** so be careful!
 
 - Delete Namespaces on a node using CLI: `ndctl destroy-namespace all --force`
 
-Going from DeviceMode:Direct to DeviceMode:LVM
+Going from direct device mode to LVM device mode
 ----------------------------------------------
 
 No special steps are needed to clean up Namespaces state.
@@ -249,7 +249,7 @@ No special steps are needed to clean up Namespaces state.
 If PMEM-CSI driver has been operating correctly, there should not be
 existing Namespaces as CSI Volume lifecycle should have been deleted
 those after end of life of Volume. If there are, you can either keep
-those (DeviceMode:LVM does honor "foreign" Namespaces and leaves those
+those (LVM device mode does honor "foreign" Namespaces and leaves those
 alone) if you have enough space, or you can choose to delete those
 using `ndctl` on node.
 
@@ -264,9 +264,8 @@ Read-only access to /sys
 ------------------------
 
 In some deployment schemes /sys remains mounted read-only in the
-container running pmsm-csi-driver. This is not problem in
-DeviceMode:LVM, but is a blocking problem in DeviceMode:Direct where
-the driver needs write access to /sys for Namespaces management
+container running pmsm-csi-driver. This creates problem for the
+driver which needs write access to /sys for Namespaces management
 operations. There is start-time check for read-write mount of /sys in
 the code. An error in pod log `pmem-driver: Failed to run driver:
 FATAL: /sys mounted read-only, can not operate` is the sign of such
@@ -276,8 +275,7 @@ Access to /dev of host
 ----------------------
 
 Containers runtime may not pass /dev from host into the
-container. This is, again, problem in DeviceMode:Direct. If the /dev/
-of the host is not accessible in the PMEM-CSI container, there will be
+container. If the /dev/ of the host is not accessible in the PMEM-CSI container, there will be
 failure in accessing of newly created block device /dev/pmemX.Y which
 will not be visible inside container. The driver does not detect the
 root cause of that problem during start-up, but only when a volume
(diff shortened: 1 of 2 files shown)


okartau:
I added a table that covers differences in shorter sentences. The longer ones I still kept  in chapters below, these now kind of repeat the message, so should we delete some text from chapters below which repeats the message of the table?

The other minor changes below in file are made by editor I used (ghostwriter which is nice for MD), they fix minor invisible artifacts we have. I can make separate commit from those if these disturb the clean change flow here.

README.md
```diff
@@ -530,7 +542,7 @@ on the Kubernetes version.
 - **Label the cluster nodes that provide persistent memory device(s)**
 
 ```sh
-    $ kubectl label node <your node> storage=pmem
```
pohly: What kind of whitespace change is this?

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
 
 ### Architecture and Operation
 
-The PMEM-CSI driver can operate in two different DeviceModes: LVM and Direct
+The PMEM-CSI driver can operate in two different DeviceModes: LVM and Direct. This tables describes shows overview and comparison of those modes. There is more detailed modes explanation down in following paragraphs.
```
pohly: This is the first time that we use the special "DeviceMode" spelling. We either need to explain why it is spelled this way or just spell it as "device modes".

As it stands, it looks like a special term that readers are supposed to be familiar with.

pohly: "this tables describes shows overview and comparison" -> "this table contains an overview and comparison"

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+
+|                   |LVM                    |Direct              |
+|:--                |:--                    |:--                 |
+|Purpose            |avoids fragmentation   |simpler, somewhat faster, but storage may get fragmented   |
```
pohly: "Purpose" -> "Main advantage"?

"fragmentation" is not obvious. Add a footnote that explains it?

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+|What is served     |LVM Logical Volume     |pmem block device   |
+|Region-affinity    |yes:One Volume Group per Region  |yes: namespace can belong to one Region only  |
```
pohly: "Region-affinity" is another technical term which needs a footnote.

Why the special spelling of "One Volume Group per Region"? I would stick to plain English as much as possible. "yes: one volume group per region" looks fine to me.

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+|Startup            |Two extra stages: pmem-ns-init(creates namespaces), vgm(creates Volume Groups)   |No extra steps at startup |
```
pohly: Space after pmem-ns-init and vgm?

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+|Namespace modes    |fsdax, sector -mode namespaces precreated as pools   |Namespace of required mode created directly, no need to pre-create pools   |
```
pohly: Why the hyphen before "mode"?

pohly: "precreated" -> "pre-created"

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+|Limiting space use | Can leave part of device unused during pools creation  | No limits, create namespaces on device until runs out of space  |
```
pohly: "use" -> "usage"

README.md
```diff
@@ -60,7 +60,19 @@ listening for API requests and provisioning volumes accordingly.
...
+| *Name* field in Namespace | *Name* gets set to 'pmem-csi' to achieve own vs. foreign marking | *Name* gets set to VolumeID, without attempting own vs. foreign marking  |
+|Minimum volume size| 4MB                   | 2GB (See also alignment adjusting below) |
```
pohly: "adjusting" -> "adjustment"

pohly: Why 2GB? From the description below it sounds like 1GB is the minimum.

README.md
okartau: 'whitespace' may not be the best term here. There is something that does not show properly and is not clean char. In emacs/X I see underscore. In emacs/text I see violet-color block. In vi I see nothing i.e. whitespace.
When I look what bytes really are there, there are actually two bytes 0xc2 and 0xa0.
These somehow appeared already some time ago, was it with version from Mary's editor? (just a guess). Since then, we have gradually replaced some of those. Now my editor saved the contents by overwriting last 4 instances with space.

okartau: C2A0 has something to do with unicode whitespace.
https://github.github.com/gfm/#unicode-whitespace
But nevertheless, we should not have it I think

README.md
okartau: during test and trials, I developed knowledge that I should not attempt allocation less than 2 GB. Thats because I learned at some point that libndctl allocates me 1-step less than I asked, and libndctl aligns down. But that was during the times when pmem-csi itself also aligned down, so worth rechecking now.

README.md
pohly: Yes, I fully agree, I just wondered what it was.

README.md
pohly: Okay, so it is a minimum size that has nothing to do with alignment. The comment in brackets made me belief that.

We should indeed investigate whether it's still needed (would help our CI!) and/or rephrase the text: "at least 2GB, potentially more depending on alignment (see next row)".

okartau: good that this triggered re-checking. It appears recent alignment related change actually broke direct-mode size calc, we now may allocate **less** than request (for example we now create 2 GB if we request 2.X GB).
Thats because in recent change I mixed alignment and size compensation together and over-simplified, taking away important "size += 2" line which was doing "combined align and compensate".
Compensate is needed because libndctl serves us 1GB less than we ask.
The more correct way, separating align and compensate layers, is to align up by 1GB and then add 1GB to compensate. (I will create separate PR).
But this compensation thing is entirely driver & libnsdctl-internal, so in user-facing terms our minimal is actually 1GB (what we should show in this table here), and that will be the served minimal actual size if we ask typical small test volume, like 1MB.

pohly: Can factor out the math into a helper function and add unit tests? It looks like this code is tricky enough that it can't be done correctly without better test coverage.

"to compensate" - I don't understand what that means. Compensate for what?

> in user-facing terms our minimal is actually 1GB (what we should show in this table here)

The table should show the smallest volume size that can be created in this mode.

okartau: > Compensate for what?

libndctl creates namespace which is one alignment step smaller than we ask. Thats I learned on path, but I should write it in big letters somewhere. It's easy to demonstrate on cmdline:
```
# ndctl create-namespace -s 2G -a 1G
{
  "dev":"namespace0.0",
  "mode":"fsdax",
  "map":"dev",
  "size":"1024.00 MiB (1073.74 MB)",
  "uuid":"35f3dad0-4650-48f5-93c4-0ebb2560e6b9",
  "sector_size":512,
  "align":1073741824,
  "blockdev":"pmem0"
}
```

and accordingly, attempt to create 1G will fail:
```
# ndctl create-namespace -s 1G -a 1G
libndctl: ndctl_pfn_enable: pfn0.0: failed to enable
  Error: namespace0.1: failed to enable
failed to create namespace: No such device or address
```

okartau: > Can you factor out the math into a helper function and add unit tests?

good idea. yes, the pure math part can be extracted as separate function. However, this will not cover all interesting parts, like alignment forcing in CreateNamespace and size & alignment extra handling by libndctl, so this test will miss some quite important parts of the path. But sure, it is good idea to add such test and cover at least some part. Also, such test is convenient place to document and explain the rules behind the math through various sizes as test source data.

README.md
okartau: should we explain this by text we make up, or point to generic definition like https://en.wikipedia.org/wiki/Fragmentation_(computing)#Data_fragmentation

README.md
okartau: hyphen is because I was thinking "fsdax-mode and sector-mode namespaces", then used comma to combine those, and then single hyphen remained. I will remove hyphen, lets see how it looks

okartau:
pushed new state. All DeviceMode -> "device mode" changes are in separate commit. Note that I changed the words, and removed extra colon in most places, but I did not change words order. In some case, it may be better to say "LVM device mode" instead of current "device mode LVM" (the latter form was used always because there was this colon-notation)

README.md
```diff
@@ -60,14 +60,26 @@ listening for API requests and provisioning volumes accordingly.
 
 ### Architecture and Operation
 
-The PMEM-CSI driver can operate in two different DeviceModes: LVM and Direct
+The PMEM-CSI driver can operate in two different device modes LVM and Direct. This table contains an overview and comparison of those modes. There is more detailed modes explanation down in following paragraphs.
```
pohly: Here "two different device modes LVM and Direct" needs to be "two different device modes: LVM and Direct".

Do we really want to capitalize "Direct"? LVM is upper case because it is an abbreviation, but "direct" is a normal English word.

pohly: In plain text I would use "direct device mode" and "LVM device mode".

README.md
pohly: I would provide a short example.

pohly: Users are probably aware what "fragmentation" means, just not how it applies in this context.

README.md
```diff
@@ -60,14 +60,26 @@ listening for API requests and provisioning volumes accordingly.
...
 
-### DeviceMode:LVM
+|                   |LVM                    |Direct              |
+|:--                |:--                    |:--                 |
+|Main advantage     |avoids fragmentation   |simpler, somewhat faster, but storage may get fragmented   |
+|What is served     |LVM Logical Volume     |pmem block device   |
+|Region-affinity    |yes:one volume group per Region  |yes: namespace can belong to one Region only  |
```
pohly: Space missing after "yes:".

README.md
```diff
@@ -60,14 +60,26 @@ listening for API requests and provisioning volumes accordingly.
...
+|Startup            |Two extra stages: pmem-ns-init (creates namespaces), vgm (creates volume groups)   |No extra steps at startup |
+|Namespace modes    |fsdax, sector mode namespaces pre-created as pools   |Namespace of required mode created directly, no need to pre-create pools   |
```
pohly: Can we link to a definition of "fsdax" and "sector" modes?

okartau:
pushed new version. 3rd commit added that mentions recommended minimum kernel 4.15, also I added another newer docker version (looked it up from CI job log).
Tried to address all comments so far.
"Device mode" wording checked and changed in most places (these changes a separate commit)

okartau:
pushed another as TOC was bad in previous push. Note (as I explained in commit msg) that TOC got regenerated using emacs plugin now, should be more in harmony with other users than previously pushed pandoc-generated whitespaces-variant.

okartau:
should we also add 1.15 in table of supported k8s versions?

pohly:
Yes, let's add 1.15.

okartau:
Added 1.15 line in "verified k8s" table

pohly:
Is this still "WIP"?

okartau:
I expected more good comments after I added new text parts. But I remove WIP now

okartau:
we need more changes. DeviceMode: -> "device mode" change should cover DEVELOPMENT.md as well.
Also I noticed, we do not give example how to use sector more. I verified sector mode  (which is in bit of orphaned role, no much use and testing of that) on recent code, good news is it works. But we could add a note in README showing example of change in deploy file.

okartau:
when do we remove Alpha code warning in README, can do it as part of this change?

okartau:
what is opinion about commented-out parts in README that originally were in template?
Should we fill/create some new sections based on placeholders, or is it better to delete, or perhaps some of both.

okartau:
pushed new versions with changes:
- DeviceMode changes in DEVELOPMENT, plus few other minor changes: clarified that option name "ndctl" selects direct mode (this is one remaining naming inconsistencies we have), simplified /sys and /dev chapters as we started to pass explicit mounts in all modes not just direct.
- added chapter with note how to deploy using sector mode. I placed it at the end of deployment steps as I dont easily see better location. Is this long enough text that we need Mary's help to improve it?
- removed Alpha warning (as part of versions change commit). Here I got idea/question: is it correct to remove warning, or should we replace this with Beta-level warning something like "this is verified in testing and demos but has got little real-life use in production"

pohly:
> when do we remove Alpha code warning in README, can do it as part of
> this change?

We decided to keep it for now. Perhaps a 1.0.0 will remove it.

> what is opinion about commented-out parts in README that originally
> were in template?
> Should we fill/create some new sections based on placeholders, or is
> it better to delete, or perhaps some of both.

Needs to be decided on a case-by-case basis. If we have the
corresponding information, we can remove the placeholder. If we can add
the information, let's do so now. If we can't (for example, instructions
for an admin is something we postponed), then keep the placeholder.

pohly:
> - removed Alpha warning (as part of versions change commit). Here I
> got idea/question: is it correct to remove warning, or should we
> replace this with Beta-level warning something like "this is verified
> in testing and demos but has got little real-life use in production"

Please restore the alpha warning. This needs further discussion.

okartau:
yes, I restored it after I saw your response. Also, rebased over another recent commit that modified README.

DEVELOPMENT.md
```diff
@@ -160,9 +160,9 @@ Command line arguments
 ----------------------
 
 Note that different set of programs is used in different
-DeviceModes. Three stages: *pmem-ns-init*, *pmem-vgm*,
-*pmem-csi-driver* run in DeviceMode:LVM. Only *pmem-csi-driver* runs
-in DeviceMode:Direct.
+device modes. Three stages: *pmem-ns-init*, *pmem-vgm*,
+*pmem-csi-driver* run in lvm device mode. Only *pmem-csi-driver* runs
```
pohly: "lvm" is an acronym. It must be in capital letters here. Might also have to fixed elsewhere, but only in plain text. If it is used as a keyword or in a path name, it can be lowercase (similar to PMEM-CSI).

For example,

    `direct` or `lvm`: one uses direct device mode, the other LVM device mode

looks good to me.

DEVELOPMENT.md
```diff
@@ -220,17 +220,17 @@ Environment variables
 TEST_WORK is used by registry server unit-test code to specify path to certificates in test system. 
 Note, THIS IS NOT USED IN PRODUCTION
 
-Notes about switching DeviceMode
+Notes about switching device mode
 ================================
 
-If DeviceMode is switched between LVM and Direct(ndctl), please keep
+If device mode is switched between LVM and direct(ndctl), please keep
```
pohly: "direct(ndctl)" -> "direct (aka ndctl)"

DEVELOPMENT.md
```diff
@@ -220,17 +220,17 @@ Environment variables
...
 in mind that PMEM-CSI driver does not clean up or reclaim Namespaces,
-therefore Namespaces plus other related context (possibly LVM state)
+therefore Namespaces plus other related context (LVM state)
```
pohly: As we are now starting to work on capitalization, let me also bring up "Namespace" - why with a capital letter? https://github.com/pmem/ndctl/blob/master/Documentation/ndctl/ndctl-create-namespace.txt uses it as a normal English word, i.e. lowercase.

We can also defer this - your choice.

README.md
```diff
@@ -802,7 +802,7 @@ Multiple different clusters can be brought up in parallel by changing
 the default `clear-govm` cluster name via the `CLUSTER` env variable.
 
 For example, this invocation sets up a cluster using the non-default
-direct DeviceMode:
+device mode direct:
```
pohly: "device mode direct" -> "direct device mode"

DEVELOPMENT.md
okartau: good point, let's change Namespace and Region. I recall I was thinking to stress that these are objects that are created. But I agree, this probably adds more confusion than it helps. I will change.

okartau:
pushed new version, addressed recent comments. Also, added separate commit un-capitalizing Namespace and Region, except where starting the sentence or table heading.
Once we started de-capitalizing, do you think Volume Groups also should get changed?

pohly:
> Once we started de-capitalizing, do you think Volume Groups also
> should get changed?

Yes. The man page (https://linux.die.net/man/8/vgcreate) is a bit
inconsistent itself, but it mostly uses "volume group", not "Volume
Group".

okartau:
pushed new state with more de-capitalized words (new changes all in decapit-commit).
Few cases are gray area, like those Cache/Ephemeral/Persistent volumes in own headings. is Volume there such a term that deserves capital. I changed those (but can revert back).

DEVELOPMENT.md
```diff
@@ -205,7 +205,7 @@ argument name        | meaning                                           | type
 -clientCertFile string | Client SSL certificate file to use for authenticating peer connections | string | | certFile
 -clientKeyFile string | Client private key associated to client certificate | string |              | keyFile
 -controllerEndpoint string | internal node controller endpoint              | string |              |
--deviceManager string      | device manager to use to manage pmem devices   | string | lvm or ndctl | lvm
+-deviceManager string      | device mode to use. **ndctl** selects mode which is described as direct mode in documentation. | string | lvm or ndctl | lvm
```
marcemq: **ndctl** shouldn't be formatted as code?

README.md
```diff
@@ -60,17 +60,42 @@ listening for API requests and provisioning volumes accordingly.
 
 ### Architecture and Operation
 
-The PMEM-CSI driver can operate in two different DeviceModes: LVM and Direct
+The PMEM-CSI driver can operate in two different device modes: *LVM* and *direct*. This table contains an overview and comparison of those modes. There is more detailed modes explanation down in following paragraphs.
 
-### DeviceMode:LVM
+|                   |LVM                    |direct              |
```
marcemq: Nitpick comment: some paragraphs in table start with uppercase and others don't, can we homogenize this?

README.md
```diff
@@ -694,6 +721,29 @@ one with ext4-format and another with xfs-format file system.
     /dev/ndbus0region0fsdax/5cc9b19e-551d-11e9-a584-928299ac4b17 on /data type xfs (rw,relatime,attr2,dax,inode64,noquota)
 ```
 
+#### Note about using the sector mode
+
+PMEM-CSI can create a namespace in the *sector* (alias safe) mode instead of default *fsdax* mode. See [Persistent Memory Programming](https://pmem.io/ndctl/ndctl-create-namespace.html) for more details about namespace modes. The main difference in PMEM-CSI context is that a sector-mode volume will not get 'dax' mount option. The deployment examples do not describe sector mode to keep the amount of combinations smaller. Here are the changes to be made to deploy volumes in sector mode:
+
+- add **nsmode: sector** line in parameters section in the storageclass definition pmem-storageclass-XYZ.yaml:
```
marcemq: **nsmode: sector** shouldn't be formatted as code?

README.md
```diff
@@ -694,6 +721,29 @@ one with ext4-format and another with xfs-format file system.
...
+```
+parameters:
+  csi.storage.k8s.io/fstype: ext4
+  eraseafter: "true"
+  nsmode: "sector"                   <-- add this
+```
+
+* Only if using LVM device mode: Modify pmem-ns-init options to create sector-mode pools in addition to fsdax-mode pools. Add **-useforfsdax** and **-useforsector** options to pmem-ns-init arguments in pmem-csi-lvm.yaml: (select the percentage values that fit your needs)
```
marcemq: **-useforfsdax** and **-useforsector** shouldn't be formatted as code too?

okartau:
I addressed recent comments in new pushed version. 
Starting of table paragraphs with upper vs lower case: I am not really convinced what is better, most of these are not full sentences, so I changed all to lowercase (except left side column which is topic). The last row however contains longer text with multiple sentences, so this could deserve capital letter as start? Should we the change all to uppercase? (and perhaps make longer to form complete sentences) Please see how it looks to your eye and taste.

Another inconsistency I spotted, some lowercase pmem is still not justified. should cover that too(?) this path tends to get longer and longer...

pohly:
> The last row however contains longer text with multiple sentences, so this could deserve capital letter as start?

Probably yes, but ...

> Should we the change all to uppercase?

... I would say it is the only one. The rest look fine in lowercase.

> Another inconsistency I spotted, some lowercase pmem is still not justified.

My proposal is to merge this PR as-is and do further cleanups as we find them.

PR #367, opened 2019-07-22
reduce duplication in Jenkinsfile

pohly:
With the help of some functions we can remove the code duplication.

Closures for lazy variable evaluation didn't work, for some reason the result always was the same.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,12 +1,9 @@
 pipeline {
-
     agent {
-
         label "pmem-csi"
     }
 
     environment {
-
         /*
           For each major Kubernetes release we need one version of Clear Linux
           which had that release. Installing different Kubernetes releases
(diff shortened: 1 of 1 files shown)


okartau:
1. CI shows error "not enough space". The signs are similar to state where same storage that was already used for LVM-mode testing, is attempted to use with direct testing. Then the pre-allocated namespace(s) exist and block further creations. 
2. There are also errors near direct mode log start about kube* objects already present, a sign the driver deployment was already running when test started? So it's not clean-room test as likely was planned?

But overall this is great change to Jenkinsfile!

pohly:
I am reusing the same CLUSTER name now, which should be fine if `make stop` properly cleans up. But perhaps it doesn't?

pohly:
> I am reusing the same CLUSTER name now, which should be fine if make stop properly cleans up. But perhaps it doesn't?

Found it. `rm -rf <path>` in `stop.sh` wasn't doing anything because that  path wasn't exposed inside the container. Because of the `-f` that wasn't treated as an error. I think it should have been, because how can there be a valid situation where that path doesn't exist while we run the `stop.sh` script that is inside it?

But that's a different fix.

pohly:
@okartau it's ready now. Can you review and merge?

PR #369, opened 2019-07-23
Direct mode Namespace creation: compensate for smaller result

okartau:
Recent alignment change broke size calculation,
and compensation for smaller creation was left out.
As the result, current code would create for example 2GB
namespace per 2.x GB request, because we would
align 2.x up to 3 GB, then libndctl creatng 3-1=2GB namespace.
We have to compensate for libndctl creating one alignment step smaller
namespace. Current change should handle all: check for
too small (i.e. zero), compensate for smaller creation,
and that it all works together with alignment.

### DIFF
--- a/pkg/pmem-device-manager/pmd-ndctl.go
+++ b/pkg/pmem-device-manager/pmd-ndctl.go
@@ -97,16 +97,25 @@ func (pmem *pmemNdctl) CreateDevice(name string, size uint64, nsmode string) err
 		glog.V(4).Infof("Device with name: %s already exists, refuse to create another", name)
 		return fmt.Errorf("CreateDevice: Failed: namespace with that name exists")
 	}
-	// Increase to minimum size supported by libndctl.
-	const minsize uint64 = 2 * 1024 * 1024 * 1024
-	if size < minsize {
-		size = minsize
+	if size <= 0 {
+		// Allocating volumes of zero size isn't supported.
+		// We use some arbitrary small minimum size instead.
+		// It will get rounded up by libndctl to meet the alignment.
+		size = 1
 	}
-	// use align by 1 GB in creation request
+	// Pass align = 1 GB into creation request as that has proven to be reliable.
 	const align uint64 = 1024 * 1024 * 1024
+	// libndctl needs to store meta data and will use some of the allocated
+	// space for that (https://github.com/pmem/ndctl/issues/79).
+	// We don't know exactly how much space that is, just
+	// that it should be a small amount. But because libndctl
+	// rounds up to the alignment, in practice that means we need
+	// to request `align` additional bytes.
+	compensatedsize := size + align
+	glog.V(4).Infof("CreateDevice:%s: Compensate for libndctl creating one alignment step smaller: change size %d to %d", name, size, compensatedsize)
 	ns, err := pmem.ctx.CreateNamespace(ndctl.CreateNamespaceOpts{
 		Name:  name,
-		Size:  size,
+		Size:  compensatedsize,
 		Align: align,
 		Mode:  ndctl.NamespaceMode(nsmode),
 	})


pohly:
This "libndctl creating one alignment step smaller namespace" sounds like a bug to me. Has this behavior been discussed with the maintainers of it?

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -97,16 +97,21 @@ func (pmem *pmemNdctl) CreateDevice(name string, size uint64, nsmode string) err
...
-	// Increase to minimum size supported by libndctl.
-	const minsize uint64 = 2 * 1024 * 1024 * 1024
-	if size < minsize {
-		size = minsize
+	if size <= 0 {
+		// If called with zero size, we force small nonzero, otherwise size becomes equal to
+		// align after compensation, remains same after up-aligning and causes CreateNamespace to fail.
+		// In other words, combined compensate-and-align logic works for nonzero values.
+		glog.V(4).Infof("CreateDevice:%s: size %d is too low, adjusting to 4096", name, size)
+		size = 4096
```
pohly: Why 4096 and not, say, 1?

pohly: Why is that only relevant for size == 0 and not for size == align?

Can you document the behavior of libndctl or, if we don't know exactly, the observations that have been made about it? Perhaps then it becomes clearer why all of these size adjustments are made.

okartau:
> This "libndctl creating one alignment step smaller namespace" sounds like a bug to me. Has this behavior been discussed with the maintainers of it?

yes, here:
https://github.com/pmem/ndctl/issues/79
Got explained having to do with difficulty to know the overhead beforehand, and owners were "not sure what to to do about this". So, after some time I thought, we can work around compensating, and let the issue get closed. Seems ndctl does not follow same policy as CSI forces on pmem-csi: "serve at least requested or fail", instead they try to have "something close to requested is ok". But let's note that no other user has complained about it, and issue 79 also did not get much feedback, so majority seems to be happy with current style.

pkg/pmem-device-manager/pmd-ndctl.go
okartau: > Why is that only relevant for size == 0 and not for size == align?

There will lead to different result:
size == 0
1. we compensate up by one align step --> size will be 1GB
2. align (inside CreateNamespace) leaves it same
3. libndctl creation of 1GB will fail

size == align
1. we compensate up by one align step --> size will be 2GB
2. align (inside CreateNamespace) leaves it same
3. libndctl creation of 2GB will create 1 GB namespace 

Thats why we have to force-set zero size to any small value to get:
size = 4096
1. we compensate up by one align step --> size will be 1GB + 4096bytes
2. align (inside CreateNamespace) aligns it up to 2 GB
3. libndctl creation of 2GB will create 1 GB namespace

It's true that exactly 4096 is not needed, 1 would do it same way (should perhaps add comment).
I thought that 1 is too weird value for volume size.

pohly:
So if I understand https://github.com/pmem/ndctl/issues/79 correctly, asking for 1GB size with 1GB alignment fails because libndctl allocates 1GB, then uses some of that space for meta data, and then not enough space is left for the actual data volume?

Then I would calculate the requested size as follows (same code as yours, just different comments):
```
// Pass align = 1 GB into creation request as that has proven to be reliable.
const align uint64 = 1024 * 1024 * 1024
if size <= 0 {
    // Allocating volumes of zero size isn't supported.
    // We use some arbitrary small minimum size instead.
    // It will get rounded up by libndctl to meet the alignment.
    size = 4096
}
// libndctl needs to store meta data and will use some of the allocated
// space for that (https://github.com/pmem/ndctl/issues/79).
// We don't know exactly how much space that is, just
// that it should be a small amount. But because libndctl
// rounds up to the alignment, in practice that means we need
// to request `align` additional bytes.
compensatedsize := size + align
```

pkg/pmem-device-manager/pmd-ndctl.go
pohly: If the goal is to be larger than zero, then 1 is closer to the goal than 4096. But that' subjective and both need to be explained.

okartau:
pushed new, changed 4096 to 1, replaces comment text with proposed, these are indeed better.

okartau:
extra space of 1 seems to work both in CI and in my local test. Other recent changes were comments text.
OK to merge?

PR #373, opened 2019-07-31
set-version.sh: also update version in test-config.sh

pohly:
The version there also needs to be updated, otherwise E2E testing
fails.


pohly:
This was found when I tried to create v0.5.0-rc3, see https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/tags/job/v0.5.0-rc3/1/execution/node/60/log/

v0.5.0-rc4 has this fix and is currently passed that stage: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/tags/job/v0.5.0-rc4/1/execution/node/60/log/

PR #374, opened 2019-07-31
QEMU: enhance documentation and check for /dev/kvm

pohly:
Not having KVM enabled in a GCE machine is a real issue that a user
ran into. Error handling in this case is poor, GoVM doesn't print any
error and all that the user gets is an error about timing out waiting
for ssh.

The updated README.md explains the real requirements better and warns
about this pitfall. The script itself checks very early for /dev/kvm
and bails out with an error message when it is missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -830,6 +830,17 @@ Note that cloud providers often don't offer KVM support on their
 regular machines. Search for "nested virtualization" for your provider
 to determine whether and how it supports KVM.
 
+Nested virtualization is also needed when using Kata Containers inside
+the cluster. On Intel-based machines it can be enabled by loading the
+`kvm_intel` module with `nested=1` (see
+https://wiki.archlinux.org/index.php/KVM#Nested_virtualization). At
+this time, Kata Containers up to and including 1.9.1 is [not
+compatible with
+PMEM-CSI](https://github.com/intel/pmem-csi/issues/303) because
+volumes are not passed in as PMEM, but Kata Containers [can be
+installed](https://github.com/kata-containers/packaging/tree/master/kata-deploy#kubernetes-quick-start)
+and used for applications that are not using PMEM.
+
 The `clear-cloud` image is downloaded automatically. By default,
 four different virtual machines are prepared. Each image is pre-configured
 with its own hostname and with network.
--- a/README.md
+++ b/README.md
@@ -821,11 +821,14 @@ available.
 E2E testing is known to work on a Linux development host system. The user
 must be allowed to use Docker.
 
-KVM must be enabled and the user must be allowed to use it. Usually this
-is done by adding the user to the `kvm` group. The
-["Install QEMU-KVM"](https://clearlinux.org/documentation/clear-linux/get-started/virtual-machine-install/kvm)
-section in the Clear Linux documentation contains further information
-about enabling KVM.
+KVM must be enabled. Usually this is the case when `/dev/kvm` exists.
+The current user does not need the privileges to use KVM and QEMU
+doesn't have to be installed because GoVM will run QEMU inside a
+container with root privileges.
+
+Note that cloud providers often don't offer KVM support on their
+regular machines. Search for "nested virtualization" for your provider
+to determine whether and how it supports KVM.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -783,11 +783,14 @@ available.
 This is known to work on a Linux development host system. The user
 must be allowed to use Docker.
 
-KVM must be enabled and the user must be allowed to use it. Usually this
-is done by adding the user to the `kvm` group. The
-["Install QEMU-KVM"](https://clearlinux.org/documentation/clear-linux/get-started/virtual-machine-install/kvm)
-section in the Clear Linux documentation contains further information
-about enabling KVM.
+KVM must be enabled. Usually this is the case when `/dev/kvm` exists.
+The current user does not need the privileges to use KVM and QEMU
```
marcemq: Nitpick question, do we need to decide to use **does not** or **doesn't** and stick to it trough the `md` documents? or does not matter that much? 
It came to my attention because in this line **does not** is used and the next one is **doesn't**
(And I think this will apply for **do not** and **don't** too and all the shorter version of ENG words)

pohly:
Marcela Morales Q <notifications@github.com> writes:

> marcemq commented on this pull request.
>
>
>
>> @@ -783,11 +783,14 @@ available.
>  This is known to work on a Linux development host system. The user
>  must be allowed to use Docker.
>  
> -KVM must be enabled and the user must be allowed to use it. Usually this
> -is done by adding the user to the `kvm` group. The
> -["Install QEMU-KVM"](https://clearlinux.org/documentation/clear-linux/get-started/virtual-machine-install/kvm)
> -section in the Clear Linux documentation contains further information
> -about enabling KVM.
> +KVM must be enabled. Usually this is the case when `/dev/kvm` exists.
> +The current user does not need the privileges to use KVM and QEMU
>
> Nitpick question, do we need to decide to use **does not** or
> **doesn't** and stick to it trough the `md` documents? or does not
> matter that much?

I learned that "doesn't" and "don't" are colloquial and not to be used
in articles or scientific papers. But it seems that documentation
nowadays leans more towards colloquial than scientific.

I think it doesn't matter that much in the end. In this case, spelling
it out has the additional advantage that it emphasizes the "not", as in
"does *not* need the privileges".

pohly:
Don't merge, the `/dev/kvm` check actually fails in the CI machines. I lost access to those, so I don't know yet why that is.

pohly:
I'm still not sure why the `/dev/kvm` check failed. I just looked at a running CI worker and the device was there. Let's try again after rebasing.

I also added a commit which enables nested virtualization. It's a small additional change that touches the same documentation section; I didn't want to create another conflicting PR for that.

pohly:
@avalluri please also check the second commit.

pohly:
@avalluri: ping? We currently have clean test results.

pohly:
> somehow ReadME section (Automated testing) links got broken with this change. Can you please fix that.

It was broken on "devel" by an an earlier PR (https://github.com/intel/pmem-csi/pull/473) and has already been fixed there by https://github.com/intel/pmem-csi/pull/481. This change here is fine, it was just based on that broken revision of "devel".

I rebased, so now "view file" shows a clean version.

README.md
```diff
@@ -821,11 +821,25 @@ available.
...
+to determine whether and how it supports KVM.
+
+Nested virtualization is also needed when using Kata Containers inside
+the cluster. On Intel-based machines it can be enabled by loading the
+`kvm_intel` module with `nested=1` (see
+https://wiki.archlinux.org/index.php/KVM#Nested_virtualization). At
+this time, Kata Containers up to and including 1.9.1 is [not
+compatible with
+PMEM-CSI](https://github.com/intel/pmem-csi/issues/303) because
+volumes are not passed in as PMEM, but it [can be
```
avalluri: > ` but it can be installed...`

I am not clear what exactly 'it' refers to in this sentence? Can you please check if any rephrase required or just my in-ablity to understand :-}

pohly: it = kata containers. I'll replace the "it".

PR #383, opened 2019-08-12
Ephemeral/generic

avalluri:
Added ephemeral volume provisiong


pohly:
Let's consider next steps...

We need to make changes for 1.16 (https://github.com/kubernetes-csi/docs/pull/209/files):
- extend driver info to include `volumeLifecycleModes: persistent, ephemeral` in a new deployment for 1.16
- enable pod info
- in the code which decides between "ephemeral" and "persistent" check podinfo for `csi.storage.k8s.io/ephemeral` and if set, use that (and only that); if not set, use the heuristic

We need automated testing on 1.16. @okartau, you were working on an extension of our test cluster setup for Cent OS + Kubernetes 1.16, right? Is that ready for merging?

avalluri:
> 
> extend driver info to include volumeLifecycleModes: persistent, ephemeral in a new deployment for 1.16

We can deal this `volumeLifecycleModes` change as part of Kubernetes 1.16 deployment change. This shouldn't effect v1.15.

> enable pod info

our deployment already  using `podInfoOnMount: true`,

> in the code which decides between "ephemeral" and "persistent" check podinfo for csi.storage.k8s.io/ephemeral and if set, use that (and only that); if not set, use the heuristic

This check is already part of this change. In addition to make it work for v1.15, we provision ephemeral only volume if *all* below conditions are met:
  - Volume with given volume-id is not found
  - Missing `StagingTargetPath `in the `NodePublishRequest`
  - No provisioner information in `NodePlubishRequest.VolumeContext`

pohly:
>> extend driver info to include volumeLifecycleModes: persistent, ephemeral in a new deployment for 1.16
>
> We can deal this `volumeLifecycleModes` change as part of Kubernetes
> 1.16 deployment change. This shouldn't effect v1.15.

Instead of merging a feature where we know that additional changes will
be needed soon, I prefer to do it all at once. Then we only need to go
through the code once and know that the current code really works in all
cases instead of assuming that it does.

pohly:
All PRs are now getting tested against 1.16 (on Fedora) and against 1.15 (on Clear). Can you update the PR such that it works with both?

To start a cluster with 1.16 locally, use `TEST_DISTRO=fedora TEST_KUBERNETES_VERSION=1.16 make start`.

avalluri:
> Let's support 1.16 from the start, i.e. avoid accumulating technical debt.

I have pushed the 1.16 deployment changes.

deploy/kustomize/patches/driverinfo-ephemeral-kubernetes-1-16-patch.yaml
```diff
@@ -0,0 +1,6 @@
+# Append volumeLifeCycltModes to DervierInfo valid only for Kubernetes v1.16+
```
pohly: Typos.

pohly:
Do we have tests? Probably we need to vendor Kubernetes 1.16.0 first, then we can enable these tests: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/ephemeral.go

deploy/kustomize/patches/driverinfo-ephemeral-kubernetes-1-16-patch.yaml
avalluri: > Instead of copying quite a few files into `deploy/kustomize/kubernetes-1.16-*` dirs, would it be possible to take the existing Kubernetes 1.15 deployment and just patch the `volumeLifecycleField` in?

Do you mean something this: https://github.com/avalluri/pmem-CSI/commit/099d91ddd135232b12608ee31cf79640edf3bec1

avalluri: > Typos.

Fixed

avalluri:
> > > Instead of copying quite a few files into `deploy/kustomize/kubernetes-1.16-*` dirs, would it be possible to take the existing Kubernetes 1.15 deployment and just patch the `volumeLifecycleField` in?
> > 
> > 
> > Do you mean something this:
> > [avalluri@099d91d](https://github.com/avalluri/pmem-CSI/commit/099d91ddd135232b12608ee31cf79640edf3bec1)
> 
> Yes, exactly what I had in mind - thanks for turning it into code :-)
> 
> But comments are currently inconsistent, some of them look like copy-and-paste leftovers. I think it is enough to have one README.md in the `deploy/kubernetes-1.16` directory that explains the difference and then drop all other comments from the directories and `kustomization.yaml` files underneath it.
> 
> Another nit: also check the file endings in those file - sometimes you have two line breaks, sometimes one, sometimes none.
> 
> Ending a file with a line break looks best IMHO.

I just pushed a new revision, that covers all these review comments.

pohly:
> devicemanager: Make GetDevice() return pointer 

Why?

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -88,57 +101,92 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-	attrib := req.GetVolumeContext()
-	mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
+	var mountOptions []string
 
-	klog.V(3).Infof("NodePublishVolume: id:%s targetpath:%v stagingtargetpath:%v readonly:%v attributes:%v mountflags:%v",
-		req.GetVolumeId(), targetPath, stagingtargetPath, readOnly, attrib, mountFlags)
+	if ephemeral {
+		val, ok := req.GetVolumeContext()["size"]
+		if !ok {
+			return nil, status.Error(codes.InvalidArgument, "size not found in volume context")
```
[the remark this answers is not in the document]
pohly: Yes, it shows up:
```
  Warning  FailedMount  0s (x6 over 17s)  kubelet, pmem-csi-pmem-govm-worker3  MountVolume.SetUp failed for volume "my-csi-volume" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = InvalidArgument desc = size not found in volume context
```

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -88,57 +101,92 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+			VolumeCapabilities: []*csi.VolumeCapability{req.VolumeCapability},
+			Name:               req.VolumeId,
+			CapacityRange:      &csi.CapacityRange{RequiredBytes: size.Value()},
+			Parameters:         params,
+		}
 
-	options := []string{"bind"}
-	if readOnly {
-		options = append(options, "ro")
+		createResp, err := ns.cs.CreateVolume(ctx, createReq)
```
pohly: We need full validation of the volume parameters. Unknown (or miss-spelled) parameters must trigger an error. Otherwise user errors may silently lead to a different than intended outcome.

As a testcase, I tried with `foo: bar` added to `pmem-app-ephemeral.yaml` and that was silently ignored.

pohly: And we also need an automated test case for this.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -88,57 +101,92 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to create new ephemeral volume: %s", err.Error()))
+		}
+
+		device, err := ns.cs.dm.GetDevice(createResp.Volume.VolumeId)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Could not retrieve newly created volume with id '%s': %s", createResp.Volume.VolumeId, err.Error()))
+		}
+
+		// Create filesystem
+		err = ns.provisionDevice(device, req.GetVolumeCapability().GetMount().GetFsType())
```
pohly: We don't have tests which use a non-default filesystem, do we? Let's not make this a blocker for merging this PR, but keep it in mind: https://github.com/intel/pmem-csi/issues/431

pkg/pmem-csi-driver/nodeserver.go
pohly: Beware that we enable pod info on mount and thus have to ignore *all* `csi.storage.k8s.io/*` entries. Unknown fields may get added under that namespace in the future.

pkg/pmem-csi-driver/nodeserver.go
avalluri: > We don't have tests which use a non-default filesystem, do we?

I am not sure about ephemral volumes, But at least for persistent volumes, Volumes test-suite covers this. Here is the needed change in our test code: https://github.com/avalluri/pmem-CSI/commit/bd357fbd757e12f32cadcc5d915aa0b2e6a18264

pohly: I was thinking of ephemeral volumes here.
_

README.md
```diff
@@ -344,8 +344,8 @@ the node cannot start until resources become available.
 
 #### Usage on Kubernetes
 
-Kubernetes cluster administrators can expose above mentioned [volume
-persistency types](#volume-persistency) to applications using
+Kubernetes cluster administrators can expose above mentioned Persistent and Cache volumes
```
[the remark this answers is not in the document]
avalluri: Done

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -88,57 +101,92 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
 		}
+		ephemeral = b
+	} else {
+		// If met all below conditions we still treat as a ephemeral volume request:
+		// 1) No Volume found with given volume id
+		// 2) No provisioner info found in VolumeContext "storage.kubernetes.io/csiProvisionerIdentity"
+		// 3) No StagingPath in the request
+		device, _ := ns.cs.dm.GetDevice(req.VolumeId)
```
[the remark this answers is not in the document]
avalluri: > > devicemanager: Make GetDevice() return pointer
> 
> Why?

Its easier/cleaner to deal with pointers than values for a struct type, especially while check if a device is valid using `nil`

README.md
```diff
@@ -400,6 +400,9 @@ application](deploy/common/pmem-app-cache.yaml) example.
 * A node is only chosen the first time a pod starts. After that it will always restart
   on that node, because that is where the persistent volume was created.
 
+Volume requests embedded in Pod spec are provisioned as Ephemeral volumes. Check with
```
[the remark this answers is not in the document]
avalluri: Done

README.md
```diff
@@ -515,7 +518,7 @@ Kubernetes cluster has been verified on:
 |-------------------|--------------------------------|------------------------------- |
```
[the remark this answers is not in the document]
avalluri: Done, added the link.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -76,7 +89,7 @@ func (ns *nodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 	return nil, status.Error(codes.Unimplemented, "")
 }
 
-func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) { // nolint(gocyclo)
```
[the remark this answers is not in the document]
avalluri: Done, I just pushed the ephemeral part to a separate function.

pkg/pmem-csi-driver/nodeserver.go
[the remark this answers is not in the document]
avalluri: > This ignores the error from `GetDevice`. Wouldn't it be cleaner to introduce a "not found" error type and check for that?

Yes, I agree to introduce error codes. But in this case, the only error returned(by both implementations) is  "not found" so we are not missing/ignoring any other errors.

pkg/pmem-csi-driver/nodeserver.go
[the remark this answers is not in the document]
avalluri: Done, I rephrased the error test as you suggested.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -88,57 +101,92 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-		req.GetVolumeId(), targetPath, stagingtargetPath, readOnly, attrib, mountFlags)
+	if ephemeral {
+		val, ok := req.GetVolumeContext()["size"]
+		if !ok {
+			return nil, status.Error(codes.InvalidArgument, "size not found in volume context")
+		}
+
+		size, err := resource.ParseQuantity(val)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Failed to parse given size '%s': %s", val, err.Error()))
```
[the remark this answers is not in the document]
avalluri: Done

README.md
```diff
@@ -400,6 +400,9 @@ application](deploy/common/pmem-app-cache.yaml) example.
 * A node is only chosen the first time a pod starts. After that it will always restart
   on that node, because that is where the persistent volume was created.
 
+Volume requests embedded in Pod spec are provisioned as Ephemeral volumes. Check with
+provided [example application](deploy/kubernetes-1.15/pmem-app-ephemeral.yaml) for
+ephemeral volume usage.
```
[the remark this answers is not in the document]
avalluri: I added a table that lists all the possible `volumeAttributes` keys and their meaning.

pkg/pmem-csi-driver/nodeserver.go
avalluri: Done, Now we validate the volume context and raise an error if occurred any unexpected keys. Testcase is pending.

pkg/pmem-csi-driver/nodeserver.go
avalluri: Yes, current ephemeral tests are quite limited and not supported to choose the filesystem type for volume.

avalluri:
Enabled ephemeral testsuite and made changes to ListVolumes() implementation so that it does not return ephemeral volumes.

pkg/pmem-csi-driver/nodeserver.go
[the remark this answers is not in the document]
pohly: So here the caller is allowed to ignore the error code because the only error is "not found", which isn't documented. That looks rather fragile to me - what if we introduce a second error later? Are you sure that this call site will then be updated?

If you don't want to check the error here and continue to test for nil, then better remove the error return from `GetDevice` entirely. Then it is obvious from the function signature that the function always succeeds (returns the device) or returns nil when not found.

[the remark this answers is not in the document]
avalluri: GetDevice() now returns os.ErrExist, and the same check added here.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -89,29 +102,68 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		ephemeral = b
+	} else {
+		// If met all below conditions we still treat as a ephemeral volume request:
+		// 1) No Volume found with given volume id
+		// 2) No provisioner info found in VolumeContext "storage.kubernetes.io/csiProvisionerIdentity"
+		// 3) No StagingPath in the request
+		device, err := ns.cs.dm.GetDevice(req.VolumeId)
+		if err != nil && err != os.ErrNotExist {
```
pohly: Isn't `GetDevice` wrapping `os.ErrNoExist`? Then a direct comparison will always fails. You have to unwrap with `errors.Cause`, then compare.

avalluri: Yes, you are right, I should use `os.IsNotExist()`. This check was added when initially `GetDevice()` returning direct `os.ErrNotExist` without wrapping, later I added wrapping.

deploy/kustomize/patches/driverinfo-ephemeral-kubernetes-1-16-patch.yaml
```diff
@@ -0,0 +1,6 @@
+# Append volumeLifeCycleModes to DriverInfo valid only for Kubernetes v1.16+
```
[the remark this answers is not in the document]
avalluri: It's my bad, result of the wrong commit. This commit should be replaced with https://github.com/avalluri/pmem-CSI/commit/f37b3b51e0e9e0013658242da8502a0c86feb584. Now I fixed this.

deploy/kustomize/kubernetes-1.16-ephemeral/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+# Patch Kubernetes 1.16 deployment such that driver 
+# supports both persistent and ephemral volume modes
```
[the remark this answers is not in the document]
avalluri: fixed in https://github.com/avalluri/pmem-CSI/commit/f37b3b51e0e9e0013658242da8502a0c86feb584

README.md
```diff
@@ -325,7 +325,7 @@ can reuse existing local data when restarting.
 Volume | Kubernetes | PMEM-CSI | Limitations
 --- | --- | --- | ---
 Persistent | supported | supported | topology aware scheduling<sup>1</sup>
-Ephemeral | [in design](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#proposal) | in design | topology aware scheduling<sup>1</sup>, resource constraints<sup>2</sup>
+Ephemeral | supported | supported | resource constraints<sup>2</sup>
```
pohly: We probably should add to the Kubernetes column remarks about which versions support it, something like "alpha support in 1.15 (CSIInlineVolume), beta in 1.15".

pohly: When looking at the rendered file, I noticed that the definition list (https://github.com/intel/pmem-csi#volume-persistency) is no longer rendered with the text on the next paragraph. This intentionally had two spaces (https://meta.stackexchange.com/questions/72395/is-it-possible-to-have-definition-lists-in-markdown/298872#298872) but we lost those in b30c9a051 - can you add them back?

README.md
```diff
@@ -396,6 +396,16 @@ application](deploy/common/pmem-app-cache.yaml) example.
 * A node is only chosen the first time a pod starts. After that it will always restart
   on that node, because that is where the persistent volume was created.
 
+Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The volume request could use below fields as [`volumeAttributes`](https://kubernetes.io/docs/concepts/storage/volumes/#csi):
+
+|key|meaning|optional|default value|
+|---|-------|--------|-------------|
+|size|Size of the requested ephemeral volume|No||
+|nsmode|PMEM namespace mode of requested ephemeral volume|Yes|"fsdax"|
+|eraseAfter|Should erase volume content after use|Yes|"true"|
```
pohly: "Should" or "will"? I think the meaning becomes more obvious when using a different verb in the description: "eraseAfter - clear all data after use and before deleting the volume"

README.md
```diff
@@ -396,6 +396,16 @@ application](deploy/common/pmem-app-cache.yaml) example.
 * A node is only chosen the first time a pod starts. After that it will always restart
   on that node, because that is where the persistent volume was created.
 
+Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The volume request could use below fields as [`volumeAttributes`](https://kubernetes.io/docs/concepts/storage/volumes/#csi):
+
+|key|meaning|optional|default value|
+|---|-------|--------|-------------|
+|size|Size of the requested ephemeral volume|No||
+|nsmode|PMEM namespace mode of requested ephemeral volume|Yes|"fsdax"|
```
pohly: What other values besides "fsdax" are valid?

deploy/common/pmem-app-ephemeral.yaml
```diff
@@ -0,0 +1,22 @@
+# This demonstrates how to use inline pmem csi volumes
```
pohly: Full sentence, end with a dot at the end?

deploy/kustomize/kubernetes-1.16/README.md
```diff
@@ -0,0 +1,6 @@
+# Kubernetes v1.16 specific changes
+
+There are no specific deployment changes needed for Kubernetes v1.16 except
+that the CSIDriver supports new field of specifying volume life-cycle modes
+supported by the driver. So we take Kubernetes-1.15 final
+deployments(not base overlays) and patch with this new field.
```
pohly: nit: missing space before opening bracket

avalluri:
> The question around checking for `os.ErrNotExit` is also still open.

Now the code uses os.IsErrorNotExist() call, which does the digging and checks for real error. This should address the unwrapping issue you pointed.

README.md
avalluri: > When looking at the rendered file, I noticed that the definition list (https://github.com/intel/pmem-csi#volume-persistency) is no longer rendered with the text on the next paragraph. This intentionally had two spaces (https://meta.stackexchange.com/questions/72395/is-it-possible-to-have-definition-lists-in-markdown/298872#298872) but we lost those in [b30c9a0](https://github.com/intel/pmem-csi/commit/b30c9a0512420fd51e97013ad10ef64099373b44) - can you add them back?

Done.

avalluri: > We probably should add to the Kubernetes column remarks about which versions support it, something like "alpha support in 1.15 (CSIInlineVolume), beta in 1.15".

Added a column remark as suggested. Can you please check if it's good enough.

README.md
avalluri: Replaced the description as suggested.

README.md
avalluri: Even `sector` is a valid option here. Modified the meaning column to: "PMEM namespace mode(fsdax/sector) of requested ephemeral volume"

deploy/common/pmem-app-ephemeral.yaml
avalluri: Fixed.

deploy/kustomize/kubernetes-1.16/README.md
avalluri: Fixed

pohly:
> > The question around checking for os.ErrNotExit is also still open.
> Now the code uses os.IsErrorNotExist() call, which does the digging and checks for real error. This should address the unwrapping issue you pointed.

Hmm, where do you see that `os.IsErrorNotExist` unwraps errors? The API only says "It is satisfied by ErrNotExist as well as some syscall errors" and the source code confirms that (https://golang.org/src/os/error.go#L92). Errors that wrap `ErrNotExist` are opaque to it, as the following test program confirms:
```
package main

import (
	"fmt"
	"github.com/pkg/errors"
	"os"
)

func main() {
	err := errors.Wrap(os.ErrNotExist, "wrapping")
	fmt.Printf("%q: os.IsNotExists %v, os.IsNotExists(errors.Cause) %v\n",
		err,
		os.IsNotExist(err),
		os.IsNotExist(errors.Cause(err)))
}
```

```
$ go run ./error.go 
"wrapping: file does not exist": os.IsNotExists false, os.IsNotExists(errors.Cause) true
```

avalluri:
> Hmm, where do you see that `os.IsErrorNotExist` unwraps errors?

Sorry, I did not update the comment. I later figured out that it does not unwrap, but I already updated the code to use `errors.Cause()`.

pkg/ndctl/ndctl.go
```diff
@@ -104,7 +106,7 @@ func (ctx *Context) GetNamespaceByName(name string) (*Namespace, error) {
 			}
 		}
 	}
-	return nil, fmt.Errorf("Not found")
+	return nil, errors.Wrapf(os.ErrNotExist, "namespace '%s' not found", name)
```
pohly: Sorry, still not convinced that reusing `os.ErrNotExist` is the right approach here. It means that the error occurred during some file-level operation, which isn't the case here. What if at some point this layer really wants to expose OS-level errors as well as the semantic "namespace not found"? `os.ErrNotExist` then cannot be used for both.

Please follow the example from https://blog.golang.org/go1.13-errors and create your own `NotFound` sentinel value.

Also consider what the final error text will look like when wrapping: the code above leads to "namespace 'foo' not found: file does not exist" (https://golang.org/src/internal/oserror/errors.go). "file" - which file?!

That message doesn't make sense, which shows that reusing `os.ErrNotExist` isn't suitable.

avalluri: Defined new error codes for DeviceManager interface: https://github.com/intel/pmem-csi/pull/383/commits/fe4f6e93e200bb7360244e9fa301f6b63911ad14

avalluri: This change also demands go.1.13 in CI to build.

pohly: Why does this need Go 1.13? You don't need to use the new features from Go 1.13; github.com/pkg/errors is good enough. With "follow the example rom https://blog.golang.org/go1.13-errors and create your own NotFound sentinel value" I meant this:
```
var ErrNotFound = errors.New("not found")
```
That works also in older Go.

pohly: I would have preferred to switch from Go 1.12 to 1.13 in a separate PR. Now I am not sure how to proceed: rewrite the code so that it works with 1.12 or switch to 1.13 as minimum version? Suggestions?

avalluri: > I would have preferred to switch from Go 1.12 to 1.13 in a separate PR. Now I am not sure how to proceed: rewrite the code so that it works with 1.12 or switch to 1.13 as minimum version? Suggestions?


Ok, then shall we deal this as part of #420, which have other changes to device manager and even unit tests. In this PR I would prefer ephemeral code to make use of existing `DeviceManager` API do a `_nil_` check for `GetDevice()` call in NodeServer, as it was in my initial change?

pohly: > In this PR I would prefer ephemeral code to make use of existing DeviceManager API

Okay.

avalluri: Now I reverted the error code change. Now it's ready for final review.

README.md
```diff
@@ -396,6 +399,16 @@ application](deploy/common/pmem-app-cache.yaml) example.
 * A node is only chosen the first time a pod starts. After that it will always restart
   on that node, because that is where the persistent volume was created.
 
+Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The volume request could use below fields as [`volumeAttributes`](https://kubernetes.io/docs/concepts/storage/volumes/#csi):
+
+|key|meaning|optional|default value|
+|---|-------|--------|-------------|
+|size|Size of the requested ephemeral volume|No||
+|nsmode|PMEM namespace mode(fsdax/sector) of requested ephemeral volume|Yes|"fsdax"|
```
pohly: It's not obvious that `fsdax` and `sector` are the valid values. What about this?

|key|meaning|optional|values|
|---|-------|--------|-------------|
|size|Size of the requested ephemeral volume|No||
|nsmode|PMEM namespace mode of requested ephemeral volume|Yes|`fsdax` (default), `sector`|
|eraseAfter|Clear all data after use and before deleting the volume|Yes|`true` (default), `false`|

test/e2e/storage/csi_volumes.go
```diff
@@ -266,3 +253,17 @@ func (m *manifestDriver) finalPatchOptions(f *framework.Framework) utils.PatchCS
...
+	shared := false
+	readOnly := false
+
+	return attributes, shared, readOnly
+}
+
+func (m *manifestDriver) GetCSIDriverName(config *testsuites.PerTestConfig) string {
+	// Return real driver name. 
+	// We can't use m.driverInfo.Name as its not the real driver name
+	return "pmem-csi.intel.com"
```
pohly: `manifestDriver` was meant to be a configurable implementation implementation of the test driver interface. Hard-coding "pmem-csi.intel.com" here is inconsistent with that.

Two options:
- keep the current approach an introduce a m.CSIDriverName field that gets set to "pmem-csi.intel.com"
- remove the manifestDriver struct and replace the implementation with something that is specific to PMEM-CSI

README.md
pohly: Even better, mark up the keys as `size`, `nsmode` and `eraseAfter`.

test/e2e/storage/csi_volumes.go
avalluri: Chose option 1. Done

README.md
avalluri: Done

Issue #379, opened 2019-08-06
move to '-V <version>' with swupd update


mythi:
```-m``` parameter is deprecated and it's good to get rid of it:

```
  -m, --manifest=V        NOTE: this flag has been superseded. Please use -V instead
```

Affects: Dockerfile

avalluri:
@mythi Where was this option used in our scripts? I could find on in comments part in Dockerfile, Is this the one you are referring to?

mythi:
@avalluri I noticed the same after I reported this issue. This is pretty minor. `-m` is only mentioned in the comments but the update script already uses `--version`

avalluri:
Thanks, I will fix the comments part then.

PR #384, opened 2019-08-13
Dockerfile: Remove deprecated option from comments

avalluri:
Replace deprecated swupd '-m' option in SWUPD_UPDATE_ARG comments section.

FIXES: #379


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # CLEARLINUX_BASE and SWUPD_UPDATE_ARG can be used to make the build reproducible
 # by choosing an image by its hash and updating to a certain version with -m:
```
mythi: nit: `-m` is also here

avalluri: Fixed it now :)

avalluri:
@okartau can we merge this change.

Issue #382, opened 2019-08-08
release images are not reproducible


mythi:
We have related issues #378 and #379 but I think this deserves it's own issue.

While working on QAT image reproducible builds I discovered a problem with ``` ${SWUPD_UPDATE_ARG}```:  The value is reset before it's used. The documentation confirms this:

```
An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an arg in multiple stages, each stage must include the ARG instruction.
```

This means ```swupd update``` never gets to use the locked version:

```
Step 8/42 : RUN echo "version ${SWUPD_UPDATE_ARG}" && swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} go-basic-dev && rm -rf /var/lib/swupd
 ---> Running in 385297cac894
version
Update started.
Preparing to update from 30650 to 30700
Downloading packs for:
```

avalluri:
Thanks, @mythi  for finding this, So redefining `ARG SWUPD_UPDATE_ARG` in each build stage should fix the problem right? I found an interesting discussion related to this: https://github.com/moby/moby/issues/34129

PR #385, opened 2019-08-13
Dockerfile: Redefine SWUPD_UPDATE_ARG in all build stages.

avalluri:
An ARG instruction goes out of scope at the end of the build stage where it was
defined. To use an arg in multiple stages, each stage must include the ARG
instruction.

Refer to:
https://github.com/docker/cli/blob/master/docs/reference/builder.md#understand-how-arg-and-from-interact

FIXES: #382

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ ARG SWUPD_UPDATE_ARG=
 # - up-to-date Clear Linux
 # - ndctl installed
 FROM ${CLEAR_LINUX_BASE} AS build
+ARG SWUPD_UPDATE_ARG
 
 ARG NDCTL_VERSION="65"
 ARG NDCTL_CONFIGFLAGS="--disable-docs --without-systemd --without-bash"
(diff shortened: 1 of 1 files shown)


avalluri:
> Did `update-clear-linux-base.sh` also work as expected with this change?

Yes. This script uses = suffix(`ARG SWUPD_UPDATE_ARG=`) in regex express.

mythi:
@avalluri how about updating release-0.5 branch too? `update-clear-linux-base.sh` run is also needed but that should probably wait until the `swupd` error 255 is fixed.

pohly:
Mikko Ylinen <notifications@github.com> writes:

> @avalluri how about updating release-0.5 branch too?
> `update-clear-linux-base.sh` run is also needed but that should
> probably wait until the `swupd` error 255 is fixed.

Yes, this needs to be backported.

@avalluri: are you tracking that?

avalluri:
> Yes, this needs to be backported. @avalluri: are you tracking that?

@pohly  No, I was not. Do we need a new release with this fix?

pohly:

>> Yes, this needs to be backported. @avalluri: are you tracking that?
>
> @pohly  No, I was not.

Can you take the AR to do this?

> Do we need a new release with this fix?

No, the next regular 0.5.x update will have it.

avalluri:
Done. New release [v0.5.2](https://hub.docker.com/r/intel/pmem-csi-driver/tags) with change is published.

Issue #386, opened 2019-08-20
NodeStageVolume violates CSI spec idempotency req.


okartau:
I was going through older TODO/FIXME set, and spotted that NodeStageVolume does not
fulfill spec if Stage is asked for volume which is already mounted.
Spec requires to return 0 if that Volume is already mounted on required point (idempotency).
But our current implementation does not check for existing mount, but let's mount attempt to fail,
so we will return error.
We should make this closer to spec.
Sanity test seems not to cover repeated StageVolume.

pohly:
Olev Kartau <notifications@github.com> writes:

> I was going through older TODO/FIXME set, and spotted that NodeStageVolume does not
> fulfill spec if Stage is asked for volume which is already mounted.
> Spec requires to return 0 if that Volume is already mounted on required point (idempotency).
> But our current implementation does not check for existing mount, but let's mount attempt to fail,
> so we will return error.
> We should make this closer to spec.

True.

> Sanity test seems not to cover repeated StageVolume.

Idempotency in general seems to be covered by poorly by the sanity test
suite. Before fixing our issue, it would be good to add such a test for
all functions to the sanity test suite. Then we'll know that the test
works (because our driver will fail initially) and we know once the fix
works (when the test passes).

okartau:
> it would be good to add such a test

#393 is first step on that path, it adds repeated `node stage volume` and demonstrates (expected) failure when testing current code in devel

okartau:
I think recently merged #463 covered that issue. Possible follow-up improvements to be tracked in #483

PR #388, opened 2019-08-21
WIP: NodeStageVolume: return OK if device already mounted, other misc. improvement

okartau:
CSI spec requires StageVolume to be idempotent, so we have
to check for same volume mount and capability match and
return OK if already mounted.
Also, other aspects in NodeStageVolume checked: multiple
error codes were just copy-pasted as InvalidArgument,
use better values conforming to CSI spec better.

Resolves: #386

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -202,14 +202,13 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	volumeMutex.LockKey(req.GetVolumeId())
 	defer volumeMutex.UnlockKey(req.GetVolumeId())
 
-	// showing for debug:
 	glog.V(4).Infof("NodeStageVolume: VolumeID:%v Staging target path:%v Requested fsType:%v",
 		req.GetVolumeId(), stagingtargetPath, requestedFsType)
 
 	device, err := ns.dm.GetDevice(req.VolumeId)
 	if err != nil {
 		glog.Errorf("NodeStageVolume: did not find volume %s", req.VolumeId)
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+		return nil, status.Error(codes.NotFound, err.Error())
 	}
 
 	// Check does devicepath already contain a filesystem?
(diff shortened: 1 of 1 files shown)


okartau:
let's note that green CI status does not mean this change is verified.
I believe changed code part is not covered by current testing performed by CI (i.e. sanity and e2e).
Some other (temporary) changes are needed to force driver to go into changed code path.

okartau:
I tried to test that code path, using modified NodeStageVolume so that after mounting, k8s should retry mounting, but attempts failed:
- when I mount and return error instead of OK, k8s ignores error value and goes on with Publish
- when I mount and call long sleep in function, simulating missing reply, k8s waits for some time and then goes on with Publish
Finally, I gave up on that idea and used another hack: I inserted code mounting same directory before running code we try to test. This at least gives possibility to see how the new code detects existing mount. That seems to work, and result is "return OK" in the middle just as planned.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -218,6 +218,26 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+	mounter := mount.New("")
+	mountedDev, _, err := mount.GetDeviceNameFromMount(mounter, stagingtargetPath)
+	if err != nil {
+		glog.Errorf("NodeStageVolume: Error getting device name for mount")
+		return nil, err
+	}
+	if mountedDev == device.Path {
+		glog.V(4).Infof("NodeStageVolume: Device %s already mounted to path %s", mountedDev, stagingtargetPath)
+		// Check are the capabilities compatible.
+		if existingFsType == requestedFsType {
```
avalluri: Can't we do this check just after determineFilesystemType() call?

okartau: good point, these two checks can be combined. Also it seems, I have copied the code so that "err" that we return here, has no value. I combine those checks and force-push. 
This actually is place to review the functionality here.
The block of code after comment "what to do if existing file system is detected...",
I am not sure is in entirely right thing. If we detect same fsType file system, we do nothing,
that means a pod will get file system mounted with previous contents.
The current code is based on my early assumption "lets not destroy data". But giving a pod existing (probably old) contents also does not seem right.
Not sure what CSI spec says here, but I doubt does it determine what to do.
I was thinking, because we have shred on volume deletion, and we clear first 4k on creation, we should never get old data.
But I have seen in test logs that the line "Skip mkfs as %v file system already exists on %v" gets printed,
so there is possibility we get old data?

okartau:
Change of this PR went into #393 where work continues combined with other changes.
Closing this PR

PR #390, opened 2019-08-23
setup-ca-kubernetes.sh: make cleanup working

okartau:
'ls' mostly returning zero, led to cleanup never being called,
causing /tmp/pmem-xxxx directories to pile up on test host.
There seems to be no reason to ever keep WORKDIR, so
we can call cleanup unconditionally.
cleanup was misspelled in INT trap setting so it was
never called on interrupted run either.
Combining last cleanup call being part of EXIT trap.
Using pushd/popd to get into and out from WORKDIR.
Do not treat WORKDIR as possibly coming from outside,
this script creates, owns and destroys it unconditionally.

### DIFF
--- a/test/setup-ca-kubernetes.sh
+++ b/test/setup-ca-kubernetes.sh
@@ -6,8 +6,6 @@
 # The default relies on a functional kubectl that uses the target cluster.
 : ${KUBECTL:=kubectl}
 
-# Directory to use for storing intermediate files
-WORKDIR=${WORKDIR:-$(mktemp -d -u -t pmem-XXXX)}
 
 cfssl_found=1
 (command -v cfssl 2>&1 >/dev/null && command -v cfssljson 2>&1 >/dev/null) || cfssl_found=0
@@ -63,12 +61,15 @@ EOF
     echo "Created: $CSR_NAME.crt"
 }
 
-mkdir -p $WORKDIR 2> /dev/null && cd $WORKDIR
-trap cleaup INT
+# Directory to use for storing intermediate files
+WORKDIR=`mktemp -d -t pmem-XXXX`
+cd $WORKDIR
+
+trap cleanup INT EXIT
 cleanup()
 {
-    echo "Cleaning up..."
-    cd .. && rm -rf $WORKDIR 2> /dev/null
+    echo "Cleaning up, deleting $WORKDIR"
+    rm -rf $WORKDIR 2> /dev/null
 }
 
 echo "Generating certificate files in: $WORKDIR"
@@ -133,5 +134,3 @@ echo "  $name.key: $(base64 -w 0 $name-key.pem)"
 done)
 EOF
 fi
-
-[ "$(ls -A $WORKDIR)" ] || cleanup


okartau:
About use of WORKDIR: 
This script seems initially prepared also for scenario where WORKDIR is given as parameter.
In that case, there may be cases where this script should not destroy WORKDIR, as upper level
caller may own it and expect it to remain.
But at least currently we dont do such calling.
To make things more clear about WORKDIR ownership, I propose we either:
1. handle the ownership scope fully, i.e. if WORKDIR comes as arg, we dont delete it at the end (we need to add code for that).
2. we do not handle WORKDIR as given arg, but we always set it in script, i.e. we can simplify first WORKDIR set stmt. That emphasizes point that this script owns WORKDIR and can destroy it freely.

I prefer 2.

test/setup-ca-kubernetes.sh
```diff
@@ -64,11 +64,11 @@ EOF
 }
 
 mkdir -p $WORKDIR 2> /dev/null && cd $WORKDIR
-trap cleaup INT
+trap cleanup INT EXIT
 cleanup()
 {
     echo "Cleaning up..."
-    cd .. && rm -rf $WORKDIR 2> /dev/null
```
avalluri: Just above cleanup() we CD into `$WORKDIR`, so it is safe to come out of the directory before removing it.

Saving script originated directory and moving back to that on cleanup and removing `WORKDIR` would be the right solution in my opinion.

avalluri:
> About use of WORKDIR:
> This script seems initially prepared also for scenario where WORKDIR is given as parameter.
> In that case, there may be cases where this script should not destroy WORKDIR, as upper level
> caller may own it and expect it to remain.
> But at least currently we dont do such calling.
> To make things more clear about WORKDIR ownership, I propose we either:
> 
> 1. handle the ownership scope fully, i.e. if WORKDIR comes as arg, we dont delete it at the end (we need to add code for that).
> 2. we do not handle WORKDIR as given arg, but we always set it in script, i.e. we can simplify first WORKDIR set stmt. That emphasizes point that this script owns WORKDIR and can destroy it freely.
> 
> I prefer 2.

`WORKDIR` in this script is kind of internal, but the initial idea was to make it similar to `setup-ca.sh` script. where one can pass `WORKDIR` to save generated certificates. But in this context I am ok with your proposal 2 - owning and cleaning up `WORKDIR` as the generated certificates anyways saved in Kubernetes cluster.

test/setup-ca-kubernetes.sh
okartau: ah yes, I missed the point that we changed into WORKDIR, sure it is more correct to get out, although it seems to run without error if we do not.
I think pushd/popd could be used here, I will push a new version.

pohly:
> 2. we do not handle WORKDIR as given arg, but we always set it in script, i.e. we can simplify first WORKDIR set stmt. That emphasizes point that this script owns WORKDIR and can destroy it freely.
>
> I prefer 2.

Me too.

okartau:
is it ok now, can merge? Works based on my local testing

test/setup-ca-kubernetes.sh
```diff
@@ -63,12 +61,16 @@ EOF
     echo "Created: $CSR_NAME.crt"
 }
 
-mkdir -p $WORKDIR 2> /dev/null && cd $WORKDIR
-trap cleaup INT
+# Directory to use for storing intermediate files
+WORKDIR=`mktemp -d -t pmem-XXXX`
+pushd $WORKDIR >/dev/null 2>&1
```
avalluri: Now by adding `pushd` and `popd` this script demands bash shell, would be good to add `!/bin/bash`

okartau: I missed this, true.
Instead of turning shell to bash, we can consider alternatives:
at least these 2 non-bash methods exist to get back to previous dir without using "cd ..":
1. 
```
cd some-dir
  # do something
cd - > /dev/null
```
2.
```
cd some-dir
  # do something
cd "$OLDPWD"
```

okartau: but I need to check do all these method work when combined with trap handler

avalluri: In my opinion, both should work.

okartau:
pushed a version without pushd/popd, using OLDPWD.
Speaking of bashisms:
checkbashisms script reports that `command` with any other option than -p is bashism.
But some other sources recommend `command -v` as POSIX compatible.
Which opinion is correct?

pohly:
I would ignore checkbashism and use shellcheck instead.

test/setup-ca-kubernetes.sh
pohly: The simplest solution is to do "cd" only in subshells and never attempt to return to the original dir.

okartau: yes, that is good guidance, but in current script the `cd into` part is in script body and  `cd out` is in cleanup handler function, so it becomes bit more complex. I think the original idea of cleanup handler was to provide removal in case where user stops long-running script with Ctrl-C for example. But when we think about it, script probably was made part of fully autonomous setup after that, so we dont have user any more who stops it, right, so can we simplify?

okartau: also, that whole idea of "cd out" before removal of directory is bit of purist approach, as the end result seems to be very same if we dont care too much and remove directory where we are in. It is last op in script anyway. What we try to achieve here is extra level of correctness.

pohly:
> also, that whole idea of "cd out" before removal of directory is bit
> of purist approach, as the end result seems to be very same if we dont
> care too much and remove directory where we are in. It is last op in
> script anyway. What we try to achieve here is extra level of
> correctness.

That's not correctness, it's not being sure about Unix
semantics... Deleting a directory that is the current directory of some
process, including the current one, is fine. We don't need to "step out"
before deleting.

okartau:
In that case, I took away `cd out` command. 
I think it is not needed to shift most of the script to become a subshell just because we cd into WORKDIR.

okartau:
if we are OK with this now, I will merge.

PR #391, opened 2019-08-23
cleaned alignment code, unified naming to use VolumeID more

okartau:
pmd-lvm.go: Delete old irrelevant comments about selection strategy.
LVM uses byte-aligned size arg instead or MB

In CSI spec, Name and VolumeID are different things.
In PMEM-CSI we use VolumeID for naming volumes and namespaces,
so let's stick to the style and use VolumeID in code,
otherwise it may be confusing why we suddenly switch back to "Name"
when code gets down to device-manager level.

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -84,20 +84,20 @@ func (lvm *pmemLvm) GetCapacity() (map[string]uint64, error) {
 }
 
 // nsmode is expected to be either "fsdax" or "sector"
-func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error {
+func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) error {
 	if nsmode != string(ndctl.FsdaxMode) && nsmode != string(ndctl.SectorMode) {
 		return fmt.Errorf("Unknown nsmode(%v)", nsmode)
 	}
 	lvmMutex.Lock()
 	defer lvmMutex.Unlock()
-	// Check that such name does not exist. In certain error states, for example when
+	// Check that such volume does not exist. In certain error states, for example when
 	// namespace creation works but device zeroing fails (missing /dev/pmemX.Y in container),
 	// this function is asked to create new devices repeatedly, forcing running out of space.
 	// Avoid device filling with garbage entries by returning error.
-	// Overall, no point having more than one namespace with same name.
-	_, err := lvm.getDevice(name)
+	// Overall, no point having more than one namespace with same volumeId.
+	_, err := lvm.getDevice(volumeId)
 	if err == nil {
-		return fmt.Errorf("CreateDevice: Failed: namespace with that name '%s' exists", name)
+		return fmt.Errorf("CreateDevice: Failed: volume with that name '%s' exists", volumeId)
 	}
 	vgs, err := getVolumeGroups(lvm.volumeGroups, nsmode)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/pmem-device-manager/pmd-util.go
```diff
@@ -12,6 +12,9 @@ import (
 )
 
 const (
+	// 1 GB align in ndctl creation request has proven to be reliable.
+	ndctlAlign       uint64        = 1024 * 1024 * 1024
```
avalluri: Is this still the case, As mentioned in related issue https://github.com/pmem/ndctl/issues/76 I guess the proposed patch got merged and that supposed to fix the default 2MB alignment issue. 

Or else what about this proposal: we change our code to create the namespace with default alignment, in case of alignment error, then try with 1GB alignment.

pkg/pmem-device-manager/pmd-util.go
```diff
@@ -81,3 +84,29 @@ func WaitDeviceAppears(dev PmemDeviceInfo) error {
...
+		// It will get rounded up by libndctl to meet the alignment.
+		size = 1
+	}
+	// libndctl needs to store meta data and will use some of the allocated
+	// space for that (https://github.com/pmem/ndctl/issues/79).
+	// We don't know exactly how much space that is, just
+	// that it should be a small amount. But because libndctl
+	// rounds up to the alignment, in practice that means we need
+	// to request `align` additional bytes.
+	size += ndctlAlign
```
avalluri: It shouldn't be that adding extra 1GB to size only in case when size < 2GB ?

pkg/pmem-device-manager/pmd-util.go
okartau: It would be relief if improved, we need to re-test this.

pkg/pmem-device-manager/pmd-util.go
okartau: IIRC, this was not size dependent. We generally always get one align-step less than what we asked. This is relatively easy to verify with cmdline ndctl, creating namespaces and checking what is created.

avalluri: Ya got it. Thanks.

avalluri: Then this issue exists with even default(2M) alignment, so I think we better handle this ndctl layer.

pkg/pmem-device-manager/pmd-util.go
okartau: > create the namespace with default alignment, in case of alignment error, then try with 1GB alignment.

That's likely the way we have to do it, as we can not be sure on top of which kernel code we run, the one before or after fix. If we statically encode use default alignment, we also have to require new kernel, and that is not so welcome requirement.

pkg/pmem-device-manager/pmd-util.go
okartau: If we move handling of size compensation out of here into ndctl layer, then even less code remains here, and also less gets covered by unit testing that I planned to add here.
The whole idea of having a separate function for align and size calc gets weaker. It is already quite weak to start with, because major factor affecting final values: InterleaveWidth is inside ndctl and not usable during unit test time as it's HW-specific.
So I have also thinking, should we give up idea what is in this PR altogether.
Comments cleanup seems a good part here, though, as some existing comments are out of date.

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -94,30 +99,23 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error
...
-	// Division by 1M should not result in smaller-than-asked here
-	// as lvcreate will round up to next 4MB boundary.
-	sizeM := int(size / (1024 * 1024))
-	// Asked==zero means unspecified by CSI spec, we create a small 4 Mbyte volume
-	// as lvcreate does not allow zero size (csi-sanity creates zero-sized volumes)
-	if sizeM <= 0 {
-		sizeM = 4
+	// adjust up to next alignment boundary, if not aligned already
+	lvmAlign := lvmAlignMB * 1024 * 1024
+	if (size % lvmAlign) != 0 {
```
avalluri: How about making calculation bit simple:
```
if n := size % lvmAlign ; n != 0 {
  size += (lvmAlign - n)
}
```

avalluri:
Looks good to me.

PR #392, opened 2019-08-26
Devicemanager enhancements

avalluri:
Added two new fields to PmemDeviceInfo structure:
 - Mode : holds the device namespace mode 
 - NumaNode: hods the numa node to which the device attached to

In LVM mode we use the volume group tags for storing these details.

### DIFF
--- a/pkg/ndctl/namespace.go
+++ b/pkg/ndctl/namespace.go
@@ -10,6 +10,7 @@ import "C"
 import (
 	"encoding/json"
 	"fmt"
+
 	/* needed for nullify
 	"os"
 	"syscall"*/
@@ -184,6 +185,12 @@ func (ns *Namespace) Type() NamespaceType {
 	}
 
 	return UnknownType
+}
+
+//NumaNode returns numa node number attached to this namespace
+func (ns *Namespace) NumaNode() int {
+	ndns := (*C.struct_ndctl_namespace)(ns)
+	return int(C.ndctl_namespace_get_numa_node(ndns))
 }
 
 //Enabled return if namespace is enabled
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -67,6 +67,12 @@ func (r *Region) Type() RegionType {
 	return UnknownRegion
 }
 
+//NumaNode returns numa node number attached to this region
+func (r *Region) NumaNode() int {
+	ndr := (*C.struct_ndctl_region)(r)
+	return int(C.ndctl_region_get_numa_node(ndr))
+}
+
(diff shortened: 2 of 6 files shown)


pohly:
Can you explain why this is useful?

avalluri:
> Can you explain why this is useful?

NSMode: 
As I made a note here: https://github.com/intel/pmem-csi/pull/383/files#diff-4a0caa0c2e37db715640b7cbf97d7a1fR263-R267 , the PMEM device should have a way to knowin which mode it was created, currently this information is with VolumeContext, but I feel the right place is PmemDeviceInfo structure.

NumaNode:
Its just an enabler change to make the driver numa aware, though the solution is still under investigation. Numa APIs in ndctl layer are useful to know a pmem region's numa  node, if we wants to support PMEM device creation in given numa node region.

pohly:
@avalluri: do we still need this PR?

avalluri:
> @avalluri: do we still need this PR?

Yes, I believe. It would be useful at least logging NUMA node related information of a device.

pohly:
@avalluri : is this PR still relevant?

avalluri:
> @avalluri : is this PR still relevant?

Yes if we need to track/log the numa node details of a volume. If not we can ignore and close the PR.

pohly:
> Yes if we need to track/log the numa node details of a volume

I don't think we have such a requirement. Are you aware of one? You were working with Sasha on NUMA.

avalluri:
> > Yes if we need to track/log the numa node details of a volume
> 
> I don't think we have such a requirement. Are you aware of one? You were working with Sasha on NUMA.

There is no such requirement enforced by CRI-RM. But in general, I feel like there is no easy way to know to which NUMA node a particular PMEM-CSI volume is attached to while debugging the performance of an application that using this volume.

pohly:
> I feel like there is no easy way to know to which NUMA node a particular PMEM-CSI volume is attached to while debugging the performance of an application that using this volume.

So you want to expose that to the user? That seems worthwhile, but then we also need to define how users actually get that information. Log file parsing IMHO shouldn't be the answer.

pohly:
> Log file parsing IMHO shouldn't be the answer.

Perhaps some metric? We should explore further what we can do with those.

avalluri:
> > Log file parsing IMHO shouldn't be the answer.
> 
> Perhaps some metric? We should explore further what we can do with those.

Yes, I agree.

pohly:
@avalluri can we close this PR until there is a need for it?

avalluri:
> @avalluri can we close this PR until there is a need for it?

Okay.

PR #393, opened 2019-08-28
WIP: DO NOT MERGE. Add repeated node tests (vendor) + tolerate repeated node ops

okartau:
One commit is prototype for to-be-upstreamed change that adds repetition of sanity tests, to test for idempotency.

### DIFF
--- a/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
+++ b/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
@@ -66,6 +66,208 @@ func isPluginCapabilitySupported(c csi.IdentityClient,
 	return false
 }
 
+func runControllerTest(sc *SanityContext, c  csi.NodeClient, s  csi.ControllerClient, cl *Cleanup, controllerPublishSupported bool, nodeStageSupported         bool, nodeVolumeStatsSupported   bool, count int) {
+
+	name := UniqueString("sanity-node-full")
+
+	By("getting node information")
+	ni, err := c.NodeGetInfo(
+		context.Background(),
+		&csi.NodeGetInfoRequest{})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(ni).NotTo(BeNil())
+	Expect(ni.GetNodeId()).NotTo(BeEmpty())
+
+	var accReqs *csi.TopologyRequirement
+	if ni.AccessibleTopology != nil {
+		// Topology requirements are honored if provided by the driver
+		accReqs = &csi.TopologyRequirement{
+			Requisite: []*csi.Topology{ni.AccessibleTopology},
+		}
+	}
+
+	// Create Volume First
+	By("creating a single node writer volume")
+	vol, err := s.CreateVolume(
+		context.Background(),
+		&csi.CreateVolumeRequest{
+			Name: name,
+			VolumeCapabilities: []*csi.VolumeCapability{
+				{
+					AccessType: &csi.VolumeCapability_Mount{
+						Mount: &csi.VolumeCapability_MountVolume{},
+					},
+					AccessMode: &csi.VolumeCapability_AccessMode{
+						Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+					},
+				},
+			},
+			Secrets:                   sc.Secrets.CreateVolumeSecret,
+			Parameters:                sc.Config.TestVolumeParameters,
+			AccessibilityRequirements: accReqs,
+		},
+	)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(vol).NotTo(BeNil())
+	Expect(vol.GetVolume()).NotTo(BeNil())
+	Expect(vol.GetVolume().GetVolumeId()).NotTo(BeEmpty())
+	cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetVolumeId()})
+
+	var conpubvol *csi.ControllerPublishVolumeResponse
+	if controllerPublishSupported {
+		By("controller publishing volume")
+
+		conpubvol, err = s.ControllerPublishVolume(
+			context.Background(),
+			&csi.ControllerPublishVolumeRequest{
+				VolumeId: vol.GetVolume().GetVolumeId(),
+				NodeId:   ni.GetNodeId(),
+				VolumeCapability: &csi.VolumeCapability{
+					AccessType: &csi.VolumeCapability_Mount{
+						Mount: &csi.VolumeCapability_MountVolume{},
+					},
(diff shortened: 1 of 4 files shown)


okartau:
@pohly do you think submitting 1st commit from here to upstream makes sense?

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
```diff
@@ -718,27 +718,29 @@ var _ = DescribeSanity("Node Service", func(sc *SanityContext) {
 		}
 		// NodeStageVolume
 		if nodeStageSupported {
-			By("node staging volume")
```
pohly: Better define two separate test cases: one which does idempotent calls with a configurable number of repetitions (at least two, perhaps better more to stress the driver a bit - 10?) and one which always invokes each function only once. The body of the tests can be the same function with a repetition parameter.

okartau: something like this? (pushed new version)

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
```diff
@@ -644,192 +845,11 @@ var _ = DescribeSanity("Node Service", func(sc *SanityContext) {
...
-				Secrets:  sc.Secrets.DeleteVolumeSecret,
-			},
-		)
-		Expect(err).NotTo(HaveOccurred())
+		By("runControllerTest single")
+		runControllerTest(sc, c, s, cl, controllerPublishSupported, nodeStageSupported, nodeVolumeStatsSupported, 1)
+		By("runControllerTest double")
+		runControllerTest(sc, c, s, cl, controllerPublishSupported, nodeStageSupported, nodeVolumeStatsSupported, 2)
+		By("runControllerTest quadruple")
+		runControllerTest(sc, c, s, cl, controllerPublishSupported, nodeStageSupported, nodeVolumeStatsSupported, 4)
```
pohly: These should be in separate `It()` calls, not `By()`. Each `It()` is one test that can be run by itself.

I think two cases make sense:
* `It("should work")` = count 1
* `It("should be idempotent")` = count defined via a new Config.IdempotentCount

The semantic of that field should be:
- 0: run with default repetition count, let's say 10
- -1: skip idempotent tests
- >0: repeat each call that many times

okartau: > * 0: run with default repetition count, let's say 10
> * -1: skip idempotent tests
> * larger than 0: repeat each call that many times

IMO this is too complex with multi-level encoding (zero means some other value). What if we have configured value which means just one one thing, count of repetitions. The default of that value, for example 10, gets written as init.value. User can override it in config set. If overriden with zero or negative, we skip idempot. test.
Seems what I propose is almost same with yours, but without the first line "zero means ten".

okartau: actually, if we keep the looping idea I already developed, value of 1 means simple test without repetitions, and values 2 and above will test idempotency. So running the 2nd It("should be idempotent") with count=1 would repeat the test of "should work". Probably makes sense to code it so that if lower than 2, we skip 2nd It()

okartau:
newly pushed version should get closer to what you proposed

pohly:
> IMO this is too complex with multi-level encoding (zero means some other value). What if we have configured value which means just one one thing, count of repetitions. The default of that value, for example 10, gets written as init.value. User can override it in config set. If overriden with zero or negative, we skip idempot. test.
> Seems what I propose is almost same with yours, but without the first
> line "zero means ten".

My intention was to have the idempotency test enabled by default when a
user of csi-test uses Config{}, i.e. new fields set to null. Otherwise
other projects will update to the new version of csi-test and most
likely forget to enable the new test because they won't even notice that
it was added.

If it is enabled by default and it fails, they will noticed. If it
works, test coverage gets increased automatically.

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/sanity.go
```diff
@@ -120,6 +120,8 @@ type Config struct {
 	RemoveStagingPathCmd string
 	// Timeout for the executed commands for path removal.
 	RemovePathCmdTimeout int
+	// Timeout for the executed commands for path removal.
+	IdempotentCount int
```
pohly: Cut-and-paste error: comment isn't updated.

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
```diff
@@ -644,192 +845,17 @@ var _ = DescribeSanity("Node Service", func(sc *SanityContext) {
...
-
-		_, err = s.DeleteVolume(
-			context.Background(),
-			&csi.DeleteVolumeRequest{
-				VolumeId: vol.GetVolume().GetVolumeId(),
-				Secrets:  sc.Secrets.DeleteVolumeSecret,
-			},
-		)
-		Expect(err).NotTo(HaveOccurred())
+		By("runControllerTest with Idempotent count")
```
pohly: I still think this should be a separate "It" test because then a test failure in this part is obviously related to idempotency ("normal" test passes, "test with idempotency" doesn't).

pohly:
> My intention was to have the idempotency test enabled by default when a user of csi-test uses Config{}, i.e. new fields set to null.

If there's a better way to achieve that, then I am all ears of course.

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
okartau: this code is in separate It() block now, is it diff-display that is not letting this to show clearly? or did I do it not right... I can use another By() line inside It() right?

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/sanity.go
pohly: The semantic of that field needs to be explained, in particular the special values (0 and -1).

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go
pohly: You are right, it's fine. I just missed the previous hunk in the diff.

okartau:
added comments about Count semantics

okartau:
cherry-picked commit from #408 to here to verify combined effect on CI

okartau:
Closing without merge. The landscape around these changes changed drastically after recent set of ephemeral and block-mode changes. Some of idempotency-related issues got covered there. I will try to cover others in smaller updates using combination with #453 as testing platform.

PR #405, opened 2019-09-14
Controller: Create VolumeID from name as hash function

okartau:
Use hash value of Volume Name in request as VolumeID, instead
of creating a new UUID each time a request arrives,
This avoids situations where different VolumeIDs may be associated
with same name.

Resolves: #404

### DIFF
--- a/pkg/pmem-csi-driver/controllerserver-master.go
+++ b/pkg/pmem-csi-driver/controllerserver-master.go
@@ -7,13 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package pmemcsidriver
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"strconv"
 	"sync"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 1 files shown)


okartau:
This is minimalistic approach to fix #404 by ensuring we use same VolumeID for request with same Name.
I added WIP in case we want to develop this further, few ideas:
now that VolumeID is a function of Name, we could stop storing Name in volume registry,
and change GetVolumeByName to compute hash and match by VolumeID ? Does such approach help? It would mean one entry less of context to maintain, including the stored state in host file system.

pohly:
As a quick fix this is okay. Some of the enhancements discussed in issue #404 can be added later.

As seen in that discussion, the purpose of this particular code wasn't clear. It would be useful to add source code comments, ideally when writing such code but definitely when making modifications after it became more obvious that further discussions around the code where necessary.

pohly:
Regarding the CI: I removed "WIP" from the title and then a build job started. Not sure whether it is a coincidence (perhaps the CI simply missed a message from GitHub when the PR was created) or a feature (WIP PRs now getting ignored).

PR #407, opened 2019-09-17
DEBUG: trace lvm.devices contents

okartau:
I suspect that this set grows without losing objects even
if volumes are deleted.

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -140,6 +140,8 @@ func (lvm *pmemLvm) CreateDevice(name string, size uint64, nsmode string) error 
 				}
 
 				lvm.devices[device.Name] = device
+				glog.V(5).Infof("CreateDevice: add [%+v] to lvm.devices", device)
+				glog.V(5).Infof("CreateDevice: lvm.devices now [%+v]", lvm.devices)
 
 				return nil
 			}
@@ -196,6 +198,7 @@ func (lvm *pmemLvm) GetDevice(id string) (PmemDeviceInfo, error) {
 }
 
 func (lvm *pmemLvm) getDevice(id string) (PmemDeviceInfo, error) {
+	glog.V(5).Infof("getDevice: lvm.devices now [%+v]", lvm.devices)
 	if dev, ok := lvm.devices[id]; ok {
 		return dev, nil
 	}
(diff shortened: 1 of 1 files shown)


avalluri:
@okartau Good catch, it's a bug in LVM code, #408 supposed to fix this.

okartau:
I cherry-picked commit from #408 to here to see the new traces that should prove that entries are not accumulating.

okartau:
looks better, no entries accumulating in lvm.devices

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -158,12 +158,18 @@ func (lvm *pmemLvm) DeleteDevice(name string, flush bool) error {
 	if err != nil {
 		return err
 	}
-	if err := ClearDevice(device, flush); err != nil {
+	if err = ClearDevice(device, flush); err != nil {
```
pohly: `err :=` is better style because it makes this line independent of the preceding code.

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -158,12 +158,18 @@ func (lvm *pmemLvm) DeleteDevice(name string, flush bool) error {
...
 		return err
 	}
 
-	_, err = pmemexec.RunCommand("lvremove", "-fy", device.Path)
-	return err
+	if _, err = pmemexec.RunCommand("lvremove", "-fy", device.Path); err != nil {
```
pohly: Same here.

Forget about "but reusing the same variable is more efficient" - it's impossible to tell whether that is measurable or even true. The compiler might do a lifecycle analysis and reuse the exact same memory automatically.

pkg/pmem-device-manager/pmd-lvm.go
okartau: sorry for creating confusion, but the real home of the code you review, is under #408. In this PR I cherry-picked changes from #408 to cause combined state go through CI and thus demonstrate improvement that lvm.devices is not growing (which is otherwise not detectable from CI results). The meaning of this PR originally was only to demonstrate that lvm.devices is growing, so it's not meant for merging.

okartau: @avalluri can you take that feedback into account in #408

PR #409, opened 2019-09-18
Change UUID to hash on Node side as well when creating VolumeID, logs improvement

okartau:
Change UUID to hash on Node side as well when creating VolumeID


okartau:
I am not sure how to test this code. It seems not activating in current sanity, e2e tests (I added log stmt and that never shows up in test logs)

okartau:
Added 2nd commit improving logging, inspired by recent discussion under #404 about getting Volume Name and ID visible on same log lines. In same pass, implemented few other logs improvement ideas that came along.

Issue #394, opened 2019-08-30
kubelet repeats "RegisterPlugin error" on kubernetes 1.15 nodes


okartau:
kubelet error is repeatedly logged on all nodes system log with kubernetes 1.15 deployment:
```
kubelet[2627]: E0830  goroutinemap.go:150] Operation for "/var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock" failed. No retries permitted until 2019-08-30 08:12:56.950765334 +0000 UTC m=+535.947459351 (durationBeforeRetry 32s). Error: "RegisterPlugin error -- failed to get plugin info using RPC GetInfo at socket /var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock, err: rpc error: code = Unimplemented desc = unknown service pluginregistration.Registration"
```

This has seen on systems:
4xVM govm cluster, Ubuntu 19.04, kubernetes 1.15.3
2xbaremetal cluster, Fedora-30, kubernetes 1.15.3

This msg is **not** reported on system:
4xVM govm cluster, Clear 29820, kubernetes 1.14.2

This error does not affect functionality: deployment, e2e and sanity tests work OK.
But it may leave impression something is wrong.
We should investigate the cause, possibly find a way to avoid this error msg,
and if  unavoidable, we should document it.

avalluri:
This is even present in v1.14:

`kubelet[10520]: E0830 09:38:47.257975   10520 plugin_watcher.go:120] error failed to get plugin info using RPC GetInfo at socket /var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock, err: rpc error: code = Unimplemented desc = unknown service pluginregistration.Registration when handling create event: "/var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock": CREATE`

But on 1.14 registration failures are not retried, whereas 1.15 introduced plugin-manager that keep retrying till all the plugin(socket files under` /var/lib/kubelet/{plugins,plugin_registry}`) registrations success.

PMEM-CSI is using `/var/lib/kubelet/plugins/` folder for keeping its listening socket, this folder is watched by plugin-manger for plugin changes. By using a different socket location(`/var/lib/kubelet/pmem-csi.intel.com`) could fix this issue.

okartau:
Should we change socket location in our deployment files then?
Starting with 1.14 ?

pohly:
Olev Kartau <notifications@github.com> writes:
[...]

PR #410, opened 2019-09-18
deploy: Do not use kubelet plugins location for driver socket

avalluri:
/var/lib/kubelet/{plugins,plugins_registry} is the location for plugins which
are supposed to implement kubelet's plugin registration API. This is implemented
by node-driver-registrar sidecar container in case of a CSI driver.

So PMEM-CSI should not use these locations for creating its listening socket.
No we use plugin state directory(/var/lib/pmem-csi.intel.com) for this.

Fixes: #394

### DIFF
--- a/deploy/kustomize/driver/pmem-csi.yaml
+++ b/deploy/kustomize/driver/pmem-csi.yaml
@@ -123,7 +123,7 @@ spec:
           privileged: true
         env:
         - name: CSI_ENDPOINT
-          value: unix:///csi/csi.sock
+          value: unix:///var/lib/pmem-csi.intel.com/csi.sock
         - name: KUBE_NODE_NAME
           valueFrom:
             fieldRef:
@@ -137,8 +137,6 @@ spec:
         - name: TERMINATION_LOG_PATH
           value: /tmp/termination-log
         volumeMounts:
-        - name: plugin-socket-dir
-          mountPath: /csi
         - name: mountpoint-dir
           mountPath: /var/lib/kubelet/plugins/kubernetes.io/csi
           mountPropagation: Bidirectional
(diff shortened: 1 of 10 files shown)


deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -155,18 +153,14 @@ spec:
         imagePullPolicy: Always
         image: quay.io/k8scsi/csi-node-driver-registrar:v1.X.Y
         args: [ "-v=3",
-            "--kubelet-registration-path=/var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock",
+            "--kubelet-registration-path=/var/lib/pmem-csi.intel.com/csi.sock",
             "--csi-address=/csi/csi.sock" ]
         volumeMounts:
-        - name: plugin-socket-dir
+        - name: pmem-state-dir
           mountPath: /csi
```
pohly: Would it make sense to change the `mountPath` from `/csi` to something else? `/pmem-csi` perhaps?

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -155,18 +153,14 @@ spec:
...
         - name: registration-dir
           mountPath: /registration
```
pohly: Is `/registration` used anywhere?

avalluri: No, I guess. Will remove part of this change.

deploy/kustomize/driver/pmem-csi.yaml
avalluri: Done

deploy/kustomize/driver/pmem-csi.yaml
avalluri: Done

avalluri: Oops, I was wrong, this registration path is used by node-driver-registrar to keep its listening socket. I will revert this change.

pohly:
Can you add a comment? If the node-driver-registrar had a command line option for it, we could use that to document the usage, but /registration is hard-coded in the source (https://github.com/kubernetes-csi/node-driver-registrar/blob/6ef000ae529ad4c8e2928f0803d65ac25f3f9fcf/cmd/csi-node-driver-registrar/node_register.go#L38).

avalluri:
> Can you add a comment to `/registration`? If the node-driver-registrar had a command line option for it, we could use that to document the usage, but /registration is hard-coded in the source (https://github.com/kubernetes-csi/node-driver-registrar/blob/6ef000ae529ad4c8e2928f0803d65ac25f3f9fcf/cmd/csi-node-driver-registrar/node_register.go#L38).

Done.

amgads:
is this fix available in 1.16?  if yes, which 1.16.x?

PR #412, opened 2019-09-19
Use klog for logging

avalluri:
Driver logs are not logging the real filename from where the message generated.
Because Kubernetes glog is just a wrapper around klog and klog implements the
logic to fetch the file name from it's caller. As in our case all log messages
are logged via glog, hence fails to find the real filename.

This change fixes this by using directly klog's logging API.


okartau:
nice! instead of:
```
pmem-csi-controller-0/pmem-driver: ... glog.go:58] Controller CreateVolume: Name:
```
we now have:
```
pmem-csi-controller-0/pmem-driver: ... controllerserver-master.go:164] CreateVolume: Name:
```

avalluri:
> Which code still needs glog? Can we get rid of it entirely?

None. I removed all traces of glog now from our tree.

PR #415, opened 2019-09-20
Dockerfile: fix curl "error loading libraries" error in CI

pohly:
When running in the CI, we have the "error loading libraries" error
again because the "ldconfig" workaround came after using curl.


pohly:
Not sure why this wasn't found when build-testing the PR, but at least the periodic builds now fail because of this:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi-periodic/176/execution/node/25/log

pohly:
@okartau okay?

PR #417, opened 2019-09-25
DeleteVolume has to return FailedPrecondition if volume is in use

okartau:
To achieve that, we chack for hard-coded error string returned
by pmd-util.go:FlushDevice()


avalluri:
I was also working on this:): https://github.com/avalluri/pmem-CSI/commit/f1aff5b3ad96cc819c34901b421c986e98b6c16d

This also needs a change from master controller, as it returns "Internal" error instead it should pass the same error to the caller.

okartau:
we need to think also how to test this properly. Added test in sanity suite? Initially I modified some test code to call remove() too early while device was used, and it triggered that status in Node as was expected.

avalluri:
> we need to think also how to test this properly.

Create volume, publish volume and delete volume should trigger this error I believe.

okartau:
> This also needs a change from master controller

In master the code is already GRPC so we can stop creating new code and pass error returned from Node?

avalluri:
> > This also needs a change from master controller
> 
> In master the code is already GRPC so we can stop creating new code and pass error returned from Node?

Yes, that's what my current change is having, but I myself not sure, what if we hit with "`code = Unavailable desc = transport is closing`" as in https://github.com/intel/pmem-csi/issues/404.

okartau:
> what if we hit with unexpected errors

Could we return FailedPrecondition if code from Node is FailedPrecondition, and Internal in all other cases ?

avalluri:
> > what if we hit with unexpected errors
> 
> Could we return FailedPrecondition if code from Node is FailedPrecondition, and Internal in all other cases ?

I think of something like, If the error is `Status` error then return the same, for other errors return error with `Internal` code.

okartau:
@avalluri do you want to pick this commit here and make it part of #420, plus make another change as you propose "This also needs a change from master controller".
My idea is, this here is smaller and can be combined with bigger #420 that also deals with Volume deletion.

avalluri:
> @avalluri do you want to pick this commit here and make it part of #420, plus make another change as you propose "This also needs a change from master controller".
> My idea is, this here is smaller and can be combined with bigger #420 that also deals with Volume deletion.

@okartau Yes, Your suggestion sounds fine to me as both deal with DeleteVolume() functionality, i can pick this and fix the remaining part.

avalluri:
> > @avalluri do you want to pick this commit here and make it part of #420, plus make another change as you propose "This also needs a change from master controller".
> > My idea is, this here is smaller and can be combined with bigger #420 that also deals with Volume deletion.
> 
> @okartau Yes, Your suggestion sounds fine to me as both deal with DeleteVolume() functionality, i can pick this and fix the remaining part.

How about if I take my fix: https://github.com/avalluri/pmem-CSI/commit/f1aff5b3ad96cc819c34901b421c986e98b6c16d

avalluri:
> @avalluri do you want to pick this commit here and make it part of #420,

@okartau Now #420 covers this, so I think we can close this PR and track this via #420.

PR #418, opened 2019-09-25
CI: automatic update of master branch

pohly:
This implements the policy described in
https://github.com/intel/pmem-csi/blob/devel/DEVELOPMENT.md#branching:
each time full testing of the "devel" branch has succeeded, "master"
gets moved forward to it.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -241,6 +241,27 @@ pipeline {
             }
         }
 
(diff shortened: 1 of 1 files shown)


pohly:
The pre-merge testing can only show that there are no syntax errors in the modified Jenkinsfile. Actual testing of the change can only happen after merging it.

PR #419, opened 2019-09-25
Controller: Consider topology in GetCapacity() request

avalluri:
If topology is set in GetCapacityRequest, we should return only pmem capacity
for that topology/node instead the cluster-wide capacity.

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -294,6 +294,61 @@ var _ = Describe("sanity", func() {
 			}, "3m", "5s").Should(Equal(capacity.AvailableCapacity), "total capacity after controller restart")
 		})
 
+		It("should return right capacity", func() {
+			resp, err := ncc.GetCapacity(context.Background(), &csi.GetCapacityRequest{})
+			Expect(err).Should(BeNil(), "Failed to get node initial capacity")
+			nodeCapacity := resp.AvailableCapacity
+			i := 0
+			volSize := int64(1) * 1024 * 1024 * 1024
+
+			// Keep creating volumes till there is change in node capacity
+			Eventually(func() bool {
+				volName := fmt.Sprintf("get-capacity-check-%d", i)
+				i++
+				By(fmt.Sprintf("creating volume '%s' on node '%s'", volName, nodeID))
+				req := &csi.CreateVolumeRequest{
+					Name: volName,
+					VolumeCapabilities: []*csi.VolumeCapability{
+						{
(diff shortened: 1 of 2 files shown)


test/e2e/storage/sanity.go
```diff
@@ -294,6 +294,22 @@ var _ = Describe("sanity", func() {
...
 
+		It("should return right capacity", func() {
+			resp, err := cc.GetCapacity(context.Background(), &csi.GetCapacityRequest{
+				AccessibleTopology: &csi.Topology{
+					Segments: map[string]string{
+						"pmem-csi.intel.com/node": nodeID,
+					},
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
```
pohly: This is an anti-pattern - always include an additional explanation.
https://github.com/kubernetes/kubernetes/issues/34059

To a lesser extend that also applies to other assertions.

test/e2e/storage/sanity.go
```diff
@@ -294,6 +294,22 @@ var _ = Describe("sanity", func() {
...
+			cap := resp.AvailableCapacity
+			resp, err = ncc.GetCapacity(context.Background(), &csi.GetCapacityRequest{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.AvailableCapacity).To(Equal(cap), "capacity mismatch")
```
pohly: If all nodes have the same amount of free storage, then this test cannot verify that the call with topology information really ends up querying the right node.

Can you make it a bit more strict?
- check and compare results for all nodes
- create a volume on one node
- check and compare all nodes again

avalluri: > If all nodes have the same amount of free storage, then this test cannot verify that the call with topology information really ends up querying the right node.
> 
> Can you make it a bit more strict?
> 
> * check and compare results for all nodes
> * create a volume on one node
> * check and compare all nodes again

This might not work. Capacity is the maximum available extent in a region/vgs on the node. So, creating a volume might not change in the Node capacity, if the volume is created in the smaller region/vgs.

pohly: Then repeat creating volumes until there is a change in capacity.

avalluri:
Updated the test such that:
 1) first get the node capacity by using node.ControllerServer interface
 2) create volumes on this node till we find change in capacity and mark this capacity as nodeCapacity
 3) get capacity using controller.ControllerServer by setting topology
 4) compare both capacities(step 2 vs step 3)

Issue #413, opened 2019-09-19
cache data in LVM device manager?


pohly:
In https://github.com/intel/pmem-csi/pull/408#issuecomment-532540904, @okartau pointed out that https://github.com/intel/pmem-csi/blob/e09d8dae00fcd126011c6bbe66951f90b35fec30/pkg/pmem-device-manager/pmd-lvm.go caches information about the LVM state on the node. @avalluri explained that this is to avoid expensive invocation of LVM commands and parsing of their output.

Let's continue the discussion whether that caching is worth the extra complexity. It already was the source of one bug (the one fixed by https://github.com/intel/pmem-csi/pull/408).

Do we have data how expensive the LVM command invocation+parsing really is?

okartau:
I repeat my reasoning here which I posted earlier in a PR
Currently we create entries recording that Volume when a Volume is created (in LVM mode):

0. LVM volume itself
1. cache entry in lvm.devices
2. entry in Node Volume registry
3. entry in Controller Volume registry

Node, Controller, and k8s (and possibly even more, some sidecars) are exchanging messages.
With rapid RPC, messages get repeated, delayed, lost, mutexes (in multiple players and layers) are waited (sometimes for long time during larger-volume shred).
PMEM-CSI will face so many combinations of possible errors that it is impossible to test everything and in long run those 4 entities may get out of sync. 
Debugging those problems is difficult/impossible without detailed logs (which are not enabled in production). That's why I propose to look for ways to reduce the amount of state we maintain.
If a LVM operation takes some (minor) extra time, it is no big problem, as provisioning of Volume is not generally time-critical operation. My idea is to prioritize reliability over performance.

avalluri:
> Let's continue the discussion whether that caching is worth the extra complexity.

I don't agree, it adds any extra complexity :), It's just one more data structure where we add/delete values to/from it. One can argue that it needs mutex locking but, all LVM calls are already synchronous.

> Do we have data how expensive the LVM command invocation+parsing really is?

No, I would see this more like a generic programming design, where a simple(inexpensive) data structure could solve than context switched exec call.

avalluri:
> I repeat my reasoning here which I posted earlier in a PR
> Currently we create entries recording that Volume when a Volume is created (in LVM mode):
> 
> 1. LVM volume itself
> 2. cache entry in lvm.devices
> 3. entry in Node Volume registry
> 4. entry in Controller Volume registry
> 
[...]

Issue #402, opened 2019-09-11
DeleteVolume response is incorrect if volume is in use


okartau:
CSI spec says that if a volume is in use, DeleteVolume response should be error=FAILED_PRECONDITION.
Currently, PMEM-CSI responds with codes.InternalError in all cases when lower layer (Device) returns any error, although the BUSY status is detected and is one of the reasons to fail.
To make the GRPC error correct, we have to start returning more detailed error codes by cs.dm.DeleteDevice and form a RPC return code correctly if volume/device was in-use.

okartau:
The fact that current PMEM-CSI code passed e2e and sanity tests, is a sign that this scenario is not tested.

okartau:
Changes of #420 fixed that, it seems, @avalluri do you agree? Now NodeController returns FailedPrecondition if volume is in use.
In that case, I think we can close this issue.

avalluri:
> Changes of #420 fixed that, it seems, @avalluri do you agree? Now NodeController returns FailedPrecondition if volume is in use.
> In that case, I think we can close this issue.

Yes, we can close this.

PR #420, opened 2019-09-25
Devicemanager: New error codes + Device Manager unit tests and few other fixes.

avalluri:
This PR address a couple of issues:

- Defined new error codes that are returned by DeviceManager interface methods. Made changes to implementations to use these error codes and revised the error messages using error formatting features provided by go-1.13 `fmt` package.

- Discussion on issue #413, raised a requirement that PMEM-CSI should handle the DeleteVolume() call for volumes that the underlined device already deleted manually by the admin.  So for such a volume we just ignore and return no error in DeviceManager.DeleteVolume(). This also makes DeleteVolume() idempotent.

- Issue #402 : CSI.DeleteVolume should return FailedPreCondition error in case of the volume is in use. Test case added for the same.

- Unit tests for device manager

- This PR also contains a commit that removes the unused code in DeviceManager.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -159,6 +159,26 @@ pipeline {
             }
         }
 
+        stage('make dm-test') {
+            options {
+                timeout(time: 30, unit: "MINUTES")
+            }
+
+            steps {
+                sh "docker run --rm ${DockerBuildArgs()} \
+                               --privileged=true \
+                               -e TEST_CHECK_SIGNED_FILES=false \
+                               -e TEST_DISTRO=clear \
+                               -e TEST_DISTRO_VERSION=${env.CLEAR_LINUX_VERSION_1_15} \
+                               -v `pwd`:`pwd`:rshared \
+                               -w `pwd` \
+                               ${env.BUILD_IMAGE} bash -c 'set -x; \
+                                   swupd bundle-add openssh-client && \
+                                   make run_dm_tests; \
+                                   make stop'"
+            }
+        }
+
         stage('Build test image') {
             options {
                 timeout(time: 60, unit: "MINUTES")
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -93,7 +93,9 @@ func (r *Region) ActiveNamespaces() []*Namespace {
 	return r.namespaces(true)
 }
 
-//AllNamespaces returns all namespaces in the region
+//AllNamespaces returns all non-zero sized namespaces in the region
+//as sometime a deleted namespace also lies around with size zero, we can ignore
+//such namespace
 func (r *Region) AllNamespaces() []*Namespace {
 	return r.namespaces(false)
 }
@@ -318,8 +320,15 @@ func (r *Region) namespaces(onlyActive bool) []*Namespace {
 	ndr := (*C.struct_ndctl_region)(r)
 
 	for ndns := C.ndctl_namespace_get_first(ndr); ndns != nil; ndns = C.ndctl_namespace_get_next(ndns) {
-		if !onlyActive || C.ndctl_namespace_is_active(ndns) == true {
-			namespaces = append(namespaces, (*Namespace)(ndns))
+		ns := (*Namespace)(ndns)
+		// If asked for only active namespaces return it regardless of it size
+		// if not, return only valid namespaces, i.e, non-zero sized.
+		if onlyActive {
+			if ns.Active() {
+				namespaces = append(namespaces, ns)
+			}
+		} else if ns.Size() > 0 {
+			namespaces = append(namespaces, ns)
 		}
 	}
 
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -131,11 +131,11 @@ func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) er
 				if err != nil {
 					return err
 				}
-				err = WaitDeviceAppears(device)
+				err = waitDeviceAppears(device)
 				if err != nil {
 					return err
 				}
-				err = ClearDevice(device, false)
+				err = clearDevice(device, false)
 				if err != nil {
 					return err
 				}
@@ -157,7 +157,7 @@ func (lvm *pmemLvm) DeleteDevice(volumeId string, flush bool) error {
 	if err != nil {
 		return err
 	}
-	if err := ClearDevice(device, flush); err != nil {
+	if err := clearDevice(device, flush); err != nil {
 		return err
 	}
 
@@ -169,18 +169,6 @@ func (lvm *pmemLvm) DeleteDevice(volumeId string, flush bool) error {
 	delete(lvm.devices, volumeId)
 
 	return nil
-}
-
-func (lvm *pmemLvm) FlushDeviceData(volumeId string) error {
-	lvmMutex.Lock()
-	defer lvmMutex.Unlock()
-
-	device, err := lvm.getDevice(volumeId)
-	if err != nil {
-		return err
-	}
-
-	return ClearDevice(device, true)
 }
 
 func (lvm *pmemLvm) ListDevices() ([]*PmemDeviceInfo, error) {
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -60,6 +60,10 @@ func NewPmemDeviceManagerLVM() (PmemDeviceManager, error) {
 	}
 	ctx.Free()
 
+	return NewPmemDeviceManagerLVMForVGs(volumeGroups)
+}
+
+func NewPmemDeviceManagerLVMForVGs(volumeGroups []string) (PmemDeviceManager, error) {
 	devices, err := listDevices(volumeGroups...)
 	if err != nil {
 		return nil, err
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -2,14 +2,12 @@ package pmdmanager
 
 import (
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
 	"sync"
 
 	"github.com/intel/pmem-csi/pkg/ndctl"
 	pmemexec "github.com/intel/pmem-csi/pkg/pmem-exec"
-	"github.com/pkg/errors"
 	"k8s.io/klog"
 )
 
@@ -43,7 +41,7 @@ func NewPmemDeviceManagerLVM() (PmemDeviceManager, error) {
 
 	ctx, err := ndctl.NewContext()
 	if err != nil {
-		return nil, fmt.Errorf("Failed to initialize pmem context: %s", err.Error())
+		return nil, err
 	}
 	volumeGroups := []string{}
 	for _, bus := range ctx.GetBuses() {
@@ -88,7 +86,7 @@ func (lvm *pmemLvm) GetCapacity() (map[string]uint64, error) {
 // nsmode is expected to be either "fsdax" or "sector"
 func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) error {
 	if nsmode != string(ndctl.FsdaxMode) && nsmode != string(ndctl.SectorMode) {
-		return fmt.Errorf("Unknown nsmode(%v)", nsmode)
+		return fmt.Errorf("unknown namespace mode %q: %w", nsmode, ErrInvalid)
 	}
 	lvmMutex.Lock()
 	defer lvmMutex.Unlock()
@@ -97,9 +95,8 @@ func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) er
 	// this function is asked to create new devices repeatedly, forcing running out of space.
 	// Avoid device filling with garbage entries by returning error.
 	// Overall, no point having more than one namespace with same volumeId.
-	_, err := lvm.getDevice(volumeId)
-	if err == nil {
-		return fmt.Errorf("CreateDevice: Failed: volume with that name '%s' exists", volumeId)
+	if _, err := lvm.getDevice(volumeId); err == nil {
+		return ErrDeviceExists
 	}
 	vgs, err := getVolumeGroups(lvm.volumeGroups, nsmode)
 	if err != nil {
@@ -131,13 +128,11 @@ func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) er
 				if err != nil {
 					return err
 				}
-				err = waitDeviceAppears(device)
-				if err != nil {
+				if err := waitDeviceAppears(device); err != nil {
 					return err
 				}
-				err = clearDevice(device, false)
-				if err != nil {
-					return err
+				if err := clearDevice(device, false); err != nil {
+					return fmt.Errorf("clear device %q: %v", volumeId, err)
 				}
 
 				lvm.devices[device.VolumeId] = device
@@ -146,7 +141,7 @@ func (lvm *pmemLvm) CreateDevice(volumeId string, size uint64, nsmode string) er
 			}
 		}
 	}
-	return fmt.Errorf("No region is having enough space required(%v)", size)
+	return ErrNotEnoughSpace
 }
 
 func (lvm *pmemLvm) DeleteDevice(volumeId string, flush bool) error {
@@ -195,7 +190,7 @@ func (lvm *pmemLvm) getDevice(volumeId string) (*PmemDeviceInfo, error) {
 		return dev, nil
 	}
 
-	return nil, errors.Wrapf(os.ErrNotExist, "no device found with id '%s'", volumeId)
+	return nil, ErrDeviceNotFound
 }
 
 func getUncachedDevice(volumeId string, volumeGroup string) (*PmemDeviceInfo, error) {
@@ -207,7 +202,8 @@ func getUncachedDevice(volumeId string, volumeGroup string) (*PmemDeviceInfo, er
 	if dev, ok := devices[volumeId]; ok {
 		return dev, nil
 	}
-	return nil, errors.Wrapf(os.ErrNotExist, "no device found with id '%s' in group '%s'", volumeId, volumeGroup)
+
+	return nil, ErrDeviceNotFound
 }
 
 // listDevices Lists available logical devices in given volume groups
@@ -215,9 +211,9 @@ func listDevices(volumeGroups ...string) (map[string]*PmemDeviceInfo, error) {
 	args := append(lvsArgs, volumeGroups...)
 	output, err := pmemexec.RunCommand("lvs", args...)
 	if err != nil {
-		return nil, fmt.Errorf("list volumes failed : %s(lvs output: %s)", err.Error(), output)
+		return nil, fmt.Errorf("lvs failure : %v", err)
 	}
-	return parseLVSOuput(output)
+	return parseLVSOutput(output)
 }
 
 func vgName(bus *ndctl.Bus, region *ndctl.Region, nsmode ndctl.NamespaceMode) string {
@@ -225,9 +221,9 @@ func vgName(bus *ndctl.Bus, region *ndctl.Region, nsmode ndctl.NamespaceMode) st
 }
 
 //lvs options "lv_name,lv_path,lv_size,lv_free"
-func parseLVSOuput(output string) (map[string]*PmemDeviceInfo, error) {
+func parseLVSOutput(output string) (map[string]*PmemDeviceInfo, error) {
 	devices := map[string]*PmemDeviceInfo{}
-	lines := strings.Split(string(output), "\n")
+	lines := strings.Split(output, "\n")
 	for _, line := range lines {
 		fields := strings.Fields(strings.TrimSpace(line))
 		if len(fields) != 3 {
@@ -269,12 +265,12 @@ func getVolumeGroups(groups []string, wantedTag string) ([]vgInfo, error) {
 	args := append(vgsArgs, groups...)
 	output, err := pmemexec.RunCommand("vgs", args...)
 	if err != nil {
-		return vgs, fmt.Errorf("vgs failure: %s", err.Error())
+		return vgs, fmt.Errorf("vgs failure: %v", err)
 	}
 	for _, line := range strings.SplitN(output, "\n", len(groups)) {
 		fields := strings.Fields(strings.TrimSpace(line))
 		if len(fields) != 4 {
-			return vgs, fmt.Errorf("Failed to parse vgs output line: %s", line)
+			return vgs, fmt.Errorf("failed to parse vgs output: %q", line)
 		}
 		tag := fields[3]
 		if tag == wantedTag {
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -1,6 +1,7 @@
 package pmdmanager
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -148,11 +149,21 @@ func (lvm *pmemLvm) DeleteDevice(volumeId string, flush bool) error {
 	lvmMutex.Lock()
 	defer lvmMutex.Unlock()
 
-	device, err := lvm.getDevice(volumeId)
-	if err != nil {
+	var err error
+	var device *PmemDeviceInfo
+
+	if device, err = lvm.getDevice(volumeId); err != nil {
+		if errors.Is(err, ErrDeviceNotFound) {
+			return nil
+		}
 		return err
 	}
 	if err := clearDevice(device, flush); err != nil {
+		if errors.Is(err, ErrDeviceNotFound) {
+			// Remove device from cache
+			delete(lvm.devices, volumeId)
+			return nil
+		}
 		return err
 	}
 
--- a/pkg/pmem-device-manager/pmd-manager.go
+++ b/pkg/pmem-device-manager/pmd-manager.go
@@ -25,9 +25,6 @@ type PmemDeviceManager interface {
 	// If 'flush' is 'true', then the device data is zerod beofore deleting the device
 	DeleteDevice(name string, flush bool) error
 
-	//FlushDeviceData zeros all blocks in the blocke device with given name
-	FlushDeviceData(name string) error
-
 	//ListDevices returns all the block devices information that was created by this device manager
 	ListDevices() ([]*PmemDeviceInfo, error)
 }
--- a/pkg/pmem-device-manager/pmd-manager.go
+++ b/pkg/pmem-device-manager/pmd-manager.go
@@ -1,5 +1,33 @@
 package pmdmanager
 
+import (
+	"errors"
+	"os"
+)
+
+var (
+	// ErrInvalid invalid argument passed
+	ErrInvalid = os.ErrInvalid
+
+	// ErrPermission no permission to complete the task
+	ErrPermission = os.ErrPermission
+
+	// ErrDeviceExists device with given id already exists
+	ErrDeviceExists = errors.New("device exists")
+
+	// ErrDeviceNotFound device does not exists
+	ErrDeviceNotFound = errors.New("device not found")
+
+	// ErrDeviceInUse device is in use
+	ErrDeviceInUse = errors.New("device in use")
+
+	// ErrDeviceNotReady device not ready yet
+	ErrDeviceNotReady = errors.New("device not ready")
+
+	// ErrNotEnoughSpace no space to create the device
+	ErrNotEnoughSpace = errors.New("not enough space")
+)
+
 //PmemDeviceInfo represents a block device
 type PmemDeviceInfo struct {
 	//VolumeId is name of the block device
@@ -12,19 +40,22 @@ type PmemDeviceInfo struct {
 
 //PmemDeviceManager interface to manage the PMEM block devices
 type PmemDeviceManager interface {
-	//GetCapacity returns the available maximum capacity that can be assigned to a Device/Volume
+	// GetCapacity returns the available maximum capacity that can be assigned to a Device/Volume
 	GetCapacity() (map[string]uint64, error)
 
-	//CreateDevice creates a new block device with give name, size and namespace mode
+	// CreateDevice creates a new block device with give name, size and namespace mode
+	// Possible errors: ErrNotEnoughSpace, ErrInvalid, ErrDeviceExists
 	CreateDevice(name string, size uint64, nsmode string) error
 
-	//GetDevice returns the block device information for given name
+	// GetDevice returns the block device information for given name
+	// Possible errors: ErrDeviceNotFound
 	GetDevice(name string) (*PmemDeviceInfo, error)
 
-	//DeleteDevice deletes an existing block device with give name.
-	// If 'flush' is 'true', then the device data is zerod beofore deleting the device
+	// DeleteDevice deletes an existing block device with give name.
+	// If 'flush' is 'true', then the device data is zeroed before deleting the device
+	// Possible errors: ErrDeviceInUse, ErrPermission
 	DeleteDevice(name string, flush bool) error
 
-	//ListDevices returns all the block devices information that was created by this device manager
+	// ListDevices returns all the block devices information that was created by this device manager
 	ListDevices() ([]*PmemDeviceInfo, error)
 }
--- a/pkg/pmem-device-manager/pmd-ndctl.go
+++ b/pkg/pmem-device-manager/pmd-ndctl.go
@@ -1,6 +1,7 @@
 package pmdmanager
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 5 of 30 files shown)


avalluri:
> Can you add tests for this, both for direct mode and LVM mode?

Tests for device manager is bit challenging, as they deal with devices. I stared with LVM tests: https://github.com/avalluri/pmem-CSI/commit/c7437e58140a65d6a9a6cd62015a8a9fbbdc6bd8 .
The ugly part is how we run the test, as it needs sudo permissions. I tried running test inside the container and ended up like this:https://github.com/avalluri/pmem-CSI/commit/c7437e58140a65d6a9a6cd62015a8a9fbbdc6bd8#diff-436cc358f7f160c8242374b3f820867d

avalluri:
> Can you add tests for this, both for direct mode and LVM mode?

@pohly, I added unit tests for LVM, but I am not sure yet about covering ndctl manager, which needs a bit thinking how to mimic the pmem device.

pkg/pmem-device-manager/pmme-lvm_test.go
```diff
@@ -0,0 +1,229 @@
+/*
```
okartau: name of that file seems misspelled, should be pmem-lvm_test.go ?

pkg/pmem-device-manager/pmme-lvm_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+	defer func() {
+		if err != nil {
+			By("losetup.Detach")
+			dev.Detach() // nolint errcheck
+		}
+	}()
+
+	By("Creating volume group")
+	// TODO: resuse vgm code
```
okartau: reuse?

vendor-bom.csv
```diff
@@ -52,6 +52,7 @@ golang.org/x/time;golang time;BSD 3-clause "New" or "Revised" License
 google.golang.org/appengine;Golang appengine;Apache License 2.0
 google.golang.org/genproto;Google go-genproto;Apache License 2.0
 google.golang.org/grpc;grpc grpc-go;Apache License 2.0
+gopkg.in/freddierice/go-losetup.v1; Go losetup; MIt License
```
okartau: MIT License ?

pkg/pmem-device-manager/pmme-lvm_test.go
avalluri: Yes, it should be `pmd-lvm_test.go`. I fixed it now.

pkg/pmem-device-manager/pmme-lvm_test.go
avalluri: Fixed

avalluri:
Device manager tets faile in CI while setting up loop device:
```
[2019-10-24T07:52:16.650Z] DeviceManage LVM
[2019-10-24T07:52:16.650Z] /go/src/github.com/intel/pmem-csi/pkg/pmem-device-manager/pmd-lvm_test.go:35
[2019-10-24T07:52:16.650Z]   LVM [BeforeEach]
[2019-10-24T07:52:16.650Z]   /go/src/github.com/intel/pmem-csi/pkg/pmem-device-manager/pmd-lvm_test.go:54
[2019-10-24T07:52:16.650Z]     Should create a new device
[2019-10-24T07:52:16.650Z]     /go/src/github.com/intel/pmem-csi/pkg/pmem-device-manager/pmd-lvm_test.go:55
[2019-10-24T07:52:16.650Z] 
[2019-10-24T07:52:16.650Z]     Failed to create volume group
[2019-10-24T07:52:16.650Z]     Expected
[2019-10-24T07:52:16.650Z]         <*errors.errorString | 0xc000021700>: {
[2019-10-24T07:52:16.650Z]             s: "could not open /dev/loop-control: open /dev/loop-control: no such file or directory",
[2019-10-24T07:52:16.650Z]         }
[2019-10-24T07:52:16.650Z]     to be nil
[2019-10-24T07:52:16.650Z] 
[2019-10-24T07:52:16.650Z]     /go/src/github.com/intel/pmem-csi/pkg/pmem-device-manager/pmd-lvm_test.go:42
```

Makefile
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+GO=GOOS=linux GO111MODULE=off go
```
pohly: Why GOOS=linux?

avalluri: > Why GOOS=linux?
This is not added part of this change. It was existing before, I just moved from that place.

pohly: We weren't consistent then. I was looking at the first changed line and that only had:
```
build: $(CMDS) $(TEST_CMDS)
	go test -run none ./pkg/... ./test/e2e
```

Dockerfile
```diff
@@ -18,7 +18,7 @@ ARG SWUPD_UPDATE_ARG
 ARG NDCTL_VERSION="66"
 ARG NDCTL_CONFIGFLAGS="--disable-docs --without-systemd --without-bash"
 ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
-ARG GO_VERSION="1.12.9"
+ARG GO_VERSION="1.13"
```
pohly: Isn't this unrelated to this PR?

pohly: Never mind. Now I remember that you wanted to use Go 1.13 for error handling. Please call that out in the PR subject and description, because it is a major change by itself.

test/test.make
```diff
@@ -128,12 +129,17 @@ test_e2e: start
 
 # Execute simple unit tests.
 .PHONY: run_tests
-test: run_tests
+test: run_tests run_device_manager_tests
 RUN_TESTS = TEST_WORK=$(abspath _work) \
-	$(TEST_CMD) $(shell go list $(TEST_ARGS) | sed -e 's;$(IMPORT_PATH);.;')
+	$(TEST_CMD) $(shell $(GO) list $(TEST_ARGS) | grep -v pmem-device-manager | sed -e 's;$(IMPORT_PATH);.;')
+RUN_DM_TESTS = \
+    $(TEST_CMD) ./pkg/pmem-device-manager -exec $(SUDO)
```
pohly: This makes "make test" depend on sudo, which is not currently a requirement and also shouldn't be. Developers should be able to build and test on a machine with a user account that isn't allowed to become root.

The other issue is that the test depends on `/dev/loop-control`, right?

Let me propose a different approach to running tests that need to be run in a certain environment:
- in the test code, check those preconditions (are root, have /dev/loop-control): if not met, skip the test
- integrate the running of unit tests into our E2E testing:
  - build test binaries with `go test -c`
  - copy test binaries into the first worker
  - run them there as root

This is more complicated, but it also gives us better test coverage, because then operations that may depend on the OS kernel of the cluster node actually get tested with that kernel. We can also do unit tests that depend on actual PMEM.

Dockerfile
pohly: Actually, it probably would be better to do the Go update in a separate PR.

test/test.make
avalluri: > This makes "make test" depend on sudo, which is not currently a requirement and also shouldn't be. Developers should be able to build and test on a machine with a user account that isn't allowed to become root.
> 

To avoid this root user. I initially tried ruining these tests in a container (https://github.com/intel/pmem-csi/pull/420#issuecomment-539984710).

> The other issue is that the test depends on `/dev/loop-control`, right?
> 
> Let me propose a different approach to running tests that need to be run in a certain environment:
> 
> * in the test code, check those preconditions (are root, have /dev/loop-control): if not met, skip the test
> * integrate the running of unit tests into our E2E testing:
>   
>   * build test binaries with `go test -c`
>   * copy test binaries into the first worker
>   * run them there as root
> 
> This is more complicated, but it also gives us better test coverage, because then operations that may depend on the OS kernel of the cluster node actually get tested with that kernel. We can also do unit tests that depend on actual PMEM.

Ok, Instead of running these(device manager) unit tests in full(3-node) test cluster, I would propose to run them in a single virtual machine node. I pushed changes to support this: https://github.com/intel/pmem-csi/pull/420/commits/45f21a6507f69173e57e360488fb520e2ae32939

avalluri: This single node vm installs only dependencies needed for PMEM, such as LVM2, ndctl. No other packages like Docker, Kubernetes.

pohly: > Ok, Instead of running these(device manager) unit tests in full(3-node) test cluster, I would propose to run them in a single virtual machine node.

That's a massive change to some already pretty complex scripts. What's the advantage compared to just using the normal VM cluster?

We need to consider two different use cases:
- a developer using this while making changes to the source code
- the CI jobs

Jenkinsfile
```diff
@@ -158,6 +158,20 @@ pipeline {
             }
         }
 
+        stage('make dm-test') {
```
pohly: Is this an experiment to figure out how long this stage will take?

avalluri: That is one reason, and the other is to run the docker with  `--privileged=true`  to run dm tests.

avalluri: @pohly do you have any clue why `ssh-keygen` is failing at this stage:

```
2019-11-08T12:10:18.773Z] + '[' '!' -e /go/src/github.com/intel/pmem-csi/_work/resources/id_rsa ']'

[2019-11-08T12:10:18.773Z] + ssh-keygen -N '' -f /go/src/github.com/intel/pmem-csi/_work/resources/id_rsa

[2019-11-08T12:10:18.773Z] + die 'failed to create /go/src/github.com/intel/pmem-csi/_work/resources/id_rsa'

[2019-11-08T12:10:18.773Z] + echo 'ERROR: failed to create /go/src/github.com/intel/pmem-csi/_work/resources/id_rsa'

[2019-11-08T12:10:18.773Z] ERROR: failed to create /go/src/github.com/intel/pmem-csi/_work/resources/id_rsa

[2019-11-08T12:10:18.773Z] + exit 1
```

pohly: Does `_work/resources` exist?

avalluri: It should be created by the previous command line, which does not throw any error. So I assume it exists.

avalluri: When printing `ssh-keygen` output revealed that `ssh-keygen` command not found. https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-420/20/pipeline#step-63-log-114

avalluri: > Is this an experiment to figure out how long this stage will take?

@pohly  Running device manager tests in VM took around 3 and half minutes in CI.
 **make dm-test - 3m 23s**

avalluri:
This PR is based on go-1.13 changes made in #461.

Dockerfile
avalluri: Now this change is in separate PR #461

test/test.make
avalluri: > That's a massive change to some already pretty complex scripts.

No, it's not that massive change as it looks in the PR. Most of the changes are indentations. It's just replacing few bits in the right place in `test/start-kubernetes.sh` script.

pohly:
@avalluri do you want this to be merged as-is or are you looking for further feedback?

avalluri:
> @avalluri do you want this to be merged as-is or are you looking for further feedback?

It is ready for review, would good to have a review round. other than commits covered by #461.

avalluri:
> In PMEM-CSI, we have to make targets for testing:
> 
> * "make test" - quick testing, works with just Go installed
> * "make test_e2e" - slow, thorough testing with additional dependencies on a virtual cluster
> 
I see these two as: `make test` is to run simple & quick tests which also covers unit tests, and `make test_e2e` is complete full-fledged end-2-end tests, which are expected to consume more time and compute resource.  

We cannot fit device manager tests into either of these two categories, as by nature those are unit tests, limited to `pmdmanager` package, and needs some extra setup which consumes time. So I see it as a special case. And as a developer, I expect to run these tests only when my changes touch the device manager source.

Moreover e2e test depends on different Kubernetes versions, and building docker images etc., which are not at all required to run unit tests. Considering all these points I came up with this new make target.
 
> That way, running the test will be faster because we don't need to bring up another VM. Of the roughly 3min for the new step in the CI, most of that time is spent on creating that VM. We can make the CI runs faster by not having that separate step.

We are negotiating for `3m 24s` CI time in total of `2h 10m 1s` :-)

pohly:
> We cannot fit device manager tests into either of these two categories, as by nature those are unit tests, limited to pmdmanager package, and needs some extra setup which consumes time. So I see it as a special case. And as a developer, I expect to run these tests only when my changes touch the device manager source.

So we have to agree to disagree here. My usage pattern is different than yours, therefore we need two different ways of running these tests.

@okartau: what do you think? Should we have two test targets for running all tests or add more special targets on a case-by-case basis?

> Moreover, e2e test depends on different Kubernetes versions, and building docker images etc., which are not at all required to run unit tests.

True, running these tests multiple times will be somewhat redundant. But I'd rather run a test a bit too often rather than not often enough, as long as it runs quickly (and without the VM setup, this tests is fast). For example, this test goes beyond normal unit testing because it interacts with the host kernel. We may have more of those tests going forward. Therefore it is useful to run it at least once on Fedora and once on Clear. The single CI step doesn't achieve that.

> We are negotiating for 3m 24s CI time in total of 2h 10m 1s :-)

Death of a thousand paper cuts. If we don't pay attention, we'll be at 3h soon. Also, testing on every Linux base distro will multiply that time.

If you want, we can merge this PR as it is and then I'll change the CI so that it runs faster again.

okartau:
I vote for flexibility. Adding more test targets does not force anybody to use those, so with added targets one can still keep previous use habits.
'make test' should be quick, I use it for smoke testing before commit, so this new test clearly should not go there.
I vote for keeping make test as it is, adding a new target for new "mid duration" unit tests, plus also 
embedding that new target into long e2e testing.
Another thread could be to make CI flow more parallel, with aim to shorten the total time.

avalluri:
> plus also embedding that new target into long e2e testing.

With the current change, this new target runs in a new stage in CI whereas  E2E tests are run in different Kubernetes versions and device modes. Embedding this new target in e2e testing is not required in my opinion.

> Another thread could be to make CI flow more parallel, with aim to shorten the total time.

Yes, few CI stages like testing device modes and Kubernetes versions could run in parallel.

okartau:
> runs in a new stage in CI

yes, thats what I meant, that it gets run in CI. I was not correct in wording.

pohly:
> Yes, few CI stages like testing device modes and Kubernetes versions could run in parallel.

We can't run on the same host in parallel, I tried and this caused timeouts. The host isn't capable enough. We could ask for multiple hosts to run on, but then we might get additional delays for booting up those hosts and need to copy artifacts and source around, so it will become complicated.

So here's a proposal:
- merge https://github.com/intel/pmem-csi/pull/461
- rebase this PR and merge it
- I'll make running the new tests more flexible so that overall test time is lower when running all tests.

avalluri:
> * rebase this PR and merge it

Now I rebased this PR on #461.

pkg/pmem-device-manager/pmd-util.go
```diff
@@ -33,17 +33,20 @@ func clearDevice(dev *PmemDeviceInfo, flush bool) error {
...
+
+	fd, err := unix.Open(dev.Path, unix.O_RDONLY|unix.O_EXCL|unix.O_CLOEXEC, 0)
+	defer unix.Close(fd)
+
 	if err != nil {
-		return err
-	}
-	if devOpen {
-		return fmt.Errorf("%s is in use", dev.Path)
+		return fmt.Errorf("failed to clear device %q as %w", dev.Path, ErrDeviceInUse)
```
pohly: The Go idiom is: "failed to clear device %q: %w".

That way, the resulting error string makes sense regardless of what text the error evaluates to.

pkg/pmem-device-manager/pmd-util.go
```diff
@@ -75,5 +78,5 @@ func waitDeviceAppears(dev *PmemDeviceInfo) error {
 			i, dev.Path, retryStatTimeout)
 		time.Sleep(retryStatTimeout)
 	}
-	return fmt.Errorf("device %s did not appear after multiple retries", dev.Path)
+	return fmt.Errorf("%w(%s)", ErrDeviceNotReady, dev.Path)
```
pohly: The Go idiom is: "%s: %w" (https://blog.golang.org/go1.13-errors).

pkg/pmem-device-manager/pmd-manager.go
```diff
@@ -1,5 +1,36 @@
...
+	ErrPermission = os.ErrPermission
+
+	// ErrDeviceExists device with given id already exists
+	ErrDeviceExists = errors.New("device exists")
+
+	// ErrDeviceNotFound device does not exists
+	ErrDeviceNotFound = errors.New("device not found")
+
+	// ErrDeviceInUse device is in use
+	ErrDeviceInUse = errors.New("device is in use")
```
pohly: "device in use"? Shorter, consistent with "device not found" and "device not ready" (i.e. leave out verb which isn't needed).

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -43,7 +41,7 @@ func NewPmemDeviceManagerLVM() (PmemDeviceManager, error) {
 
 	ctx, err := ndctl.NewContext()
 	if err != nil {
-		return nil, fmt.Errorf("Failed to initialize pmem context: %s", err.Error())
+		return nil, fmt.Errorf("%w occurred while initialize pmem context: %s", ErrUnknown, err.Error())
```
pohly: All of the new `fmt.Errorf` should follow the Go idiom of `<description>: <inner error>`. We don't need full sentences either.

For example, in this case `"pmem context: %w: %v", ErrUnknown, err` is better.

Note also the use of `%v` and `err` instead of `%s` and `err.Error()`: it's the idiomatic approach and shorter.

I'm not going to mark further occurrences of this. Please check the diff and update.

pkg/pmem-device-manager/pmd-lvm_test.go
```diff
@@ -0,0 +1,339 @@
...
+		// Delete call should not return any error on non-existing device
+		err = dm.DeleteDevice(name, true)
+		Expect(err).Should(BeNil(), "DeleteDevice() is not idempotent")
+	})
+	//	})
+})
+
+func precheck() {
+	if os.Geteuid() != 0 {
+		Skip("Required root used privilege to these tests", 1)
```
pohly: "root privileges are required"

test/test.make
```diff
@@ -137,7 +137,8 @@ test_e2e: start
 .PHONY: run_tests
 test: run_tests
 RUN_TESTS = TEST_WORK=$(abspath _work) \
-	$(TEST_CMD) $(shell $(GO) list $(TEST_ARGS) | sed -e 's;$(IMPORT_PATH);.;')
+	$(TEST_CMD) $(shell $(GO) list $(TEST_ARGS) | grep -v pmem-device-manager | sed -e 's;$(IMPORT_PATH);.;')
```
pohly: This is a potential trap: if we ever add unit tests for `pmem-device-manager` and forget to remove this, they won't run. It also prevents us from running them via "make test" when they can run (for example, with "sudo make test").

I'd prefer to let it run and rely on the skip check.

test/setup-fedora-govm.sh
```diff
@@ -54,14 +59,15 @@ EOF
 # For the sake of reproducibility, use fixed versions.
```
pohly: Comment no longer aligned with code.

test/setup-fedora-govm.sh
```diff
@@ -54,14 +59,15 @@ EOF
...
+        1.13) packages+=" kubelet-1.13.9-0 kubeadm-1.13.9-0 kubectl-1.13.9-0";;
+        1.14) packages+=" kubelet-1.14.7-0 kubeadm-1.14.7-0 kubectl-1.14.7-0";;
+        1.15) packages+=" kubelet-1.15.4-0 kubeadm-1.15.4-0 kubectl-1.15.4-0";;
+        1.16) packages+=" kubelet-1.16.0-0 kubeadm-1.16.0-0 kubectl-1.16.0-0";;
+        *) echo >&2 "Kubernetes version ${TEST_KUBERNETES_VERSION} not supported, package list in $0 must be updated."; exit 1;;
+    esac
+    packages+=" --disableexcludes=kubernetes"
+fi
 
 # Sometimes we hit a bad mirror and get "Failed to synchronize cache for repo ...".
```
pohly: Same here.

test/test.make
avalluri: > I'd prefer to let it run and rely on the skip check.

In this case, when run using `make tests` there will be no information for the user if the tests really run are skipped. Both the cases it shows:

`ok  	github.com/intel/pmem-csi/pkg/pmem-device-manager	0.008s`

test/setup-fedora-govm.sh
avalluri: You mean with this code change?. This change is only touching indentation.

pkg/pmem-device-manager/pmd-lvm_test.go
avalluri: Done. Rephrased the text.

pkg/pmem-device-manager/pmd-lvm.go
avalluri: Done. Rephrased all error messages as suggested.

pkg/pmem-device-manager/pmd-util.go
avalluri: Done.

pkg/pmem-device-manager/pmd-util.go
avalluri: Done.

pkg/pmem-device-manager/pmd-manager.go
avalluri: Done.

test/test.make
pohly: I don't agree, but have it your way.

test/setup-fedora-govm.sh
pohly: Yes, and the change doesn't change the indention of this comment. It should also indent the comment further to the right such that the `#` is in the same column as the code again.

avalluri: oops, I misunderstood the word 'aligned'. I will fix it.

avalluri: Fixed it now.

test/setup-fedora-govm.sh
avalluri: This is out of `If` block. I think it needs no further indentation.

test/setup-clear-govm.sh
```diff
@@ -54,100 +59,112 @@ EOF
...
-    # Reconfiguration done, start daemons. Starting kubelet must wait until kubeadm has created
-    # the necessary config files.
-    systemctl daemon-reload
-    systemctl restart $cri_daemon || (
-        systemctl status $cri_daemon || true
-        journalctl -xe || true
-        false
-    )
-    systemctl enable $cri_daemon kubelet
+function init_region(){
```
pohly: Is there a reason why this is a function and "init kubernetes" isn't? That seems inconsistent.

Just wondering, don't bother changing it.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -126,9 +124,8 @@ func (pmem *pmemNdctl) CreateDevice(volumeId string, size uint64, nsmode string)
 	if err != nil {
 		return err
 	}
-	err = ClearDevice(device, false)
-	if err != nil {
-		return err
+	if err := clearDevice(device, false); err != nil {
+		return fmt.Errorf("clear device %q: %w : %v", volumeId, ErrUnknown, err)
```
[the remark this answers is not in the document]
avalluri: > Do we need `ErrUnknown`? It doesn't add any additional information here.

I agree we can drop `ErrUnknown` totally as no caller would check if it's an unknown error. I will do the changes.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -181,7 +172,11 @@ func (pmem *pmemNdctl) ListDevices() ([]*PmemDeviceInfo, error) {
 func (pmem *pmemNdctl) getDevice(volumeId string) (*PmemDeviceInfo, error) {
 	ns, err := pmem.ctx.GetNamespaceByName(volumeId)
 	if err != nil {
-		return nil, err
+		e := ErrUnknown
+		if errors.Is(err, ndctl.ErrNotExist) {
```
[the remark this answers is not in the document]
avalluri: May be yes. Added some context to error to be returned: 
         `fmt.Errorf("error getting device %q: %v", volumeId, err)`

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -117,7 +115,7 @@ func (pmem *pmemNdctl) CreateDevice(volumeId string, size uint64, nsmode string)
 		Mode:  ndctl.NamespaceMode(nsmode),
 	})
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %v", ErrUnknown, err)
```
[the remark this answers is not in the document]
avalluri: After checking the code, I felt like the error returned by `CreateNamespace()` is good enough to return directly by `CreateDevice`. And removed the `ErrUnknown` error from device manager error list.

test/setup-clear-govm.sh
avalluri: This inconsistency is the result of moving this part of code from `test/setup-kubernetes.sh`.

pohly: This actually points towards a bigger conceptual issue that I missed before: `init_region` isn't specific to the underlying OS and thus shouldn't be in `setup-fedora-govm.sh` and `setup-clear-govm.sh`. It's not much code, but it's still some code duplication.

Can you come up with a solution that puts `init_region` into a single file?

I don't want to stop you from implementing this, but let me point out the solution that I came up with in the meantime: https://github.com/pohly/pmem-CSI/commit/38af431d2be36db0957d5199da6c92cffdbbb013

Your argument was that `make run_dm_tests` will allow a developer to test changes to the device manager quickly while working on the code. On my desktop, that takes over 4 minutes for each invocation. `make test_e2e TEST_E2E_FOCUS=gotests.*pmem-device-manager` achieves the same in 9 seconds once the VM was created (which only needs to be done once per day).

Do you still think that these changes are necessary?

avalluri: > `make test_e2e TEST_E2E_FOCUS=gotests.*pmem-device-manager` achieves the same in 9 seconds once the VM was created

As a developer now I should remember the new way how I to run the device manager tests, which is as good as running a new make target in my opinion.

If we do not kill the test VM created after running device manager tests assuming that it can be reusable or make it a bit flexible to choosing VM say `make test TEST_VM=worker1` then the current solution also shouldn't consume more time.

avalluri: > This actually points towards a bigger conceptual issue that I missed before: `init_region` isn't specific to the underlying OS and thus shouldn't be in `setup-fedora-govm.sh` and `setup-clear-govm.sh`. It's not much code, but it's still some code duplication.
> 

I myself had the same concern while doing this change, wether to duplicate code or move it a separate file or to post `create_vms()` in `start-kubernetes.sh` script. But chose to duplicate as it's not so big code and assuming that we might not need to touch this in future.

But I agree with you, should avoid code duplication.

> Can you come up with a solution that puts `init_region` into a single file?

Sure, I will work on this.

avalluri: > > Can you come up with a solution that puts `init_region` into a single file?
> 
> Sure, I will work on this.

Fixed this by moving the `init_region()`(renamed to `init_pmme_regions`)  to `start-kubernetes.sh` script and calling it after `create_vms()`.

avalluri: > make it a bit flexible to choosing VM say `make test TEST_VM=worker1` then the current solution also shouldn't consume more time.

Made these changes to `run_dm_test` target, so that one can choose a cluster where these tests should be run. It uses the `pmem-csi-${CLUSTER}-worker1` to run the tests. If this node already set up(say via `make start`) it just copies the binary and runs the tests.

pohly:
`run_dm_tests` fails in the CI: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-420/41/pipeline

pohly:
> run_dm_tests fails in the CI: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-420/41/pipeline

If it works for you, then I don't think we need to investigate this further. As discussed before, I believe it is better to run this test inside the existing steps and I may have that working. See https://github.com/intel/pmem-csi/pull/473 for the arguments and code changes. I'm currently waiting for CI tests results.

pohly:
@avalluri: so what did you change to fix the failure? It's a bit hard to tell because of the rebase. Is it https://github.com/intel/pmem-csi/pull/420/commits/8b22191bdd7cb94ccee776160e8ce8f64880cbd8 and thus https://github.com/intel/pmem-csi/issues/472?

avalluri:
Fix



> @avalluri: so what did you change to fix the failure? It's a bit hard to tell because of the rebase. Is it [8b22191](https://github.com/intel/pmem-csi/commit/8b22191bdd7cb94ccee776160e8ce8f64880cbd8) and thus #472?

I am about mention that :) Yes, This should fix.

avalluri:
> I believe it is better to run this test inside the existing steps

If this what you feel right, we can still run using `make run_dm_tests`  in CI, then it picks the first node in the default cluster to run device manager tets.

avalluri:
@pohly I took your change https://github.com/pohly/pmem-CSI/commit/a9e7136f7a7f2cb1605a62fcef63c456f88a953f and squashed to ff07726d18d97d3c7dfad4ee24697eebf53cd582. Thanks.

avalluri:
> `DestroyNamespace` will first deactivate a namespace, then destroy it. Suppose the driver crashes or gets killed after the first and before the second step. We are left with an inactive namespace that shouldn't be ignored.

Interesting question. Ya, may be true. We cannot ignore that namespace in this case.

avalluri:
> > `DestroyNamespace` will first deactivate a namespace, then destroy it. Suppose the driver crashes or gets killed after the first and before the second step. We are left with an inactive namespace that shouldn't be ignored.
> 
> Interesting question. Ya, may be true. We cannot ignore that namespace in this case.

By the way, we already dealing with only active namespaces for ListDevices, this case affects this also.

avalluri:
There is something weird going on in CI,  `1-16 Direct` tests running for 2h 28m but no results yet. Somehow got stuck. I am not sure if the issue with the code or something else.

pohly:
> Interesting question. Ya, may be true. We cannot ignore that namespace in this case.

So what do you think of the solution I have here: https://github.com/intel/pmem-csi/commit/45f7c42fe7c02981e91fd787a5b73b154aa01c2b#diff-b99dbbc6a60be19c9eacba35ad8a5526

> By the way, we already dealing with only active namespaces for ListDevices, this case affects this also.

What do we need to fix for ListDevices? Should it report inactive, non-empty namespaces and currently ignores them?

pohly:
> There is something weird going on in CI, 1-16 Direct tests running for 2h 28m but no results yet. Somehow got stuck. I am not sure if the issue with the code or something else.

You could try running the E2E tests in that mode locally.

avalluri:
> So what do you think of the solution I have here: [45f7c42#diff-b99dbbc6a60be19c9eacba35ad8a5526](https://github.com/intel/pmem-csi/commit/45f7c42fe7c02981e91fd787a5b73b154aa01c2b#diff-b99dbbc6a60be19c9eacba35ad8a5526)

Should work, I couldn't see any issues with that.

> What do we need to fix for ListDevices? Should it report inactive, non-empty namespaces and currently ignores them?

Yes, it also should have that fix. Maybe we better fix this ndctl layer.

pkg/ndctl/region.go
```diff
@@ -93,7 +93,7 @@ func (r *Region) ActiveNamespaces() []*Namespace {
 	return r.namespaces(true)
 }
 
-//AllNamespaces returns all namespaces in the region
+//AllNamespaces returns all non-zero sized namespaces in the region
```
pohly: Can you add a comment here explaining why it ignores namespaces with zero size? I know its in the commit message, but that's not necessarily where someone will look when reading this description.

pkg/ndctl/region.go
```diff
@@ -318,8 +318,9 @@ func (r *Region) namespaces(onlyActive bool) []*Namespace {
 	ndr := (*C.struct_ndctl_region)(r)
 
 	for ndns := C.ndctl_namespace_get_first(ndr); ndns != nil; ndns = C.ndctl_namespace_get_next(ndns) {
-		if !onlyActive || C.ndctl_namespace_is_active(ndns) == true {
-			namespaces = append(namespaces, (*Namespace)(ndns))
+		ns := (*Namespace)(ndns)
+		if (onlyActive && ns.Active()) || ns.Size() != 0 {
```
pohly: Doesn't this break the semantic of `onlyActive`?

If it is true, then inactive namespaces are returned if their size is > 0, which is not what the caller asked for, is it?

pohly:
@avalluri: Did you also make some more changes to the Jenkinsfile? It's no longer valid: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-420/44/pipeline

pkg/ndctl/region.go
avalluri: Yes, you are right, Fixed it now. Thanks.

pkg/ndctl/region.go
avalluri: Done.

Jenkinsfile
```diff
@@ -159,6 +159,26 @@ pipeline {
...
+                               --privileged=true \
+                               -e TEST_CHECK_SIGNED_FILES=false \
+                               -e TEST_DISTRO=clear \
+                               -e TEST_DISTRO_VERSION=${env.CLEAR_LINUX_VERSION_1_15} \
+                               -v `pwd`:`pwd`:rshared \
+                               -w `pwd` \
+                               ${env.BUILD_IMAGE} bash -c 'set -x; \
+                                   swupd bundle-add openssh-client && \
+                                   make run_dm_tests; \
+                                   make stop'"
```
pohly: This step is now treated as success even if `make run_dm_tests` fails. As I intend to remove this step in https://github.com/intel/pmem-csi/pull/473, we can merge this PR without fixing this.

PR #380, opened 2019-08-07
Adding support for block volumes

avalluri:
Signed-off-by: Amarnath Valluri <amarnath.valluri@intel.com>

PR #421, opened 2019-09-26
TRY: blockvolume without local kubectl support, combining #365 and #380

okartau:
This combines effect of #365 and #380 (plus small extra commit to polish it to get to build). so that we can see possibly running CI tests about block volumes.
In my local test I see one failure
```
Dynamic PV (block volmode)] volumeMode [It] should fail in binding dynamic provisioned PV to PVC
```
where error is expected but we get nil instead.
Let's see does CI hit the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ Kubernetes cluster has been verified on:
 
 | Branch            | Kubernetes branch/version      | Required alpha feature gates   |
 |-------------------|--------------------------------|------------------------------- |
-| devel             | Kubernetes 1.13                | CSINodeInfo, CSIDriverRegistry |
+| devel             | Kubernetes 1.13                | CSINodeInfo, CSIDriverRegistry, CSIBlockVolume |
 | devel             | Kubernetes 1.14                |                                |
 | devel             | Kubernetes 1.15                |                                |
 
@@ -743,6 +743,10 @@ parameters:
          command:
          - /go/bin/pmem-ns-init
 ```
+
+#### Note about raw block volumes
+
+Applications can use volumes provisioned by PMEM-CSI as [raw block devices](https://kubernetes.io/blog/2019/03/07/raw-block-volume-support-to-beta/). For provisioning a PMEM volume as raw block device, one has to create a `PersistentVolumeClaim` with `volumeMode: Block`. See example [PVC](deploy/common/pmem-pvc-block-volume.yaml) and [application](deploy/common/pmem-app-block-volume.yaml) for usage reference.
 
 <!-- FILL TEMPLATE:
 
--- a/deploy/common/pmem-app-block-volume.yaml
+++ b/deploy/common/pmem-app-block-volume.yaml
@@ -1,0 +1,18 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: my-csi-app
+spec:
+  containers:
+    - name: my-frontend
+      image: busybox
+      command: [ "sleep", "100000" ]
+      volumeDevices:
+      - devicePath: "/dev/xpmem"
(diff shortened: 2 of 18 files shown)


avalluri:
> This combines effect of #365 and #380 (plus small extra commit to polish it to get to build). so that we can see possibly running CI tests about block volumes.
> In my local test I see one failure
> 
> ```
> Dynamic PV (block volmode)] volumeMode [It] should fail in binding dynamic provisioned PV to PVC
> ```
> 
> where error is expected but we get nil instead.
> Let's see does CI hit the same.

This been fixed in my PR #380

okartau:
I fixed `testsuites.CapBlock: true` and pushed, should see CI result soon. In meantime, I tried it locally and I see same error as #380:
`should create sc, pod, pv, and pvc, read/write to the pv, and delete all created resources`

okartau:
(still checking my local test log, but I expect CI to show similar one when it finishes)
I found this close to where this pod was started,
is "user can not get resource" reason for failure?
```
event for security-context-64422483-e045-11e9-9393-c81f66f59e2f: {kubelet pmem-csi-clear-govm-worker1} FailedMapVolume: MapVolume.SetUp failed for volume "pvc-5e85cc69-e045-11e9-87d6-0242ac110002" : volumeattachments.storage.k8s.io "csi-3a043beb154fa6de213bc5d487d570fc4193a8bd0ac0c8e2eb470e22cd3bff96" is forbidden: User "system:node:pmem-csi-clear-govm-worker1" cannot get resource "volumeattachments" in API group "storage.k8s.io" at the cluster scope: no relationship found between node "pmem-csi-clear-govm-worker1" and this object
Sep 26 13:12:20.122: INFO: At 2019-09-26 13:09:07 +0300 EEST - event for security-context-64422483-e045-11e9-9393-c81f66f59e2f: {kubelet pmem-csi-clear-govm-worker1} FailedMount: Unable to mount volumes for pod "security-context-64422483-e045-11e9-9393-c81f66f59e2f_volumemode-9811(634d8dd2-e045-11e9-87d6-0242ac110002)": timeout expired waiting for volumes to attach or mount for pod "volumemode-9811"/"security-context-64422483-e045-11e9-9393-c81f66f59e2f". list of unmounted volumes=[volume1]. list of unattached volumes=[volume1 default-token-wsdrt]
```

okartau:
now CI is clean because it uses k8s 1.15, this is expected. So our implementation of block volumes is testable with test suites we have using >= 1.15 cluster, but is it also overall usable with same limitation?

pohly:
> now CI is clean because it uses k8s 1.15, this is expected. So our implementation of block volumes is testable with test suites we have using >= 1.15 cluster, but is it also overall usable with same limitation?

Is block volume support in general broken in older releases, or was it just the test that is broken? I've not followed closely what you found, I just remember that you said something about pods being created differently.

If it's just the test, then we can either add a skip statement for older Kubernetes (in our test.make) or fix the test - the former is probably simpler. If it is Kubernetes, we need to document the issue and also disable the test.

okartau:
> Is block volume support in general broken in older releases, or was it just the test that is broken?

in pmem-csi scope we don't have answer to that, as we never had block volume support yet, it is just coming in #380. Also what limits our visibility is that we don't have any use case on pmem-csi side, other than test suite.
To get visibility, we could try to verify support leaving test suite aside:
1. merge support for #365 (or work in branch(es) where this is present).
2. try some use case that utilizes block volume in different k8s versions.

pohly:
We can test manually with the example that comes with the PR.

okartau:
closing, job by this trial is done.

PR #422, opened 2019-09-30
Jenkinsfile: add k8s 1.15 testing, using Clear=31070

okartau:

PR #423, opened 2019-10-01
testing on Kubernetes 1.15

pohly:
This is PR #422 plus enhanced cluster setup.

### DIFF
--- a/test/start-kubernetes.sh
+++ b/test/start-kubernetes.sh
@@ -198,6 +198,30 @@ EOF
     done
 )
 
+# Wait for some pids to finish. Once the first one fails, the others
+# are killed. Slightly racy, because we also kill the process we already
+# waited for and whose pid thus might have been reused.
+function waitall() {
+    result=0
+    numpids=$#
+    while [ $numpids -gt 0 ]; do
+        if ! wait -n "$@"; then
+            if [ $result -eq 0 ]; then
+                result=1
+                # To increase the chance that the asynchronously logged output is shown first, wait a bit.
+                sleep 10
+                kill "$@" 2>/dev/null || true # We don't care about errors here, at least one process is gone.
+                echo >&2 "ERROR: some child process failed, killed the rest."
+            fi
+        fi
+        # We don't know which one terminated. We just decrement and
+        # keep on waiting for all pids.
+        numpids=$(($numpids - 1))
+    done
+
+    return $result
+}
+
 # Prints a single line of foo=<value of foo> assignments for
 # proxy variables and all variables starting with TEST_.
 function env_vars() (
@@ -215,6 +239,7 @@ function log_lines(){
         # those out, otherwise the line overwrites the prefix.
         echo "$(date +%H:%M:%S) $prefix: $line" | sed -e 's/\r//' | tee -a $logfile
     done
+    echo "$(date +%H:%M:%S) $prefix: END OF OUTPUT"
 }
 
 function init_kubernetes_cluster() (
@@ -226,6 +251,7 @@ function init_kubernetes_cluster() (
     KUBECONFIG=${WORKING_DIRECTORY}/kube.config
     echo "Installing dependencies on cloud images, this process may take some minutes"
     vm_id=0
+    pids=""
     for ip in $(print_ips); do
         vm_name=$(govm list -f '{{select (filterRegexp . "IP" "'${ip}'") "Name"}}') || die "failed to find VM for IP $ip"
         log_name=${WORKING_DIRECTORY}/${vm_name}.log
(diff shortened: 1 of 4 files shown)


pohly:
It looks like increasing ServerAliveInterval did the trick, it's now working.

test/start-kubernetes.sh
```diff
@@ -55,22 +55,21 @@ SSH_TIMEOUT=60
 SSH_ARGS="-oIdentitiesOnly=yes -oStrictHostKeyChecking=no \
         -oUserKnownHostsFile=/dev/null -oLogLevel=error \
         -i ${SSH_KEY}"
-SSH_ARGS+=" -oServerAliveInterval=1" # required for disruptive testing, to detect quickly when the machine died
+SSH_ARGS+=" -oServerAliveInterval=15" # required for disruptive testing, to detect quickly when the machine died
```
okartau: why separate += for adding ServerAliveInterval? For comment? We can still keep comment if all options are assigned at once

test/start-kubernetes.sh
```diff
@@ -101,45 +100,41 @@ EOF
...
             ubuntu)
                 base_url="https://cloud-images.ubuntu.com/disco/current/"
                 sha_file="SHA256SUMS"
-                curl -s -O ${base_url}/${CLOUD_IMAGE}.xz
-                curl -s -O ${base_url}/${sha_file}
-                if sha256sum $sha_file ${CLOUD_IMAGE}.xz; then
-                    unxz ${CLOUD_IMAGE}.xz
-                    CLOUD_IMAGE=${CLOUD_IMAGE/.xz}
-                fi
+                curl -s -O ${base_url}/${CLOUD_IMAGE}.xz || die "failed to download ${base_url}/${CLOUD_IMAGE}.xz"
```
okartau: here, and also in other sections, base_url and sha_file values are equally good without double quotes?

test/start-kubernetes.sh
```diff
@@ -101,45 +100,41 @@ EOF
...
+                curl -s -O ${base_url}/${sha_file} || die "failed to download ${base_url}/${sha_file}"
+                sha256sum $sha_file ${CLOUD_IMAGE}.xz || die "validation of ${CLOUD_IMAGE}.xz failed"
+                unxz ${CLOUD_IMAGE}.xz || die "failed to unpack ${CLOUD_IMAGE}.xz"
+                CLOUD_IMAGE=${CLOUD_IMAGE/.xz}
```
okartau: if the idea is to append "/.xz", then this does not work. should be: CLOUD_IMAGE=${CLOUD_IMAGE}/.xz
(similar few lines below)

okartau: hm no it probably wasnt the idea.

okartau: but still confusing, / is substitution in bash, so we want to delete .xz  ? but why it is added just few lines above then ?

test/start-kubernetes.sh
pohly: > For comment?

Yes.

> We can still keep comment if all options are assigned at once

It's harder to correlate the comment with the option when doing that - one basically has to repeat the "-oServerAliveInterval" in the comment.

test/start-kubernetes.sh
pohly: Yes, but it can be added to be on the safe side.

test/start-kubernetes.sh
pohly: No, that's bash syntax for *removing* the suffix.

pohly: `CLOUD_IMAGE` starts with `.xz` suffix (for example, `CLOUD_IMAGE="clear-$TEST_DISTRO_VERSION-cloud.img.xz"`) because the full file name has to be used during downloading. Then the file gets unpacked and the suffix removed. Could have been done differently, but this approach was already used before my changes, so I have kept it. :man_shrugging:

test/start-kubernetes.sh
pohly: Added some more quotation marks as part of the script cleanup commit.

test/start-kubernetes.sh
okartau: The logic does not add up in ubuntu, fedora cases (which are different from clear case).
1. If CLOUD_IMAGE has .xz suffix in these blocks, there will double suffix used in most of lines which is weird, and last line takes away suffix.
2. If CLOUD_IMAGE is without suffix, most of lines make sense by adding suffix, but last line is NOP as there was no suffix.
So the whole block of ubuntu or fedora can not be correct IMHO, or do you see otherwise? 
What does commands trace show when this code runs?
I believe it is the 2nd case and we have NOP stmt at the end, so its not causing error, but code should not be confusing nevertheless.

pohly: Ah, now I see what you mean. Yes, both code blocks are currently dead code and broken. Should I remove them instead of doing partial fixes?

At least Fedora already got fixed in my "fedora" branch:
https://github.com/pohly/pmem-CSI/blob/d1d6a8a901674bc30cb101748077c6bf90515f7a/test/start-kubernetes.sh#L94-L140

okartau: yes, removed, or marked "under construction" is more clear

pohly: I removed it.

PR #433, opened 2019-10-16
Jenkinsfile: add timestamps option

okartau:
Jobs timings have differences that are not easy to understand, for example why "Create Build Env" sometimes takes 4 and some other time 17 minutes.
Seeing timestamps of CI is overall useful IMHO, so lets see does that clause activate those.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,7 @@
 pipeline {
+    options {
+        timestamps()
+    }
     agent {
         label "pmem-csi"
     }


Jenkinsfile
```diff
@@ -45,6 +45,7 @@ pipeline {
         stage('Create build environment') {
             options {
                 timeout(time: 60, unit: "MINUTES")
+                timestamps ()
```
pohly: Do we only want that in this stage or in all of them?

The commit message should have more information about why this is useful, and "try" looks like a left-over from experimenting.

PR #435, opened 2019-10-16
cluster setup fix

pohly:
Fixes two problems:
- Feature gates not set for control plane components, which prevented pod tests from running on Kubernetes 1.13 where "skip attach" was alpha and thus by default enabled; we rely on that feature. We didn't have proper tests for that, but now we do and full CI testing fails for "devel".
- Kubernetes version not completely locked down (not nice, but not critical at this time).


pohly:
CI testing of the branch shows that the fix works: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/cluster-setup-fix/

PR #437, opened 2019-10-16
CI: enable JUnit output, enhance and fix TEST_E2E make parameters

pohly:
We have to run with a custom reporter in order to write JUnit
files. This will be useful for reporting in the CI what tests we ran.
While at it, add TEST_E2E_FOCUS and fix TEST_E2E_SKIP (variable was
spelled wrong).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -349,6 +349,12 @@ git push origin HEAD:master
             }
         }
     }
+
+    post {
+        always {
+            junit 'build/reports/**/*.xml'
+        }
+    }
 }
 
 /*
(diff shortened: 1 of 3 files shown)


okartau:
seems this makes the "Tests" tab on CI side showing list of tests, with execution times, looks nice!

pohly:
This works, but the presentation in the Jenkins UI is a bit odd because it expects Java class names in the testsuite name: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-437/9/testReport/

I'm going to experiment a bit more to see whether we can get rid of the useless "root" package.

PR #438, opened 2019-10-16
Adding support for block volumes

pohly:
This PR #380 rebased on top of PR #435, with one additional change that ensures that "block volmode.*should store data" actually executes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ Kubernetes cluster has been verified on:
 
 | Branch            | Kubernetes branch/version      | Required alpha feature gates   |
 |-------------------|--------------------------------|------------------------------- |
-| devel             | Kubernetes 1.13                | CSINodeInfo, CSIDriverRegistry |
+| devel             | Kubernetes 1.13                | CSINodeInfo, CSIDriverRegistry, CSIBlockVolume |
 | devel             | Kubernetes 1.14                |                                |
 | devel             | Kubernetes 1.15                |                                |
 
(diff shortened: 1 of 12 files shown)


pohly:
Because the source branch is in the main repo, we can track test results for all configurations here:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/blockvolume/

pohly:
No luck, the "should store data" test failed on 1.13: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/blockvolume/2/execution/node/129/log

avalluri:
> No luck, the "should store data" test failed on 1.13: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/blockvolume/2/execution/node/129/log

CSI block volume support is broken in Kubernetes 1.13. CSI block volume does not consider the driver's  `attachRequired:` flag and tries to get attachment object for the volume, this is where the failure was. [Here](https://github.com/kubernetes/kubernetes/issues/79884) is the related upstream issue.

If we want to support 1.13 block volumes, We better branch out the devel and revert https://github.com/intel/pmem-csi/commit/b59e2f33b536a18f38c944700907d8ac684be433. @pohly what is your opinion?

pohly:
> If we want to support 1.13 block volumes, We better branch out the devel and revert b59e2f3. @pohly what is your opinion?

We could also revert that just for 1.13. Skipping attach was an alpha feature in that release after all; depending on an alpha feature is always going to be risky.

But that makes supporting 1.13 even more complicated. I'm leaning towards declaring it as "unsupported/untested". With 1.16 out, it's also not supported upstream anymore (well, at least not officially - it still got the CVE-2019-11253 fix).

Issue #434, opened 2019-10-16
test "persistent volume retains data" fails on Ubuntu node


okartau:
A recently added test "persistent volume retains data" fails on Ubuntu node where a ssh login emits welcome banner, with expected data "hello" only shown after the banner, so we see test failure:

     persistent volume retains data
     go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:521

      read failure
      Expected
          <string>: Welcome to Ubuntu 19.04 (GNU/Linux 5.0.7-050007-generic x86_64)\n\n * Documentation:  https://help.ubuntu.com\n * Management:     https://landscape.canonical.com\n * Support:        https://ubuntu.com/advantage\n\n  System information as of Wed Oct 16 11:11:08 UTC 2019\n\n  System load:  0.44              Users logged in:        0\n  Usage of /:   7.6% of 48.29GB   IP address for ens5:    172.17.0.3\n  Memory usage: 19%               IP address for docker0: 172.18.0.1\n  Swap usage:   0%                IP address for cni0:    10.244.3.1\n  Processes:    149\n\n\n75 updates can be installed immediately.\n40 of these updates are security updates.\n\nFailed to connect to https://changelogs.ubuntu.com/meta-release. Check your Internet connection or proxy settings\n\n\nhello
      to equal
          <string>: hello

Hopefully there is way to get the file contents without getting ssh session info into buffer.

pohly:
Ubuntu shouldn't treat command execution the same way as executing a login or interactive shell. Is it possible to suppress that message, perhaps with ssh command line options?

okartau:
I tried to simplify.
The current method of running just ssh.N wrapper first, is equal to opening interactive shell in one command, then feeding target command into Stdin, right.
I propose to prepare command and run it in one exec call, being equal to run single "ssh cmd" command.
That seems also to get around login banner on Ubuntu.
I tested this in local Ubuntu-based tester and it works.
Created #439 to see results in CI.
If that works, we can consider  replacing also file creation part with single-cmd method, then can reuse the sshcmd setup part (which is repeated code already in existing code).

PR #439, opened 2019-10-17
test/e2e/storage/sanity.go: use single ssh cmd instead of session

okartau:
On a system where interactive ssh session emits banner,
like Ubuntu, the present ssh cmd run method would fail,
as response would contain banner + "hello",
so lets do it in one command.
Remove also extra "sh -c" as ssh already runs shell.

Resolves: #434

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -542,9 +542,10 @@ var _ = Describe("sanity", func() {
 
 				i := strings.Split(nodeID, "worker")[1]
 
+				sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%s", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), i)
 				// write some data to mounted volume
-				ssh := exec.Command(fmt.Sprintf("%s/_work/%s/ssh.%s", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), i))
-				ssh.Stdin = bytes.NewBufferString(`sudo sh -c 'echo -n "hello" > ` + v.getTargetPath() + "/target/test-file" + `'`)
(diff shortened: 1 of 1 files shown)


okartau:
or was the idea specifically to have ssh session open and run a cmd in it, so that status of remote command is passed to caller, instead of ssh command (as it becomes in changed code) ?

okartau:
if we want to keep the session idea, can/should we then kind of "drain" the output after 1st ssh cmd, thus getting the banner-part consumed, and then run the "cat" on clean table so that gets own output only.
Banner is response to log in, and should be considered and consumed as separate output.

test/e2e/storage/sanity.go
```diff
@@ -555,8 +555,9 @@ var _ = Describe("sanity", func() {
 				v.publish(volName, vol)
 
 				// ensure the data retained
-				ssh = exec.Command(fmt.Sprintf("%s/_work/%s/ssh.%s", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), i))
-				ssh.Stdin = bytes.NewBufferString(`sudo sh -c 'cat ` + v.getTargetPath() + "/target/test-file" + `'`)
+				sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%s", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), i)
+				cmd := "sudo cat " + v.getTargetPath() + "/target/test-file"
```
pohly: This looks simpler and I can't think of a reason why this command would have to be passed via stdin. It's sometimes necessary when executing complex scripts where special shell characters just don't make it through the ssh command line, but not here.

@avalluri, you wrote this - was there a reason for doing it that way?

okartau:
This works with Ubuntu tester host now, also passes CI, removed WIP, good to merge?

PR #440, opened 2019-10-17
CI: enhance JUnit output

pohly:
A trailing dot is needed to ensure that the test name is treated as
top-level package and is shown directly by the UI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -421,7 +421,7 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                   -w `pwd` \
                   ${env.BUILD_IMAGE} \
                   bash -c 'set -x; \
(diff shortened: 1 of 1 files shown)


pohly:
Yes, that looks better: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-440/lastCompletedBuild/testReport/

okartau:
looks better, but is that all we can get ? The list of test sets shows as links, so I click on one, but this takes me to empty list. Also, there appears down-pointed-triangle on the right side when I move mouse over a line, indicating a menu or list would appear, but there is none. Something still missing?

pohly:
> looks better, but is that all we can get ? The list of test sets shows
> as links, so I click on one, but this takes me to empty list. Also,
> there appears down-pointed-triangle on the right side when I move
> mouse over a line, indicating a menu or list would appear, but there
> is none. Something still missing?

I think that's because output is not recorded for successful
tests. The Ginkgo JUnit file doesn't include it and the Jenkins JUnit
also would ignore it by default even if it was present.

PR #441, opened 2019-10-18
WIP: #447 + add repeated node tests (vendor) + tolerate repeated node ops

### LINKED ISSUE #386: NodeStageVolume violates CSI spec idempotency req.

### LINKED ISSUE #403: NodeUnpublishVolume does not check for volume existence by VolumeID

### LINKED ISSUE #411: NodePublishVolume does not check volume states by VolumeID
CSI spec states that NodePublishVolume handler should check does Volume specified by VolumeID exist, and return NOT_FOUND if it does not exist.
If that Volume is already published, it should return ALREADY_EXISTS.
TargetPath and Stagingtargetpath are taken from the request, and used in bind-mount call.

The fact that current PMEM-CSI code passed e2e and sanity tests, is a sign that failure scenarios
are not tested.

okartau:
This is continuation of #393 but rebased onto csi test 2.3.0 as in #447
Resolves #386 : improves NodeStageVolume to handle repeated requests
Resolves #411 : improves NodePublishVolume to handle repeated requests
Resolves #403 : Adds Volume existence check in NodeUnpublishVolume

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -252,15 +252,15 @@
   version = "v0.6.1"
 
 [[projects]]
-  digest = "1:652c938b1510b21f775fcea8cadd523a9f7706f117e08338e9efdba7edd6662f"
+  digest = "1:e2ec36588cab12eed6abce39593398f5d37e53746801d6ebf793a608e3345a69"
   name = "github.com/kubernetes-csi/csi-test"
   packages = [
     "pkg/sanity",
     "utils",
   ]
   pruneopts = "UT"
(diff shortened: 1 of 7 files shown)


okartau:
closing, taking useful changes from here into #393

PR #442, opened 2019-10-18
drop support for Kubernetes 1.13

pohly:
It's an old version which is difficult to support due to its bugs
and/or limitations. Without clear demand for it, we better should
focus on more recent versions.


pohly:
@avalluri let's merge this first, then rebase PR #380 and merge that, then PR #383.

PR #443, opened 2019-10-18
DEVELOPMENT.md: document 'go test' requires go 1.12 or newer

okartau:
Seems 'go test' is not runnable any more with go 1.11.
Is this right way documenting?
Should we perhaps change earlier stmt about gofmt as well?
We can add some other doc changes to this PR if we get some good ideas.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -37,7 +37,8 @@ Coding style
 ------------
 
 The normal Go style guide applies. It is enforced by `make test`, which calls `gofmt`. `gofmt` should be
-from Go >= v1.11.4, older version (like v1.10) format the code slightly differently.
+from Go >= v1.11.4, older versions (like v1.10) format the code slightly differently.
+Running `go test` requires go 1.12 or newer.
 
 
 Input validation


avalluri:
Along with this would be good to add `//build go1.12` to the source file, i guess.

pohly:
> Along with this would be good to add //build go1.12 to the source file, i guess.

That then excludes the file from builds when using an older Go. It would just replace one build error (the one we get now) with another (missing symbols). Is that really useful?

I believe build constraints are meant to cover conditional compilation (i.e. ifdef at a file level), not for documenting build requirements.

DEVELOPMENT.md
```diff
@@ -37,7 +37,8 @@ Coding style
 ------------
 
 The normal Go style guide applies. It is enforced by `make test`, which calls `gofmt`. `gofmt` should be
-from Go >= v1.11.4, older version (like v1.10) format the code slightly differently.
+from Go >= v1.11.4, older versions (like v1.10) format the code slightly differently.
+Running `go test` requires go 1.12 or newer.
```
pohly: Looks like a suitable location to document this limitation to me.

avalluri:
> > Along with this would be good to add //build go1.12 to the source file, i guess.
> 
> That then excludes the file from builds when using an older Go. It would just replace one build error (the one we get now) with another (missing symbols). Is that really useful?
> 
My intention was to provide version-specific source files if the failures are in our source so that we can support older `go` versions. But, after trying myself, those failures are from vendored sources. So this solution won't work anyway.

okartau:
removed WIP

pohly:
>> That then excludes the file from builds when using an older Go. It would just replace one build error (the one we get now) with another (missing symbols). Is that really useful?
>> 
> My intention was to provide version-specific source files if the
> failures are in our source so that we can support older `go` versions.

There's no need to support older Go versions. We just need to be clear
about which version(s) of Go developers are supposed to use.

PR #444, opened 2019-10-18
test: Remove leftover traces of Kubernetes 1.13

avalluri:
In commit 2e9fde49, we dropped support for v1.13, but tests are still pointing to
storage class files in deploy/kubernetes-1.13. This change fixes this by
pointing to deploy/common.


pohly:
This wasn't caught when merging PR #444 because testing didn't run due to the CI being down, right?

It's still down, but I am going to merge anyway because it looks right and `devel` is broken without this fix.

PR #447, opened 2019-10-24
step up csi-test to 2.3.0

### LINKED ISSUE #428: Leftover volume "new-addition" after "go test"
Seeing this based on local testing with current devel + #426. 
After one round of "go test", there is one leftover on one node: "volume_id":"6a2130accf186ac16f0d4f336bd488c9a512513f"

This sha is result of name=new-addition which is created under test
"pagination should detect volumes added between ..."

Where is deletion of new-addition volume supposed to happen in that code?
There is no deletion requests in test log for 1st round.

When I repeat "go test" 2nd time, I see both deletions and creations for that volume name/VolID,
happening on multiple nodes. Is 2nd round deleting volume that was created in 1st round?
Lets note that after (recently) we moved to hashing volID from Name, that Name always results to same ID, which was not the case earlier when these tests were created.

okartau:
With csi-test 2.3.0 it becomes easier to work with "repeat node volume tests" addition that is in progress of up-streaming

Resolves #428

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -67,7 +67,6 @@ var _ = Describe("sanity", func() {
 		// and deletes all extra entries that it does not know about.
 		TargetPath:  "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.XXXXXX",
 		StagingPath: "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-staging.XXXXXX",
-		IDGen: &sanity.DefaultIDGenerator{},
 	}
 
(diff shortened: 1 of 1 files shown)


okartau:
> Please use the opportunity to revert

Added 2nd, revert commit

okartau:
this resolves #428 as "pagination should detect volumes added..." test gets disabled.
But let's wait with merging (adding WIP now), as presence of that test keeps leak happening and provides chance to see leftover volumes detection in action, what I plan to add.

okartau:
> But let's wait with merging

I take that back. We can merge this if you agree. There are other ways of simulating a volume leak for other PRs. I removed WIP from title.

PR #448, opened 2019-10-25
test/e2e/storage/csi_volumes.go: Enable VolumeIO test suite

okartau:
This also requires to change claim size from 1MB to 110 MB,
otherwise files created by test will not fit (needs 102 MB).

As we now have local kubectl installed during testing, we can enable volumeIO tests.
Later after upstream gets kubectl-avoidance change merged,
we will try to stop installing kubectl locally and see does testing still work.

### DIFF
--- a/test/e2e/storage/csi_volumes.go
+++ b/test/e2e/storage/csi_volumes.go
@@ -64,10 +64,11 @@ var _ = Describe("E2E", func() {
 					"ext4": "deploy/common/pmem-storageclass-ext4.yaml",
 					"xfs":  "deploy/common/pmem-storageclass-xfs.yaml",
 				},
-				// We use 16Mi size volumes because this is the minimum size supported
-				// by xfs filesystem's allocation group
+				// There is test in VolumeIO suite creating 102 MB of content
+				// so we use 110 MB claimSize to fit that with some margin.
+				// Without VolumeIO suite, 16Mi would be enough as smallest xfs system size.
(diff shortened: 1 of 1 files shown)


test/e2e/storage/csi_volumes.go
```diff
@@ -81,7 +81,8 @@ var _ = Describe("E2E", func() {
 				// We use 16Mi size volumes because this is the minimum size supported
 				// by xfs filesystem's allocation group
 				// Ref: http://man7.org/linux/man-pages/man8/mkfs.xfs.8.html
-				claimSize: "16Mi",
+				claimSize: "110Mi",
```
avalluri: The above comment is not valid anymore. And the filesizes are depend on `testsuites.DriverInfo.MaxFileSize`, currently we are setting it to `testpatterns.FileSizeMedium` which is 100M. I believe It would be good if `claimSzie` is calculated from that, say MaxFileSize + SomeBuffer.

pohly: Kubernetes 1.17 will dynamically calculate the volume size based on constraints of the test and of the driver.

I agree that the comment needs to be updated here.

okartau: I changed comment to mention 110MB, retaining also most of reasoning for 16 MB (is that needed now or can be deleted?).
About using base value: if base is 100 MB and we need to add some hard-coded const to it in the code, then the result is not much better than using hard coded end value of 100MB, right? 
We are just unlucky here that our case needs slightly more than base value of 100 MB.

PR #449, opened 2019-10-25
test/e2e/storage/ : add check for leftover volumes

### LINKED ISSUE #286: check for leftover volumes during testing
We had a regression in csi-sanity where it failed to delete volumes. It would be good to automatically check that each test leaves all nodes in the same state as before.

This could be done with a BeforeEach/AfterEach. However, we don't want to run expensive checks for tests that get skipped, so this might not be the best approach.

okartau:
If a tester creates volumes that are not deleted properly,
we will have leftover device volumes (not CSI volumes)
after testing. Let's try to detect those using device
volume list commands.

Resolves #286

### DIFF
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -51,10 +51,6 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var (
-	cleanup func()
-)
-
 // Run the csi-test sanity tests against a pmem-csi driver
 var _ = Describe("sanity", func() {
 	workerSocatAddresses := []string{}
@@ -72,6 +68,8 @@ var _ = Describe("sanity", func() {
 	f := framework.NewDefaultFramework("pmem")
 	f.SkipNamespaceCreation = true // We don't need a per-test namespace and skipping it makes the tests run faster.
 	var execOnTestNode func(args ...string) string
+	var cleanup func()
+	var prevVol map[string][]string
 
 	BeforeEach(func() {
 		cs := f.ClientSet
@@ -163,9 +161,13 @@ var _ = Describe("sanity", func() {
 		config.CreateStagingDir = mkdir
 		config.RemoveTargetPath = rmdir
 		config.RemoveStagingPath = rmdir
+		// Register list of volumes before test, using out-of-band host commands (i.e. not CSI API).
+		prevVol = GetHostVolumes()
 	})
 
 	AfterEach(func() {
+		// Check list of volumes after test to detect left-overs
+		CheckForLeftoverVolumes(prevVol)
 		if cleanup != nil {
 			cleanup()
 		}
@@ -994,4 +996,45 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
 			return true, nil
 		})
 	return pods, err
+}
+
+// Register list of volumes before test, using out-of-band host commands (i.e. not CSI API).
+func GetHostVolumes() map[string][]string {
+	var cmd string
+	var hdr string
+	switch os.Getenv("TEST_DEVICEMODE") {
+	case "lvm":
+		// lvs adds many space (0x20) chars at end, we could squeeze
+		// repetitions using tr here, but TrimSpace() below strips those away
+		cmd = "sudo lvs --foreign --noheadings"
+		hdr = "LVM Volumes"
+	case "direct":
+		// ndctl produces multiline block. We want one line per namespace.
+		// Remove double quotes, delete lines dev:xyz and blockdev:xyz as these elems may change after reboot,
+		// remove newlines, then insert one at the end, clean some more.
+		cmd = "sudo ndctl list |tr -d '\"' |grep -v '^    dev:' |grep	-v '^    blockdev:' |tr -d '\n' |tr ']' '\n' |tr -d '[{}' |tr -d ' '"
+		hdr = "Namespaces"
+	}
+	result := make(map[string][]string)
+	// Instead of trying to find out number of hosts, we trust the set of
+	// ssh.N helper scripts matches running hosts, which should be the case in
+	// correctly running tester system. We run ssh.N commands until a ssh.N
+	// script appears to be "no such file".
+	for worker := 1; ; worker++ {
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		ssh := exec.Command(sshcmd, cmd)
+		out, err := ssh.CombinedOutput()
+		if err != nil && os.IsNotExist(err) {
+			break
+		}
+		buf := fmt.Sprintf("%s on Node %d", hdr, worker)
+		result[buf] = strings.Split(strings.TrimSpace(string(out)), "\n")
+	}
+	return result
+}
+
+// CheckForLeftovers lists volumes again after test, diff means leftovers.
+func CheckForLeftoverVolumes(volBefore map[string][]string) {
+	volNow := GetHostVolumes()
+	Expect(volNow).To(Equal(volBefore), "same volumes before and after the test")
 }
--- a/test/e2e/storage/csi_volumes.go
+++ b/test/e2e/storage/csi_volumes.go
@@ -123,6 +123,7 @@ var _ = Describe("E2E", func() {
 		var (
 			storageClassLateBindingName = "pmem-csi-sc-late-binding" // from deploy/common/pmem-storageclass-late-binding.yaml
 			claim                       v1.PersistentVolumeClaim
+			prevVol                     map[string][]string
 		)
 		f := framework.NewDefaultFramework("latebinding")
 		BeforeEach(func() {
@@ -132,6 +133,8 @@ var _ = Describe("E2E", func() {
 				framework.Skipf("storage class %s not found, late binding not supported", storageClassLateBindingName)
 			}
 			framework.ExpectNoError(err, "get storage class %s", storageClassLateBindingName)
+			// Register list of volumes before test, using out-of-band host commands (i.e. not CSI API).
+			prevVol = GetHostVolumes()
 
 			claim = v1.PersistentVolumeClaim{
 				ObjectMeta: metav1.ObjectMeta{
@@ -150,6 +153,11 @@ var _ = Describe("E2E", func() {
 					StorageClassName: &storageClassLateBindingName,
 				},
 			}
+		})
+
+		AfterEach(func() {
+			// Check list of volumes after test to detect left-overs
+			CheckForLeftoverVolumes(prevVol)
 		})
 
 		It("works", func() {


okartau:
This covers some possible after-test space.
More to be added.
It is expected to fail because we have test active that leaves a volume, as described in #428 
This testing condition will go away after #447 gets merged.

okartau:
confirmed, fails in CI as expected

test/e2e/storage/sanity.go
```diff
@@ -996,3 +997,21 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+	case "lvm":
+		cmd = "sudo lvs --foreign --noheadings"
+	}
+	for worker := 1; worker <= numWorkers; worker++ {
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		ssh := exec.Command(sshcmd, cmd)
+		out, _ := ssh.CombinedOutput()
+		// TODO: remove this debug print before merging:
+		framework.Logf("checkforleftover-DEBUG(%d): cmd:[%s] out:[%s]", worker, cmd, out)
+		Expect(string(out)).To(BeEmpty())
```
pohly: This introduces the new requirement that the cluster must be completely unused (for example, none of the example apps running), otherwise testing will fail. That's okay for the CI, but may be too limited for testing on a development system.

I'm also worried that the commands (`ndctl list` and `lvs`) will report things that should be ignored on a more complex cluster.

We can either make this check optional and only activate it in the CI or enhance the check: list volumes before the test runs and again afterwards, then expect the results to be the same.

okartau: > list volumes before the test runs and again afterwards

That is good method IMO. 
It also makes the failure point standing out better. Right now, we get series of failures, only 1st being significant and all remaining are flagged only because the list cmd keeps indicating that one leftover entry. By recording state before and after test, we get rid of any previous state impact. And instead of 29 failures, we will see 1.

pohly: I should mention another drawback, though: we cannot run tests in parallel anymore. We don't do that anyway at the moment because some tests are disruptive and would interfere with other tests, but with volume checking we are forced to run sequentially also for other tests. I'm fine with that.

okartau:
developed idea further, now we store volumes/namespaces list in Before handler and compare to After state. If we assume non-parallel testing, one global variable storing the list should be enough. This code worked in my local testing, I also tried with previously existing volumes in LV mode, which were tracked correctly.

okartau:
previous state showed OK in CI tests.
Now I added artificial leak as separate commit, to check how do we catch a leak

okartau:
The combined string comparison that I use is powerful in sense that it does not need lot of code, but it leads to state where it's not easy to see on which node the leak is.
So I was thinking, we could insert headings/tags into buffer marking the output from nodes, something like:
== volumes on node 1 ==
== volumes on node 2 ==
== volumes on node 3 ==

test/e2e/storage/sanity.go
```diff
@@ -995,3 +999,34 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+		buf += string(out)
+	}
+	framework.Logf("getVolumeList-DEBUG: combined:[%s]", buf)
+	return buf
+}
+
+func checkForLeftoverVolumes(numWorkers int, volBefore string) {
+	volNow := getVolumesList(numWorkers)
+	framework.Logf("checkForLeftoverVolumes-DEBUG: now:[%s] before:[%s]", volNow, volBefore)
+	Expect(volNow).To(Equal(volBefore))
```
pohly: This results in:
```
 Expected

[2019-11-04T10:06:27.311Z]           <string>:   6a2130accf186ac16f0d4f336bd488c9a512513f ndbus0region0fsdax -wi-a----- 4.00m                                                    \n

[2019-11-04T10:06:27.311Z]       to equal

[2019-11-04T10:06:27.311Z]           <string>: 
```

Can you add a description to `To`, something like `.To(Equal(volBefore), "same volumes before and after the test"`)?

Comparing strings is okay, but if we have many active volumes finding the one that was left over may become harder. Does the output become nicer when comparing lists of volumes (i.e. split at `\n` in `getVolumesList`)? I don't know how well gomega handles this, but there are toolkits which then nicely highlight changes.

Gopkg.toml
```diff
@@ -42,7 +42,7 @@
 
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-test"
-  version = "2.3.0"
```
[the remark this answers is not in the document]
okartau: yes, I reverted 2 commits to get to state before csi-2-3-0 change was merged, just to create leak I knew was there.
My original attempt to create artificial leak by commenting out one remove() did not work, I guess because there is Delete() in After handler.
New attempt seems have worked as Ci failed

okartau:
force pushed with changes: moved var declrs to Describe() block, added descriptive text, added Node %d headers in output.

test/e2e/storage/sanity.go
```diff
@@ -996,3 +996,36 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+		buf += string(out)
+	}
+	framework.Logf("getVolumeList-DEBUG: combined:[%s]", buf)
+	return buf
+}
+
+func checkForLeftoverVolumes(numWorkers int, volBefore string) {
+	volNow := getVolumesList(numWorkers)
+	framework.Logf("checkForLeftoverVolumes-DEBUG: now:[%s] before:[%s]", volNow, volBefore)
+	Expect(volNow).To(Equal(volBefore), "same volumes before and after the test")
```
pohly: The log output does not help when the comparison fails because it is printed separately from the test failure:
```
[2019-11-04T13:24:50.878Z]   /go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:168
[2019-11-04T13:24:50.878Z] Nov  4 13:24:50.545: INFO: getVolumesList-DEBUG: cmd:[sudo lvs --foreign --noheadings]
[2019-11-04T13:24:51.641Z] Nov  4 13:24:51.500: INFO: getVolumeList-DEBUG: combined:[=== Volumes on Node 1:
[2019-11-04T13:24:51.641Z] === Volumes on Node 2:
[2019-11-04T13:24:51.641Z]   6a2130accf186ac16f0d4f336bd488c9a512513f ndbus0region0fsdax -wi-a----- 4.00m                                                    
[2019-11-04T13:24:51.641Z] === Volumes on Node 3:
[2019-11-04T13:24:51.641Z] ]
[2019-11-04T13:24:51.641Z] Nov  4 13:24:51.500: INFO: checkForLeftoverVolumes-DEBUG: now:[=== Volumes on Node 1:
[2019-11-04T13:24:51.641Z] === Volumes on Node 2:
[2019-11-04T13:24:51.641Z]   6a2130accf186ac16f0d4f336bd488c9a512513f ndbus0region0fsdax -wi-a----- 4.00m                                                    
[2019-11-04T13:24:51.641Z] === Volumes on Node 3:
[2019-11-04T13:24:51.641Z] ] before:[=== Volumes on Node 1:
[2019-11-04T13:24:51.641Z] === Volumes on Node 2:
[2019-11-04T13:24:51.641Z] === Volumes on Node 3:
[2019-11-04T13:24:51.641Z] ]
[2019-11-04T13:24:51.641Z] 
[2019-11-04T13:24:51.641Z] [91m[1m• Failure in Spec Teardown (AfterEach) [3.416 seconds][0m
[2019-11-04T13:24:51.641Z] sanity
[2019-11-04T13:24:51.641Z] [90m/go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:55[0m
[2019-11-04T13:24:51.641Z]   [91m[1mController Service [Controller Server] [AfterEach][0m
[2019-11-04T13:24:51.641Z]   [90m/go/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44[0m
[2019-11-04T13:24:51.641Z]     ListVolumes
[2019-11-04T13:24:51.641Z]     [90m/go/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:166[0m
[2019-11-04T13:24:51.641Z]       pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted
[2019-11-04T13:24:51.641Z]       [90m/go/src/github.com/intel/pmem-csi/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:262[0m
[2019-11-04T13:24:51.641Z] 
[2019-11-04T13:24:51.641Z]       [91msame volumes before and after the test
[2019-11-04T13:24:51.641Z]       Expected
[2019-11-04T13:24:51.641Z]           <string>: "...e 2:
[2019-11-04T13:24:51.641Z]             6a21..."
[2019-11-04T13:24:51.641Z]       to equal               |
[2019-11-04T13:24:51.641Z]           <string>: "...e 2:
[2019-11-04T13:24:51.641Z]           === Vo..."[0m
[2019-11-04T13:24:51.641Z] 
[2019-11-04T13:24:51.641Z]       /go/src/github.com/intel/pmem-csi/test/e2e/storage/sanity.go:1030
```

The strings also got truncated.

Please try comparing string lists and put the node number into the `To` description.

okartau:
Log output is meant to be for debug-stage only and removed in final version, I was hoping that the Equal() check shows the strings making problem visible. But seems it considers string too long or too multi-line, or some decision it does to truncate. BTW, I recall seen much longer texts printed in test comparison (pod, volume lists?). Is there some other check form more suitable?
But you are right, we should make the result display easier for person reading the test output.

pohly:
> Is there some other check form more suitable?

I would try comparing two lists of lines. But I am not sure how well that will be handled, so you'll have to experiment a bit.

pohly:
> Log output is meant to be for debug-stage only and removed in final version, I was hoping that the Equal() check shows the strings making problem visible.

I'm aware that the log output will be gone. I was just pointing out the information contained in them (the node number) is indeed useful and thus needs to be part of the assertion output when the comparison fails.

okartau:
There are flags and limits in gomega format pkg to control truncate operation for long strings, but as you pointed out, the result will be better if we create structured output from start, so lets try now with slice instead of single string. At least with one leaked volume this looks nice in my local test.
Gomega shows elems of slice as single line even if those contain newline chars, so the risk here is that output of one node may get wide with multiple elements. To avoid that, we may consider split at newlines and create deeper structure, per volumes.

pohly:
> To avoid that, we may consider split at newlines and create deeper structure, per volumes.

`[string][]string` (node name to list of volumes hash) might work best.

okartau:
I changed to map type you suggested and it seems to work, and it looks good in LVM mode.
Direct mode failure is not easy to get in CI as job stops after first failure which is LVM (if we want to see in CI we could change Jenkinsfile temporarily to start with Direct mode)
I tried a local run in Direct mode and this shows interesting result.
(ndctl namespace is shown as block of text, so I use some tr-commands to make it one-liner)
And testing shows one additional failure in reboot test, as it seems device gets different device name
for same namespace after reboot: all fields match expect device name _blockdev_.
For this change, I could work around  this by excluding "blockdev" line when composing volumes list.
But another question arises: is our state-saving safe against block device name change in reboot?

Here is the failure using current state of code:
```
can mount again after reboot
    same volumes before and after the test
    Expected
        <map[string][]string | len:3>: {
            "Namespaces on Node 1": [
                "dev:namespace0.0,mode:fsdax,map:dev,size:1073741824,uuid:f0edfa6b-0084-11ea-bddc-0242ac110003,blockdev:pmem0,name:6a2130accf186ac16f0d4f336bd488c9a512513f",
                "",
            ],
            "Namespaces on Node 2": [""],
            "Namespaces on Node 3": [""],
        }
    to equal
        <map[string][]string | len:3>: {
            "Namespaces on Node 1": [
                "dev:namespace0.3,mode:fsdax,map:dev,size:1073741824,uuid:f0edfa6b-0084-11ea-bddc-0242ac110003,blockdev:pmem0.3,name:6a2130accf186ac16f0d4f336bd488c9a512513f",
                "",
            ],
            "Namespaces on Node 2": [""],
            "Namespaces on Node 3": [""],
        }
```

okartau:
> work around this by excluding "blockdev" line

Actually, as can be seen in output, _dev_ changes as well, so it needs to get excluded also.

test/e2e/storage/sanity.go
```diff
@@ -995,3 +996,40 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+		hdr = "Namespaces"
+	}
+	result := make(map[string][]string)
+	//framework.Logf("getHostVolumes-DEBUG: cmd:[%s]", cmd)
+	for worker := 1; worker <= numWorkers; worker++ {
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		ssh := exec.Command(sshcmd, cmd)
+		out, _ := ssh.CombinedOutput()
+		buf := fmt.Sprintf("%s on Node %d", hdr, worker)
+		result[buf] = strings.Split(string(out), "\n")
```
pohly: I think if you trim the output before splitting, then we'll get rid of one redundant empty entry.

okartau: Thanks, indeed TrimSpace takes away that extra trailing newline.
Pushed new version with that change, also added a separate tmp-commit where I try to modify Jenkinsfile not to stop after LVM stage failure

okartau:
Hm, the Jenkinsfile change that I tried, has effect that pipeline does not advance after LVM test (my plan was to get it ti continue into Direct mode test), but status in GH does not change to failure.
Perhaps easier is just to switch blocks order in Jenkinsfile instead of wanting both.
The LVM mode failure output looks reasonable.

okartau:
no, it actually worked as I wanted, just the pipeline GUI did not indicate clearly after LVM stage that Direct stage was running (I waited for multiple minutes). After Direct stage, it shows both stages with logs so the end result is what is needed. Pushing new state where dev: and blockdev: lines are filtered out from namespace entry.

okartau:
I expanded test to e2e side in csi_volumes.go, re/using the functions already implemented. But these new calls now (quick trial) use hard-coded num of nodes, as I don't see easy way in  that file to access num of tester nodes. Any ideas?
Could the code on sanity side store num of workers in some variable which is alive for both scopes? The sanity side seems to run before e2e side.

test/e2e/storage/sanity.go
```diff
@@ -995,3 +998,41 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+		hdr = "LVM Volumes"
+	case "direct":
+		// ndctl produces multiline block. We want one line per namespace.
+		// Remove double quotes, delete lines dev:xyz and blockdev:xyz as these elems may change after reboot,
+		// remove newlines, then insert one at the end, clean some more.
+		cmd = "sudo ndctl list |tr -d '\"' |grep -v '^    dev:' |grep	-v '^    blockdev:' |tr -d '\n' |tr ']' '\n' |tr -d '[{}' |tr -d ' '"
+		hdr = "Namespaces"
+	}
+	result := make(map[string][]string)
+	for worker := 1; worker <= numWorkers; worker++ {
```
pohly: Instead of taking a `numWorkers` argument, this function can iterate over all `_work/<cluster>/ssh.*` files. Then the caller doesn't have to figure out how many nodes there are in the cluster.

okartau: I was having the same line of thinking. This approach may fail if the running VM set is not same as set of ssh.N files, but then something is wrong or system was manually altered, i.e. edge cases which are already out of ordinary, so I think let's try this.

okartau:
Force-pushed new version that relies on presence of ssh.N scripts to determine where to run host commands.

okartau:
If we are satisfied about the detected leaked entries display now, and also places where checking is added,
I will remove debug-related parts, i.e. reversion of csi-test that triggers leak, and some debug logging.

pohly:
> If we are satisfied about the detected leaked entries display now, and also places where checking is added, I will remove debug-related parts, i.e. reversion of csi-test that triggers leak, and some debug logging.

Yes, please. One of the open questions for me is to @avalluri (https://github.com/intel/pmem-csi/pull/449#pullrequestreview-312388432), but that can also become an issue once we merge this PR.

okartau:
Removed debug-related commits and log messages, force-pushed

okartau:
CI ok, removing WIP

test/e2e/storage/sanity.go
```diff
@@ -995,3 +997,46 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
 		})
 	return pods, err
 }
+
+// Register list of volumes before test, using out-of-band host commands (i.e. not CSI API).
+// TODO: remove debug prints before merging
```
pohly: TODO: remove TODO :grin:

test/e2e/storage/sanity.go
```diff
@@ -995,3 +997,46 @@ func WaitForPodsWithLabelRunningReady(c clientset.Interface, ns string, label la
...
+		cmd = "sudo ndctl list |tr -d '\"' |grep -v '^    dev:' |grep	-v '^    blockdev:' |tr -d '\n' |tr ']' '\n' |tr -d '[{}' |tr -d ' '"
+		hdr = "Namespaces"
+	}
+	result := make(map[string][]string)
+	// Instead of trying to find out number of hosts, we trust the set of
+	// ssh.N helper scripts matches running hosts, which should be the case in
+	// correctly running tester system. We run ssh.N commands until a ssh.N
+	// script appears to be "no such file". Have some quite-high upper
+	// limit for loop "just in case".
+	for worker := 1; worker < 1024; worker++ {
```
pohly: Can you leave out this artificial upper limit? Just start at 1 and run until there's no corresponding ssh script.

okartau: yes, can do. I added it with idea to not become looking as potential forever loop. But in practice, it will not be, so we can leave out indeed. Having a upper limit introduces another artificial limit of max num of workers supported here.

test/e2e/storage/sanity.go
okartau: thanks, removed

test/e2e/storage/sanity.go
okartau: removed upper limit checking and force-pushed

PR #451, opened 2019-10-28
Use ndctl version 67

okartau:
New version has just been released, lets see how this goes in our CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ FROM ${CLEAR_LINUX_BASE} AS build
 ARG CLEAR_LINUX_BASE
 ARG SWUPD_UPDATE_ARG
 
-ARG NDCTL_VERSION="66"
+ARG NDCTL_VERSION="67"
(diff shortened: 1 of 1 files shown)


pohly:
Looks fine to me, let's merge after the CI is done with testing it.

pohly:
@okartau did you force-push to trigger a CI build? That can also be done by logging into the Jenkins UI. Then reviewers don't need to check the diff again.

okartau:
> @okartau did you force-push to trigger a CI build? That can also be done by logging into the Jenkins UI. Then reviewers don't need to check the diff again.

it was rebase to latest devel and force push. I know rebase is not mandatory most of time but I sometimes I do it anyway to keep  things up to date. I need to comment then to avoid extra checking need.

pohly:
> it was rebase to latest devel and force push. I know rebase is not
> mandatory most of time but I sometimes I do it anyway to keep  things
> up to date.

I don't think that this is necessary unless there really is a code
conflict. It just causes extra churn.

PR #452, opened 2019-10-30
README: minor correction and improvement

okartau:
Starting block with "This" is bit fuzzy as there are multiple
preceding statements where "This" could relate to.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,11 +1,13 @@
 Table of Contents
 ----------------
 
+- [Setup](#setup)
+    - [Build PMEM-CSI](#build-pmem-csi)
 - [Code quality](#code-quality)
     - [Coding style](#coding-style)
     - [Input validation](#input-validation)
 - [Release management](#release-management)
--   - [Branching](#branching)
+    - [Branching](#branching)
     - [Tagging](#tagging)
 - [APIs](#apis)
     - [CSI API](#csi-api)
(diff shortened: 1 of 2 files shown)


okartau:
Few things stinging the eye while I made a read-through after some time passed from previous read.
Small but I get them documented in this PR while I still remember.
We can add more if we find.
One possible location to check fix is the explicit Docker versions that we list. This lists moves along and doc gets out of date, so we should either ad more "Docker tested" versions there, or simplify is to mention only major versions, or something else (what about removing the whole version list, does it help at all?)

README.md
```diff
@@ -516,7 +516,7 @@ ls: cannot access '/dev/pmem*': No such file or directory
 
 ## Supported Kubernetes versions
 
-PMEM-CSI driver implements CSI specification version 1.0.0, which only
+PMEM-CSI driver implements CSI specification version 1.0.0, which is only
```
pohly: While we are at it: "PMEM-CSI implements the CSI specification version 1.x, which is..."

pohly:
> Few things stinging the eye while I made a read-through after some time passed from previous read.
> Small but I get them documented in this PR while I still remember.

We should simply merge them each time there is something. No need to wait...

> One possible location to check and fix is the list of explicit Docker versions. This set moves along and README gets out of date, so we should either add more "Docker tested" versions there, or simplify mentioning only major versions, or something else (what about removing the whole version list, does it help at all?)

Let's remove it. The README.md is for users and they don't need Docker.

okartau:
> The README.md is for users and they don't need Docker.

but we have chapter in README "Build PMEM-CSI" which states "Use `make build-images` to produce Docker container images.", means we document building which requires Docker.
The statement we are removing, implicitly documents that dependency now. 
What about we replace current stmt stating Docker versions, with something like "Docker is required to build PMEM-CSI" under "Software required" ?

pohly:
> "Build PMEM-CSI" 

Let's move that to DEVELOPMENT.md. I don't think we need to document a Docker version there, because in practice it doesn't matter and we don't want to force developers to install some specific version just because that's the one we tested with.

DEVELOPMENT.md
```diff
@@ -1,11 +1,14 @@
 Table of Contents
 ----------------
 
+- [Setup](#setup)
+    - [Get source code](#get-source-code)
```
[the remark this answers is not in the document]
okartau: I half-reverted so that we move "Build" part only. It is bit awkward to have "get source" and "build" chapters in different documents, but thats what we get now unless we want to repeat.

okartau:
ready for merging now?

README.md
pohly: This is still open.

okartau: ah, did not notice all diffs in your line. Changed now.

PR #453, opened 2019-11-01
WIP: Repeat node volume test operations

okartau:
sanity/node.go: repeat tests to test idempotency
test/e2e/storage/sanity.go: set IdempotentCount: 4

Trying with smaller set of changes, to see how repeated tests work in CI with current HEAD.
This PR implements vendor-part of #393.
I am seeing some trouble in #393 with other changes combined, so lets isolate.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -212,7 +212,7 @@ pipeline {
                 }
             }
         }
-
+/**
         stage('make test') {
             options {
                 timeout(time: 20, unit: "MINUTES")
@@ -222,7 +222,7 @@ pipeline {
                 sh "${RunInBuilder()} ${env.BUILD_CONTAINER} make test"
             }
         }
-
+**/
         stage('Build test image') {
             options {
                 timeout(time: 60, unit: "MINUTES")
--- a/test/e2e/storage/sanity.go
+++ b/test/e2e/storage/sanity.go
@@ -57,7 +57,8 @@ import (
 var _ = Describe("sanity", func() {
 	workerSocatAddresses := []string{}
 	config := sanity.Config{
-		TestVolumeSize: 1 * 1024 * 1024,
+		TestVolumeSize:  1 * 1024 * 1024,
+		IdempotentCount: 4,
 		// The actual directories will be created as unique
 		// temp directories inside these directories.
 		// We intentionally do not use the real /var/lib/kubelet/pods as
--- a/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
+++ b/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
@@ -114,6 +114,10 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *Sanity
 	AfterEach(func() {
 		cl.DeleteVolumes()
 	})
+	count := sc.Config.IdempotentCount
+	if count == 0 {
+		count = 10
+	}
 
 	Describe("ControllerGetCapabilities", func() {
 		It("should return appropriate capabilities", func() {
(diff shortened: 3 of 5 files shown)


okartau:
This code change failed formatting check that was recently merged, so rebased, fixed and f-pushed

okartau:
next try, added dm-test to exclusion set together with 'make test'

okartau:
One of failures here is this:
```
[2019-11-27T13:50:05.094Z] Nov 27 13:50:04.728: INFO: ExecWithOptions {Command:[/bin/sh -c tar -xf - &&if _work/test.test -h 2>&1 | grep -q ginkgo; then TEST_WORK=_work _work/test.test -test.v -ginkgo.v; else TEST_WORK=_work _work/test.test -test.v; fi] Namespace:default PodName:pmem-csi-node-2fbbq ContainerName:pmem-driver Stdin:0xc000466930 CaptureStdout:true CaptureStderr:true PreserveWhitespace:false}
[2019-11-27T13:50:05.094Z] Nov 27 13:50:04.728: INFO: >>> kubeConfig: /src/pmem-csi/_work/clear/kube.config
[2019-11-27T13:50:05.450Z] tar: _work/evil-ca: Cannot stat: No such file or directory
[2019-11-27T13:50:05.450Z] tar: _work/pmem-ca: Cannot stat: No such file or directory
[2019-11-27T13:50:05.450Z] tar: Exiting with failure status due to previous errors
```
I don't quite understand what is the role of tar command there.
Is it problem that space is missing in `&&if` ?
Something goes wrong with arguments to `tar`.
But what adds confusion, locally run e2e seems to work (ubuntu-govm cluster).

pohly:
> I don't quite understand what is the role of tar command there.

It copies from the local filesystem to the container filesystem.

> Is it problem that space is missing in `&&if`

No, shell doesn't need spaces around `&&`, `||`, `;`, etc.

The problem is that these files are missing unless they were created by something else, like a preceding `make test`. A missing make dependency is causing that. I'll fix it.

okartau:
But this make dependency issue likely is not causing other direct-mode failures we see that seem more low level: pfn errors, or "seed for namespace" as I have seen few times.
I can also repeat many errors `CreateVolume: failed to create volume: Failed to get seed namespace in region region0` in a local e2e tests in fedora-30 cluster, direct mode. But then, a repeated run in same cluster (after recreating VMs) was passing OK.
Has some pmem-csi recent change triggered those, or is there change in outer parts, like Fedora-30 image, kernel etc. ? These issues started very recently.

okartau:
Rebased and force-pushed after many changes in devel, to verify that current head of devel still tolerates repeated tests.
I guess troubles mentioned in recent comments can now be ruled as root-caused by recent "pfn: no seed" issue.

okartau:
I went on based on idea once expressed by @pohly, and enabled repetition of Controller related test operations in addition to existing repeated Node volume operations.
But, it appears controller volume related operations that make sense to repeat for idempotency (which I added in recently added commit) do not run because get skipped by capability check.
So the added part does not make sense in pmem-csi context and I take those away again.
I don't see test cases that we can repeat more, so we are left with added node tests as it is now.

okartau:
rebase and force push to see CI result after fix as previous CI job suffered from #518

okartau:
The code here loses relevance as this functionality (with newer code) was now committed in csi-test, and becomes usable by pmem-csi after we switch to more recent version of csi-test.
So I guess we can close this now.

okartau:
Closing.
The repeating code is now in csi-test, and although not yet running (I think we use older version), it will come in future, and this PR now has developed conflicts which do not make sense to resolve/rebase any more.

PR #454, opened 2019-11-04
WIP: #453 + NodeVolumeUnpublish checks volume existence first

### LINKED ISSUE #403: NodeUnpublishVolume does not check for volume existence by VolumeID

okartau:


okartau:
@avalluri what do you think of 2nd commit here? Your ephemeral change improved the idempotency aspect, but there is still issue #403 which needs addressing. I tried in 2nd commit.

test/e2e/storage/sanity.go
```diff
@@ -60,6 +60,7 @@ var _ = Describe("sanity", func() {
 	workerSocatAddresses := []string{}
 	config := sanity.Config{
 		TestVolumeSize: 1 * 1024 * 1024,
+		IdempotentCount: 4,
```
avalluri: Setting the count to 2 is not enough. I just concerned about the test time it increases. @pohly your opinion.

pohly: > Setting the count to 2 is not enough.

Enough for what?

> I just concerned about the test time it increases.

How much slower does it get and why? The additional calls shouldn't do any real work, so it would be surprising if that had any big effect.

okartau: 4 is value @pohly suggested to use, IIRC. I also think using 2 rounds very likely can expose same problems as value of 4 rounds. 
I can run few tries of local tests using different values, to get some data to back up guesses.

avalluri: > > Setting the count to 2 is not enough.
> 
> Enough for what?

To test idempotency.

> 
> > I just concerned about the test time it increases.
> 
> How much slower does it get and why? The additional calls shouldn't do any real work, so it would be surprising if that had any big effect.

I just try to understand if any reason(i missed) behind choosing this number 4.

okartau: Thanks for valuable discussion, but can we move discussion about repeated tests implementation under #453 ?
The point of this PR is to discuss the changed return value of NodeUnpublishVolume which is present in another commit here. The "repeat tests" commit is here only to provide tests repetition feature so that actual change can be seen in action in CI.
Once we can agree about changed return code (and code moved to earlier to make this happen),
I plan to take out commit of #453 from here, then it becomes ready for merging.

pohly: @avalluri: did you mean to write "Setting the count to 2 is enough"? You wrote "Setting the count to 2 is *not* enough."

avalluri: > @avalluri: did you mean to write "Setting the count to 2 is enough"? You wrote "Setting the count to 2 is _not_ enough."

Oops, Sorry for the confusion. Yes, my intention was "Setting count to 2 should be enough to capture idempotency errors."

okartau:
closing without merge, role for this is completed by merging actual change in #457

PR #455, opened 2019-11-05
e2e tests: Run ephemeral testsuites only supported Kubernetes versions

avalluri:
Recent ephemeral change broke the e2e tests when running on k8s v1.14, as there
is no support for ephemeral volumes. So, with this change we conditionally add
ephmeral testsuite to driver supported testsuites.


test/e2e/storage/csi_volumes.go
```diff
@@ -84,7 +85,15 @@ var _ = Describe("E2E", func() {
 		testsuites.InitVolumeIOTestSuite,
 		testsuites.InitVolumeModeTestSuite,
 		testsuites.InitVolumesTestSuite,
-		testsuites.InitEphemeralTestSuite,
+	}
+	if k8sVersion := os.Getenv("TEST_KUBERNETES_VERSION"); k8sVersion != "" {
+		var major, minor int
+		_, err := fmt.Sscanf(k8sVersion, "%d.%d", &major, &minor)
+		framework.ExpectNoError(err, "failed to parse kubernetes version %s", k8sVersion)
```
pohly: This assertion triggers: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-455/1/pipeline

Spelling error ("kubernetes").

avalluri: Seems like we cannot write Fail() statement inside `Describe/Context`. So now I changed `framework.Logf`, so that we just log the failure and continue running tests.

pohly: But why does it fail? It shouldn't.

avalluri: > But why does it fail? It shouldn't.

You mean adding `framework.ExpectNoErro()` should not lead to an assertion _or_ errors while parsing version string should not fail e2e tests as I was trying to do ?.
If ti's about assertion - Unlike `It()` The function body in Describe/Context call will be executed by ginkgo immediately. So calling ExpectNoError(), without setting failure handler leads to the assertion.

avalluri: @pohly How to enable 1.14 testing in CI?

pohly: The Gomega assertion failed, i.e. `err` was non-nil. Reporting that failure then itself failed and panicked (-> stack trace) instead of recording a test failure

That second part makes sense because there is no running test and failure reporting hasn't been configured yet, but `err` should be nil.

pohly: Or at least that was my interpretation of the failure report. It looks like I was wrong, as there is no log message in the newer test run.

test/e2e/storage/csi_volumes.go
```diff
@@ -84,7 +85,18 @@ var _ = Describe("E2E", func() {
 		testsuites.InitVolumeIOTestSuite,
 		testsuites.InitVolumeModeTestSuite,
 		testsuites.InitVolumesTestSuite,
-		testsuites.InitEphemeralTestSuite,
+	}
+	if k8sVersion := os.Getenv("TEST_KUBERNETES_VERSION"); k8sVersion != "" {
+		var major, minor int
+		if _, err := fmt.Sscanf(k8sVersion, "%d.%d", &major, &minor); err != nil {
+			framework.Logf("Failed to parse Kubernetes version '%s': %s", k8sVersion, err.Error())
```
pohly: Can you skip this log message when `TEST_KUBERNETES_VERSION` is empty?

test/e2e/storage/csi_volumes.go
```diff
@@ -84,7 +85,18 @@ var _ = Describe("E2E", func() {
...
+		}
+		// Run ephemeral test only on supported Kubernetes versions, i.e., 1.14+
+		if major >= 1 && minor > 14 {
```
pohly: Please invert the logic: always add the test unless `TEST_KUBERNETES_VERSION` was set to something <= 1.14. Otherwise they might get skipped accidentally.

test/e2e/storage/csi_volumes.go
avalluri: Added log message in case of empty version string.

test/e2e/storage/csi_volumes.go
avalluri: Rewrite the logic to support your suggestion.

PR #457, opened 2019-11-06
NodeUnpublishVolume: check for VolumeID existence first

### LINKED ISSUE #403: NodeUnpublishVolume does not check for volume existence by VolumeID

okartau:
By CSI spec, NodeUnpublish has to return NOT_FOUND if
volume does not exist, but currently we don't check it
before umounting, so we will return different status.

Resolves #403


avalluri:
Looks good to me.

okartau:
I forgot to ask for review, but I think we can consider above comment as accepting review and merge, right.

avalluri:
> I forgot to ask for review, but I think we can consider above comment as accepting review and merge, right.

Yes, we can merge.

PR #460, opened 2019-11-08
test/test.make: add test/ to gofmt checking path set

okartau:
test/ was left out of gofmt checking and some formatting
errors have been slipped in there.


okartau:
ATM there are 2 files failing gofmt check, I can create another commit here fixing those.
Presence of format issues means CI should fail on with current commit alone, right.

okartau:
Fixed two files that failed  using gofmt, force pushed, so this PR should become OK in CI now

okartau:
CI ok, removing WIP

PR #461, opened 2019-11-09
Go 1.13 + Moving to go modules

avalluri:
Set of changes to move to go-1.13 release and introducing go modules for dependency tracking.  With this change we no more use 'dep' tool going forward.


Dockerfile
```diff
@@ -18,7 +18,7 @@ ARG SWUPD_UPDATE_ARG
 ARG NDCTL_VERSION="67"
 ARG NDCTL_CONFIGFLAGS="--disable-docs --without-systemd --without-bash"
 ARG NDCTL_BUILD_DEPS="os-core-dev devpkg-util-linux devpkg-kmod devpkg-json-c"
-ARG GO_VERSION="1.12.9"
+ARG GO_VERSION="1.13"
```
pohly: That's old. Let's use 1.13.4 instead.

go.mod
```diff
@@ -0,0 +1,75 @@
...
+	k8s.io/client-go v0.0.0
+	k8s.io/component-base v0.0.0
+	k8s.io/klog v1.0.0
+	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d // indirect
+	k8s.io/kubernetes v1.17.0-alpha.0.0.20191109043140-c3f2a6524ed8
+	k8s.io/utils v0.0.0-20190801114015-581e00157fb1
+)
+
+replace (
+	k8s.io/api => k8s.io/api v0.0.0-20191016110408-35e52d86657a
```
pohly: How are we going to maintain this section in the future?

Two options:
- use a script like the one I wrote for kubernetes-csi: https://github.com/kubernetes-csi/csi-release-tools/blob/master/go-get-kubernetes.sh
- wait for Kubernetes 1.17, which will have proper tags that'll show up here instead of "v0.0.0-20191016110408-35e52d86657a"

Jenkinsfile
```diff
@@ -48,7 +48,7 @@ pipeline {
         /* last version before the 1.14 update in 28630 */
         CLEAR_LINUX_VERSION_1_13 = "28620"
 
-        PMEM_PATH = "/go/src/github.com/intel/pmem-csi"
```
pohly: The error message says "go mod tidy; go mod vendor". Isn't that the real problem, not where the source is located? I'm not sure (doing it makes some changes to go.sum, but that may be due to slightly different Go versions - I am using 1.13.4).

The code should still build inside GOPATH. Moving it elsewhere smells like a workaround that doesn't address the real issue.

avalluri:
> Something is odd. Despite `-mod=vendor`, `make test` ends up downloading some stuff:
> 
> ```
> GOOS=linux GO111MODULE=on GOFLAGS=-mod=vendor go test -run none ./pkg/... ./test/e2e
> ?   	github.com/intel/pmem-csi/pkg/coverage	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/ndctl	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/pmem-common	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/pmem-csi-driver	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/pmem-device-manager	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/pmem-exec	[no test files]
> ?   	github.com/intel/pmem-csi/pkg/pmem-grpc	[no test files]
> ok  	github.com/intel/pmem-csi/pkg/pmem-ns-init	0.011s [no tests to run]
> ?   	github.com/intel/pmem-csi/pkg/pmem-registry	[no test files]
> ok  	github.com/intel/pmem-csi/pkg/pmem-state	0.006s [no tests to run]
> ?   	github.com/intel/pmem-csi/pkg/pmem-vgm	[no test files]
> ok  	github.com/intel/pmem-csi/pkg/registryserver	0.012s [no tests to run]
> ok  	github.com/intel/pmem-csi/test/e2e	0.035s [no tests to run]
> go: downloading github.com/docker/go-units v0.4.0
> go: downloading golang.org/x/net v0.0.0-20191027233614-53de4c7853b5
> go: downloading k8s.io/utils v0.0.0-20190801114015-581e00157fb1
> go: downloading k8s.io/klog v1.0.0
> go: downloading google.golang.org/grpc v1.24.0
> ...
> ```
> 

These ad hoc package downloads are coming from `make test-kustomize` while building `kustomize` binary.
We have 2 options if we want to avoid these package downloads:
 1) Download `kustomize` [release binary](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md#quickly-curl-the-latest-binary) instead of building from source
 2) Maintain `kustomize` dependencies part of vendor tree.

The same applies for `govm` binary while `make start`.

go.mod
pohly: BTW, how was this section created?

avalluri: > BTW, how was this section created?

Using the same release script you pointed.

avalluri: I was in the impression that we might not need to do more update rounds before going to Kubernetes v1.17

pohly:
> These ad hoc package downloads are coming from make test-kustomize while building kustomize binary.

That's okay. I just wanted to be sure that we are vendoring correctly (i.e. `vendor` is up-to-date) and really use it. Speaking of "up-to-date": we probably need a test for that. Otherwise we might end up making source code changes or dependency changes (`go get`) that aren't followed by "go mod tidy && go mod vendor" and/or committed without the updated vendor dir. Can you copy the test that I wrote for kubernetes-csi and integrate it into "make test"? We don't need the dep part and the if check for skipping the test must be modified to work in Jenkins.

https://github.com/kubernetes-csi/csi-release-tools/blob/master/verify-vendor.sh

> We have 2 options if we want to avoid these package downloads:
>
> 1. Download kustomize release binary instead of building from source
> 2. Maintain kustomize dependencies part of vendor tree.

I found that downloading from github.com can be pretty slow, sometimes even slower than building from source. I'm just pointing that out as something that we need to check before deciding. But in this case the actual binaries seems to be hosted on AWS and getting them was fast for me, so this might be the best solution.

I'd prefer to not add tool sources to our vendor dir if we can avoid it because it makes our BOM larger and harder to maintain.

go.mod
pohly: Okay, so let's just keep in mind that this script was used and if one of us three developers needs to change the Kubernetes code we'll know what to do.

Jenkinsfile
avalluri: It was not GOPATH. GOROOT is the real cause:  https://github.com/golang/go/issues/34657

pohly: So the problem was that the Jenkinsfile and the Dockerfile both used /go, one for the source code (aka GOPATH), the other for the Go compiler (and thus GOROOT)? I wonder where GOPATH and/or HOME are set. It would be nice to avoid that overlap.

In the meantime, choosing some other directory is fine. However, can you add a comment explaining that PMEM_PATH is used as location for the source to build outside of GOPATH?

avalluri: > So the problem was that the Jenkinsfile and the Dockerfile both used /go, one for the source code (aka GOPATH), the other for the Go compiler (and thus GOROOT)? I wonder where GOPATH and/or HOME are set. It would be nice to avoid that overlap.

As of my understanding with go modules, GOPATH is irrelevant. That means there is no more hart need of keeping source code inside GOPATH. The only problem with GOROOṬ. In our both(Jenkinsfile & Dockerfile) cases GOROOT is pointed(I guess) to /go.

Makefile
```diff
@@ -189,5 +185,5 @@ clean-kustomize:
 .PHONY: test-kustomize $(addprefix test-kustomize-,$(KUSTOMIZE_OUTPUT))
 test: test-kustomize
 test-kustomize: $(addprefix test-kustomize-,$(KUSTOMIZE_OUTPUT))
-$(addprefix test-kustomize-,$(KUSTOMIZE_OUTPUT)): test-kustomize-%: _work/kustomize-$(KUSTOMIZE_VERSION)
+$(addprefix test-kustomize-,$(KUSTOMIZE_OUTPUT)): test-kustomize-%: _work/kustomize
```
pohly: Can you restore the `-$(KUSTOMIZE_VERSION)` suffix? Without it, "make kustomize" might continue to use an old, incorrect version. The sequence of events that can lead to that is:
- run `make kustomize`
- check out some PMEM-CSI source code where `${KUSTOMIZE_VERSION}` is different
- run `make kustomize` again

avalluri: @pohly, I tried to mimic the sequence of commands as you mentioned, but couldn't reproduce the issue you pointed:
 1) run `make test-kustomize` with KUSTOMIZE_VERSION=3.4.0
 2) change to KUSTOMIZE_VERSION=3.3.0
 3) rerun `make test-kustomize` - now it downloads the v3.3.0 binary.

```
$ _work/kustomize version
{Version:kustomize/v3.4.0 GitCommit:2c9635967a2b1469d605a91a1d040bd27c73ca7d BuildDate:2019-11-12T05:00:57Z GoOs:linux GoArch:amd64}

$ make test-kustomize
curl -L https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v3.3.0/kustomize_v3.3.0_linux_amd64.tar.gz -o /home/avalluri/work/go/src/github.com/intel/pmem-csi/_work/kustomize_v3.3.0_linux_amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   626    0   626    0     0    578      0 --:--:--  0:00:01 --:--:--   579
100 8192k  100 8192k    0     0   674k      0  0:00:12  0:00:12 --:--:-- 1311k
tar xzf _work/kustomize_v3.3.0_linux_amd64.tar.gz -C _work

$ _work/kustomize version
Version: {Version:kustomize/v3.3.0 GitCommit:7050c6a7b692fdba6e831e63c7b83920ab03ad76 BuildDate:2019-10-24T17:54:30Z GoOs:linux GoArch:amd64}
```

pohly: Never mind, I missed the dependency of `kustomize` on the version (`_work/kustomize: _work/kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz`). That's different from before, but also works.

pohly:
There's a new failure in the CI (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-461/8/pipeline):
```
[2019-11-14T10:09:09.184Z] PATH="/mnt/workspace/pmem-csi_PR-461/_work/bin:$PATH" test/start-kubernetes.sh
[2019-11-14T10:09:09.184Z] get current user: user: Current requires cgo or $USER set in environment
```

avalluri:
> There's a new failure in the CI (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-461/8/pipeline):
> 
> ```
> [2019-11-14T10:09:09.184Z] PATH="/mnt/workspace/pmem-csi_PR-461/_work/bin:$PATH" test/start-kubernetes.sh
> [2019-11-14T10:09:09.184Z] get current user: user: Current requires cgo or $USER set in environment
> ```

I suspect due to change in `govm` version, I will check.

pohly:
> Can you copy the test that I wrote for kubernetes-csi and integrate it into "make test"? We don't need the dep part and the if check for skipping the test must be modified to work in Jenkins.

@avalluri: are you working on that or shall I add that later?

avalluri:
> > Can you copy the test that I wrote for kubernetes-csi and integrate it into "make test"? We don't need the dep part and the if check for skipping the test must be modified to work in Jenkins.
> 
> @avalluri: are you working on that or shall I add that later?

I will do that, part of this PR.

test/test.make
```diff
@@ -22,14 +22,19 @@ fmt:
...
+# actually use any package from it. "go mod vendor" only vendors
+# the packages we really need and we can limit the
+# check to content that we really have in vendor.
 .PHONY: test_vendor_bom
 test: test_vendor_bom
 test_vendor_bom:
 	@ if ! diff -c \
 		<(tail -n +2 vendor-bom.csv | sed -e 's/;.*//') \
-		<((grep '^  name =' Gopkg.lock  | sed -e 's/.*"\(.*\)"/\1/') | LC_ALL=C LANG=C sort); then \
+		<(for path in $$(go list -f '{{.Path}}' -m all); do if [ -d "vendor/$$path" ]; then echo "$$path"; fi; done | LC_ALL=C LANG=C sort); then \
```
pohly: I started wondering whether we need `$(GO)` (and thus `GOFLAGS=-mod=vendor`) here instead of just go. Sure enough, using `go` results in downloading packages (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-461/14/pipeline/51).

There's just one snag: with `GOFLAGS=-mod=vendor`, some modules aren't listed anymore (for example, k8s.io/kubectl) even though we have that module in `vendor`. This needs further investigation, don't merge.

pohly: Raised the question upstream: https://github.com/golang/go/issues/35589

pohly: @avalluri: please remove this commit and instead cherry-pick this one: https://github.com/pohly/pmem-CSI/commit/43dfe7fabfcd116b947ef6986eda8d5f4ccc3648

I switched to a different approach. Does that make sense?

avalluri: Yes, sound more promising than depending on `go list`. I will cherry-pick.

avalluri:
> PMEM-CSI uses Go modules and thus can be checked out and (if that should be desired) built anywhere in the filesystem. Pre-built container images are available and thus users don't need to build from source, but they will still need some additional files. To get the source code, use:
> 
> git clone https://github.com/intel/pmem-csi

Good point. Done

test/test.make
avalluri: > @avalluri: please remove this commit and instead cherry-pick this one: [pohly@43dfe7f](https://github.com/pohly/pmem-CSI/commit/43dfe7fabfcd116b947ef6986eda8d5f4ccc3648)
> 
Done

avalluri:
Added new [commit](https://github.com/intel/pmem-csi/pull/461/commits/585afabad081ee445e6972952a5f6aa820cfef11) that increases the ssh timeout while creating test cluster.

test/start-kubernetes.sh
```diff
@@ -78,7 +78,7 @@ case ${TEST_DISTRO} in
         ;;
 esac
 
-SSH_TIMEOUT=60
```
pohly: So the cluster now comes up slower than before? That's worrisome, but not much that we can do about it. :man_shrugging:

avalluri: To my surprise, It was not observed in CI but on my dev machine, the failure was quite frequent.

avalluri:
A test failure in e2e tests, I suspect unrelated this change:

```
19-11-15T12:26:56.220Z]   late binding
[2019-11-15T12:26:56.220Z]   [90m/src/pmem-csi/test/e2e/storage/csi_volumes.go:122[0m
[2019-11-15T12:26:56.220Z]     [91m[1mworks [AfterEach][0m
[2019-11-15T12:26:56.220Z]     [90m/src/pmem-csi/test/e2e/storage/csi_volumes.go:163[0m
[2019-11-15T12:26:56.220Z] 
[2019-11-15T12:26:56.220Z]     [91msame volumes before and after the test
[2019-11-15T12:26:56.220Z]     Expected
[2019-11-15T12:26:56.220Z]         <map[string][]string | len:3>: {
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 1": [""],
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 2": [""],
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 3": [""],
[2019-11-15T12:26:56.220Z]         }
[2019-11-15T12:26:56.220Z]     to equal
[2019-11-15T12:26:56.220Z]         <map[string][]string | len:3>: {
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 1": [""],
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 2": [""],
[2019-11-15T12:26:56.220Z]             "LVM Volumes on Node 3": [
[2019-11-15T12:26:56.221Z]                 "WARNING: Failed to connect to lvmetad. Falling back to device scanning.",
[2019-11-15T12:26:56.221Z]             ],
[2019-11-15T12:26:56.221Z]         }[0m
[2019-11-15T12:26:56.221Z] 
[2019-11-15T12:26:56.221Z]     /src/pmem-csi/test/e2e/storage/sanity.go:1039
```

okartau:
> A test failure in e2e tests, I suspect unrelated this change:

That is from recently merged volume leak test.
Seems to be a LVM subsystem problem getting LV volums list
Such error did not happen during "baking time" of those tests, so we need to add code to deal with such condition. (probably retries ?)

okartau:
actually, as this is just added Warning text and real operations did succeed, it is just question of filtering out warning text, no retry needed.

avalluri:
> actually, as this is just added Warning text and real operations did succeed, it is just question of filtering out warning text, no retry needed.

We need to disable `lvmetad` checking on VM hosts as we did in our docker images:
```
RUN echo "global { use_lvmetad = 0 }" >> /etc/lvm/lvm.conf && \
    echo "activation { udev_sync = 0 udev_rules = 0 }" >> /etc/lvm/lvm.conf
```

okartau:
> We need to disable `lvmetad` checking

Good idea! 
lets move that thread into freshly created #468

avalluri:
I am :confused:. How come CI results are green. Clearly, there is one test failure.

okartau:
you mean one test in "production 1.15 Clear" failed?
But there is retry(2) option that makes it try again, and 2nd attempt worked.

avalluri:
@pohly Is it good enough to merge.

avalluri:
> you mean one test in "production 1.15 Clear" failed?
> But there is retry(2) option that makes it try again, and 2nd attempt worked.

Ok, got it. Thanks.

Issue #462, opened 2019-11-13
Improve NodePublishVolume return codes


okartau:
Return codes are not always correct in NodePublishVolume,
and it needs added checks for attributes compatibility if volume already mounted.
If StagingTargetPath is not set, the response now is INVALID_ARG but should be FAILED_PRECONDITION.
If a volume is already published, the response now is OK,
but should be: check for path, capability and readonly attributes match, and return OK only if these match, and ALREADY_EXISTS error otherwise.

okartau:
I have code prototype already developed in pre-ephemeral times, I will modify this to fit into current code.

PR #463, opened 2019-11-13
Improve NodePublishVolume and NodeStageVolume

okartau:
NodePublishVolume performs checks for attributes compatibility if volume already mounted.
If StagingTargetPath is not set, return FAILED_PRECONDITION.

Resolves: #462 
Resolves: #465

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -182,7 +182,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		}
 	case *csi.VolumeCapability_Mount:
 		if !ephemeral && len(srcPath) == 0 {
-			return nil, status.Error(codes.InvalidArgument, "Staging target path missing in request")
+			return nil, status.Error(codes.FailedPrecondition, "Staging target path missing in request")
 		}
 
 		notMnt, err := mount.IsNotMountPoint(ns.mounter, targetPath)
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -302,6 +302,9 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	volumeMutex.LockKey(req.GetVolumeId())
 	defer volumeMutex.UnlockKey(req.GetVolumeId())
 
+	klog.V(4).Infof("NodeStageVolume: VolumeID:%v Staging target path:%v Requested fsType:%v",
+		req.GetVolumeId(), stagingtargetPath, requestedFsType)
+
 	device, err := ns.cs.dm.GetDevice(req.VolumeId)
 	if err != nil {
 		if errors.Is(err, pmdmanager.ErrDeviceNotFound) {
@@ -346,7 +349,6 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	volumeMutex.LockKey(req.GetVolumeId())
 	defer volumeMutex.UnlockKey(req.GetVolumeId())
 
-	// showing for debug:
 	klog.V(4).Infof("NodeUnStageVolume: VolumeID:%v Staging target path:%v",
 		req.GetVolumeId(), stagingtargetPath)
 
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -345,8 +345,24 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Errorf(codes.Internal, "failed to get device details for volume id %q: %v", req.VolumeId, err)
 	}
 
-	if err = ns.provisionDevice(device, requestedFsType); err != nil {
+	// Check does devicepath already contain a filesystem?
+	existingFsType, err := determineFilesystemType(device.Path)
+	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// what to do if existing file system is detected;
+	if existingFsType != "" {
+		// Is existing filesystem type same as requested?
+		if existingFsType == requestedFsType {
+			klog.V(4).Infof("Skip mkfs as %v file system already exists on %v", existingFsType, device.Path)
+		} else {
+			return nil, status.Error(codes.AlreadyExists, "File system with different type exists")
+		}
+	} else {
+		if err = ns.provisionDevice(device, requestedFsType); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	// FIXME(avalluri): we shouldn't depend on volumecontext to determine the device mode,
@@ -480,44 +496,25 @@ func (ns *nodeServer) createEphemeralDevice(ctx context.Context, req *csi.NodePu
 // and mounts at given targetPath.
 func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType string) error {
 	if fsType == "" {
-		// Default to ext4 filesystem
+		// Empty FsType means "unspecified" and we pick default, currently hard-coded to ext4
 		fsType = defaultFilesystem
 	}
 
-	// Check does devicepath already contain a filesystem?
-	existingFsType, err := determineFilesystemType(device.Path)
-	if err != nil {
-		return err
-	}
-
-	// what to do if existing file system is detected and is different from request;
-	// forced re-format would lead to loss of previous data, so we refuse.
-	if existingFsType != "" {
-		// Is existing filesystem type same as requested?
-		if existingFsType == fsType {
-			klog.V(4).Infof("Skip mkfs as %v file system already exists on %v", existingFsType, device.Path)
-		} else {
-			return fmt.Errorf("File system with different type(%s) exists, whereas requested type is '%s'", existingFsType, fsType)
-		}
+	cmd := ""
+	var args []string
+	// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
+	if fsType == "ext4" {
+		cmd = "mkfs.ext4"
+		args = []string{"-b 4096", "-F", device.Path}
+	} else if fsType == "xfs" {
+		cmd = "mkfs.xfs"
+		args = []string{"-b", "size=4096", "-f", device.Path}
 	} else {
-		// no existing file system, make fs
-		// Empty FsType means "unspecified" and we pick default, currently hard-codes to ext4
-		cmd := ""
-		var args []string
-		// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
-		if fsType == "ext4" {
-			cmd = "mkfs.ext4"
-			args = []string{"-b 4096", "-F", device.Path}
-		} else if fsType == "xfs" {
-			cmd = "mkfs.xfs"
-			args = []string{"-b", "size=4096", "-f", device.Path}
-		} else {
-			return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
-		}
-		output, err := pmemexec.RunCommand(cmd, args...)
-		if err != nil {
-			return fmt.Errorf("mkfs failed: %s", output)
-		}
+		return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
+	}
+	output, err := pmemexec.RunCommand(cmd, args...)
+	if err != nil {
+		return fmt.Errorf("mkfs failed: %s", output)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


okartau:
I just realized that if Capability/mount_flags are present in NodePublishVolume request, current code never uses those in mounting.
I think in reality mount options are not set in request, at least we have not seen such yet (?)
The current PR contains log message to trace given mount option, so we will at least see and log what is given in request.
If we start comparing those mount flags, it means we also need to start to use those in mount.
The spec says these are "The mount options that can be used for the volume",
does the "can be used" actually mean "must be used" ?

okartau:
Removed test-repeater part which was  present only for testing, and WIP.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -343,7 +346,6 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	volumeMutex.LockKey(req.GetVolumeId())
 	defer volumeMutex.UnlockKey(req.GetVolumeId())
 
-	// showing for debug:
 	klog.V(4).Infof("NodeUnStageVolume: VolumeID:%v Staging target path:%v",
```
avalluri: In this case, we can remove this logging?

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -42,7 +43,12 @@ type nodeServer struct {
 	nodeCaps []*csi.NodeServiceCapability
 	cs       *nodeControllerServer
 	// Driver deployed to provision only ephemeral volumes(only for Kubernetes v1.15)
-	mounter mount.Interface
+	mounter       mount.Interface
+	volReadOnly   map[string]bool
```
avalluri: How about introducing a new structure and keeping all these new params as fields in it. Then we can maintain a single `VolumeId:Params` map, easy to maintain. If I remember correctly we used to have such type earlier but removed as found no real need that time.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -108,11 +118,15 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	var device *pmdmanager.PmemDeviceInfo
-	var mountOptions []string
 	var err error
 
 	srcPath := req.GetStagingTargetPath()
 	targetPath := req.GetTargetPath()
+	mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
+	readOnly := req.GetReadonly()
+	fsType := req.GetVolumeCapability().GetMount().GetFsType()
+	klog.V(3).Infof("NodePublishVolume request: VolID:%s targetpath:%v sourcepath:%v readonly:%v Capab/mountflags:%v Capab/fsType:%v",
```
avalluri: Can you change `Capab/mountflags` to `mount flags:` in the log message. Capab is not informative.

pkg/pmem-csi-driver/nodeserver.go
okartau: My idea was to have log msg in all handling functions near start, so that we see the start of handling. There is log message later, but in the middle there may be return with another msg. So my reasoning has been, lets start each handler with message. Level is 4 so we see this only in debug and test cases. I wanted to remove comment above log line because this does not add more value, level 4 of logging shows already it's for dev. purposes.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -209,14 +236,21 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	if err := ns.mount(srcPath, targetPath, mountFlags); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	ns.volReadOnly[req.VolumeId] = readOnly
+	ns.volTargetPath[req.VolumeId] = targetPath
+	// before storing: sort mountFlags slice and join to become single string, as this is what we use to compare
+	askedFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
+	sort.Strings(askedFlags)
+	joinedAskedFlags := strings.Join(askedFlags[:], ",")
```
avalluri: Some code duplication, that we can avoid by reusing the already prepared mount flags array here: https://github.com/intel/pmem-csi/pull/463/commits/fd89169bb0b4c245157c40d554035d0a5353bae8#diff-4a0caa0c2e37db715640b7cbf97d7a1fR212-R214

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -374,6 +413,7 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 		return nil, err
 	}
 
+	ns.volFstype[req.VolumeId] = ""
```
avalluri: When exactly we clean/remove the cache cahced map(s) elements?

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -199,14 +199,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 		if !notMnt {
-			// TODO(https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/95):
 			// Check if mount is compatible. Return OK if these match:
 			// 1) Requested target path MUST match the published path of that volume ID
 			// 2) VolumeCapability MUST match
-			//   Note that we do not check VolumeCapability/fsType as fsType may not be present (is used in NodeStageVolume scope).
-			//   Means, we check only mountflags part of Capability.
-			//   Mount flags are coded as []string, so we sort and join mount flags to single string
-			//   for easier comparing.
+			//    We do not check VolumeCapability/fsType as fsType may be missing (is used in NodeStageVolume).
```
avalluri: These remvoed comments are added part of this PR, I feel like you can squash with your earlier commit which introduced these comments. I am ok to completely sqashing this commit too.

pkg/pmem-csi-driver/nodeserver.go
avalluri: > My idea was to have log msg in all handling functions near start, so that we see the start of handling.

I fully agree with your point. My point was we can remove the existing log message in line 349, which is a duplicate of this newly added logging line. Or I missing something?

avalluri: Opps sorry, both are diffrent functions. Nevermind

pkg/pmem-csi-driver/nodeserver.go
okartau: yes, good idea. I started with that, but I realized now the creation and deletion happen in two places, so we can not just conveniently get rid of whole map entry in one place.
Another related question arises: I added map for fsType, which is there now for completeness, but is not actually used. This part is created and used in different function.
Can we perhaps leave fsType out, the we can have one map handling in single place.

pkg/pmem-csi-driver/nodeserver.go
okartau: Changed in current code, just pushed

pkg/pmem-csi-driver/nodeserver.go
okartau: The code above is inside if-block so it is not guaranteed it has created that value.

pkg/pmem-csi-driver/nodeserver.go
okartau: thanks, indeed I should use delete() instead of just overwriting with empty, added delete() calls in latest code.

pkg/pmem-csi-driver/nodeserver.go
okartau: thanks, yes I had changed in later commit that I had created in previous one. I now combined the changes of text-changing commit into previous bigger commit.

okartau:
addressed recent commits, force-pushed, also e2e testing that code locally while CI is not yet functioning.

pkg/pmem-csi-driver/nodeserver.go
okartau: what about we move that computation above the "if" so that it happens always once?

pohly:
We have good CI results for this one. @avalluri, is it okay to merge?

pkg/pmem-csi-driver/nodeserver.go
avalluri: > creation and deletion happen in two places, so we can not just conveniently get rid of whole map entry in one place.

Yes, If I understood correctly the change, the two possible places depend on type of volume: ephemeral or persistent.  So for persistent volumes the point of creation/deletion of map entry in Stage/Unstage and for ephemeral its publish/unpublish.

pkg/pmem-csi-driver/nodeserver.go
avalluri: > what about we move that computation above the "if" so that it happens always once?

Yes I think, as we always store that info anyway.

okartau:
I pushed a version where mountflags sorting for storing is made above "if" to have single instance of that code. 
This PR needs more work to handle fsType values storing, based on off-line discussion we had with @avalluri, thus restoring WIP in title

okartau:
pushed with added code to compare also fsType, but only if present in request.
But this now happens only in NodePublish. I wonder should we also add fsType check request vs existing in NodeStage.

okartau:
pushed new version where main improvement commit is changed so that it get fsType from device instead of storing in map.

okartau:
pushed a new version again, checking of existing file system is moved out from ProvisionDevice.

okartau:
One more thing, let's combine the change here once more with #453 and see the CI result. It used to be combined, then I removed #453 part, but then we changed it some more, so the latest state or PR change is not verified combined with repeated tests.

okartau:
As this CI failure here demonstrates (and also failure on #453 after rebase), recent change to strictly forcing vendor/ state takes away method that I use for development purposes: having local changes in devel/, but only in PR, not to be merged. I combine such changes with other PRs to generate a state with those changes that are not (yet) upstreamed. Is this how we want it, (i.e. we cant put changes through CI with intentional vendor/ changes that we know of but we plan not to check in), or could we have some method to work around that check?
I understand this protects us against unintentional vendor/ changes, but what about intentional.

pohly:
> I understand this protects us against unintentional vendor/ changes,
> but what about intentional.

Remove the "make test" step in the Jenkinsfile?

okartau:
> Remove the "make test" step in the Jenkinsfile?

Tried that in #453 but it fails in dm-test. Does dm-test depend on something that happens in 'make test' ? It's not that obvious from output what went wrong, or is it for experienced eye...
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-453/4/pipeline

avalluri:
```
[2019-11-25T15:05:08.372Z] + make run_dm_tests
[2019-11-25T15:05:08.372Z] curl -L https://github.com/govm-project/govm/releases/download/0.9-alpha/govm_0.9-alpha_Linux_x86_64.tar.gz -o /mnt/workspace/pmem-csi_PR-453/_work/govm_0.9-alpha_Linux_amd64.tar.gz
[2019-11-25T15:05:08.372Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2019-11-25T15:05:08.372Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2019-11-25T15:05:08.372Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   625    0   625    0     0   2192      0 --:--:-- --:--:-- --:--:--  2192
[2019-11-25T15:05:08.372Z] Warning: Failed to create the file 
[2019-11-25T15:05:08.372Z] Warning: /mnt/workspace/pmem-csi_PR-453/_work/govm_0.9-alpha_Linux_amd64.tar.gz
[2019-11-25T15:05:08.372Z] Warning: : No such file or directory
```
faile at the `govm `installation step.

okartau:
you mean failure was independent of skip of 'make test' step ? and it could work when retried?

avalluri:
> you mean failure was independent of skip of 'make test' step ? and it could work when retried?

It should but, `_work/govm_$(GOVM_VERSION)_Linux_amd64.tar.gz:` target is not creating `_work` dir. This was not visible as `make test`(maybe `setup-ca.sh`) crates that.

pohly:
> Tried that in #453 but it fails in dm-test. Does dm-test depend on
> something that happens in 'make test' ?

Let's just get rid of it, okay?

See https://github.com/intel/pmem-csi/pull/473

okartau:
CI success in combination with repeated tests, removing WIP and repeating-dev-debug commits, force pushed

pohly:
Is this ready for merging or does it need another round of review by @avalluri?

avalluri:
> Is this ready for merging or does it need another round of review by @avalluri?

I can be merged., but needs a follow-up PR to cover the points mentioned here https://github.com/intel/pmem-csi/pull/463#pullrequestreview-322116750

okartau:
> but needs a follow-up PR

I created issue #483, will merge this.

PR #470, opened 2019-11-18
Go version

pohly:
Fixes a typo (G0 instad of GO) and adds a warning that tells developers when they are using a different Go version.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,7 +6,7 @@
 ! pkg/**
 ! test/test-config.d/**
 ! test/test-config.sh
-! hack/copy-modules-license.sh
+! hack/**
 ! vendor/**
 ! go.mod
 ! go.sum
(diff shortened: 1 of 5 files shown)


okartau:
makes sense, I see go version warning on a system that has older go 1.12.
Doc change also makes sense.
CI is not functional right now, I think we want to wait for that result

pohly:
Added one change to `.dockerignore` to ensure that `hack/verify-go-version.sh` is available in Docker - see updated commit message.

pohly:
@okartau: do you agree with the additional change (https://github.com/intel/pmem-csi/pull/470/files#diff-f7c5b4068637e2def526f9bbc7200c4e): 
> The "hack" directory should be small enough that we can make all of it
> available for Docker builds. That way we don't need to whitelist
> individual files.

okartau:
> do you agree with the additional change

yes!

PR #473, opened 2019-11-21
device manager tests + kustomize support for end-users

pohly:
This builds on top of https://github.com/intel/pmem-csi/pull/420 and enables that new test in the CI by running it and all other Go tests in the existing steps and VMs. This makes the CI run a bit faster (3:24min, see https://github.com/intel/pmem-csi/pull/420#issuecomment-553835763), keeps the Jenkins UI a bit more manageable (the current number of steps already require scrolling), and last but not least, we get better code coverage (all Go tests run on all distros, in exactly the same environment (= container) as the PMEM-CSI driver itself).

To run the device manager inside the normal test cluster, we need some free PMEM. To achieve that, kustomization of the PMEM-CSI driver deployment with `kubectl kustomize` gets enabled and used during the cluster setup. This feature should also be useful for admins because it is an easy way to move the driver into its own namespace.

Finally, this adds some testing of the `pmem-ns-init` parameters, something that we didn't have before.


pohly:
@okartau, @avalluri: can you both please review this?

The 1.15 Clear Linux production cluster fails to come up and I'm investigating that right now, but other than that this PR is ready.

pohly:
> The 1.15 Clear Linux production cluster fails to come up and I'm investigating that right now

Fixed. The symlinks in the 1.15 directory were incomplete.

test/e2e/gotests/gotests.go
```diff
@@ -0,0 +1,66 @@
...
+	framework.ExpectNoError(err, "list PMEM-CSI pods")
+	Expect(pods.Items).NotTo(BeEmpty(), "have PMEM-CSI pods")
+	pmem := pods.Items[0]
+
+	By(fmt.Sprintf("Running in PMEM-CSI pod %s", pmem.Name))
+	pod.RunInPod(f, root,
+		[]string{"_work/test.test", "_work/evil-ca", "_work/pmem-ca"},
+		"if _work/test.test -h 2>&1 | grep -q ginkgo; then "+
+			"TEST_WORK=_work _work/test.test -test.v -ginkgo.v; else "+
+			"TEST_WORK=_work _work/test.test -test.v -ginkgo.v; fi",
```
avalluri: Nothing changes in `if` and `else` cases.

avalluri:
@pohly I try to understand the comments you made in commit https://github.com/intel/pmem-csi/pull/473/commits/43f3735d3a37c9b731f97c27c280ff7011939144: 
> This has multiple advantages:
> - we test in exactly the environment in which the code will run

Other than the device manager what other unit tests in your opinion require(benefit from) such an environment.

> Because it runs so quickly, it's okay to simply repeat this testing in
> each CI step which runs E2E tests.

Currently, running all(registry server, device manager, and state) unit tests taking around 27s in each run. Do you think this argument is still valid once we improve our unit tests?

test/e2e/gotests/gotests.go
```diff
@@ -0,0 +1,66 @@
...
+// runGoTest builds and copies the Go test binary into the PMEM-CSI
+// driver container and executes it there. This way it runs in exactly
+// the same environment as the driver (the distro's kernel, our
+// container user space), which may or may not expose bugs that are
+// not found when running those tests on the build host.
+func runGoTest(f *framework.Framework, pkg string) {
+	root := os.Getenv("REPO_ROOT")
+	var err error
+
+	build := exec.Command("/bin/sh", "-c", os.Getenv("TEST_CMD")+" -c -o _work/test.test "+pkg)
```
avalluri: How about making a single binary for all unit tests and run in a single stretch.

test/e2e/gotests/gotests.go
pohly: Good catch, that `else` case must drop the `-ginkgo.v`. Apparently all our current unit tests are Ginkgo tests suites.

test/e2e/gotests/gotests.go
pohly: We would have to reorganize the test source code. I prefer to keep it decentralized (one test per package).

test/e2e/gotests/gotests.go
pohly: Fixed.

PR #474, opened 2019-11-22
RFC: #420 + Device manager tests in same cluster as e2e tests

avalluri:
To avoid extra time consumed to create a new VM to run device manager tests in CI, Now we run them at the time of running e2e tests so that we can reuse the same cluster.


avalluri:
@pohly Does [this change](https://github.com/intel/pmem-csi/pull/474/commits/8ea3940d0120ba455a5c2079fcdfa55df4ccaada) makes any sense for your concern about the increase in CI time.

pohly:
> Does this change makes any sense for your concern about the increase in CI time.

I don't want special cases for individual tests. If `run_dm_tests` is how you intend to run this test that's fine, but I prefer a uniform way of running tests.

avalluri:
Closing this in favour of #473

Issue #450, opened 2019-10-28
test that MAP_SYNC works


pohly:
We currently lack tests that verify that the volumes provided by PMEM-CSI really are backed by PMEM. What we promise is that `mmap(MAP_SYNC)` works, so we should also test exactly that.

We need to test this also with raw block volumes. If possible, just mmap the special device file.

pohly:
We can implement this as a sanity or E2E test. I'm leaning towards doing it as a proper E2E test. Because we have to run in a cluster, sanity tests aren't that much simpler than a proper E2E test.

PR #476, opened 2019-11-22
dax tests + raw block volumes

pohly:
We want to know whether mmap(MAP_SYNC) really works.

This builds on top of the infrastructure in PR #473, so that needs to be merged first.

Having to fork Kubernetes is unfortunate, but simpler than copying the entire code that we need. Let's wait for some feedback on https://github.com/kubernetes/kubernetes/pull/85540 before merging this.

Fixes: #450


pohly:
Rebased and updated to reflect that `mmap(MAP_SYNC)` does not work directly for the raw block device. At least I didn't know that - learned something new this week. :man_shrugging: 

This changes the value proposition for raw block volumes. I'm not sure anymore how useful they really are.

pohly:
> Having to fork Kubernetes is unfortunate, but simpler than copying the entire code that we need. Let's wait for some feedback on kubernetes/kubernetes#85540 before merging this.

It turned out that Google made the same change in their own fork of Kubernetes, so there is support upstream for this API change. It's now just a matter of getting it reviewed and tested. I think we could merge this PR if we agree that the test is worth it (IMHO it is).

pohly:
I started extending the test to raw block volumes. Good that I did that, because there are additional complications:
- https://github.com/kubernetes/kubernetes/issues/85624 has to be worked around before mkfs+mount work inside the privileged application container
- with the default 110Mi claim size the resulting raw block volume doesn't support `-odax`. 8Gi works. To be investigated further...

pohly:
This has passed CI testing, please review.

pohly:
> with the default 110Mi claim size the resulting raw block volume doesn't support -odax. 8Gi works. To be investigated further...

The reason was the default block size chosen by mkfs.ext4 when `-b` is not given. Adding `-b` as we do in PMEM-CSI itself made it work also for smaller volumes.

README.md
```diff
@@ -775,13 +775,33 @@ parameters:
...
+Applications can use volumes provisioned by PMEM-CSI as [raw block
+devices](https://kubernetes.io/blog/2019/03/07/raw-block-volume-support-to-beta/). Such
+volumes use the same "fsdax" namespace mode as filesystem volumes
+and therefore are block devices. That mode only supports dax (=
+`mmap(MAP_SYNC)`) through a filesystem. Pages mapped on the raw block
+device go through the Linux page cache. Applications have to format
+and mount the raw block volume themselves if they want dax. The
+advantage then is that they have full control over that part.
+
+Volumes with "devdax" namespace mode are not supported. That mode
```
avalluri: I feel like mentioning here in Blockvolumes section about `devdax` is not appropriate as we have not mentioned `devdax` as supported namespace mode anywhere in our documentation

cmd/pmem-dax-check/main.go
```diff
@@ -0,0 +1,107 @@
+/*
```
avalluri: I believe, the better place for this code would be under `test/` as it's only used for E2E tests.

README.md
pohly: Which is exactly the problem that I am trying to solve here: if we don't mention it anywhere, how does a reader know that it really isn't supported?

This felt like a good place because in some ways it would be more natural to use "devdax" for raw block devices. Some users might come here with that expectation.

cmd/pmem-dax-check/main.go
pohly: Currently yes, but this might change. It's a rather useful tool that I have used quite a lot lately also outside of the automated tests. I'd prefer to keep it here, but if you disagree then I'll move it.

README.md
avalluri: We have listed supported namespace modes under `Architecture and Operation` section.
> |Namespace modes    |*fsdax, sector* mode<sup>3</sup> namespaces pre-created as pools   |namespace in required mode created directly, no need to pre-create pools   |

If you would like to add a note about why we are not supporting `devdax` that would be the right place in my opinion.

cmd/pmem-dax-check/main.go
avalluri: I would prefer to keep in `test/`, It's good to know what @okartau feels.

README.md
pohly: Moved.

cmd/pmem-dax-check/main.go
pohly: I moved it.

test/e2e/storage/dax/dax.go
```diff
@@ -0,0 +1,247 @@
...
+			Namespace: f.Namespace.Name,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					SecurityContext: &v1.SecurityContext{
+						Privileged: &privileged,
+					},
+					Name:    containerName,
+					Image:   os.Getenv("PMEM_CSI_IMAGE"),
```
avalluri: I think we may use any generic(maybe ubuntu) image here which has `mknod` and other filesystem utilities installed right. `PMEM_CSI_IMAGE` here is bit misleading. Running this dax test outside our driver image is preferable in my opinion.

test/e2e/storage/dax/dax.go
```diff
@@ -0,0 +1,247 @@
...
+	podErr := e2epod.WaitForPodRunningInNamespace(f.ClientSet, createdPod)
+	framework.ExpectNoError(podErr, "running pod")
+	if volumeMode == v1.PersistentVolumeBlock {
+		By("mounting raw block device")
+		// TODO: remove the workaround above and script invocation here.
+		pmempod.RunInPod(f, l.root, nil, "/data/create-dax-dev.sh && mkfs.ext4 -b 4096 /dax-dev && mkdir -p /mnt && mount -odax /dax-dev /mnt", ns, pod.Name, containerName)
+	}
+
+	By("checking that missing DAX support is detected")
+	pmempod.RunInPod(f, l.root, []string{l.daxCheckBinary}, l.daxCheckBinary+" /no-dax; if [ $? -ne 1 ]; then echo should have reported missing DAX >&2; exit 1; fi", ns, pod.Name, containerName)
```
avalluri: @pohly I couldn't get this point. Where we are copying the test binary to the target container? Can you please help me to understand.

test/e2e/storage/dax/dax.go
pohly: Then we need to pull another, potentially large image into the cluster. We also don't have control over the content of that image and would have to lock onto a specific, soon out-dated version to avoid a potential regression should "ubuntu:latest" ever drop required tools (like mkfs.ext4).

I chose the PMEM-CSI image because it's already on the cluster and we know that it has the necessary tools.

test/e2e/storage/dax/dax.go
pohly: It's done by `RunInPod`, see https://github.com/intel/pmem-csi/blob/46259b59ace37e69b2678a1500a65c2372e2d069/test/e2e/pod/exec.go#L24-L41

avalluri: Oops, sorry I was bit lazy :). Thanks.

test/e2e/storage/dax/dax.go
avalluri: Ok, let's keep this way.

PR #477, opened 2019-11-25
CI: run under normal user inside Docker

pohly:
The recently added vendor check rewrites "vendor" inside Docker, which
later broken working with the repo because those files then became
owned by root.

We could run "sudo chown" after each "docker run" invocation, but it
is safer and faster to not run as root inside Docker in the first
place.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,11 +23,7 @@ ARG GO_VERSION="1.13.4"
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN echo "Updating build image from ${CLEAR_LINUX_BASE} to ${SWUPD_UPDATE_ARG:-the latest release}."
-# openssh-client and sudo are needed for CI testing. They are small enough that it doesn't
-# affect developers that they get installed here, but it is a small saving in the CI
-# because then the installation can be cached.
-RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic openssh-client sudo && rm -rf /var/lib/swupd /var/tmp/swupd
-
+RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic && rm -rf /var/lib/swupd /var/tmp/swupd
 # Workaround for "pkg-config: error while loading shared libraries" when using older Docker
 # (see https://github.com/clearlinux/distribution/issues/831)
 RUN ldconfig
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,11 @@ ARG GO_VERSION="1.13.4"
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN echo "Updating build image from ${CLEAR_LINUX_BASE} to ${SWUPD_UPDATE_ARG:-the latest release}."
-RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic && rm -rf /var/lib/swupd /var/tmp/swupd
+# openssh-client and sudo are needed for CI testing. They are small enough that it doesn't
+# affect developers that they get installed here, but it is a small saving in the CI
+# because then the installation can be cached.
+RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic openssh-client sudo && rm -rf /var/lib/swupd /var/tmp/swupd
+
 # Workaround for "pkg-config: error while loading shared libraries" when using older Docker
 # (see https://github.com/clearlinux/distribution/issues/831)
 RUN ldconfig
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,8 +48,6 @@ pipeline {
         /* last version before the 1.14 update in 28630 */
         CLEAR_LINUX_VERSION_1_13 = "28620"
 
-        // path for placing the source to build outside of GOROOT
-        PMEM_PATH = "/src/pmem-csi"
         REGISTRY_NAME = "cloud-native-image-registry.westus.cloudapp.azure.com"
 
         // Per-branch build environment, marked as "do not promote to public registry".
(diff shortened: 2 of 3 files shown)


pohly:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/devel/109/console shows the problem:
```
15:22:25  + git checkout FETCH_HEAD
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/.gitignore': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/LICENSE': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/README.md': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/constants.go': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/device.go': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/format.go': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/info.go': Permission denied
15:22:25  warning: unable to unlink 'vendor/gopkg.in/freddierice/go-losetup.v1/losetup.go': Permission denied
15:22:25  error: unable to unlink old 'vendor/modules.txt': Permission denied
15:22:25  Note: checking out 'FETCH_HEAD'.
```

pohly:
@avalluri: please have a look. This was rebased and should now work (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-477/12 is passed the point that was previously failing mysteriously - the suppressed sshkey-gen stderr made it a bit harder to understand the root cause = missing user entry in /etc/passwd).

pohly:
Spoke too soon. The cluster started, but didn't actually work and I still had a debug "sleep" in it.

Looks like "sudo" and root-privileges will be needed after all:
```
15:33:33  test/start-kubernetes.sh: line 177: sudo: command not found
15:33:33  ERROR: failed to mount tmpfs with size 1073741824 on /mnt/workspace/pmem-csi_PR-477/_work/clear/data/pmem-csi-clear-master/etcd-volume
```

pohly:
This took a while to converge on a solution...

I was experimenting for a while with a combination of "docker create + docker start + docker exec" instead of "docker exec --rm" (i.e. start build container only once for the entire CI run, then run different commands in it), but that had drawbacks (like "docker exec" not supporting `--group-add`) and ultimately didn't work (despite `--privileged`, a command running under sudo was unable to execute privileged operations like filesystem mounting).

Now I am updating the build container image once at the beginning with changes that shouldn't be in our Dockerfile (like configuring sudo) and continued to use "docker exec --rm". To avoid repeated downloading, the required tools still get installed in the Dockerfile.

@avalluri: please have a look.

Dockerfile
```diff
@@ -23,7 +23,11 @@ ARG GO_VERSION="1.13.4"
 #pull dependencies required for downloading and building libndctl
 ARG CACHEBUST
 RUN echo "Updating build image from ${CLEAR_LINUX_BASE} to ${SWUPD_UPDATE_ARG:-the latest release}."
-RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic && rm -rf /var/lib/swupd /var/tmp/swupd
+# openssh-client and sudo are needed for CI testing. They are small enough that it doesn't
+# affect developers that they get installed here, but it is a small saving in the CI
+# because then the installation can be cached.
+RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add ${NDCTL_BUILD_DEPS} c-basic openssh-client sudo && rm -rf /var/lib/swupd /var/tmp/swupd
```
avalluri: @pohly Is not enough installing these packages new packages in `env,.BUILD_IMAGE`? I would prefer if we can avoid ending-up these CI specific packages in the final image.

Jenkinsfile
```diff
@@ -125,6 +123,29 @@ pipeline {
...
+                    sh "echo 'jenkins:*:0:0:99999:0:::' | docker exec -i builder tee --append /etc/shadow >/dev/null"
+
+                    // Now commit those changes for use below and clean up.
+                    sh "docker commit builder ${env.BUILD_IMAGE}"
+                    sh "docker stop builder"
+                    sh "docker rm builder"
+
+                    // Verify that docker and mount work under sudo in the updated image (needed for etcd tmpfs, see start-kubernetes.sh).
+                    // sudo depends on --privileged.
+                    sh "docker run --rm ${DockerBuildArgs()} ${env.BUILD_IMAGE} docker ps"
```
avalluri: I guess these below steps are added as part of debugging, I think we can remove these checks from the final version?

Jenkinsfile
```diff
@@ -342,7 +363,13 @@ git push origin HEAD:master
  "docker run" parameters which:
  - make the Docker instance on the host available inside a container (socket and command)
  - set common Makefile values (cachebust, cache populated from images if available)
- - source in $GOPATH as current directory
+ - source in current directory
+ - GOPATH alongside it
+ - HOME above it
```
avalluri: I couldn't understand why we choose this restriction of `HOME` and `GOPATH` relative to the current directory.

Dockerfile
pohly: > in the final image

This isn't the final image. It's the build image which gets only used for preparing the binaries which then later get copied into the final image.

This approach is a tradeoff between "cleanliness" and "performance". I agree that it would be cleaner to do the installation in the Jenkinsfile, but then we have to run `swupd bundle-add` every single time that a Jenkins job runs. The way it is now, the build image gets cached in the private Docker registry.

Having said that, at least on my desktop machine `swupd bundle-add sudo openssh-client` is fast (just a few seconds). But I'd prefer to keep it here nonetheless because performance and reliability of remote network access is a bit unpredictable - it might be slower in the CI.

Jenkinsfile
pohly: I kept them intentionally because these are fast, local operations which ensure that a CI job gets aborted early if (for whatever reason) these crucial aspects of the build container no longer work.

Jenkinsfile
pohly: Jenkins is configured to invoke us with the current directory on a fast, local disk. I could have hard-coded `/mnt/workspace` here, but that is only true for the current Jenkins configuration, so I use the current directory instead to find that path.

`HOME` gets set to the parent for code which uses `$HOME/.cache` - we want that to be fast.

`GOPATH` is then redundant because it defaults to something in `$HOME` (don't remember what exactly it is). It gets explicitly merely to document what it is.

Dockerfile
avalluri: > I agree that it would be cleaner to do the installation in the Jenkinsfile, but then we have to run `swupd bundle-add` every single time that a Jenkins job runs.

Does installing them in the `Create build environment` stage helps? As this change altering the build image anyway.

pohly: Hmm, a bit. We still run "swupd bundle-add", but the result should end up in the cached image. Okay, let's move it.

pohly: Done.

Jenkinsfile
avalluri: Thanks for the clarification.

PR #478, opened 2019-11-27
update Clear Linux and CRI support

pohly:
We should better test with a recent Clear Linux. On that release we can also enable "containerd" as runtime.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,8 +126,8 @@ pipeline {
                     sh "env; echo Building BUILD_IMAGE=${env.BUILD_IMAGE} for BUILD_TARGET=${env.BUILD_TARGET}, CHANGE_ID=${env.CHANGE_ID}, CACHEBUST=${env.CACHEBUST}."
                     sh "docker build --cache-from ${env.BUILD_IMAGE} --label cachebust=${env.CACHEBUST} --target build --build-arg CACHEBUST=${env.CACHEBUST} -t ${env.BUILD_IMAGE} ."
                     // Create a running container (https://stackoverflow.com/a/38308399).
-                    sh "docker create --name=builder -it ${env.BUILD_IMAGE}"
-                    sh "docker start builder"
+                    sh "docker create --name=builder ${env.BUILD_IMAGE} sleep infinity"
+                    sh "docker start builder && \
                         timeout=0; \
                         while [ \$(docker inspect --format '{{.State.Status}}' builder) != running ]; do \
                             docker ps; \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -128,6 +128,17 @@ pipeline {
                     // Create a running container (https://stackoverflow.com/a/38308399).
                     sh "docker create --name=builder -it ${env.BUILD_IMAGE}"
                     sh "docker start builder"
+                        timeout=0; \
+                        while [ \$(docker inspect --format '{{.State.Status}}' builder) != running ]; do \
+                            docker ps; \
+                            if [ \$timeout -ge 60 ]; then \
+                               docker inspect builder; \
+                               echo 'ERROR: builder container still not running'; \
+                               exit 1; \
+                            fi; \
+                            sleep 10; \
+                            timeout=\$((timeout + 10)); \
+                       done"
 
                     // Install additional tools:
                     // - ssh client for govm
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -164,6 +164,10 @@ pipeline {
                     sh "docker run --rm ${DockerBuildArgs()} ${env.BUILD_IMAGE} docker ps"
                     sh "docker run --rm --privileged ${DockerBuildArgs()} ${env.BUILD_IMAGE} bash -c \
                             'mkdir /tmp/mnt && sudo mount -osize=4096 -t tmpfs none /tmp/mnt && sudo umount /tmp/mnt'"
+
+                    // Run a per-test registry on the build host.  This is where we
+                    // will push images for use by the cluster during testing.
+                    sh "docker run -d -p 5000:5000 --restart=always --name registry registry:2"
                 }
             }
         }
@@ -205,7 +209,17 @@ pipeline {
             }
 
             steps {
+                // This builds images for REGISTRY_NAME with the version automatically determined by
+                // the make rules.
                 sh "docker run --rm ${DockerBuildArgs()} ${env.BUILD_IMAGE} make build-images"
+
+                // For testing we have to have those same images also in a registry. Tag and push for
+                // localhost, which is the default test registry.
+                sh "imageversion=\$(docker run --rm ${DockerBuildArgs()} ${env.BUILD_IMAGE} make print-image-version) && \
+                    for suffix in '' '-test'; do \
+                        docker tag ${env.REGISTRY_NAME}/pmem-csi-driver\$suffix:\$imageversion localhost:5000/pmem-csi-driver\$suffix:\$imageversion && \
+                        docker push localhost:5000/pmem-csi-driver\$suffix:\$imageversion; \
+                    done"
             }
         }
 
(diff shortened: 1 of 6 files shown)


pohly:
This has failed for "1.16 direct", which shouldn't be affected by this PR at all: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-478/1/pipeline

@okartau, @avalluri: there are multiple "pfn: no seed" errors, which comes from https://github.com/intel/pmem-csi/blob/09b139dbb08181896793716d2e33c828735adbcb/pkg/ndctl/namespace.go#L360

What is a "pfn" and what does this error mean?

okartau:
I am also worried why CI testing started to fail more frequently recently. My 2 PRs both fail in CI, one with similar pfn errors, but other just getting different errors in gotests.
BTW, with recent addition to gotests, the usual way of checking number of errors (I used to read end of test to see number of failures) does not work any more, as 
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi/branches/PR-453/runs/6/nodes/64/steps/69/log/?start=0
reports 4 FAILs at the end, but there are more in the middle, like 8 gotests failures
at [2019-11-27T13:50:05.451Z]

There are different gotests failures like tar: _work/evil-ca: Cannot stat: No such file or directory
None of new failures seem to be related to changes in PR.

But the again, I tried to run e2e locally in direct mode, it passed once just OK.

pfn is Page Frame Number and is used by libndctl to manage the actual namespaces and pages phys. mapping. "no pfn seed " means ndctl_region_get_pfn_seed(ndr) failed, but to get more info, we need to enable more tracing in libndctl code.
Is the setup of ndctl storage happening properly, the labeling steps?
Is there been some recent changes in related areas?

pohly:
It looks like Docker on recent Clear Linux is no longer stable. In https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-478/3/consoleFull, the cluster came up, but then in the middle of testing containerd, docker and kubelet all went down (probably in this order) and thus the node disappeared.

I propose we switch to cri-o as default runtime. For that I need to extend the setup-fedora.sh script, because there we only support Docker at the moment.

pohly:
Hrmmm. There are no pre-built crio-binaries except some oldish ones from CentOS (https://github.com/cri-o/cri-o/issues/1141). Perhaps containerd (directly, without Docker) is the better choice?

avalluri:
> This has failed for "1.16 direct", which shouldn't be affected by this PR at all: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-478/1/pipeline
> 
> @okartau, @avalluri: there are multiple "pfn: no seed" errors, which comes from
> 
> https://github.com/intel/pmem-csi/blob/09b139dbb08181896793716d2e33c828735adbcb/pkg/ndctl/namespace.go#L360
> 
> What is a "pfn" and what does this error mean?

We fail to get pfn seed device pointer to the given region, pfn holds some metadata of fsdax namespace. According to [nvdimm documentation](https://nvdimm.wiki.kernel.org/)  `persistence of the PFN metadata is not important; this is just convenient because it scales with the persistent memory capacity`. But I am not sure what are the side effects if we ignore these failures and use such namespace.

pohly:
This code only runs in the CI and assumes that Docker is active: https://github.com/intel/pmem-csi/blob/9fda32212d8219e8b36244c8da82197852161ef6/test/setup-kubernetes.sh#L21

That's not the case anymore when choosing cri-o... maybe something can be done with crictl instead.

pohly:
@avalluri, @okartau: please review. This now also includes the fix for "build container not running".

The revised solution with running the Docker registry on the build host is much simpler and closer to how we work during development. Not sure why that wasn't done in the first place.

Jenkinsfile
```diff
@@ -439,10 +453,9 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                   --privileged \
                   -e CLUSTER=clear \
                   -e GOVM_YAML=`pwd`/_work/clear/deployment.yaml \
-                  -e TEST_BUILD_PMEM_REGISTRY=${env.REGISTRY_NAME} \
+                  -e TEST_LOCAL_REGISTRY=\$(ip addr show dev docker0 | grep ' inet ' | sed -e 's/.* inet //' -e 's;/.*;;'):5000 \
```
avalluri: @pohly I am not clear at this point. Above you are tagging the images as "localhost:5000/..." and here using docker ip as `TEST_LOCAL_REGISTRY`.  Is it that we have to tag the images also with same ip, something like: `172.17.0.1:5000/.`.

Jenkinsfile
```diff
@@ -127,7 +127,18 @@ pipeline {
                     sh "docker build --cache-from ${env.BUILD_IMAGE} --label cachebust=${env.CACHEBUST} --target build --build-arg CACHEBUST=${env.CACHEBUST} -t ${env.BUILD_IMAGE} ."
                     // Create a running container (https://stackoverflow.com/a/38308399).
                     sh "docker create --name=builder -it ${env.BUILD_IMAGE}"
-                    sh "docker start builder"
+                    sh "docker start builder && \
+                        i=0; \
+                        while [ \$(docker inspect --format '{{.State.Status}}' builder) != running ]; do \
+                            docker ps; \
+                            if [ \$i -gt 7 ]; then \
```
avalluri: Nit, If the intention here is to wait for a minute(60sec) would better use `-gt 6`.

Jenkinsfile
pohly: `localhost:5000/pmem-csi-driver:canary" and `172.17.0.1:5000/pmem-csi-driver:canary` are the same image. They just get accessed via different IP addresses.

I could have used `172.17.0.1:5000/pmem-csi-driver:canary` also when pushing, but using `localhost` was easier.

avalluri:
> But as we immediately kille

A small typo in [Commit messge](https://github.com/intel/pmem-csi/pull/478/commits/cbc7adca7af445ab6122d1ad017f0118d9076e72)

Jenkinsfile
pohly: No, of course the intention was to wait for 70s. :grin: 

Will fix together with the commit message typo.

pohly: `-gt 6` is still not correct, has to be `-ge 6`. It doesn't matter, though, which is also why I didn't bother thinking about this earlier.

Jenkinsfile
```diff
@@ -126,18 +126,18 @@ pipeline {
...
                             docker ps; \
-                            if [ \$i -gt 7 ]; then \
+                            if [ \$timeout -ge 60 ]; then \
                                docker inspect builder; \
                                echo 'ERROR: builder container still not running'; \
                                exit 1; \
                             fi; \
                             sleep 10; \
-                            i=\$((\$i + 1)); \
+                            timeout=\$((\$timeout + 10)); \
```
okartau: Doesn't such double-parentheses expression work without inner $-sign equally?

Jenkinsfile
avalluri: >  sh "docker run -d -p 5000:5000 --restart=always --name registry registry:2"

So you mean the above docker registry could be accessed by both `localhost` and `172.17.0.1`? Ok.

Jenkinsfile
pohly: What do you mean with `inner $-sign`?

```
$ echo ((1 + 1))
-bash: syntax error near unexpected token `('
$ echo $((1 + 1))
2
```

Jenkinsfile
pohly: Yes.

Jenkinsfile
okartau: I mean, one can use $((timeout + 10)) instead of $(($timeout + 10)) ?
"inner" was referring to $-sign in the parenthesis

pohly: Apparently one can. Learned something new again...

However, I'd prefer to not start another test run with this change and just get it merged as-is.

pohly: It failed because of https://github.com/intel/pmem-csi/issues/485, so I might as well update it once more...

PR #482, opened 2019-11-28
Use Fedora-31 cloud image for testers

okartau:
Trial to use Fedora-31 cloud image.
In my local setup, after I made these changes, kubernetes setup does not complete on master but fails with timeout. There is kubelet running on master, but networking seems to have problem:
Container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady
Let's see does we see same in CI.

### DIFF
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -20,6 +20,8 @@ function error_handler(){
 }
 trap 'error_handler ${LINENO}' ERR
 
+# add own name to /etc/hosts to avoid external name lookup(s) that produce warning by kubeadm
+echo "127.0.0.1 `hostname`" >> /etc/hosts
 
 # For PMEM.
 packages+=" ndctl"
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -37,7 +37,7 @@ enabled=1
 gpgcheck=1
 gpgkey=https://download.docker.com/linux/centos/gpg
 EOF
-    packages+=" docker-ce-3:19.03.2-3.el7"
+    packages+=" docker-ce-3:19.03.5-3.el7"
 
     # Install according to https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/
     modprobe br_netfilter
@@ -59,10 +59,10 @@ EOF
     # List generated with:
     # for v in 1.13 1.14 1.15 1.16; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo yum --showduplicates list kubelet | grep " $v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u  | tail -n 1)"; done; done
     case ${TEST_KUBERNETES_VERSION} in
-        1.13) packages+=" kubelet-1.13.9-0 kubeadm-1.13.9-0 kubectl-1.13.9-0";;
-        1.14) packages+=" kubelet-1.14.7-0 kubeadm-1.14.7-0 kubectl-1.14.7-0";;
-        1.15) packages+=" kubelet-1.15.4-0 kubeadm-1.15.4-0 kubectl-1.15.4-0";;
-        1.16) packages+=" kubelet-1.16.0-0 kubeadm-1.16.0-0 kubectl-1.16.0-0";;
+        1.13) packages+=" kubelet-1.13.12-0 kubeadm-1.13.12-0 kubectl-1.13.12-0";;
+        1.14) packages+=" kubelet-1.14.10-0 kubeadm-1.14.10-0 kubectl-1.14.10-0";;
+        1.15) packages+=" kubelet-1.15.8-0 kubeadm-1.15.8-0 kubectl-1.15.8-0";;
+        1.16) packages+=" kubelet-1.16.5-0 kubeadm-1.16.5-0 kubectl-1.16.5-0";;
         *) echo >&2 "Kubernetes version ${TEST_KUBERNETES_VERSION} not supported, package list in $0 must be updated."; exit 1;;
     esac
     packages+=" --disableexcludes=kubernetes"
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -90,8 +90,13 @@ if $INIT_KUBERNETES; then
     # Upstream kubelet looks in /opt/cni/bin, actual files are in
     # /usr/libexec/cni from
     # containernetworking-plugins-0.8.1-1.fc30.x86_64.
-    mkdir -p /opt/cni
-    ln -s /usr/libexec/cni /opt/cni/bin
+    # There is reason why we don't create symlink /opt/cni/bin -> /usr/libexec/cni.
+    # Some CNI variants (weave) try to add binaries, which fails if bin/ is symlink
+    # We create /opt/cni/bin containing symlinks for every binary:
+    mkdir -p /opt/cni/bin
+    for i in /usr/libexec/cni/*; do
+        ln -vs $i /opt/cni/bin/
+    done
 
     # Testing may involve a Docker registry running on the build host (see
     # TEST_LOCAL_REGISTRY and TEST_PMEM_REGISTRY). We need to trust that
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -80,9 +80,9 @@ while ! dnf install -y $packages; do
     fi
     cnt=$(($cnt + 1))
     # If it works, proceed immediately. If it fails, sleep and try again without aborting on an error.
-    if ! dnf update --refresh; then
+    if ! dnf update -y --refresh; then
         sleep 20
-        dnf update --refresh || true
+        dnf update -y --refresh || true
     fi
 done
 
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -57,7 +57,7 @@ EOF
 
     # For the sake of reproducibility, use fixed versions.
     # List generated with:
-    # for v in 1.13 1.14 1.15 1.16; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo yum --showduplicates list kubelet | grep " $v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u  | tail -n 1)"; done; done
+    # for v in 1.13 1.14 1.15 1.16; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo dnf --showduplicates list kubelet | grep " $v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u  | tail -n 1)"; done; done
     case ${TEST_KUBERNETES_VERSION} in
         1.13) packages+=" kubelet-1.13.12-0 kubeadm-1.13.12-0 kubectl-1.13.12-0";;
         1.14) packages+=" kubelet-1.14.10-0 kubeadm-1.14.10-0 kubectl-1.14.10-0";;
@@ -73,9 +73,9 @@ fi
 # suggests to try again after a `dnf update --refresh`, so that's what we do here for
 # a maximum of 5 attempts.
 cnt=0
-while ! yum install -y $packages; do
+while ! dnf install -y $packages; do
     if [ $cnt -ge 5 ]; then
-        echo "yum install failed repeatedly, giving up"
+        echo "dnf install failed repeatedly, giving up"
         exit 1
     fi
     cnt=$(($cnt + 1))
--- a/test/start-kubernetes.sh
+++ b/test/start-kubernetes.sh
@@ -351,7 +351,28 @@ function init_kubernetes_cluster() (
     echo "Installing dependencies on cloud images, this process may take some minutes"
     vm_id=0
     pids=""
-    for ip in $(print_ips); do
+    IPS=$(print_ips)
+    # in Fedora-31 case, add a flag to kernel cmdline and reboot
+    if [ "$TEST_DISTRO" = "fedora" -a "$TEST_DISTRO_VERSION" = "31" ]; then
+        SYSTEMD_UNIFIED_CGROUP_HIERARCHY=systemd.unified_cgroup_hierarchy
+        for ip in ${IPS}; do
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo sed -i 's/n8\"/n8 ${SYSTEMD_UNIFIED_CGROUP_HIERARCHY}=0\"/g' /etc/default/grub"
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo grub2-mkconfig -o /boot/grub2/grub.cfg"
+            # use --force to speed up shutdown as nothing important running yet to wait for
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo systemctl --force reboot"
+        done
+        for ip in ${IPS}; do
+            SECONDS=0
+            echo "After reboot: Waiting for ssh connectivity on vm with ip $ip"
+            # Verify that system runs with added cmdline parameter
+            while ! ssh $SSH_ARGS ${CLOUD_USER}@${ip} "grep -q ${SYSTEMD_UNIFIED_CGROUP_HIERARCHY} /proc/cmdline"; do
+                if [ "$SECONDS" -gt "$SSH_TIMEOUT" ]; then
+                    die "timeout waiting for ${ip} to reboot with changed kernel parameters"
+                fi
+            done
+        done
+    fi
+    for ip in ${IPS}; do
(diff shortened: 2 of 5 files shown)


okartau:
About use of TEST_DISTRO_VERSION in tester scripts:
I notice we use same variable for both Clear and Fedora, is this intentional?
It is bit misleading, and that variable is documented in test/test-config.sh to refer to Clear only.
I also added variable to minor release numbering part which changes in F-31 but is hard-coded.

pohly:
> About use of TEST_DISTRO_VERSION in tester scripts:
> I notice we use same variable for both Clear and Fedora, is this intentional?

Yes.

> It is bit misleading,

Why? Both Clear and Fedora are distros, and both have versions. The values are different, but the meaning is the same.

> and that variable is documented in test/test-config.sh to refer to Clear only.

That needs to be fixed.

okartau:
> Why? Both Clear and Fedora are distros, and both have versions. The values are different, but the meaning is the same.

I was thinking of scenario where I want to maintain different override versions, say I want to test with Clear 30080 and Fedora 31 and both are different than default, so I want to set those in test/test-config.sh,
but the name of variable is shared so I cant set two values, so I have to override in different code blocks in test/start-kubernetes.sh, and that's not so clean.

pohly:
>  so I want to set those in test/test-config.sh

In that file itself or in a `test/test-config.d` file? There should not be a need to modify `test/test-config.sh` unless you want those changes to be merged.

But that aside, permanently overriding the default Clear and Fedora versions in such a file is indeed not possible. I'm just wondering how important that is - using something else than colleagues on a permanent basis is potentially problematic, so I would prefer to make that obvious by setting `TEST_DISTRO_VERSION` on the command line when starting a cluster.

okartau:
> There should not be a need to modify `test/test-config.sh` unless you want those changes to be merged.

I have this not-to-be-recommended habit to edit that file in such a tree where state gets dirty from trace/debug/trial changes and I dont plan to create any change requests anyway, so I am too lazy to create new files under test-config.d/.
BTW, when one reads our README doc about that file, we say:
`Several aspects of the cluster and build setup can be configured by overriding the settings in the test-config.sh file.`
Note that at reaching end of that sentence, natural reaction of impatient person is to go and edit that file. Only in following few statements there is the method described how one should change settings without altering that defaults file. But the wording of 1st sentence leave it up to interpreter, do we mean either 1) overriding happens in that file or 2) overriding of defaults present in that file, happens in another place.

okartau:
> I'm just wondering how important that is

not important, true. I just got idea to point out that shared var.name takes away to consistently override both at same time. I understand the way it is implemented now, makes sense internally.

okartau:
> using something else than colleagues on a permanent basis is potentially problematic

true, but OTOH it also widens the coverage. The rest of the world is not guaranteed to stick to one foreseen practice, so if we make side-tracks, we can also discover more problems.

pohly:
>> using something else than colleagues on a permanent basis is potentially problematic
>
> true, but OTOH it also widens the coverage. The rest of the world is
> not guaranteed to stick to one foreseen practice, so if we make
> side-tracks, we can also discover more problems.

True, but then one needs to be aware that one is on a side track. When
putting those changes into a test-config.d file it's easy to forget
about them.

okartau:
By my current understanding, getting k8s up on Fedora-31 requires added kernel cmdline option systemd.unified_cgroup_hierarchy=0.
I was able to get F-31 k8s cluster (1.17) up locally with that change only, keeping the remaining config same. But this is not achievable with current tester setup scripts as we don't support reboot-and-continue before k8s init step. I was thinking about another way, avoiding reboot: one can give kernel options as qemu options. But this requires then using kernel from host side, which would need another not-so-simple set of added steps.
Do we want to complicate our VM setup script with reboot step?
At some point we will be forced to step forward in Fedora releases line.

pohly:
> Do we want to complicate our VM setup script with reboot step?

Isn't that something that we'll also need for Ubuntu to get a dax-enabled kernel?

Adding a reboot step looks like the least bad solution. If possible, keep it optional so that we only do it for those configurations which need it.

okartau:
I added reboot step, but k8s cluster does not start correctly on F-31 when using k8s 1.16.4. I can reproduce that as well on local setup. CNI is not started correctly on all nodes (in recent try I saw one node getting cni running but 3 nodes did not). The reason seems to be outside of pmem-csi scope and symptoms are similar to https://github.com/coreos/flannel/issues/1236 .
k8s on F-31 starts (in local case) when using 1.17, so I force-push such changes for cross-checking on CI (bumped version to 1.17 and added symlink in deploy)

okartau:
I could not get cni0 interface appearing on all nodes, and there is no log info why this happens. I tried to switch to different CNI, After unsuccessful calico tour I tried weave and it worked in local version, so let's see  how it goes in CI. I removed 1.17 related changes from change set that I already had inserted there, as problem we debug seems not to be k8s version dependent after all.
As side effect, I discovered few instances of pkg install cmds without -y option that cause user input prompt and "Aborted" commands, adding fix to those as well here.
Also, new F-31 moves further away from yum so lets use dnf instead.

okartau:
Finally found reason why Weave networking did not start, it was /opt/bin as single symlink. In my local setup that worked, I had used older code for that which created separate symlinks for elements under /opt/cni/bin/. The difference why flannel works, is because flannel already is an element under /opt/cni/bin, but weave wants to add it's binaries there, which fails when bin is symlink and not directory

okartau:
The testin should now at least start (and even expected to succeed).
A remaining issue causes failure in some runs with:
`server.go:271 failed to run Kubelet: mountpoint for cpu not found`

Is this another issue caused by changes in Fedora-31 ?
Search shows older cases mentioning such error, and stating
that mount of  type cgroup on `/sys/fs/cgroup/cpu,cpuacct` would avoid such error.
At least in working case, we have such mount in VM.
But I haven't checked on failing case yet.

okartau:
force-pushed, rebased manually due to conflict in Makefile NO_PROXY line

test/setup-fedora-govm.sh
```diff
@@ -80,9 +80,9 @@ while ! dnf install -y $packages; do
     fi
     cnt=$(($cnt + 1))
     # If it works, proceed immediately. If it fails, sleep and try again without aborting on an error.
-    if ! dnf update --refresh; then
+    if ! dnf update -y --refresh; then
```
okartau: Could the "aborting on an error" mentioned in above comment have been caused by lack of `-y` option ?

pohly: No. I think there were simply some intermittent network issues.

test/setup-fedora-govm.sh
```diff
@@ -90,8 +90,11 @@ if $INIT_KUBERNETES; then
     # Upstream kubelet looks in /opt/cni/bin, actual files are in
     # /usr/libexec/cni from
     # containernetworking-plugins-0.8.1-1.fc30.x86_64.
-    mkdir -p /opt/cni
-    ln -s /usr/libexec/cni /opt/cni/bin
+    mkdir -p /opt/cni/bin
+    # There is reason why we don't create symlink /usr/libexec/cni <- /opt/cni/bin
+    # but instead we create /opt/cni/bin containing symlinks for ever binary:
+    # another CNI variant, like weave wants to add binaries, and that will not work if bin/ is symlink
+    for i in /usr/libexec/cni/*;do ln -vs $i /opt/cni/bin/; done
```
pohly: nit: can you use three lines for the loop?

test/start-kubernetes.sh
```diff
@@ -263,6 +263,21 @@ EOF
...
             fi
         done
+	# in Fedora-31 case, add 'systemd.unified_cgroup_hierarchy=0' to kernel cmdline and reboot
+        if [ "$TEST_DISTRO" = "fedora" -a "$TEST_DISTRO_VERSION" = "31" ]; then
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo sed -i 's/n8\"/n8 systemd.unified_cgroup_hierarchy=0\"/g' /etc/default/grub"
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo grub2-mkconfig -o /boot/grub2/grub.cfg"
+            # use --force to speed up shutdown as nothing important running yet to wait for
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo systemctl --force reboot"
+            # sleep, letting VM go down. I have seen ssh returning zero if running immediately so while-loop does not wait.
+            sleep 5
```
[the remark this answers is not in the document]
okartau: > Also, can trigger the reboot for all machines (i.e. in this while loop) in parallel, then wait for them to come back?

in the place where this code is now, this is not easy, as the code is already inside serial loop over hosts, so I cant introduce parallel operation there. In that loop there is wait for 1st ssh connection, so it also does not seem feasible to move the code ahead of that loop.
But, the 2nd point here makes sense: whole boot-param addition could be moved into `init_kubernetes_cluster` as it makes sense only if INIT_CLUSTER=true,
and there the game is different, so we can try to achieve more parallelism there.
I will try that change in next draft.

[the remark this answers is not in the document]
pohly: You could have stopped the loop after the shutdown request, then restart it afterwards.

okartau:
Tried to address recent comments with newly pushed state. The cmdline option addition and reboot  is moved into cluster setup function, and uses two separate loops as proposed. The 2nd loop checks does cmdline contain new parameter (one more thing to improve there: should avoid two spell/outs of same string, but as 1st one is sed-line inside single quotes, how to use string there..

okartau:
CI job failed in 1st packages install stage, seems unrelated to change. Force pushed for retry (fixed a commit msg).

pohly:
> Force pushed for retry

You can also do that in the Jenkins UI.

test/start-kubernetes.sh
```diff
@@ -351,7 +351,27 @@ function init_kubernetes_cluster() (
...
+            # use --force to speed up shutdown as nothing important running yet to wait for
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo systemctl --force reboot"
+        done
+        for ip in ${IPS}; do
+            SECONDS=0
+            echo "After reboot: Waiting for ssh connectivity on vm with ip $ip"
+            # Verify that system runs with added cmdline parameter
+            while ! ssh $SSH_ARGS ${CLOUD_USER}@${ip} "grep -q systemd.unified_cgroup_hierarchy /proc/cmdline"; do
+                if [ "$SECONDS" -gt "$SSH_TIMEOUT" ]; then
+                    die "timeout accessing ${ip} through ssh"
```
pohly: This is the same error as above, and it can be a bit misleading if (for whatever reason) ssh works, but the `/proc/cmdline` change didn't. Can you change it into `timeout waiting for ${ip} to reboot with new kernel parameters`?

test/start-kubernetes.sh
```diff
@@ -351,7 +351,27 @@ function init_kubernetes_cluster() (
     echo "Installing dependencies on cloud images, this process may take some minutes"
     vm_id=0
     pids=""
-    for ip in $(print_ips); do
+    IPS=$(print_ips)
+    # in Fedora-31 case, add 'systemd.unified_cgroup_hierarchy=0' to kernel cmdline and reboot
+    if [ "$TEST_DISTRO" = "fedora" -a "$TEST_DISTRO_VERSION" = "31" ]; then
+        for ip in ${IPS}; do
+            ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo sed -i 's/n8\"/n8 systemd.unified_cgroup_hierarchy=0\"/g' /etc/default/grub"
```
pohly: > should avoid two spell/outs of same string

I think copying the string is more readable and thus wouldn't use a variable.

> as 1st one is sed-line inside single quotes, how to use string there

This should work:
```
SYSTEMD_UNIFIED_CGROUP_HIERARCHY=systemd.unified_cgroup_hierarchy

ssh $SSH_ARGS ${CLOUD_USER}@${ip} "sudo sed -i 's/n8\"/n8 ${SYSTEMD_UNIFIED_CGROUP_HIERARCHY}=0\"/g' /etc/default/grub"
```

That's because the `${SYSTEMD_UNIFIED_CGROUP_HIERARCHY}` gets expanded as part of the outer double-quoted string, not as part of the inner single-quoted string.

If that wasn't the case, one could combine both inside a single parameter:
```
foo="   world   "
echo 'hello '"${foo}"', how are you?'
```

pohly:
> CI job failed in 1st packages install stage, seems unrelated to change. 

This is 
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-482/21/execution/node/97/log/ right?

```
22:57:25  Waiting for ssh connectivity on vm with ip 172.17.0.4
22:57:53  Waiting for ssh connectivity on vm with ip 172.17.0.5
22:57:53  21:57:50 pmem-csi-govm-master: ssh: connect to host 172.17.0.4 port 22: Connection refused
22:57:53  21:57:50 pmem-csi-govm-master: END OF OUTPUT
```

Master is 172.17.0.4 (can be seen in the post-failure debug output, `e8cb42d33a pmem-csi-govm-master  jenkins   172.17.0.4`). Why did ssh pass the "waiting for ssh" loop and then fail with "Connection refused"?

This a new failure mode and seems related to switching to Fedora 31. If it's now more flaky than before, we need to do something and can't just ignore it.

okartau:
> Why did ssh pass the "waiting for ssh" loop and then fail with "Connection refused"?

My logical guess was, sshd must accept login once, and then going into refusal mode (or stop running).
And indeed, if we read same log further, the explanation is below where dmesg of VM is printed:
Seems the 2nd ssh command hits the window when sshd is stopped (related to keygen?)
and restarted. Is this something happening in F-31 cloud image and is not happening in F-30 ?

```
 govm.jenkins.pmem-csi-govm-master: OK  Started OpenSSH server daemon.

## here user login succeeds, ending the wait-loop. (for some reason, 2 sessions??)
govm.jenkins.pmem-csi-govm-master: OK  Started Session 1 of user fedora.
govm.jenkins.pmem-csi-govm-master: OK  Started Session 3 of user fedora.

## but then, sshd is stopped:
govm.jenkins.pmem-csi-govm-master:          Stopping OpenSSH server daemon...
govm.jenkins.pmem-csi-govm-master:   OK  Stopped OpenSSH server daemon.

## seems, stopping was related to sshd.keygen.target ??
govm.jenkins.pmem-csi-govm-master:  OK  Stopped target msshd-keygen.target.
govm.jenkins.pmem-csi-govm-master:  OK  Reached target msshd-keygen.target.

## and sshd is started again:
govm.jenkins.pmem-csi-govm-master:          Starting OpenSSH server daemon...
govm.jenkins.pmem-csi-govm-master:   OK  Started OpenSSH server daemon.
```

okartau:
In a system where sshd gets enabled, disabled, enabled again, we probably need to retry after short wait.
(which is bit troublesome if we want to add retry loop in all commands, but in current case really only the  next one after initial detection hits the problem).
Or, another idea, lets make initial detection more robust so that it repeats few times with 1s interval and decides only after few OK attempts in a row, that system is up.

pohly:
> Is this something happening in F-31 cloud image and is not happening
> in F-30 ?

At least I haven't seen it there.

> Or, another idea, lets make initial detection more robust so that it
> repeats few times with 1s interval and decides only after few OK
> attempts in a row, that system is up.

That seems more promising.

Or even better (?), use a wrapper around *every* ssh command which can
distinguish between "connection refused" and "command itself failed" and
retries only for the former. The problem will be to determine that based
on stderr and/or exit codes.

okartau:
> Or even better (?), use a wrapper around *every* ssh command

I think thats a promising plan. If we can have wrapper, then we dont need to have those special "wait ssh responds" loops, but we can just call ssh wrapper which waits when needed (and we need to wrap also scp and use similar checker there before command.

But meanwhile, as I had already made that change, I pushed another try with existing method,
where the first "wait for ssh" loop is re-arranged so that it waits for all VMs, then makes another loop over VMs. If the issue happens on master only because of no-delay continuation of next ssh, this re-arranged order may save us from hitting this dead window.

okartau:
I think the trial of separated wait-loop for first connection fixes (or at least improves significantly) the issue: In logs it can be seen that now there is approx. 2.5 seconds delay from 1st VM (master) getting ssh-responsive, to the point when next ssh is attempted to same VM (the time spent during wait for other 3 nodes to become responsive). This delay most probably covers the short window of sshd service break.
Let's monitor this approach through some other improvements that I try to make here (like replacing that cmdline string with variable for example).
The more proper way of course is to introduce ssh/scp wrapper that can check, wait and retry. This is also worth developing as overall improvement.

test/start-kubernetes.sh
```diff
@@ -254,15 +255,17 @@ EOF
...
         SECONDS=0
-        NO_PROXY+=",$ip"
         #Wait for the ssh connectivity in the vms
         echo "Waiting for ssh connectivity on vm with ip $ip"
         while ! ssh $SSH_ARGS ${CLOUD_USER}@${ip} exit 2>/dev/null; do
             if [ "$SECONDS" -gt "$SSH_TIMEOUT" ]; then
                 die "timeout accessing ${ip} through ssh"
             fi
         done
+    done
```
pohly: It's not obvious why the loop is split here. Please add a comment explaining it, otherwise we might end up reverting this again in the future.

okartau: Good point,  thanks! Comment added and pushed (no functional changes)

PR #484, opened 2019-11-29
test: add individual machines to NO_PROXY

pohly:
Extending NO_PROXY in create_vms only had an effect inside that
subshell. We explicitly need to make that change also in the outer
shell, otherwise kubeadm will fail to reach the API server because the
IP address is not on the exception list.

### DIFF
--- a/test/start-kubernetes.sh
+++ b/test/start-kubernetes.sh
@@ -219,6 +219,13 @@ function print_ips() (
     govm list -f '{{select (filterRegexp . "Name" "^'$(node_filter ${NODES[@]})'$") "IP"}}' | tac
 )
 
+function extend_no_proxy() (
+    for ip in $(print_ips); do
+        NO_PROXY+=",$ip"
+    done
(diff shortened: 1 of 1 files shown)


pohly:
@okartau, @avalluri: this is something that @devimc ran into when trying "make start". I assume you are not hitting this issue because you already added the Docker IP range to your `no_proxy` env variable?

I've not verified that this patch fixes the issue. Perhaps you can try `NO_PROXY= no_proxy make start` in your environment and test whether the cluster comes up?

okartau:
> I assume you are not hitting this issue because you already added the Docker IP range to your `no_proxy` env variable?

That's correct at least with me. I use the practice that env. already has NO_PROXY with (to soon become a) cluster address range before I get into stage of "kubeadm init". This practice has been in place from times before pmem-csi, for cases of manual creation of clusters. In "make start" case, it also takes care of NO_PROXY without need to explicit steps.

PR #486, opened 2019-12-03
test: enable nested virtualization

pohly:
Nested virtualization is needed for Kata Containers. This works on
hosts which have kvm_intel loaded with nested=1. Might also need a
recent enough host CPU.


pohly:
Let's merge via #374.

PR #487, opened 2019-12-04
CI: speed up execution

pohly:
The last remaining reason for requiring sudo and --privileged was that
the scripts themselve set up a tmpfs for etcd. By moving that step out
into the build host we can remove the need for sudo inside the
container.

The etcd volume gets used repeatedly, therefore "mkfs.ext4" must be
forced to ignore an already existing filesystem.

We can switch back and forth between LVM and direct mode on a cluster merely by applying a different deployment. If we run two tests back-to-back which use the same cluster configuration (for example, Fedora with Kubernetes 1.16), then we reuse the existing cluster, which saves around 5 minutes for the second test.

We have some tests marked as '[Slow]': https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-487/32/testReport/clear-31760-lvm-production/E2E/

For example, the `xfs` variant (somehow, haven't checked) is marked as slow. It seems okay to skip those tests in pre-merge testing. This cuts down the time down by several minutes (1:13h https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-487/32/testReport/ vs. 0:59h https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-487/33/testReport/).

Finally, some issues found while testing get fixed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -457,6 +457,7 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                   -e TEST_DEVICEMODE=${deviceMode} \
                   -e TEST_DEPLOYMENTMODE=${deploymentMode} \
                   -e TEST_CHECK_SIGNED_FILES=false \
+                  -e TEST_CHECK_KVM=false \
                   -e TEST_DISTRO=${distro} \
                   -e TEST_DISTRO_VERSION=${distroVersion} \
                   -e TEST_KUBERNETES_VERSION=${kubernetesVersion} \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,6 +56,10 @@ pipeline {
         // Set below via a script, must *not* be set here as it can't be overwritten.
         // BUILD_IMAGE = ""
 
+        // A running container based on BUILD_IMAGE, with volumes for everything that we
+        // need from the build host.
+        BUILD_CONTAINER = "builder"
+
         // Tag or branch name that is getting built, depending on the job.
         // Set below via a script, must *not* be set here as it can't be overwritten.
         // BUILD_TARGET = ""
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -142,15 +156,12 @@ pipeline {
...
 
-                    // Allow non-root users to become root via sudo when they are part of the root group.
-                    // ${DockerBuildArgs()} includes the necessary parameters for that.
-                    sh "docker exec builder mkdir /etc/sudoers.d"
-                    sh "echo '%root ALL=(ALL) NOPASSWD: ALL' | docker exec -i builder tee /etc/sudoers.d/nopasswd >/dev/null"
+                    // Make /usr/local/bin writable for all users. Used to install kubectl.
+                    sh "docker exec builder sh -c 'mkdir -p /usr/local/bin && chmod a+wx /usr/local/bin'"
 
-                    // sudo needs a user entry for the jenkins user.
+                    // Some tools expect a user entry for the jenkins user (like govm?)
```
okartau: not strictly in scope of this change, but changed comment directed my attention to following two lines. Isnt creation of lines with hard-coded values and number of fields directly in /etc/passwd and /etc/shadow brutal and non-portable way? There is portable and cleaner method called useradd, is there reason we can not use it?

pohly: Yes, it might be cleaner, but the current approach worked and I didn't want to invest further time into making it better.

Regarding "portable": this isn't that important, because neither the OS we use this on nor the format of `/etc/passwd` or `/etc/group` are likely to change.

Jenkinsfile
```diff
@@ -485,7 +484,13 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                            loggers=; \
                            atexit () { set -x; kill \$loggers; kill \$( ps --no-header -o %p ); }; \
                            trap atexit EXIT; \
+                           echo CLUSTER=\$CLUSTER TEST_LOCAL_REGISTRY=\$TEST_LOCAL_REGISTRY TEST_DISTRO=\$TEST_DISTRO TEST_DISTRO_VERSION=\$TEST_DISTRO_VERSION TEST_KUBERNETES_VERSION=\$TEST_KUBERNETES_VERSION >_work/new-cluster-config && \
```
avalluri: @pohly Can't we reuse the cluster in case of LCOAL_REGISTRY changes? Is it because it needs a change in docker config(insecure registry)?

avalluri:
Overall looks good to me.

Jenkinsfile
pohly: Correct.

pohly:
This caused two regressions in code that only runs after merging a PR:
- empty `-ginkgo.focus` causes all tests to be skipped in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/devel/122/execution/node/106/log/
- image pushing failed: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/devel/122/execution/node/228/log/

I'll revert in "devel" and create a new PR which gets tested completely before merging by creating the branch in the main repo.

pohly:
Correction: it's the `||` in
```
-ginkgo.skip='no-such-test||should.access.volume.from.different.nodes'
```
which caused tests to be skipped.

Issue #480, opened 2019-11-28
direct mode: pfn: no seed


pohly:
Recently (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-477/27/execution/node/104/log/?consoleFull for https://github.com/intel/pmem-csi/pull/477, https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-478/1/pipeline for https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-478/1/pipeline) we started to get intermittent failures:

```
14:16:38  pmem-csi-controller-0/pmem-driver: W1128 13:16:38.702163       1 controllerserver-master.go:237] failed to create volume name:pmem-csi-clear-worker2 id:sanity-controller-create-twice-5AB71DB6-10958FB2 on 89bcd0f0a3119dd538e3cf25e6111d3267bfb503: rpc error: code = Internal desc = Node CreateVolume: failed to create volume: pfn: no seed
...
[91m[1m[Fail] [0m[90mE2E [0m[0m[Driver: pmem-csi] [0m[90m[Testpattern: inline ephemeral CSI volume] ephemeral [0m[91m[1m[It] should support two pods which share the same volume [0m
15:22:16  [37m/mnt/workspace/pmem-csi_PR-477/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/ephemeral.go:259[0m
15:22:16  
15:22:16  [91m[1m[Fail] [0m[90mE2E [0m[0m[Driver: pmem-csi] [0m[90m[Testpattern: inline ephemeral CSI volume] ephemeral [0m[91m[1m[It] should support multiple inline ephemeral volumes [0m
15:22:16  [37m/mnt/workspace/pmem-csi_PR-477/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/ephemeral.go:259[0m
15:22:16  
15:22:16  [91m[1m[Fail] [0m[90mE2E [0m[0m[Driver: pmem-csi] [0m[90m[Testpattern: inline ephemeral CSI volume] ephemeral [0m[91m[1m[It] should create read-only inline ephemeral volume [0m
[...]

PR #488, opened 2019-12-10
ndctl: use fresh context for each operation

pohly:
It is uncertain whether libndctl supports concurrent operations while
a context is open. To minimize the risk it is better to start each
operation with a new context.

Update: Dan Williams confirmed that "one libndctl instance expects it is the only agent manipulating the kernel state. A single instance can make any changes it wants without refreshing the context" - this implies that if there are multiple instances (like PMEM-CSI and ndctl command line), the context must be refreshed, as we do here.

### DIFF
--- a/pkg/pmem-device-manager/pmd-ndctl.go
+++ b/pkg/pmem-device-manager/pmd-ndctl.go
@@ -17,7 +17,6 @@ const (
 )
 
 type pmemNdctl struct {
-	ctx *ndctl.Context
 }
 
 var _ PmemDeviceManager = &pmemNdctl{}
@@ -30,13 +29,6 @@ var ndctlMutex = &sync.Mutex{}
 
 //NewPmemDeviceManagerNdctl Instantiates a new ndctl based pmem device manager
 func NewPmemDeviceManagerNdctl() (PmemDeviceManager, error) {
-	ndctlMutex.Lock()
-	defer ndctlMutex.Unlock()
-
-	ctx, err := ndctl.NewContext()
-	if err != nil {
-		return nil, err
-	}
 	// Check is /sys writable. If not then there is no point starting
 	mounts, _ := mount.New("").List()
 	for _, mnt := range mounts {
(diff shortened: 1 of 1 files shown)


pohly:
Let's see how this goes in the CI. I'm also testing locally.

pohly:
@okartau: CI and local testing in direct mode was successful. Shall we merge and consider issue #480 resolved by this?

okartau:
right now, CI is going (again?) although I dont see new push here..
I believe this is right and good solution, but I want to see my 10 local rounds clean before I am sure.
While I was debugging that issue, it sometimes took >5 cycles to see one "no seed" case, so first OK-round is no proof yet.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -33,10 +32,6 @@ func NewPmemDeviceManagerNdctl() (PmemDeviceManager, error) {
 	ndctlMutex.Lock()
```
okartau: Now in this function there is nothing to protect with mutex any more?

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -54,19 +49,23 @@ func NewPmemDeviceManagerNdctl() (PmemDeviceManager, error) {
...
-		ctx: ctx,
-	}, nil
+	return &pmemNdctl{}, nil
 }
 
 func (pmem *pmemNdctl) GetCapacity() (map[string]uint64, error) {
 	ndctlMutex.Lock()
 	defer ndctlMutex.Unlock()
 
+	ndctx, err := ndctl.NewContext()
```
okartau: In all remaining cases (if we agree we can delete use of mutex in NewPmemDeviceManagerNdctl()),
there is both "get lock" and "create context" as pair of operations, so can we perhaps combine those 2 in same function, and use defer in other functions to call another common release-delete function ?
Not a big deal though, its quite small and clear already now.

pohly:
> right now, CI is going (again?) although I dont see new push here..

I restarted manually to get another data point.

pkg/pmem-device-manager/pmd-ndctl.go
pohly: True, can be removed.

pkg/pmem-device-manager/pmd-ndctl.go
pohly: I was debating that with myself and decided to go with separate code for mutex and context because it is clearer and more flexible.

okartau:
looking good, 16 rounds of local testing without "no seed" error. I take away mutex locking in startup function and run few more rounds.

pkg/pmem-device-manager/pmd-ndctl.go
pohly: Done.

Issue #377, opened 2019-08-05
remove entrypoint from Dockerfile


pohly:
It's very annoying that the `pmem-csi-driver` is marked as entry point for the `pmem-csi-driver` image when using it as the container for running tools that aren't installed on a host, like ndctl or LVM. For example:

```sh
$ docker run --privileged  --rm intel/pmem-csi-driver:v0.5.0 -- ndctl list -Nu
failed to initialize driver: Invalid driver mode: unified
```

This works:
```sh
docker run --privileged  --rm -ti --entrypoint /bin/sh intel/pmem-csi-driver:v0.5.0
```

PR #489, opened 2019-12-10
remove entry point for PMEM-CSI image

pohly:
It's very annoying that the pmem-csi-driver is marked as entry point
for the pmem-csi-driver image when using it as the container for
running tools that aren't installed on a host, like ndctl or LVM. For
example:

     $ docker run --privileged  --rm intel/pmem-csi-driver:v0.5.0 -- ndctl list -Nu
     failed to initialize driver: Invalid driver mode: unified

It's better to not specify an entry point and instead select the
command in the deployment. This is then also consistent for all three
commands that we run from the image.

Fixes: #377

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -116,4 +116,3 @@ RUN echo "global { use_lvmetad = 0 }" >> /etc/lvm/lvm.conf && \
     echo "activation { udev_sync = 0 udev_rules = 0 }" >> /etc/lvm/lvm.conf
 
 ENV LD_LIBRARY_PATH=/usr/lib
-ENTRYPOINT ["/usr/local/bin/pmem-csi-driver"]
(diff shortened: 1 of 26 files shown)


okartau:
`starting container process caused "exec: \"-deviceManager=lvm\": executable file not found in $PATH"`
this part comes from kustomize patch, something goes wrong in patching now?

pohly:
Ah, my local testing (where it worked) was flawed because I hadn't actually pushed the image without the entry point. My assumption that "command" and "args" will be combined is incorrect. It looks like "command" is ignored if there is an "args" field. Hmm, but then did that work for our init containers? They also mixed "command" and "args". Upstream documentation is silent about how those two fields interact.

I'll move the "args" into "command". I started with that but stopped going that way because it means touching also all patch files, but it's the cleaner solution.

pohly:
@okartau: okay to merge?

PR #490, opened 2019-12-11
Remove sector mode related code and doc

okartau:


okartau:
This is a draft to get rid of sector mode. I think we haven't decided on removal yet, but we can keep the code change ready. I started with mechanical removal of related parts in code and doc.
Open question: what to do with parameter name `useforfsdax`. The functionality is still useful, as it allows to leave some part unused. (I assume we want to still keep this). But what do to with variable name now, should we keep it or change it to something else, like `useforcsi` ?
Or, should we drop this limiting functionality from pmem-csi altogether and move it elsewhere? (this relates to init stage functionality plans).
Also, with `useforsector` gone, the unit test part of that code does not do much, so should we remove it?

pohly:
> Open question: what to do with parameter name `useforfsdax`. The
> functionality is still useful, as it allows to leave some part
> unused. (I assume we want to still keep this). But what do to with
> variable name now, should we keep it or change it to something else,
> like `useforcsi` ?

`usefordax` still makes sense to me as it's for a "fsdax" namespace.

> Or, should we drop this limiting functionality from pmem-csi
> altogether and move it elsewhere? (this relates to init stage
> functionality plans).

Eventually yes, but not now.

> Also, with `useforsector` gone, the unit test part of that code does
> not do much, so should we remove it?

I'm not sure which test you mean, but yes, tests for code that no longer
exists should also be removed.

okartau:
> `usefordax` still makes sense

ok, good, lets keep it that way, less mods to API

> I'm not sure which test you mean 

I mean pkg/pmem-ns-init/main_test.go

> tests for code that no longer exists should also be removed.

yes that's for sure (already done in PR), but I mean that after removal of irrelevant tests, the remaining part dilutes to be so small that it may seem silly to have it. In code supporting both modes, we have limits checking for both modes, plus limits combinations, totaling 9 cases. After removal of 2nd mode, we take away tests specific to that mode, but also combination-specifics, and what remains is just trivial bounds-checking for one mode, with 4 cases.

okartau:
rebased, resolving merge conflict, and force-pushed. What do you think, does it make sense to keep unit test in pkg/pmem-ns-init/ after reduction by this PR, or is it now too trivial to have?

pohly:
> What do you think, does it make sense to keep unit test in pkg/pmem-ns-init/ after reduction by this PR, or is it now too trivial to have?

A test, no matter how trivial, is better than no test. Let's keep it.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -365,14 +362,7 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	// FIXME(avalluri): we shouldn't depend on volumecontext to determine the device mode,
```
pohly: Apropos these FIXME notes: these have been in the code for a while. We should turn them into issues and that way actually track progress towards solving them.

At least we got rid of this one :smile:

PR #494, opened 2019-12-12
Operator docker image

avalluri:
Support for building the docker image for Operator

Also contains a minor change to remove unneeded Namespace field from DeploymentSpec

Issue #261, opened 2019-04-30
prevent CSI driver from being evicted


pohly:
Without any resource specifications, the CSI driver pod is among the first that gets evicted when resources are running low on a node. I was filling up /var by pulling a large image on host-1 of the QEMU virtual cluster and the result was:

```
pmem-csi-node-testing-dml4m               0/1     Evicted   0          2m4s    <none>        host-1   <none>           <none>
pmem-csi-node-testing-fn86q               1/1     Running   0          74m     192.168.8.8   host-3   <none>           <none>
pmem-csi-node-testing-pz64d               1/1     Running   0          74m     192.168.8.6   host-2   <none>           <none>
pmem-csi-node-w56mn                       0/2     Evicted   0          4m15s   <none>        host-1   <none>           <none>
...
```

I did not catch the description, but some other evicted pod had this event:
```
 Warning  Evicted    6m28s  kubelet, host-1    The node was low on resource: ephemeral-storage. Container .... was using 52Ki, which exceeds its request of 0.
```

We need to ensure that we properly declare the resource usage of our containers, including all sidecars.

pohly:
The other evicted pod was `rfs-redisfailover-pmem-75d4b4d856-2qwzf`, from the spotahome redis operator. That operator also needs to be fixed to create pods with resource settings.

PR #495, opened 2019-12-12
Operator Documentation 

avalluri:


operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+| logLevel | integer | PMEM-CSI driver logging level | 3 |
+| controller | CotrollerDriver | configuration of PMEM-CSI driver running as controller | |
+| node | NodeDriver | configuration of PMEM-CSI driver running at compute node | |
+
+### ControllerDriver
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| provisionerImager | ImageInfo | CSI provisioner container image information | `quay.io/k8scsi/csi-provisioner:v1.2.1` |
+| resources | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core) | describes the compute resource requirements for controller pod |
```
pohly: Instead of blank "default value", put "none"?

We probably should change that (see https://github.com/intel/pmem-csi/issues/261), but not now.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+| pullPolicy | string | Image pull policy. either one of `Always`, `Never`, `IfNotPresent` | `IfNotPresent` |
+
+## Pre-requisites
+
+### Build operator image
+
+Build PMEM-CSI operator image using the provided [Dockerfile](https://github.com/intel/pmem-csi/tree/devel/operator/build/Dockerfile) and push the image to local docker registry.
+
+```sh
+docker build --pull -t pmem-csi-operator:canary  --label revision=$(VERSION) -f operator/build/Dockerfile .
```
pohly: Have you considered just adding the operator to the existing image?

Not having to maintain and publish another image is going to be simpler. The downside is that one additional file will be put onto each node where PMEM-CSI runs.

I think I prefer to start simple (i.e. add to existing image) and later optimize for overall size if needed.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+```
+
+### Deploy PMEM-CSI Deployment CRD
+
+```sh
+kubectl create -f https://raw.githubusercontent.com/avalluri/pmem-CSI/operator/operator/deploy/crds/pmem-csi.intel.com_deployments_crd.yaml
+```
+> FIXME: **Should this be handled with in operator**?
```
pohly: Do you know how this is handled in other operators?

pohly: The external-snapshotter used to create the CRD.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+Build PMEM-CSI operator image using the provided [Dockerfile](https://github.com/intel/pmem-csi/tree/devel/operator/build/Dockerfile) and push the image to local docker registry.
+
+```sh
+docker build --pull -t pmem-csi-operator:canary  --label revision=$(VERSION) -f operator/build/Dockerfile .
+```
+
+### Deploy PMEM-CSI Deployment CRD
+
+```sh
+kubectl create -f https://raw.githubusercontent.com/avalluri/pmem-CSI/operator/operator/deploy/crds/pmem-csi.intel.com_deployments_crd.yaml
```
pohly: One general question: how useful is it to have the "operator" directory? Do we need it (for example, because the SDK expects it) or can we get rid of it and move the content into the root?

I find it confusing that we now have duplicated "pkg" and "deploy" directories. I'd prefer to have "pkg/operator" (if needed) instead of "operator/pkg".

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+
+## Usage
+
+### Create a PMEM-CSI driver Deployment
+
+```sh
+kubectl create -f https://raw.githubusercontent.com/avalluri/pmem-CSI/operator/operator/deploy/crds/pmem-csi.intel.com_v1alpha1_deployment_cr.yaml
+```
+
+Once the deployment resource installation successful, we can see all the driver pods in Running state:
```
pohly: Long term we should add a "deployment.Status" field where the operator itself reports whether PMEM-CSI is running properly.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+
+Current PMEM-CSI Deployment object supports below API:
+
+### Deployment
+
+|Field | type | Description |
+|---|---|---|
+| apiVersion | string  | `pmem-csi.intel.com/v1alpha1`|
+| kind | string | `Deployment`|
+| metaData | [ObjectMeta](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status) | Standard objects metadata. |
```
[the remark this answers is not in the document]
pohly: We need to document whether we support more than one deployment.

I think we should, in which case the CSI driver name also has to be made configurable. It may be useful at least for testing.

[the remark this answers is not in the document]
pohly: We could specify that the name of the deployment is used as CSI driver name. Then there's no risk of getting two deployments with conflicting driver names. OTOH, a user may prefer a shorter name for the deployment.

I'm undecided...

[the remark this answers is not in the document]
avalluri: The name of the deployment is relevant, only if we plan to support more than one driver deployments, which is undecided by myself.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+| spec | [DeploymentSpec](#deployment-spec) | Specification of the desired behavior of the deployment |
+
+### DeployemtSpec
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| image | ImageInfo | PMEM-CSI driver container image information | `intel/pmem-csi-driver:canary` |
```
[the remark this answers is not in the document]
avalluri: > If I do a `go get -o yaml deployment.pmem-csi.intel.com/v1alpha1`, will it show what the default is or only the explicitly set values?

You mean `kubectl get -o yaml deployment.pmem-csi.intel.com example-deployment`?
No, it does not show the defaults, it only shows the values set by the user as we do not update the Deployment object. Instead, we prepare(/update?) the secondary resources using the `DeploymentSpec` generated with the combined values(defaults + user set).

Do you feel we should update the Deployment itself with the defaults we used?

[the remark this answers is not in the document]
avalluri: > I'm not sure how that is normally handled. What don't want is that defaults are stored as if the user had set them intentionally. For example, if we change the default here from "intel/pmem-csi-driver:v0.6.0" to "intel/pmem-csi-driver:v0.6.1", then a deployment where the user hasn't explicitly selected the older version should get updated,

If we consider this, to make this work smoothly, I guess we should expose these defaults as Operator configuration, say `ConfigMap`. So that admin alters the defaults by updating ConfigMap, then the Operator catches these changes and redeploys the driver for existing Deployment(s).

operator/README.md
avalluri: Current code is using the below defaults, which should be reviewed:

```
	defaultControllerResourceCPU    = int64(100) // MilliSeconds
	defaultControllerResourceMemory = int64(250) // MB
	defaultNodeResourceCPU          = int64(100) // MilliSeconds
	defaultNodeResourceMemory       = int64(250) // MB
```

I guess I should update the `Default Value` column with these values.

operator/README.md
avalluri: > Have you considered just adding the operator to the existing image?

No.

I felt like this deserves its own image :-)

operator/README.md
avalluri: > Do you know how this is handled in other operators?

I didn't check. But, my feeling is that the CRD is owned by this Operator and the end-user is no way concerned about altering this. So, I see no extra benefit of asking user to first install CRD yourself before starting Operator.

operator/README.md
avalluri: I am not sure though but I believe it's not so easy/clean to regenerate the operator code if we alter the directory structure.

operator/README.md
avalluri: Exactly, It needs some brainstorming time for deciding what this Status should hold some useful information,

operator/README.md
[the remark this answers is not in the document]
pohly: > Do you feel we should update the Deployment itself with the defaults we used?

No, not at all ("What don't want is that defaults are stored as if the user had set them intentionally."). We must be able to distinguish between behavior chosen by the user and behavior selected because it is the default. Then the default can change.

> If we consider this, to make this work smoothly, I guess we should expose these defaults as Operator configuration, say ConfigMap.

I don't think we need this. The defaults can be built into the operator. One usage pattern that was mentioned for OpenShift/Operator Hub is that an admin opts into auto-updating operators. Then the updated defaults for versions also cause the corresponding software to be updated *without* the admin being involved.

My question was just whether those defaults then can be observed somehow or must be documented.

operator/README.md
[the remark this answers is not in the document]
pohly: So we need to decide it before we can proceed with the implementation...

operator/README.md
pohly: I agree that letting the operator install and/or update the CRD seems better.

operator/README.md
pohly: So considering the pros and cons, what do you prefer and why?

operator/README.md
pohly: Which code needs to be (re)generated? Let's test whether we can do that with everything at the top level.

operator/README.md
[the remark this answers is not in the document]
pohly: Let's support multiple deployments, but with a warning that these deployments must not run multiple drivers on the same node. Later we can investigate whether that can be supported, too.

This implies that the API needs a field for the driver name, and that field has to be unique for all deployments.

operator/README.md
pohly: Discussed just now, let's use one image.

Once we have something working we can discuss whether we want to optimize for size by splitting into two images.

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
-Operator for PMEM-CSI
+# PMEM-CSI-Operator
+
+Kubernetes operator for [PMEM-CSI](https://github.com/intel/pmem-csi) driver deployment. Using this operator one can easily deploy and manage PMEM-CSI driver in a Kubernetes cluster.
+
+The PMEM-CSI operator is based on the CoreOS [operator-sdk](https://github.com/operator-framework/operator-sdk) tools and APIs.
+
+## PMEM-CSI Deployment custom resource definition(CRD)
+
+Current PMEM-CSI Deployment object supports below API:
```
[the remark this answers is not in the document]
avalluri: Ok, Done

operator/README.md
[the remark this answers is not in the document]
avalluri: Done. Changed the description wording as suggested.

operator/README.md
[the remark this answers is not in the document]
avalluri: Added "warning" for multi deployments.

operator/README.md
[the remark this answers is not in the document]
pohly: Not sure where "usage" was originally, but what I meant was that the flow should be:
- summary of what the operator is ("about")
- usage:
  - pre-requisites
  - deployment
  - API + explanation of that API
- example

I'm undecided about

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+| spec | [DeploymentSpec](#deployment-spec) | Specification of the desired behavior of the deployment |
+
+### DeployemtSpec
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| image | ImageInfo | PMEM-CSI driver container image information | `intel/pmem-csi-driver:canary` |
+| logLevel | integer | PMEM-CSI driver logging level | 3 |
+| controller | CotrollerDriver | configuration of PMEM-CSI driver running as controller | |
+| node | NodeDriver | configuration of PMEM-CSI driver running at compute node | |
```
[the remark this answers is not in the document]
pohly: @avalluri any comments on this?

operator/README.md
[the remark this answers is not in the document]
avalluri: > Not sure where "usage" was originally, but what I meant was that the flow should be:

Initially, Usage was after API description. like:
 * About
 * Deployment API
 * Usage
   * Pre-requisites
   * Deploy operator
   * Example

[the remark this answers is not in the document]
avalluri: > * summary of what the operator is ("about")
> * usage:
>   
>   * pre-requisites
>   * deployment
>   * API + explanation of that API
> * example

Now I changed to this.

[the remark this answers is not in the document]
avalluri: Add ToC section

operator/README.md
```diff
@@ -1,2 +1,92 @@
...
+| kind | string | `Deployment`|
+| metaData | [ObjectMeta](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status) | Standard objects metadata. |
+| spec | [DeploymentSpec](#deployment-spec) | Specification of the desired behavior of the deployment |
+
+### DeployemtSpec
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| image | ImageInfo | PMEM-CSI driver container image information | `intel/pmem-csi-driver:canary` |
+| logLevel | integer | PMEM-CSI driver logging level | 3 |
```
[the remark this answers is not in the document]
avalluri: Currently, we have no description of logging levels in our documentation.

operator/README.md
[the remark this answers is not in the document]
avalluri: Ok, you suggest flat API. I will change to that.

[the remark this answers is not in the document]
avalluri: Done.

operator/README.md
```diff
@@ -1,2 +1,104 @@
...
+A running Kubernetes cluster with one or more nodes installed with PMEM hardware(NVDIMM). The persistent memory on these nodes must be [pre-provisioned](../README.md#persistent-memory-pre-provisioning).
+
+### Deploy PMEM-CSI operator
+
+<!-- The assumptions:
+  a) pmem-csi-operator binary is part of released intel/pmem-csi-driver image and no additional steps required to build the images
+-->
+
+```sh
+$ kubectl create -f https://raw.githubusercontent.com/intel/pmem-CSI/operator/operator/deploy/operator.yaml
```
okartau: `pmem-CSI/operator/operator/deploy/`
I think the path with double `operator` points to code in branch=operator, is that intentional to remain so in final version?

avalluri: I kept this as the PR is against the operator branch. But when merging to devel this should change.

operator/README.md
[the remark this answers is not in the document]
okartau: good point, we could add note in README (or better DEVELOPMENT ?) about our style, which seems to be off-by-one from guidelines at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
i.e. we use (Info levels) 3 for production level and 4,5 for extended/development.
I recall the history was, when we cleaned our logging levels, we had most of statements at level 3 already so we wanted to keep those and we used 4 and 5 as more verbose levels.
 We do not use info levels 0,1,2 currently at all.
At some point, do we wish to change all levels so that we conform better to convention in abovementioned page?

operator/README.md
```diff
@@ -1,2 +1,104 @@
...
+
+Current PMEM-CSI Deployment object supports below API:
+
+#### Deployment
+
+|Field | type | Description |
+|---|---|---|
+| apiVersion | string  | `pmem-csi.intel.com/v1alpha1`|
+| kind | string | `Deployment`|
+| metadata | [ObjectMeta](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status) | Standard objects metadata. |
```
okartau: table entry which is not full sentence, so should be without dot

operator/README.md
```diff
@@ -1,2 +1,104 @@
...
+| spec | [DeploymentSpec](#deployment-spec) | Specification of the desired behavior of the deployment |
+
+#### DeploymentSpec
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| driverName | string | Unique CSI driver name to use. | `pmem-csi.intel.com` |
```
okartau: table entry which is not full sentence, so should be without dot

operator/README.md
okartau: also, should pmem-CSI be without mixed uppercase, like pmem-csi

operator/README.md
avalluri: Done

operator/README.md
avalluri: Done

operator/README.md
avalluri: > also, should pmem-CSI be without mixed uppercase, like pmem-csi

Fixed.

operator/README.md
avalluri: > Which code needs to be (re)generated? Let's test whether we can do that with everything at the top level.

I was referring to the auto-generation of deep copy(zz_generated_deepcoy.go) code. But anyway I gave it a try to meld operator code to pme-csi directory tree in my [branch](https://github.com/avalluri/pmem-CSI/blob/operator-direcotry-tree). Could resolve this issue: https://github.com/avalluri/pmem-CSI/blob/operator-direcotry-tree/operator/operator.make#L15-L24

operator/README.md
avalluri: Made needed changes to auto-install the CRD within operator: https://github.com/avalluri/pmem-CSI/commit/c574c4b9661c56cdba42446db9b27aa4c2285822

operator/README.md
pohly: https://github.com/avalluri/pmem-CSI/blob/operator-direcotry-tree/operator/operator.make#L15-L24 is even more complicated than before.

Why can't we just run `operator-sdk generate k8s` directly in the `pmem-csi` directory and put all files where it expects them?

`pkg/apis` at the top level looks good to me. We may end up with more than just the operator API there, but that's fine.

`cmd/manager` is less nice. Why does code generation care about the code that uses the generated code? As far as I can tell, `cmd/manager/main.go` was initially bootstrapped with the SDK, but now is entirely maintained by hand. Correct?

operator/README.md
pohly: Please remove  https://github.com/avalluri/pmem-CSI/blob/dc89540733bd126ac20924590340e6989cbfb5bf/operator/build/Dockerfile#L1-L15

This is not an image that we would be able to ship anyway (`FROM registry.access.redhat.com/ubi8/ubi-minimal:latest`...)

operator/README.md
avalluri: > `cmd/manager` is less nice. Why does code generation care about the code that uses the generated code?

It tries to find if the operator is Go operator or not by checking `cmd/manager/main.go`.
The code is here: https://github.com/operator-framework/operator-sdk/blob/6b0412be9d0bb3b3d715c87ddb3afbb2b4f9080e/internal/util/projutil/project_util.go#L93

operator/README.md
avalluri: > This is not an image that we would be able to ship anyway (`FROM registry.access.redhat.com/ubi8/ubi-minimal:latest`...)

This is what PR #494 about :-)  https://github.com/intel/pmem-csi/pull/494/files#diff-04ea00daa9ec469d5b78e40cf8e7649a

But anyway we decided to ship both operator and driver with a single image, so we might not need this Dockerfile in this case.

pohly: Exactly, so let's get rid of it.

operator/README.md
pohly: And it's hard-coded:  https://github.com/operator-framework/operator-sdk/blob/6b0412be9d0bb3b3d715c87ddb3afbb2b4f9080e/internal/util/projutil/project_util.go#L39

That's unfortunate and probably worth a feature request to allow reconfiguring that. But for now we can either use `cmd/manager` (as expected) or make it a permanent symlink to `cmd/pmem-csi-operator`.

avalluri: > That's unfortunate and probably worth a feature request to allow reconfiguring that. But for now we can either use `cmd/manager` (as expected) or make it a permanent symlink to `cmd/pmem-csi-operator`.

This not ends here, operatro-sdk is more strict about directory structure :(
```
make operator-generate-k8s 
./_work/bin/operator-sdk-v0.12.0 generate k8s
FATA[0000] must run command in project root dir: project structure requires build/Dockerfile 
operator/operator.make:16: recipe for target 'operator-generate-k8s' failed
make: *** [operator-generate-k8s] Error 1
```

pohly: This begs the question: is operator SDK really the right tool for us?

If all that we needs is code generation, can't we use the underlying tooling (https://github.com/kubernetes/code-generator, if I'm not mistaken) directly or (if that is too low-level) https://github.com/kubernetes-sigs/kubebuilder instead?

pohly:
Moved to PR #525.

Issue #498, opened 2019-12-18
huge pages


pohly:
https://nvdimm.wiki.kernel.org/2mib_fs_dax explains how memory mapping a file on a dax-capable filesystem can be done with huge pages (2MiB) instead of the normal 4KiB.

We should write a test that verifies whether an application can use huge pages. The Wiki has instructions for testing this.

This test might currently fail, because the Wiki mentions a "-E stride=512" option for mkfs.ext4 and some other changes for xfs which we currently don't use. We might have to add that.

pohly:
The test can be an extension of the existing https://github.com/intel/pmem-csi/blob/devel/test/e2e/storage/dax/dax.go

okartau:
I tried the steps (no k8s involved, just command line for starter) from above-mentioned page in a VM we use for CI tests (running Clear) , but I see no traces in `/sys/kernel/debug/tracing/trace`. I wonder is that trace supposed to appear in VM. If no trace on VM, then it's not easy to use that technique in CI tests.
Next, I will try on bare metal for comparison.
Another side question is, what determines the page size for given region: is there some setting (to be set where?), or is this supposed to be transparent to user? Or is it so that bigger page is tried first with fallback to smaller size?

pohly:
I don't know. Please ask our Intel colleagues who work on the kernel support for PMEM.

pohly:
Resolved via https://github.com/intel/pmem-csi/pull/499

PR #499, opened 2019-12-18
Create filesystem with 2MB aligned and sized block allocations

okartau:
Give additional options to mkfs.ext4 and mkfs.xfs.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -28,6 +28,7 @@ There is a more detailed explanation in the following paragraphs.
 | *Name* field in namespace | *Name* gets set to 'pmem-csi' to achieve own vs. foreign marking | *Name* gets set to VolumeID, without attempting own vs. foreign marking  |
 |Minimum volume size| 4 MB                   | 1 GB (see also alignment adjustment below) |
 |Alignment requirements |LVM creation aligns size up to next 4MB boundary  |driver aligns  size up to next alignment boundary. The default alignment step is 1 GB. Device(s) in interleaved mode will require larger minimum as size has to be at least one alignment step. The possibly bigger alignment step is calculated as interleave-set-size multiplied by 1 GB |
+|Huge pages supported<sup>4</sup> | maybe| yes|
 
 <sup>1 </sup> **Free space fragmentation** is a problem when there appears to
 be enough free capacity for a new namespace, but there isn't a contiguous
@@ -55,6 +56,14 @@ Programming](https://pmem.io/ndctl/ndctl-create-namespace.html) for
 details. `devdax` mode is not supported. Though a
 raw block volume would be useful when a filesystem isn't needed, Kubernetes
 cannot handle [binding a character device to a loop device](https://github.com/kubernetes/kubernetes/blob/7c87b5fb55ca096c007c8739d4657a5a4e29fb09/pkg/volume/util/util.go#L531-L534).
+
+<sup>4 </sup> **Huge pages supported**: ext4 and XFS filesystems are created using the options that should enable huge
+ page support, as explained in section "Verifying IO Alignment" in
+ ["Using Persistent Memory Devices with the Linux Device Mapper"]
+(https://pmem.io/2018/05/15/using_persistent_memory_devices_with_the_linux_device_mapper.html).
+[Testing that support by observing page faults](/test/cmd/pmem-access-hugepages/main.go) confirmed that
+this worked for direct mode. It did not work for LVM mode in the QEMU virtual machines, but it cannot be
+ruled out that it works elsewhere.
 
 ## LVM device mode
 
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -270,6 +270,17 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if ephemeral && fsType == "xfs" {
+		// FS was created only in ephemeral case.
+		// Only if created filesytem and it was XFS:
+		// Tune XFS file system to serve huge pages:
+		// Set file system extent size to 2 MiB sized and aligned block allocations.
+		_, err := pmemexec.RunCommand("xfs_io", "-c", "extsize 2m", hostMount)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
+
 	if !volumeParameters.GetKataContainers() {
 		// A normal volume, return early.
 		return &csi.NodePublishVolumeResponse{}, nil
@@ -530,6 +541,15 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 
 	if err = ns.mount(device.Path, stagingtargetPath, mountOptions, false /* raw block */); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if existingFsType == "" && requestedFsType == "xfs" {
+		// Only if created a new filesytem and it was XFS:
+		// Align XFS file system for hugepages
+		_, err := pmemexec.RunCommand("xfs_io", "-c", "extsize 2m", stagingtargetPath)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	return &csi.NodeStageVolumeResponse{}, nil
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -666,18 +666,20 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 	cmd := ""
 	var args []string
 	// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
-	if fsType == "ext4" {
+	switch fsType {
+	case "ext4":
 		cmd = "mkfs.ext4"
 		args = []string{"-b", "4096", "-E", "stride=512,stripe_width=512", "-F", device.Path}
-	} else if fsType == "xfs" {
+	case "xfs":
 		cmd = "mkfs.xfs"
 		// reflink=0: reflink and DAX are mutually exclusive
 		// (http://man7.org/linux/man-pages/man8/mkfs.xfs.8.html).
 		// su=2m,sw=1: use 2MB-aligned and -sized block allocations
 		args = []string{"-b", "size=4096", "-m", "reflink=0", "-d", "su=2m,sw=1", "-f", device.Path}
-	} else {
+	default:
 		return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
 	}
+
 	output, err := pmemexec.RunCommand(cmd, args...)
 	if err != nil {
 		return fmt.Errorf("mkfs failed: output:[%s] err:[%v]", output, err)
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -648,12 +648,13 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 	// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
 	if fsType == "ext4" {
 		cmd = "mkfs.ext4"
-		args = []string{"-b 4096", "-F", device.Path}
+		args = []string{"-b", "4096", "-E", "stride=512,stripe_width=512", "-F", device.Path}
 	} else if fsType == "xfs" {
 		cmd = "mkfs.xfs"
-		// reflink and DAX are mutually exclusive
+		// reflink=0: reflink and DAX are mutually exclusive
 		// (http://man7.org/linux/man-pages/man8/mkfs.xfs.8.html).
-		args = []string{"-b", "size=4096", "-m", "reflink=0", "-f", device.Path}
+		// su=2m,sw=1: use 2MB-aligned and -sized block allocations
+		args = []string{"-b", "size=4096", "-m", "reflink=0", "-d", "su=2m,sw=1", "-f", device.Path}
 	} else {
 		return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
 	}
--- a/test/e2e/storage/dax/dax.go
+++ b/test/e2e/storage/dax/dax.go
@@ -29,6 +29,8 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework/volume"
 	storageframework "k8s.io/kubernetes/test/e2e/storage/framework"
 
+	api "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1beta1"
+	"github.com/intel/pmem-csi/test/e2e/deploy"
 	"github.com/intel/pmem-csi/test/e2e/ephemeral"
 	pmempod "github.com/intel/pmem-csi/test/e2e/pod"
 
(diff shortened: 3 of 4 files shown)


pohly:
Let's first implement a test (https://github.com/intel/pmem-csi/issues/498) and then add those options. As it stands we have no idea whether these options improve anything. All we'll learn is if they break a test, which is of course good to know, but IMHO unlikely.

okartau:
I am confused, this form created args parsing error in mkfs in initial CI pass:
`args = []string{"-b 4096", "-E stride=512", "-F", device.Path}`
which means, we are supposed to give cmd line elems as separate strings, right?

So, this one worked in CI
`args = []string{"-b 4096", "-E", "stride=512", "-F", device.Path}`

But question, how come the "-b 4096" gets processed even as two args given in single string?
Is there something different how mkfs processes these arguments?
Based on test output, the file system uses 4k blocks so the arg seems to have desired effect.

okartau:
Passing run in CI shows these
`Warning: RAID stripe-width 256 not an even multiple of stride 512.`
seems we also have to add `stripe_width` to make mkfs.ext4 happier.

pohly:

> I am confused, this form created args parsing error in mkfs in initial CI pass:
> `args = []string{"-b 4096", "-E stride=512", "-F", device.Path}`
> which means, we are supposed to give cmd line elems as separate strings, right?
>
> So, this one worked in CI
> `args = []string{"-b 4096", "-E", "stride=512", "-F", device.Path}`

Better change "-b 4096" into "-b", "4096" - each argv entry has to be a
separate string.

> But question, how come the "-b 4096" gets processed even as two args given in single string?
> Is there something different how mkfs processes these arguments?

For that you would have to consult the source code.

okartau:
rebased after recent XFS option change had modified same lines with mkfs.xfs options

okartau:
I added 2nd commit with dax-specific hugepagefault tests.
This is still a draft and not ready, I want to get it into PR to avoid getting lost if only presnt in local state.
The testing is based on method described/pointed in #498, and currently added to existing dax test.
What makes hugepagefault testing more complex, is that monitoring of debug trace events has to run in host context, and testing in a pod/container. But AFAIK, we dont know the host during the scheduling on that pod, so current code solves this by starting monitor on all hosts, running test code, then checking which host has debug traces saved. This is bit clumsy but it works. I had idea to reduce "scan for results" need if tester code could somehow indicate in its output, on which node it is running, but this appears not so easy (?). Perhaps on /proc or /sys there is something that can be used for this.
About detecting page faults. This appears to be not deterministic and that bothers me. I cant implement short test and then verify in yes/no style. Seems heuristics are needed, like running 10 tries and checking did we get any page fault events. But this is not so clean.
Overall -- does it make sense to run such testing in every CI cycle, or should we perhaps run this off-line to verify mount options actually work?
About those mount options that we add in initial commit here: Based on my manual testing:
expected page fault pattern correlates with added mount options in xfs case, but not in ext4 case where additional mount options seem to have no effect, and huge page faults happen even without added option.

okartau:
I added xfs_io use in both driver and testing code, and the results with current draft code are promising: all direct-mode tests appear to  trigger NOPAGE event as we want, and all LVM-mode tests appear to fail (i.e. produce FALLBACK event instead of NOPAGE). Thats encouraging. Is it so that getting expected NOPAGE in LVM-mode is too much asked, and we should limit our case to direct mode only, also documenting this.

test/e2e/storage/dax/dax.go
```diff
@@ -359,3 +361,130 @@ func testDaxOutside(
...
+			VolumeAttributes: map[string]string{
+				"size": "110Mi",
+			},
+		},
+	}
+	It("should cause hugepage paging event with default fs", func() {
+		init()
+		testHugepageInPod(f, l.root, &vsource, config)
+	})
+	/* there is issue in kubelet causing panic and retry loop when fsType is set to ext4 or xfs.
```
okartau, line 410: This now mentions the issue in kubelet causing panic and retries in pod creation (panic trace in kubelet log on corresponding worker). We should probably take that up on kubelet side.

pohly: Please file an issue in kubernetes/kubernetes and link to it here.

okartau:
numerous testing failures, but these are not related to hugepage testing change, at least not in 1.20 logs that I examined.
Also, I added describing comment in helper program source which is used to cause a page fault event.

okartau:
tests keep failing, although the 1st (and only enabled) hugepage test succeeds. I have believed the following failures are not related, but aren't they really? Is there something in kubelet that gets into wrong state because of first inline ephemeral creation? As at least some of failures seem to relate to inline volume creation as well.
I tried to find some explanations in kubelet logs, and then I discovered we dont save kubelet logs correctly, shell redirection misuse means there is log from last worker only, and no pod state. So I added one more commit in this PR to fix this. It makes more sense to have this commit here compared to separate PR, as regular PR likely will not fail and exception handler code will node get covered

okartau:
Now that we get repeatable builds in CI, there is important additional fact to be added to "kubelet panics" story: the panic happens in 1.20 testing but not in 1.19. I will mention that in kubelet-side issue. The jenkinsfile fixes and modifications are generic so I take these out from here into separate PR.

test/e2e/storage/dax/dax.go
pohly: Ping?

okartau: I filed an issue two days ago as I planned, the link is 2 lines below. Did I create issue in correct location?

pohly (changes requested):

pohly (changes requested):
Please also document in the LVM vs. direct mode comparison that huge pages only work in direct mode.

okartau:
CI test failed for unrelated reason (connectivity failures) and I will re-trigger. But how does the doc change look now? If wording needs to changed, better to do it before re-run.

docs/design.md
```diff
@@ -28,6 +28,7 @@ There is a more detailed explanation in the following paragraphs.
 | *Name* field in namespace | *Name* gets set to 'pmem-csi' to achieve own vs. foreign marking | *Name* gets set to VolumeID, without attempting own vs. foreign marking  |
 |Minimum volume size| 4 MB                   | 1 GB (see also alignment adjustment below) |
 |Alignment requirements |LVM creation aligns size up to next 4MB boundary  |driver aligns  size up to next alignment boundary. The default alignment step is 1 GB. Device(s) in interleaved mode will require larger minimum as size has to be at least one alignment step. The possibly bigger alignment step is calculated as interleave-set-size multiplied by 1 GB |
+|Huge pages supported by `mkfs` options<sup>4</sup>| no| yes|
```
pohly, line 31: Why mention "mkfs" here? Are the options different, or is the result different (huge pages work with one mode but not the other)?

docs/design.md
```diff
@@ -56,6 +57,8 @@ details. `devdax` mode is not supported. Though a
 raw block volume would be useful when a filesystem isn't needed, Kubernetes
 cannot handle [binding a character device to a loop device](https://github.com/kubernetes/kubernetes/blob/7c87b5fb55ca096c007c8739d4657a5a4e29fb09/pkg/volume/util/util.go#L531-L534).
 
+<sup>4 </sup> **Huge pages supported by mkfs options** means that filesystem is created using options supporting huge pages. See [Using Persistent Memory Devices with the Linux Device Mapper](https://pmem.io/2018/05/15/using_persistent_memory_devices_with_the_linux_device_mapper.html). The options supporting huge pages are used in both modes, but page faults related to huge pages occur in the direct mode and not in the LVM mode.
```
pohly, line 60: Can you make the link more specific or describe what I should look for in that page? It's not immediately obvious (probably the mkfs parameters and `xfs_io -c "extsize 2m"`.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -270,6 +270,16 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if ephemeral && fsType == "xfs" {
+		// FS was created only in ephemeral case.
+		// Only if created filesytem and it was XFS:
+		// Align XFS file system for hugepages
+		_, err := pmemexec.RunCommand("xfs_io", "-c", "extsize 2m", hostMount)
```
pohly, line 277: Can you add a comment explaining why 2m is used here?

docs/design.md
okartau: I planned to have it first as "Huge pages supported" but then I was thinking, for technically oriented reader, I could add few words and make the information much richer, i.e. what exactly enables that support. The diff is really only in mkfs options. Actually, if we want to be really precise, it is added mkfs options and added xfs_io call in xfs case. So in that sense, yes perhaps we should not be too specific here, a technical person will find that information if really wants

pohly: Let's use "Huge pages supported" here. But looking at your explanation below, it doesn't really "explain" this difference. It just states that we observed it during testing.

So under different circumstances it might also work in direct mode (alignment perhaps?) - we simply don't know, do we?

docs/design.md
pohly: I would rewrite this as:
```
**Huge pages supported**: ext4 and XFS filesystems are created using the options that should enable huge page support, as explained in ["Using Persistent Memory Devices with the Linux Device Mapper"](https://pmem.io/2018/05/15/using_persistent_memory_devices_with_the_linux_device_mapper.html). [Testing that support by observing page faults](/test/cmd/pmem-access-hugepages/main.go) confirmed that this worked for LVM mode. It did not work for direct mode in the QEMU virtual machines, but it cannot be ruled out that it works elsewhere.
```

And then turn "no" into "maybe" in the table...

pohly: Please preserve line breaks. A full paragraph in a single line makes it harder to comment on a change future diffs will be larger.

test/e2e/storage/dax/dax.go
```diff
@@ -359,3 +361,127 @@ func testDaxOutside(
...
+		   The following 2 items can be enabled after that gets fixed.
+	           https://github.com/kubernetes/kubelet/issues/28
```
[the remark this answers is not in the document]
pohly: Is there now a bug in kubernetes/kubernetes?

docs/design.md
```diff
@@ -28,6 +28,7 @@ There is a more detailed explanation in the following paragraphs.
 | *Name* field in namespace | *Name* gets set to 'pmem-csi' to achieve own vs. foreign marking | *Name* gets set to VolumeID, without attempting own vs. foreign marking  |
 |Minimum volume size| 4 MB                   | 1 GB (see also alignment adjustment below) |
 |Alignment requirements |LVM creation aligns size up to next 4MB boundary  |driver aligns  size up to next alignment boundary. The default alignment step is 1 GB. Device(s) in interleaved mode will require larger minimum as size has to be at least one alignment step. The possibly bigger alignment step is calculated as interleave-set-size multiplied by 1 GB |
+|Huge pages supported<sup>2</sup> | no| yes|
```
pohly, line 31: I just noticed that the footnote number doesn't match the new footnote. This needs to be 4, not 2.

Also, can you replace "no" with "maybe" to match our current understanding of the problem?

test/e2e/storage/dax/dax.go
```diff
@@ -359,3 +361,127 @@ func testDaxOutside(
...
+) {
+	ns := f.Namespace.Name
+	containerName := pod.Spec.Containers[0].Name
+
+	// run trace monitor on all workers
+	for worker := 1; ; worker++ {
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		ssh := exec.Command(sshcmd, "sudo sh -c 'echo 1 > /sys/kernel/debug/tracing/events/fs_dax/dax_pmd_fault_done/enable; echo 1 > /sys/kernel/debug/tracing/tracing_on; cat /sys/kernel/debug/tracing/trace_pipe > /tmp/tracetmp 2>&1 &'")
+		_, err := ssh.Output()
+		if err != nil && os.IsNotExist(err) {
```
pohly, line 455: We've changed that elsewhere already so that the existence of ssh.%d file is checked before using it instead of relying on a specific error after usage.

Can you also do that here?

test/e2e/storage/dax/dax.go
```diff
@@ -359,3 +361,127 @@ func testDaxOutside(
...
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		ssh := exec.Command(sshcmd, "sudo sh -c 'echo 0 > /sys/kernel/debug/tracing/events/fs_dax/dax_pmd_fault_done/enable; echo 0 > /sys/kernel/debug/tracing/tracing_on; pkill -f cat\\ /sys/kernel/debug/tracing/trace_pipe'")
+		_, err := ssh.Output()
+
+		// There may be garbage (zero values) at start, we get better results by counting from end, thats why we use NF-relative fields in awk
+		ssh = exec.Command(sshcmd, "cat /tmp/tracetmp|awk '{print $(NF-13) $(NF-9)}'")
+		traceOutput, err := ssh.Output()
+		if err != nil && os.IsNotExist(err) {
+			// if we are here, no worker node had tracer output
+			By("Unexpected: no worker has tracer output?")
```
pohly, line 472: Shouldn't that be a test failure?

No-one is going to read the CI output to find this message unless the test is marked as failed...

okartau: turned that into failure

test/e2e/storage/dax/dax.go
okartau: changed to check existence as was in leak check code

test/e2e/storage/dax/dax.go
```diff
@@ -358,3 +360,137 @@ func testDaxOutside(
...
+	root string,
+	source *v1.VolumeSource,
+) {
+	ns := f.Namespace.Name
+	containerName := pod.Spec.Containers[0].Name
+
+	// run trace monitor on all workers
+	for worker := 1; ; worker++ {
+		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
+		if _, err := os.Stat(sshcmd); err == nil {
```
pohly, line 452: Nit: check for NotFound.

test/e2e/storage/dax/dax.go
```diff
@@ -358,3 +360,137 @@ func testDaxOutside(
...
+			ssh := exec.Command(sshcmd, "sudo sh -c 'echo 1 > /sys/kernel/debug/tracing/events/fs_dax/dax_pmd_fault_done/enable; echo 1 > /sys/kernel/debug/tracing/tracing_on; cat /sys/kernel/debug/tracing/trace_pipe > /tmp/tracetmp 2>&1 &'")
+			_, err = ssh.Output()
+			if err != nil {
+				framework.Failf("Failed to start pagefault tracing: %v", err)
+			}
+		} else {
```
pohly, line 458: Nit: `if/else` is non-idiomatic. Normally one would do `if ... { break }` and then reduce the indention of the alternative case, without using `else`.

pohly (approved):
Some nits, but those can be ignored.

Let's merge after waiting for the test results.

okartau:
Latest run failed with 108 cases under fedora-1_19.olm.API.deployment and more, seems unrelated, so there is chance to address error checking and if/else parts now. Also, I got idea to add worker number to fail messages, probably not that important but can help in debugging if needed.

pohly:
> Latest run failed with 108 cases under fedora-1_19.olm.API.deployment and more, seems unrelated, so there is chance to address error checking and if/else parts now.

Work on the alignment problem is more important. Let's merge this PR as-is.

okartau:
hm, just got it edited and tested meanwhile.. also added few more little improvements like ExpextNoError instead of two instances of if-err-check. I can throw it away, or create another PR as further improvement.

pohly:
If you want those enhancements to be merged, then please create another PR.

test/e2e/storage/dax/dax.go
```diff
@@ -358,3 +360,137 @@ func testDaxOutside(
...
+
+// Hugepage page fault testing part starts here
+const (
+	accessHugepagesBinary = "_work/pmem-access-hugepages"
+)
+
+var _ = deploy.DescribeForSome("dax", func(d *deploy.Deployment) bool {
+	// Run these tests for all driver deployments that were created
+	// through the operator, where device mode is Direct.
+	return !d.HasOperator && d.HasDriver && d.Mode == api.DeviceModeDirect
```
pohly, line 372: @okartau: I just noticed that the comment doesn't match the code. Should this run for deployments done with or without the operator? Why that criteria?

Issue #501, opened 2019-12-19
CreateVolume: reject unknown parameters


pohly:
The following code should be changed to iterate over key/value pairs in the parameters and reject unknown keys:
- https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/controllerserver-master.go#L180-L198
- https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/controllerserver-node.go#L151-L166

PR #504, opened 2019-12-20
CSI driver: unify and enhance volume parameter handling

pohly:
Validation of user input was missing, with the result that misspelled
keys were silently ignored.

`persistencyModel: none` was supported, but not documented. `none` is
a misleading name, so it gets renamed to `normal` (a breaking
change!).

Support for units in the ephemeral inline volume size was undocumented
and (against expectations) was implemented such that 1G expanded to
1000 * 1000 * 1000 (decimal unit). In this context binary units make
more sense, so those are used now (another breaking change).

There was support for `persistencyModel: ephemeral` in the master
controller, but that value is invalid there, so now this triggers an
error and all code around ephemeral volumes in the master controller
could be removed.

Fixes: #501


pohly:
Before investing more effort into this I'd like to get some review feedback whether this makes sense.

To do before merging:
- add more tests
- switch to Kubernetes resource for parsing the size

README.md
```diff
@@ -404,7 +404,7 @@ Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The v
 
 |key|meaning|optional|values|
 |---|-------|--------|-------------|
-|`size`|Size of the requested ephemeral volume|No||
+|`size`|Size of the requested ephemeral volume, binary units are support ("1k" == 1024 bytes)|No||
```
avalluri: support => supported

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -140,62 +133,95 @@ func (cs *nodeControllerServer) CreateVolume(ctx context.Context, req *csi.Creat
...
 	}
 
-	asked := req.GetCapacityRange().GetRequiredBytes()
-
-	if vol = cs.getVolumeByName(req.Name); vol != nil {
+	asked := capacity.GetRequiredBytes()
+	if vol := cs.getVolumeByName(volumeName); vol != nil {
 		// Check if the size of existing volume can cover the new request
-		klog.V(4).Infof("Node CreateVolume: Volume exists, name:%s id:%s size:%v", req.Name, vol.ID, vol.Size)
+		klog.V(4).Infof("create volume: volume exists, name: %q id: %s size: %v", volumeName, vol.ID, vol.Size)
```
okartau: You think spelling out "Node" in log msg is not needed?
I added Node and Controller to many log msgs some time ago, and I find those still useful, even though logging change later made file names also visible (which were not when I added the prefixes). The prefixes help to see diff Node vs Controller msgs which in multiple cases have exact same wording.
You want to get rid of those prefixes now?
But this change creates non-uniform landscape as many other places still have those.

README.md
```diff
@@ -353,10 +353,10 @@ to applications using
 [`StorageClass
 Parameters`](https://kubernetes.io/docs/concepts/storage/storage-classes/#parameters). An
 optional `persistencyModel` parameter differentiates how the
-provisioned volume can be used.
+provisioned volume can be used:
 
-* if no `persistencyModel` parameter specified in `StorageClass` then
-  it is treated as normal Kubernetes persistent volume. In this case
+* no `persistencyModel` parameter or `persistencyModel: normal` in `StorageClass`
```
okartau: extra spaces (2 of them) at the end of line. Note that GH diff view shows me just one (is it so that one gets added by this change?)
 `git diff` on cmdline shows two.
Note also there are few other occurrences in the file of extra space at end of bullet-line (and others), but outside of scope of this change.

pkg/pmem-csi-driver/controllerserver-node.go
pohly: I can add back "node" if needed.

The main intention was to make the capitalization simpler (there's no reason to write "node" with upper case here, same for "Volume") and make it easier to read with some extra spaces (but that's of course subjective).

I can take out the format string change.

README.md
pohly: Two spaces are intentional, please don't remove them: https://meta.stackexchange.com/questions/72395/is-it-possible-to-have-definition-lists-in-markdown/298872#298872

okartau: Thanks for pointer, I did not know that feature. IMHO it is not so good idea to use trailing spaces for formatting, but if it's so then let's try to maintain those.
Not looking so good via `git diff` which indicates trailing whitespace as problem.

README.md
pohly: Other than this and the log message changes that @okartau pointed out, is this PR conceptually okay?

avalluri: I didn't check fully yet but it sounds like useful change. I will review it now.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -129,7 +124,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		req.GetVolumeId(), targetPath, srcPath, readOnly, mountFlags, fsType)
 
 	// Kubernetes v1.16+ would request ephemeral volumes via VolumeContext
-	val, ok := req.GetVolumeContext()[volumeContextEphemeral]
+	val, ok := req.GetVolumeContext()[parameterEphemeral]
```
avalluri: `"csi.storage.k8s.io/ephemeral"`is not a parameter passed by the end-user. Better not to mix with parameter handling.

pkg/pmem-csi-driver/parameters.go
```diff
@@ -0,0 +1,272 @@
...
+
+type persistencyModel string
+type parameterOrigin int
+
+// Beware of API and backwards-compatibility breaking when changing these string constants!
+const (
+	parameterCacheSize        = "cacheSize"
+	parameterEraseAfter       = "eraseafter"
+	parameterName             = "name"
+	parameterPersistencyModel = "persistencyModel"
```
avalluri: The `"persistencyModel"` parameter was a result of the initial thought of supporting different volume models using storage class, which I believe is not valid anymore. The `"cacheSize" `itself suffices for differentiating a cache volume. So we better to deprecate this parameter.

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -247,23 +262,20 @@ func (cs *nodeControllerServer) DeleteVolume(ctx context.Context, req *csi.Delet
...
-				eraseafter = bVal
-			}
-		}
-	} else {
+	vol := cs.getVolumeByID(req.VolumeId)
+	if vol == nil {
+		// Already deleted.
 		return &csi.DeleteVolumeResponse{}, nil
 	}
+	parameters, err := parseVolumeParameters(nodeVolumeParameters, vol.Params)
```
avalluri: `vol.Params` are pre-validated and re-validating them here makes not much sense to me. Instead would prefer to change `Params` in `nodeVolume` struct from `map[stirng]string` to `volumeParameters`.

pkg/pmem-csi-driver/parameters.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+// validParameters is a whitelist of which parameters are valid in which context.
+var validParameters = map[parameterOrigin][]string{
+	// Parameters from Kubernetes and users for a persistent volume.
+	createVolumeParameters: []string{
+		parameterCacheSize,
+		parameterEraseAfter,
+		parameterPersistencyModel,
+		parameterPodInfoPrefix,
```
avalluri: Pod info is not valid in this context I guess.

pkg/pmem-csi-driver/parameters.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Parameters from Kubernetes and users.
+	ephemeralVolumeParameters: []string{
+		parameterEraseAfter,
+		parameterPodInfoPrefix,
+		parameterSize,
+	},
+
+	// The volume context prepared by CreateVolume.
+	persistentVolumeParameters: []string{
```
avalluri: How is it different from `createVolumeParameters`

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -271,7 +266,15 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	}
 
 	if vol == nil {
-		return nil, status.Error(codes.NotFound, fmt.Sprintf("No volume found with volume id '%s'", req.VolumeId))
+		return nil, status.Errorf(codes.NotFound, "no volume found with volume id %q'", req.VolumeId)
+	}
+
+	parameters, err := parseVolumeParameters(nodeVolumeParameters, vol.Params)
```
avalluri: Same here as mentioned at https://github.com/intel/pmem-csi/pull/504/files#r362738437

pkg/pmem-csi-driver/parameters.go
```diff
@@ -0,0 +1,272 @@
...
+		parameterPersistencyModel,
+		parameterCacheSize,
+		parameterEraseAfter,
+	},
+
+	// Internally we store everything except the volume ID
+	// which is handled separately.
+	nodeVolumeParameters: []string{
```
avalluri: It can be removed if we agree with https://github.com/intel/pmem-csi/pull/504/files#r362738437

pkg/pmem-csi-driver/controllerserver-node.go
pohly: `vol.Params` contains volume parameters that were potentially serialized by some older version of PMEM-CSI. We can't assume that they always match exactly with what currently is stored in `volumeParameters`.

If we allow backwards-incompatible changes, then this validation would be were we give up and return an error.

But as the next comment says, that would be very bad and force an admin to get involved. It'll be better to add support for old parameters into `parseVolumeParameters` such that they can still be converted into `volumeParameters` even if the parameter itself is deprecated and/or not allowed anymore for new volumes.

pkg/pmem-csi-driver/nodeserver.go
pohly: "parameter" does not mean that is has to come directly from an end-user. I was using it in the more general meaning of "attributes" of a volume. With that in mind, the name `parameterEphemeral` makes sense to me. :shrug:

Regarding "mixing": we don't have a choice here. This `req.GetVolumeContext()` can contain parameters set by us (in `CreateVolume` and thus indirectly by the user), parameters set by a user (the inline volume parameters) and by Kubernetes (pod info, parameterEphemeral).

I think it makes sense to use the same `parameter` prefix for all of those strings to document that they all appear as keys in the same map.

pkg/pmem-csi-driver/parameters.go
pohly: So you are proposing to deprecate `parameterPersistencyModel` and instead say that any volume with `cacheSize` > 1 is a cache volume?

That could work.

pkg/pmem-csi-driver/parameters.go
pohly: This is what we get in `CreateVolume`. We have enabled pod info, so we get those additional parameters.

pkg/pmem-csi-driver/parameters.go
pohly: We have filtered out the pod info.

pkg/pmem-csi-driver/parameters.go
pohly: Let's discuss and implement this separately, i.e. not make that change in this PR.

pohly:
@avalluri: can you check my replies? Have they addressed your concerns?

@okartau: I pushed a revised commit which changes log messages as little as possible.

I also addressed all todo items from https://github.com/intel/pmem-csi/pull/504#issuecomment-568011952.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: In my opinion `parseVolumeParameters()` are supposed to call in two places:
 - while creating a (persisten/ephemeral/internal) volume - to validate params/context
 - while restoring volume info from state - to validate any backward-incompatibilites

and save the results in our cached voulume structures, Once after we should deal as Parameters.

pohly: I started looking into that, but got stuck with the design of the statemanager: it deals with opaque objects that can be serialized to and from JSON. So `parseVolumeParameters` cannot be called there.

That means we have to keep `nodeVolume` as it is and validate at the level above statemanager. But then we have to be sure that we really catch all places where  `nodeVolume` is retrieved from the statemanager. The Go compiler will not warn us if we miss one.

I'd rather validate too often (i.e. before each use) instead of not often enough.

pohly: Another advantage of validating before each use is that a failure is clearly associated with a certain volume and operation.

If we validate in `NewNodeControllerServer` (which is one place were JSON files are loaded), then all we can do is log some warning and ignore the existing volume. I'm almost certain that this would lead to leaked volume (as in `DeleteVolume` called, volume state couldn't be loaded and thus call succeeds, Kubernetes and the user think that the volume is gone whereas the actual state is "volume could not be deleted").

pohly: > But then we have to be sure that we really catch all places where nodeVolume is retrieved from the statemanager.

... and then decide where to store the parsed parameters. Convert back to `map[string]string` and store them in the existing `nodeVolume` or create a `nodeInternalVolume` where `Params` is of type `volumeParameters` instead of `map[string]string`? Both is more complicated than what's currently in this PR.

avalluri: > That means we have to keep `nodeVolume` as it is and validate at the level above statemanager. But then we have to be sure that we really catch all places where `nodeVolume` is retrieved from the statemanager.

You can just provide json marshal interface methods to this new type `volumeParameters`:
```
func (p volumeParameters) MarshalJSON() ([]byte, error) {
    return json.Marshal(p.toVolumeContext())
}
```

This should be enough to save params using state manager.

pkg/pmem-csi-driver/parameters.go
avalluri: The PodInfo info is passed to `NodePublishVolume` only I guess, not to `CreateVolume`. Please correct me if I am wrong.

pkg/pmem-csi-driver/parameters.go
```diff
@@ -0,0 +1,276 @@
...
+	persistencyEphemeral persistencyModel = "ephemeral" // only used internally
+
+	// Parameters are from storage class in controller CreateVolume.
+	createVolumeParameters parameterOrigin = iota
+	// Parameters are from master in node CreateVolume.
+	createVolumeInternalParameters
+	// Parameters are for an ephemeral volume in NodePublishVolume.
+	ephemeralVolumeParameters
+	// Parameters are for a persistent volume in NodePublishVolume.
+	persistentVolumeParameters
```
avalluri: Origin of this(`persistentVolumeParameters`) type is defined here but not used anywhere in the code. I think we can remove.

pkg/pmem-csi-driver/parameters.go
avalluri: No problem to take it as separate PR, As this code is touching parameters so I thought its the right time to deprecate persistency model param.

okartau:
> I pushed a revised commit which changes log messages as little as possible.

thanks, I see you even added one "Node", making log msgs more consistent!

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: > ... and then decide where to store the parsed parameters. Convert back to `map[string]string` and store them in the existing `nodeVolume`

This is exactly what my proposal is. Deal everywhere as `volumeParameters` once after parsing.

pohly: > You can just provide json marshal interface methods to this new type volumeParameters

That would work, except that it then has the problem I mentioned in https://github.com/intel/pmem-csi/pull/504#discussion_r363800540: suppose there is one "old" volume that for whatever reason cannot be handled anymore by the current PMEM-CSI version, which will be indicated by `parseVolumeParameters` by returning an error. Other volumes are fine.

In my current proposal, this will be noticed and reported in operations that deal with that volume. Those operations then fail in an obvious way and an admin can try to resolve the problem manually.

How do you propose to handle this aspect?

pkg/pmem-csi-driver/parameters.go
pohly: You are probably right, I was confused about where pod info is provided. I'll fix this.

I still prefer to keep `persistentVolumeParameters` and `createVolumeParameters` separate even if they (currently) have the same content because conceptually they are different.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: If the driver is working with such an in-compatible state it would be good to fail as early as possible while node-driver initialization with an appropriate error message.

pohly: But the *driver* is not in an incompatible state. Only some *volumes* are. It doesn't make sense to me to then also refuse to do any work for those volumes that are still okay.

pkg/pmem-csi-driver/parameters.go
pohly: It's actually the opposite: the code which should have used it (the persistent volume code path in `NodePublishVolume`) didn't validate the content of the volume context. That's bad because robust code should reject unknown and/or unexpected input. I've fixed and at the same time added the volume name to the volume context because that is actually more useful than replicating the `CreateVolume` parameters.

pohly:
I've pushed an updated version. Somehow my editors stopped calling gofmt, so I had to fix some formatting errors.

The changes besides that are:
```
diff --git a/pkg/pmem-csi-driver/controllerserver-master.go b/pkg/pmem-csi-driver/controllerserver-master.go
index 5622bddf..3a40fb8c 100644
--- a/pkg/pmem-csi-driver/controllerserver-master.go
+++ b/pkg/pmem-csi-driver/controllerserver-master.go
@@ -242,6 +242,11 @@ func (cs *masterController) CreateVolume(ctx context.Context, req *csi.CreateVol
                })
        }
 
+       // Volume ID and name are not the same. Store the original
+       // name in the volume context for logging purposes.
+       name := req.GetName()
+       parameters.name = &name
+
        return &csi.CreateVolumeResponse{
                Volume: &csi.Volume{
                        VolumeId:           vol.id,
diff --git a/pkg/pmem-csi-driver/controllerserver-node.go b/pkg/pmem-csi-driver/controllerserver-node.go
index cc84691c..3405fe12 100644
--- a/pkg/pmem-csi-driver/controllerserver-node.go
+++ b/pkg/pmem-csi-driver/controllerserver-node.go
@@ -272,7 +272,7 @@ func (cs *nodeControllerServer) DeleteVolume(ctx context.Context, req *csi.Delet
                // This should never happen because PMEM-CSI itself created these parameters.
                // But if it happens, better fail and force an admin to recover instead of
                // potentially destroying data.
-               return nil, status.Errorf(codes.Internal, "previously stored volume parameters: %v", err)
+               return nil, status.Errorf(codes.Internal, "previously stored volume parameters for volume with ID %q: %v", req.VolumeId, err)
        }
 
        if err := cs.dm.DeleteDevice(req.VolumeId, parameters.getEraseAfter()); err != nil {
diff --git a/pkg/pmem-csi-driver/nodeserver.go b/pkg/pmem-csi-driver/nodeserver.go
index 175bf3be..6bdeafb8 100644
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -120,8 +120,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
        mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
        readOnly := req.GetReadonly()
        fsType := req.GetVolumeCapability().GetMount().GetFsType()
-       klog.V(3).Infof("NodePublishVolume request: VolID:%s targetpath:%v sourcepath:%v readonly:%v mount flags:%v fsType:%v",
-               req.GetVolumeId(), targetPath, srcPath, readOnly, mountFlags, fsType)
+       volumeContext := req.GetVolumeContext()
+       // volumeContext contains the original volume name for persistent volumes.
+       klog.V(3).Infof("NodePublishVolume request: VolID:%s targetpath:%v sourcepath:%v readonly:%v mount flags:%v fsType:%v context:%q",
+               req.GetVolumeId(), targetPath, srcPath, readOnly, mountFlags, fsType, volumeContext)
 
        // Kubernetes v1.16+ would request ephemeral volumes via VolumeContext
        val, ok := req.GetVolumeContext()[parameterEphemeral]
@@ -152,6 +154,11 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
                srcPath = device.Path
                mountFlags = append(mountFlags, "dax")
        } else {
+               // Validate parameters. We don't actually use any of them here, but a sanity check is worthwhile anyway.
+               if _, err := parseVolumeParameters(persistentVolumeParameters, req.GetVolumeContext()); err != nil {
+                       return nil, status.Error(codes.InvalidArgument, "persistent volume context: "+err.Error())
+               }
+
                if device, err = ns.cs.dm.GetDevice(req.VolumeId); err != nil {
                        if errors.Is(err, pmdmanager.ErrDeviceNotFound) {
                                return nil, status.Errorf(codes.NotFound, "no device found with volume id %q: %v", req.VolumeId, err)
@@ -274,7 +281,7 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
                // This should never happen because PMEM-CSI itself created these parameters.
                // But if it happens, better fail and force an admin to recover instead of
                // potentially destroying data.
-               return nil, status.Errorf(codes.Internal, "previously stored volume parameters: %v", err)
+               return nil, status.Errorf(codes.Internal, "previously stored volume parameters for volume with ID %q: %v", req.VolumeId, err)
        }
 
        // Check if the target path is really a mount point. If its not a mount point do nothing
@@ -429,7 +436,7 @@ func (ns *nodeServer) NodeExpandVolume(context.Context, *csi.NodeExpandVolumeReq
 func (ns *nodeServer) createEphemeralDevice(ctx context.Context, req *csi.NodePublishVolumeRequest) (*pmdmanager.PmemDeviceInfo, error) {
        parameters, err := parseVolumeParameters(ephemeralVolumeParameters, req.GetVolumeContext())
        if err != nil {
-               return nil, status.Error(codes.InvalidArgument, "ephemeral inline volume: "+err.Error())
+               return nil, status.Error(codes.InvalidArgument, "ephemeral inline volume parameters: "+err.Error())
        }
 
        // Create new device, using the same code that the normal CreateVolume also uses,
@@ -472,7 +479,12 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType
                args = []string{"-b 4096", "-F", device.Path}
        } else if fsType == "xfs" {
                cmd = "mkfs.xfs"
-               args = []string{"-b", "size=4096", "-f", device.Path}
+               // reflink and DAX are mutually exclusive
+               // (http://man7.org/linux/man-pages/man8/mkfs.xfs.8.html).
+               // Recent kernels started to check for this instead of
+               // allowing both to be active
+               // (https://www.spinics.net/lists/linux-xfs/msg15448.html).
+               args = []string{"-b", "size=4096", "-m", "reflink=0", "-f", device.Path}
        } else {
                return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
        }
diff --git a/pkg/pmem-csi-driver/parameters.go b/pkg/pmem-csi-driver/parameters.go
index 076ec962..c74922f2 100644
--- a/pkg/pmem-csi-driver/parameters.go
+++ b/pkg/pmem-csi-driver/parameters.go
@@ -33,6 +33,9 @@ const (
        // Additional, unknown parameters that are okay.
        parameterPodInfoPrefix = "csi.storage.k8s.io/"
 
+       // Added by https://github.com/kubernetes-csi/external-provisioner/blob/feb67766f5e6af7db5c03ac0f0b16255f696c350/pkg/controller/controller.go#L584
+       parameterProvisionerID = "storage.kubernetes.io/csiProvisionerIdentity"
+
        persistencyNormal    persistencyModel = "normal" // In releases <= 0.6.x this was called "none", but not documented.
        persistencyCache     persistencyModel = "cache"
        persistencyEphemeral persistencyModel = "ephemeral" // only used internally
@@ -56,7 +59,6 @@ var validParameters = map[parameterOrigin][]string{
                parameterCacheSize,
                parameterEraseAfter,
                parameterPersistencyModel,
-               parameterPodInfoPrefix,
        },
 
        // These parameters are prepared by the master controller.
@@ -75,14 +77,22 @@ var validParameters = map[parameterOrigin][]string{
                parameterSize,
        },
 
-       // The volume context prepared by CreateVolume.
+       // The volume context prepared by CreateVolume. We replicate
+       // the CreateVolume parameters in the context because a future
+       // version of PMEM-CSI might need them (the current one
+       // doesn't) and add the volume name for logging purposes.
+       // Kubernetes adds pod info and provisioner ID.
        persistentVolumeParameters: []string{
-               parameterPersistencyModel,
                parameterCacheSize,
                parameterEraseAfter,
+               parameterPersistencyModel,
+
+               parameterName,
+               parameterPodInfoPrefix,
+               parameterProvisionerID,
        },
 
-       // Internally we store everything except the volume ID
+       // Internally we store everything except the volume ID,
        // which is handled separately.
        nodeVolumeParameters: []string{
                parameterCacheSize,
@@ -183,6 +193,7 @@ func parseVolumeParameters(origin parameterOrigin, stringmap map[string]string)
                                p := persistencyEphemeral
                                result.persistency = &p
                        }
+               case parameterProvisionerID:
                default:
                        if !strings.HasPrefix(key, parameterPodInfoPrefix) {
                                return result, fmt.Errorf("unknown parameter: %q", key)
diff --git a/pkg/pmem-csi-driver/parameters_test.go b/pkg/pmem-csi-driver/parameters_test.go
index 90dc7eff..a6e656d7 100644
--- a/pkg/pmem-csi-driver/parameters_test.go
+++ b/pkg/pmem-csi-driver/parameters_test.go
@@ -41,7 +41,6 @@ func TestParameters(t *testing.T) {
                                parameterCacheSize:        "5",
                                parameterEraseAfter:       "false",
                                parameterPersistencyModel: "cache",
-                               "csi.storage.k8s.io/foo":  "bar",
                        },
                        parameters: volumeParameters{
                                cacheSize:   &five,
@@ -103,11 +102,16 @@ func TestParameters(t *testing.T) {
                                parameterCacheSize:        "5",
                                parameterEraseAfter:       "false",
                                parameterPersistencyModel: "cache",
+
+                               parameterName:            name,
+                               "csi.storage.k8s.io/foo": "bar",
+                               parameterProvisionerID:   "provisioner XYZ",
                        },
                        parameters: volumeParameters{
                                cacheSize:   &five,
                                eraseAfter:  &no,
                                persistency: &cache,
+                               name:        &name,
                        },
                },
                {
@@ -158,9 +162,9 @@ func TestParameters(t *testing.T) {
                        name:   "invalid-persistent-context",
                        origin: persistentVolumeParameters,
                        stringmap: map[string]string{
-                               parameterPodInfoPrefix + "foo": "bar",
+                               "foo": "bar",
                        },
-                       err: "parameter \"csi.storage.k8s.io/foo\" invalid in this context",
+                       err: "parameter \"foo\" invalid in this context",
                },
                {
                        name:   "invalid-node-context",
@@ -215,7 +219,9 @@ func TestParameters(t *testing.T) {
                                                value = "normal"
                                        }
                                }
-                               if key != parameterVolumeID && !strings.HasPrefix(key, parameterPodInfoPrefix) {
+                               if key != parameterVolumeID &&
+                                       key != parameterProvisionerID &&
+                                       !strings.HasPrefix(key, parameterPodInfoPrefix) {
                                        result[key] = value
                                }
                        }
```

I'm including a commit which fixes XFS mounting on the Fedora test cluster. I can also submit that in a separate PR.

pkg/pmem-csi-driver/nodeserver.go
avalluri: @pohly Ok, I got your idea of treating volume contexts also like parameters. This case better to parse/validate the volume context bit beforehand so that we can avoid special treatment for `parameterEphemeral` and `volumeProvisionerIdentity/parameterProvisionerID`.

I would propose to change `ephemeralVolumeParameters` origin to something like `publishVolumeParameters` and parse them beforehand and use `params.Ephemeral()` params.ProvisionerID()` calls

pohly:
Rebased on top of the merged XFS PR.

pkg/pmem-csi-driver/nodeserver.go
pohly: But parameters for publishing of ephemeral volumes and for persistent volumes are not the same. With your proposal, `publishVolumeParameters` would have to be a superset of both and then we won't catch when a user erroneously puts "cacheSize" into the inline volume spec. That's why we need to validate differently after determining what the volume is.

avalluri: > we won't catch when a user erroneously puts "cacheSize" into the inline volume spec.

Yes, agree.

PR #505, opened 2020-01-08
XFS: disable reflink

pohly:
An update of Clear Linux from 32020 to 32050 replaces xfstools 5.0.0
with 5.4.0 and that enables reflink by default. "-o dax" and reflink
are mutually exclusive, so we now have to override the default.

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -494,7 +494,9 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 		args = []string{"-b 4096", "-F", device.Path}
 	} else if fsType == "xfs" {
 		cmd = "mkfs.xfs"
-		args = []string{"-b", "size=4096", "-f", device.Path}
+		// reflink and DAX are mutually exclusive
+		// (http://man7.org/linux/man-pages/man8/mkfs.xfs.8.html).
+		args = []string{"-b", "size=4096", "-m", "reflink=0", "-f", device.Path}
 	} else {
 		return fmt.Errorf("Unsupported filesystem '%s'. Supported filesystems types: 'xfs', 'ext4'", fsType)
 	}


pohly:
I encountered this problem during `TEST_DISTRO=fedora make test_e2e` locally with a newly built CSI driver image and cluster brought up by that make invocation.

pohly:
Why it occurs now and not earlier is a bit unclear. Perhaps mkfs.xfs in recent Clear Linux started to enable reflink by default only recently?

pohly:
I'm still not sure how I ended up with reflink enabled. It should be off by default. I tried with recent Clear Linux and (if I read the output correctly), it was still off:
```
root@fbaf3f6d7149/ # mkfs.xfs -f /loop
meta-data=/loop                  isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=0
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
```

That was just a quick manual test in Docker. Perhaps it would have behaved differently on a Fedora kernel.


Either way, explicitly disabling reflink is better than relying on some potentially changing default.

pohly:
Here's the smoking gun, from the PMEM-CSI driver log for the XFS example:
```
I0109 07:52:36.009450       1 pmem-exec.go:15] Output: meta-data=/dev/ndbus0region0/a7a42d83dddec209b5dcac372b9b5374be0e5d24 isize=512    agcount=4, agsize=262144 blks
         =                       sectsz=4096  attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1
data     =                       bsize=4096   blocks=1048576, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=4096  sunit=1 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
```

This is with current `devel`, cluster and driver image built from scratch. I still don't know why the default changed or why I wasn't seeing that in the manual test before, but the fix is the right one.

okartau:
I can confirm that. In my locally-run F-31-based govm cluster, I saw reflink=0 with driver image (branch=devel) which was built few days ago. After fresh rebuild, I see reflink=1 and same failure: refusal of mount. Something changed very recently.

pohly:
My previous manual test was with the base clearlinux image. After updating with "swupd update" I get a more recent xfstools an that changes the reflink default. Mystery solved... I've updated the PR accordingly.

@okartau please review and approve.

PR #513, opened 2020-01-13
XFS: disable reflink

pohly:
A update of Clear Linux from 32020 to 32050 replaces xfstools 5.0.0
with 5.4.0 and that enables reflink by default. "-o dax" and reflink
are mutually exclusive, so we now have to override the default.

(cherry picked from commit 77374b875527d463d4f875116954b49ad4403596)


okartau:
CI failure "docker login failed" is not related to change but caused by something else, I believe.

okartau:
code change seems ok, but as this is about code, we should see CI OK before merge.

pohly:
@okartau testing now was successful, okay to merge?

PR #514, opened 2020-01-13
docs: describe latest release and how to create a new one

pohly:
It's worthwhile to state right at the beginning of our main landing
page (= README.md) what we currently support and how we handle tags
because users need to know that when picking something to install.

Doing a release involves certain steps. We had documentation for most
of that, but not a single simple check list.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -9,6 +9,7 @@ Table of Contents
 - [Release management](#release-management)
     - [Branching](#branching)
     - [Tagging](#tagging)
+    - [Release checklist](#release-checklist)
 - [APIs](#apis)
     - [CSI API](#csi-api)
     - [Network ports](#network-ports)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +54,13 @@ The PMEM-CSI driver follows the [CSI
 specification](https://github.com/container-storage-interface/spec) by
 listening for API requests and provisioning volumes accordingly.
 
+The [v0.6.0
+release](https://github.com/intel/pmem-csi/releases/tag/v0.6.0) is the
+latest feature release and is [getting
+updated](./DEVELOPMENT.md#release-management) with newer base images
+regularly and with bug fixes as needed, so the latest [v0.6.x
```
okartau: There is a convenient link https://github.com/intel/pmem-csi/releases/latest to get to latest release, should we mention it here? It's not so good to point to latest link while we explicitly list releases by number though, but latest link exists and is helpful for user.

DEVELOPMENT.md
```diff
@@ -144,6 +144,24 @@ release: the actual image only gets pushed when there is a tag that
 corresponds to the version embedded in the source code. The
 Jenkinsfile ensures that.
 
+Release Checklist
```
okartau: The title says Checklist but listed is more like "steps to execute", which for my eye is different thing. Checklist sounds more like things to check after something is already performed?

DEVELOPMENT.md
```diff
@@ -144,6 +144,24 @@ release: the actual image only gets pushed when there is a tag that
...
+-----------------
+
+* Create a new `release-x.y` branch.
+* Run `hack/update-clear-linux-base.sh`.
+* Run `hack/set-version.sh vx.y.z` and commit the modified files.
+* Push to `origin`.
+* [Create a draft
+  release](https://github.com/intel/pmem-csi/releases/new) for that
+  new branch, including a change log gathered from new commits.
+* Review change log.
```
okartau: I think more correct is "Review the change log"

DEVELOPMENT.md
pohly: It's normal for a "checklist" to also include things that need to be done. See for example https://wiki.mozilla.org/Releases/Checklist which contains tasks like "Email release-drivers" or "Push actual bits".

README.md
pohly: The intention is to document what the latest release is, not just describe how to find it.

How would you add the link (full sentence)?

DEVELOPMENT.md
pohly: Indeed.

pohly: Fixed. Please have another look and check also my replies.

README.md
okartau: yeah, it may not be so good idea after all to duplicate link to latest release, it;'s not that difficult to locate by anyone who is interested in releases. Let's not add it.

DEVELOPMENT.md
okartau: ok, good as it is then.

okartau: another thing about section header: should it be added in TOC?

pohly: Yes, good catch. Fixed.

At some point we need to automate this. The kubernetes/enhancements repo has a script and check for it that we could copy.

PR #515, opened 2020-01-15
test Kubernetes 1.17 on Clear Linux 32080

pohly:
We don't need to change anything for Kubernetes 1.17 except adding a
symlink so that "make start" knows which deployment files to use.


pohly:
This should be backported to release-0.6 because otherwise using the latest Clear doesn't work there (missing kubernetes-1.17 symlink).

pohly:
@okartau: testing passed, okay to merge?

PR #517, opened 2020-01-15
build improvements

pohly:
One cosmetic change and one fix.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,13 +86,14 @@ ARG BIN_SUFFIX
 # which flavor (production or testing). The actual binary name in the
 # image is going to be the same, to avoid unnecessary deployment
 # differences.
-RUN make VERSION=${VERSION} pmem-csi-driver${BIN_SUFFIX} pmem-vgm${BIN_SUFFIX} pmem-ns-init${BIN_SUFFIX} && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -86,13 +86,14 @@ ARG BIN_SUFFIX
 # which flavor (production or testing). The actual binary name in the
 # image is going to be the same, to avoid unnecessary deployment
 # differences.
-RUN make VERSION=${VERSION} pmem-csi-driver${BIN_SUFFIX} pmem-vgm${BIN_SUFFIX} pmem-ns-init${BIN_SUFFIX} && \
+RUN set -x && \
```
avalluri: @pohly Is `set -x` is intentional here? I guess by default all RUN statements are echoed by `docker build`.

avalluri: @pohly Ignore my earlier comment. Here I see adding `set -x` seems useful for debugging individual commands inside RUN

avalluri:
Looks good to me.

Dockerfile
pohly: Exactly. In the output without `-x` it was impossible to see which command actually caused Go to fetch dependencies.

PR #519, opened 2020-01-16
test: dump docker logs for failed govm cluster

pohly:
When the machines don't come up, we need the "docker logs" output to
understand why.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -460,6 +460,13 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
         up ourselves. "make stop" was hanging and waiting for these processes to
         exit even though there were from a different "docker exec" invocation.
 
+        The default QEMU cpu enables nested virtualization with "-cpu
+        host".  However, that fails on some Azure machines
+        (`qemu-system-x86_64: error: failed to set MSR 0x48b to
+        0x1582e00000000`,
+        https://www.mail-archive.com/qemu-devel@nongnu.org/msg665051.html),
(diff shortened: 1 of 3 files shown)


pohly:
The change itself should be fine, but let's run the PR test job a few times to see whether it really catches something.

pohly:
And here's the root cause (first PR test succeeded, second, third and forth failed):
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-519/2/execution/node/97/log/
```
21:41:20  qemu-system-x86_64: -realtime mlock=off: warning: '-realtime mlock=...' is deprecated, please use '-overcommit mem-lock=...' instead
21:41:20  2020-01-16T20:39:18.186048Z qemu-system-x86_64: error: failed to set MSR 0x48b to 0x1582e00000000
21:41:20  Assertion failed: ret == cpu->kvm_msr_buf->nmsrs (/home/buildozer/aports/community/qemu/src/qemu-4.2.0/target/i386/kvm.c: kvm_put_msrs: 2947)
```

Searching for that error leads to https://www.mail-archive.com/qemu-devel@nongnu.org/msg665051.html where @devimc  ran into the same problem.

@devimc: any idea why it sometimes works?

I'll try adding the `-vmx` workaround.

pohly:
`-cpu host,-vmx` worked reliably in three builds, so I made it a permanent workaround for our CI.

devimc:
@pohly seems like this is a kvm bug in hyperv

PR #520, opened 2020-01-17
test: include docker0 interface address in no_proxy

pohly:
The default IP address for reaching the developer Docker registry
comes from "ip addr show dev docker0" which happens to be a /16
subnet:

```
9: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether 02:42:79:6d:15:b3 brd ff:ff:ff:ff:ff:ff
    inet 172.17.42.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
    inet6 fe80::42:79ff:fe6d:15b3/64 scope link
       valid_lft forever preferred_lft forever
```

Because the code which extended no_proxy only checked for /24 subnets,
this address 172.17.42.1 was not automatically included.

Reported by @devimc in https://github.com/kata-containers/runtime/issues/2262#issuecomment-575620216.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ endif
 # tests using that address try to go through the proxy.
 HTTP_PROXY=$(shell echo "$${HTTP_PROXY:-$${http_proxy}}")
 HTTPS_PROXY=$(shell echo "$${HTTPS_PROXY:-$${https_proxy}}")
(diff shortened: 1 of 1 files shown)


okartau:
This change adds docker addr to no_proxy, good. Note that it also picks up 127.0.0.1/8 and adds it, but I think there is no harm adding that to no_proxy.

Issue #503, opened 2019-12-20
handle volume ID hash conflicts


pohly:
https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/controllerserver-node.go#L172-L179 creates a new volume ID from the `CreateVolumeRequest.Name`.

While unlikely, it is possible that two different names hash to the same ID. We should either catch that and throw an error (just to be safe) or add a comment why ignoring this potential conflict is fine.

pohly:
Same code exists in master controller: https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/controllerserver-master.go#L171-L176

This should be moved into a common utility function.

pohly:
We also should add a comment explaining *why* this hashing is done instead of just using the `Name` string.

okartau:
> We should either catch that and throw an error (just to be safe) or add a comment why ignoring this potential conflict is fine.

What if we try to handle it in 3rd way: trying to continue without error?
We don't have requirement that VolumeID has to be hash of name, it is up to driver to generate some unique ID. And we dont have other code path generating hash where we must get same value. So, what about this: We generate hash and see did we get unique VolumeID. If not, we try again, for example by getting new hash from Name2+hash. (if we get collision again, it is already much-much smaller chance, but we can continue that loop until we get unique hash). Then we return new hash as VolumeID. Would such method work? The resulting VolumeID will be repeatable if the set of existing VolumeNames/VolumeIDs remains, but I think the main point is, we dont even need it to be repeatable as no other code path generates VolumeIDs.
There is however similar hash generation in node part of code, that may complicate things for this scenario.

pohly:
> The resulting VolumeID will be repeatable if the set of existing VolumeNames/VolumeIDs remains, but I think the main point is, we dont even need it to be repeatable as no other code path generates VolumeIDs.

The "if the set of existing VolumeNames/VolumeIDs remains" part is crucial here. We can't assume that, so with this proposal we may end up creating different volume IDs for the same volume name.

Is that a problem? Looking at the code I suspect that it wouldn't be a problem. Idempotency uses the volume name to find some already existing volume.

[...]

PR #523, opened 2020-01-23
Handle hash collision

okartau:
Address points listed in #503


okartau:
I added a commit changing sha-1 use to sha256, with resulting 64-char VolumeID. And that points immediately out failure in Direct mode. I somehow assumed that ndctl can use long names. But I was wrong, ndctl namespace name max size seems to be 64, including zero terminator. So we can not attempt creation with name of 64 chars.
We can switch to next-shorter sha224, generating 56-char hash. Then there will be space of 7 more chars where we can add starting characters from Volume Name, as planned.

pohly:
@okartau: is this still WIP?

okartau:
> @okartau: is this still WIP?

thanks for pinging, that PR has been left starving by me. I think I have addressed points raised and discussed in the issue, so please take a look, asking for review.

Issue #522, opened 2020-01-21
Front-page diagrams describe sector mode which has been removed from code


okartau:
We should modify the diagrams so that sector-mode namespaces are not shown.

PR #524, opened 2020-01-23
Update front-page diagrams to match current code

okartau:
Recently the sector-mode support got removed, but diagrams
still contained sector mode related information.

Resolves: #522


okartau:
@pohly @avalluri what do you think, should we keep the /usr/bin/ndctl (with arrow) on right side of those diagrams? I added this in early days, with idea to describe the command-line use of ndctl which is useful for debug and maintaining the pmem devices on a development host. But it may be misleading in the clean context, which the final doc should present.
So I was thinking, I could a) remove ndctl binary name and that big arrow, or b) add some text on diagram to explain why it is there. 
I like a) more, as use of `ndctl` is described in examples and DEVELOPMENT file anyway

avalluri:
> @pohly @avalluri what do you think, should we keep the /usr/bin/ndctl (with arrow) on right side of those diagrams?

I feel like it is better to remove `/usr/bin/ndctl` reference here in this diagram. As the diagram supposed to illustrate how our driver works in `direct` mode where we do not use `ndctl` binary. So better avoid mentioning this in the diagram.

okartau:
I share same opinion. I pushed new versions where ndctl binary name and arrow are moved. I also modified text near that sysctl use arrow to be different in LVM and direct mode, to reflect the different stages it gets used.

pohly:
> should we keep the /usr/bin/ndctl (with arrow) on right side of those diagrams

I agree that it makes sense to remove them. The simpler we can make these diagrams, the better.

Do we need `ndctl` in the https://github.com/intel/pmem-csi/blob/6784d96fa05ff1581116d47c6540d88da451817c/docs/images/devicemodes/pmem-csi-direct.png at all? it looks like pmem-csi calls the `ndctl` binary which then calls `libndctl` but that's not the case (anymore?).

Something that has always confused me: for LVM mode (https://github.com/intel/pmem-csi/blob/6784d96fa05ff1581116d47c6540d88da451817c/docs/images/devicemodes/pmem-csi-lvm.png), we show `libndctl` being used as part of "pmem-csi plugin", underneath the CSI API. It's just the comment on the arrow which explains what really happens. It would be better to also adapt the diagram, by using one box inside the bigger eclipse for `pmem-csi-driver` and other boxes for `pmem-ns-init` and `pmem-vgm`.

Either way, let's get rid of "pmem-csi plugin" in the diagrams. Either replace with "PMEM-CSI driver" (the term we use elsewhere now) or with the individual binary names.

okartau:
> Do we need `ndctl` in the https://github.com/intel/pmem-csi/blob/6784d96fa05ff1581116d47c6540d88da451817c/docs/images/devicemodes/pmem-csi-direct.png at all? it looks like pmem-csi calls the `ndctl` binary which then calls `libndctl` but that's not the case (anymore?).

If you mean `ndctl` between horiz.lines, then good point, we should remove it. It does not mean ndctl as binary name, I just used "ndctl mode" as alias to "direct mode", but after creation of those diagrams, we unified the device mode names so better not to mention "ndctl mode" any more.
Calling ndctl directly from pmem-csi was never the case, we have always used libndctl.

> It would be better to also adapt the diagram, by using one box inside the bigger eclipse for pmem-csi-driver and other boxes for pmem-ns-init and pmem-vgm

yes, I will add separate entities to show that init.stages are performed by InitContainers, leaving libndctl out of driver serving CSI API.

okartau:
pushed new versions, Initcontainers shown as separate entities, arrows, names and some texts changed to add clarity and to conform to latest naming convention.

okartau:
pushed with these changes. I also swapped the lines so that name of initContainer is at the top.

PR #525, opened 2020-01-24
Operator changes

avalluri:
The carried changes by this PR are:

- Existing changes in Operator branch with rebased against latest devel branch
- ReadMe doc for operator
- Made changes to operator-sdk generated directory structure to embed within the existing pmem-csi directory tree
- Made changes to Deployment API as discussed in #495
- Manage Deployment CRD from within the operator
- Split driver deployment into phases, updating the development status with the current deployment phase.
- Auto-generation(CSR creation & approval) of needed driver certificates
- Support for multiple deployments
- Support for updates to running/pending deployments

The current operator implementation supports the below features:

- Docker images for the operator via pmem-csi-driver images
- YAMLfor operator deployment(one needs to provide the right operator image, currently it uses intel/pmem-csi-driver:canary)
- Supports deploying PMEM-CSI driver by creating a Deployment object in pmem-csi.intel.com/v1alpha1 API group.
- Supports choosing driver namespace, name, image, device-mode and compute resources for both controller & node.
- Supports deploying multiple PMEM-CSI drivers(both on different nodes, and on same node(s))
- Auto-generating of TLS certificates needed for PMEM-CSI driver communication, hence no manual intervention.
- Support to make changes to running PMEM-CSI deployment: say could change compute resources, driver image etc,

**Known issues/limitations/TODOs**:

- Not tested/supported on pre v1.16.0 clusters.
- No E2E tests for operator
- Driver deployed in test cluster(make start) not using the operator to deploy driver.


pohly:
@avalluri: there have been test failures: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-525/3/execution/node/67/log/

Can you check those?

avalluri:
> @avalluri: there have been test failures: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-525/3/execution/node/67/log/
> 
> Can you check those?

Yes, I realized this(caused by missing changes while rebasing) and pushed the fix and waiting for CI results

pohly:
Speaking of rebasing: can I review each commit one-by-one or are there cases where mistakes in one commit get fixed by a later commit?

avalluri:
> Speaking of rebasing: can I review each commit one-by-one or are there cases where mistakes in one commit get fixed by a later commit?

I tried my best to squash such changes and keep the commits independent.

avalluri:
Pushed new commit that adds support for watching for deployments in cluster level than just operator namespace or preconfigured watch namespace via an environment variable.

pohly:
The actual E2E testing is now failing: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-525/5/console

avalluri:
> The actual E2E testing is now failing: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-525/5/console

Both of the failures were the result of mount failures, which are not related to this PR changes:

First failures:
```
[2020-01-27T14:53:20.266Z] • Failure [321.173 seconds]
[2020-01-27T14:53:20.266Z] E2E
[2020-01-27T14:53:20.266Z] /mnt/workspace/pmem-csi_PR-525/test/e2e/storage/csi_volumes.go:44
[2020-01-27T14:53:20.266Z]   [Driver: pmem-csi]
[2020-01-27T14:53:20.266Z]   /mnt/workspace/pmem-csi_PR-525/test/e2e/storage/csi_volumes.go:117
[2020-01-27T14:53:20.266Z]     [Testpattern: Dynamic PV (xfs)][Slow] volumes
[2020-01-27T14:53:20.266Z]     /mnt/workspace/pmem-csi_PR-525/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/base.go:92
[2020-01-27T14:53:20.266Z]       should allow exec of files on the volume [It]
[2020-01-27T14:53:20.266Z]       /mnt/workspace/pmem-csi_PR-525/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/volumes.go:187
[2020-01-27T14:53:20.266Z] 
[2020-01-27T14:53:20.266Z]       Unexpected error:
[2020-01-27T14:53:20.267Z]           <*errors.errorString | 0xc00032b7f0>: {
[2020-01-27T14:53:20.267Z]               s: "expected pod \"exec-volume-test-pmem-csi-dynamicpv-vwd8\" success: Gave up after waiting 5m0s for pod \"exec-volume-test-pmem-csi-dynamicpv-vwd8\" to be \"success or failure\"",
[2020-01-27T14:53:20.267Z]           }
[2020-01-27T14:53:20.267Z]           expected pod "exec-volume-test-pmem-csi-dynamicpv-vwd8" success: Gave up after waiting 5m0s for pod "exec-volume-test-pmem-csi-dynamicpv-vwd8" to be "success or failure"
[2020-01-27T14:53:20.267Z]       occurred
```
Failure casue:
```
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.808931       1 tracing.go:19] GRPC call: /csi.v1.Node/NodeStageVolume
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.808953       1 tracing.go:20] GRPC request: {"staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-56c76d6e-3f6d-49db-9943-6485344ab398/globalmount","volume_capability":{"AccessType":{"Mount":{"fs_type":"xfs"}},"access_mode":{"mode":1}},"volume_context":{"eraseafter":"false","storage.kubernetes.io/csiProvisionerIdentity":"1580136316002-8081-pmem-csi.intel.com"},"volume_id":"4d2ce6ed7cf8a6a3b509211834e6a556d9189926"}
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.813423       1 nodeserver.go:337] NodeStageVolume: VolumeID:4d2ce6ed7cf8a6a3b509211834e6a556d9189926 Staging target path:/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-56c76d6e-3f6d-49db-9943-6485344ab398/globalmount Requested fsType:xfs Requested mount options:[]
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.813442       1 pmem-exec.go:12] Executing: file -bsL /dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.853992       1 pmem-exec.go:15] Output: data
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.854077       1 pmem-exec.go:12] Executing: mkfs.xfs -b size=4096 -f /dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926
[2020-01-27T14:48:01.914Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.887212       1 pmem-exec.go:15] Output: meta-data=/dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926 isize=512    agcount=4, agsize=7168 blks
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver:          =                       sectsz=4096  attr=2, projid32bit=1
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver:          =                       crc=1        finobt=1, sparse=1, rmapbt=0
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver:          =                       reflink=1
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: data     =                       bsize=4096   blocks=28672, imaxpct=25
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver:          =                       sunit=0      swidth=0 blks
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: log      =internal log           bsize=4096   blocks=1221, version=2
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver:          =                       sectsz=4096  sunit=1 blks, lazy-count=1
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: realtime =none                   extsz=4096   blocks=0, rtextents=0
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.887399       1 nodeserver.go:551] mount args: [[-c -o dax /dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926 /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-56c76d6e-3f6d-49db-9943-6485344ab398/globalmount]]
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:01.887425       1 pmem-exec.go:12] Executing: mount -c -o dax /dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926 /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-56c76d6e-3f6d-49db-9943-6485344ab398/globalmount
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: I0127 14:48:02.092241       1 pmem-exec.go:15] Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-56c76d6e-3f6d-49db-9943-6485344ab398/globalmount: wrong fs type, bad option, bad superblock on /dev/ndbus0region0fsdax/4d2ce6ed7cf8a6a3b509211834e6a556d9189926, missing codepage or helper program, or other error.
[2020-01-27T14:48:02.268Z] pmem-csi-node-n4vqs/pmem-driver: E0127 14:48:02.093612       1 tracing.go:23] GRPC error: rpc error: code = Internal desc = mount filesystem failed: exit status 32
```

Second failure:
```
[2020-01-27T15:22:22.912Z] 
[2020-01-27T15:22:22.912Z] • Failure [922.330 seconds]
[2020-01-27T15:22:22.912Z] E2E
[2020-01-27T15:22:22.912Z] /mnt/workspace/pmem-csi_PR-525/test/e2e/storage/csi_volumes.go:44
[2020-01-27T15:22:22.912Z]   [Driver: pmem-csi]
[2020-01-27T15:22:22.912Z]   /mnt/workspace/pmem-csi_PR-525/test/e2e/storage/csi_volumes.go:117
[2020-01-27T15:22:22.912Z]     [Testpattern: Dynamic PV (xfs)][Slow] volumes
[2020-01-27T15:22:22.912Z]     /mnt/workspace/pmem-csi_PR-525/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/base.go:92
[2020-01-27T15:22:22.912Z]       should store data [It]
[2020-01-27T15:22:22.912Z]       /mnt/workspace/pmem-csi_PR-525/vendor/k8s.io/kubernetes/test/e2e/storage/testsuites/volumes.go:146
[2020-01-27T15:22:22.912Z] 
[2020-01-27T15:22:22.912Z]       Jan 27 15:17:02.128: Failed to create injector pod: timed out waiting for the condition
[2020-01-27T15:22:22.912Z] 
[2020-01-27T15:22:22.912Z]       /mnt/workspace/pmem-csi_PR-525/vendor/k8s.io/kubernetes/test/e2e/framework/volume/fixtures.go:565
```

Failure cause:

```
[2020-01-27T15:07:37.323Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.234938       1 tracing.go:19] GRPC call: /csi.v1.Node/NodeStageVolume
[2020-01-27T15:07:37.323Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.235017       1 tracing.go:20] GRPC request: {"staging_target_path":"/var/lib/k
ubelet/plugins/kubernetes.io/csi/pv/pvc-ae97a14d-8fa4-4d9a-b427-8c23804c4640/globalmount","volume_capability":{"AccessType":{"Mount":{"fs_type":"xfs"}},"
access_mode":{"mode":1}},"volume_context":{"eraseafter":"false","storage.kubernetes.io/csiProvisionerIdentity":"1580137534222-8081-pmem-csi.intel.com"},"
volume_id":"264cc8917b96fe5e5648da392e42a68f78a535be"}
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.242654       1 nodeserver.go:337] NodeStageVolume: VolumeID:264cc8917b96fe5e56
48da392e42a68f78a535be Staging target path:/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-ae97a14d-8fa4-4d9a-b427-8c23804c4640/globalmount Requested f
sType:xfs Requested mount options:[]
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.242717       1 pmem-exec.go:12] Executing: file -bsL /dev/ndbus0region0fsdax/2
64cc8917b96fe5e5648da392e42a68f78a535be
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.257120       1 pmem-exec.go:15] Output: SGI XFS filesystem data (blksz 4096, i
nosz 512, v2 dirs)
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.257166       1 pmem-exec.go:12] Executing: blkid -c /dev/null -o full /dev/ndb
us0region0fsdax/264cc8917b96fe5e5648da392e42a68f78a535be
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.264870       1 pmem-exec.go:15] Output: /dev/ndbus0region0fsdax/264cc8917b96fe
5e5648da392e42a68f78a535be: UUID="5d1fd557-e988-4385-9f96-291e22fcb411" BLOCK_SIZE="4096" TYPE="xfs"
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.264923       1 nodeserver.go:358] Skip mkfs as xfs file system already exists 
on /dev/ndbus0region0fsdax/264cc8917b96fe5e5648da392e42a68f78a535be
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.275963       1 nodeserver.go:551] mount args: [[-c -o dax /dev/ndbus0region0fs
dax/264cc8917b96fe5e5648da392e42a68f78a535be /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-ae97a14d-8fa4-4d9a-b427-8c23804c4640/globalmount]]
[2020-01-27T15:07:37.324Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.276143       1 pmem-exec.go:12] Executing: mount -c -o dax /dev/ndbus0region0f
sdax/264cc8917b96fe5e5648da392e42a68f78a535be /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-ae97a14d-8fa4-4d9a-b427-8c23804c4640/globalmount
[2020-01-27T15:07:37.680Z] pmem-csi-node-7l6vs/pmem-driver: I0127 15:07:37.299949       1 pmem-exec.go:15] Output: mount: /var/lib/kubelet/plugins/kubern
etes.io/csi/pv/pvc-ae97a14d-8fa4-4d9a-b427-8c23804c4640/globalmount: wrong fs type, bad option, bad superblock on /dev/ndbus0region0fsdax/264cc8917b96fe5
e5648da392e42a68f78a535be, missing codepage or helper program, or other error.
[2020-01-27T15:07:37.680Z] pmem-csi-node-7l6vs/pmem-driver: E0127 15:07:37.300177       1 tracing.go:23] GRPC error: rpc error: code = Internal desc = mo
unt filesystem failed: exit status 32
```

pohly:
> Both of the failures were the result of mount failures, which are not related to this PR changes:

So we have a problem that we need to fix before we can add more new code. Can you look into it?

pohly:
Also, why did this not occur earlier? Recent regular testing of "devel" had problems, but the E2E testing always worked: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi-periodic/

For this PR, it hasn't passed at all yet (currently on the 5th retry): https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-525/

avalluri:
> > Both of the failures were the result of mount failures, which are not related to this PR changes:
> 
> So we have a problem that we need to fix before we can add more new code. Can you look into it?

I am not sure if these mount issues are fixed in devel branch, I could see below commit in devel explicit to "xfs" filesystem:
https://github.com/intel/pmem-csi/commit/77374b875527d463d4f875116954b49ad4403596

pohly:
Discussed this with @avalluri: in contrast to what I was expecting based on earlier comments, this PR is *not* proposing to merge the operator into "devel". That combined with "operator" (the real target branch) being out-of-date is causing the test failures.

The intention of the "operator" branch was to do incremental, shared development by merging into it regularly. This hasn't happened, but as @avalluri is also not sure yet either whether we really want this in "devel", let's do something else instead:
- reset "operator" to "devel"
- rebase this PR and the commits previously in "operator" on top of "devel"
- re-review everything in this PR

avalluri:
> * reset "operator" to "devel"

@pohly For rebasing the existing Operator branch on to devel needs a force push, as it has conflicting changes. so we have two options:
 1) First, rebase and force push, and then resubmit this PR.
 2) Keep the current operator implementation in "devel"

I leaned towards (2) and created a PR #527 to merge into devel. I mentioned the current features and known issues/todos, hope that helps you to decide if the operator is ready for merging into devel.

pohly:
My proposal was to make `operator` equal to `devel`, i.e. `git push origin +refs/heads/devel:refs/heads/`.

> I mentioned the current features and known issues/todos, hope that helps you to decide if the operator is ready for merging into devel.

Given some of the known todos, in particular "Not tested/supported on pre v1.16.0 clusters." and "No E2E tests for operator", let's keep it out of devel for now.

If you agree, then force-push `devel` and update your `operator-direcotry-tree` branch such that it has the rebased code and we'll review in this PR.

pohly:
@avalluri: you still need to force-push `operator-direcotry-tree`.

avalluri:
> @avalluri: you still need to force-push `operator-direcotry-tree`.

rearranging and squashing of few commits took more time :) Now its there.

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+| kind | string | `Deployment`|
+| metadata | [ObjectMeta](https://git.k8s.io/community/contributors/devel/api-conventions.md#spec-and-status) | Standard objects metadata |
+| spec | [DeploymentSpec](#deployment-spec) | Specification of the desired behavior of the deployment |
+
+#### DeploymentSpec
+
+|Field | type | Description | Default Value |
+|---|---|---|---|
+| driverName | string | Unique CSI driver name to use | `pmem-csi.intel.com` |
+| image | string | PMEM-CSI docker image name used for the deployment | `intel/pmem-csi-driver:canary` |
```
pohly: Instead of specifying "canary" here, can we say "the same version as the operator" and ensure that this is really the case for releases?

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+| provisionerImage | string | [CSI provisioner](https://github.com/kubernetes-csi/external-provisioner) docker image name | `quay.io/k8scsi/csi-provisioner:v1.2.1` |
```
pohly: Same here: the default may change between releases. Instead of copying that information, just point the user to where they can find the latest information.

Speaking of that, how does the operator know which version to use? It may depend on the Kubernetes version. Does it keep a builtin copy of the "deploy" files and then modify those? That's currently our canonical source of truth for how to deploy the driver for the different supported Kubernetes versions.

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+a `phase` field. The phase of a Deployment is high-level summary of where the
+Deployment is in it's lifecycle.
+
+The possible `phase` values and their meaning are as below:
+
+| value | meaning |
+|---|---|---|
+| Pending | The `Deployment` has been accepted by the system, but waiting for one or more of it's pre-requisites to be completed. Such as waiting for required [TLS certificates](../README.md#security) to be generated |
+| Initializing | The required sub-resources of the `Deployment` are being initialized |
+| Running | All the required sub-resources of the `Deployment` are created |
```
pohly: Does that mean that the driver is really operational, i.e. all nodes are up and ready? Probably not. This is fine for now, but should be called out here.

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+| Failed | For some reason the state of the `Deployment` failed and cannot be progressed |
```
pohly: If it fails, will there be some kind of explanation? How?

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+### TLS Certificate approval
+
+All PMEM-CSI driver communication is protected by mutual TLS. So the driver
+deployment needs to generate those certificates. The operator uses Certificate
+Authority(CA) provided by the Kubernetes cluster to get those certificates. On a
```
pohly: "Authority (CA)"

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+new deployment, the operator creates Kubernetes `CertificateSigningRequest`
+objects and keeps the deployment in `pending` state till those CSRs get approved
+by an authorized user so that Kubernetes CA could generate the certificates.
+Once the CSRs are approved and certificates are ready, then the deployment
+process gets progressed.
```
pohly: "gets progressed" -> "proceeds"

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+| Failed | For some reason the state of the `Deployment` failed and cannot be progressed |
+
+### TLS Certificate approval
+
+All PMEM-CSI driver communication is protected by mutual TLS. So the driver
+deployment needs to generate those certificates. The operator uses Certificate
+Authority(CA) provided by the Kubernetes cluster to get those certificates. On a
+new deployment, the operator creates Kubernetes `CertificateSigningRequest`
+objects and keeps the deployment in `pending` state till those CSRs get approved
+by an authorized user so that Kubernetes CA could generate the certificates.
```
pohly: "so that Kubernetes CA could generate" -> "and the Kubernetes CA generates"

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+Once the CSRs are approved and certificates are ready, then the deployment
+process gets progressed.
+
+If the operator is configured with appropriate RBAC rules with enough
```
pohly: Isn't that always going to be the case when deploying the driver as intended?

As it stands now, this sentence begs the question how an admin can configure this, which isn't explained.

operator/README.md
pohly: What happens if nodes are added later on that match the label for the PMEM-CSI daemon set?

It's not enough to create the CSRs once. This has to be a control loop, like everything else.

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+permissions to approve CSRs, the operator itself runs a dedicated certificate
+approval control-loop, that detects and approves the certificate signing
+requests generated by a PMEM-CSI Deployment. For validating if a CSR is
+generated by the operator it uses the `Username` field of
+`CertificateSigningRequestSpec`.
```
pohly: And that `Username` cannot be faked by someone else?

pkg/pmem-csi-operator/controller/add_deployment.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller/certificates"
+	"github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller/deployment"
+)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	AddToManagerFuncs = append(AddToManagerFuncs, certificates.Add)
+	AddToManagerFuncs = append(AddToManagerFuncs, deployment.Add)
```
pohly: Nit: this can be a single `append()`.

operator/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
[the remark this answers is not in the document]
avalluri: This is not the intention. I should have dropped this file, just kept for reference. But the right commit related to operator image is: https://github.com/intel/pmem-csi/pull/525/commits/5c4401eace8c58780a2f9aed66cbef5011b3f491 , where the operator binary is kept part of actual driver image.

operator/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+ENV OPERATOR=/usr/local/bin/pmem-csi-operator \
+    USER_UID=1001 \
+    USER_NAME=pmem-csi-operator
+
+# install operator binary
+COPY build/_output/bin/pmem-csi-operator ${OPERATOR}
```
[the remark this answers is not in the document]
avalluri: > If do end up using a separate image

This is the intention. This auto-generated Dockerfile kept for reference.

operator/README.md
avalluri: Sounds good, though a bit tricky to implement this.

operator/README.md
avalluri: > Speaking of that, how does the operator know which version to use? It may depend on the Kubernetes version

Though not implemented to support < 1.16 Kubernetes versions part of this PR, It supposed to [determine the running cluster version](https://github.com/intel/pmem-csi/pull/525/commits/7187383845b38f7495d47588b328a0bcb76c5fde) and prepares the appropriate driver deployment manifests.

operator/build/Dockerfile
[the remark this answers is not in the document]
pohly: Please remove it. Having it around is just causing confusion. The git history is for archiving things, not the current tree.

operator/README.md
avalluri: Whether the failure is supposed to visible in [status conditions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#podcondition-v1-core) or in a simple string field called `reason` is not yet decided and not implemented in this PR.

operator/README.md
avalluri: The certificates are not generated per node. We create sets of certificates  `peme-registry` and `pmem-node-controller`. `pmem-node-controller` certificate is shared by all the nodes. So we do not generate a  new CSR upon a new pmem node.

operator/README.md
avalluri: > And that `Username` cannot be faked by someone else?

Oops, this was talking about the initial idea, and no more valid now. That means we are not using `Username` check anymore, Just checking if the `CSR` is owned by a `Deployment`.

I will fix the documentation.

operator/README.md
pohly: Please clarify that in the docs.

operator/README.md
pohly: So that's the current status and it's kind of okayish for manual deployment. But do you agree that the operator must handle also dynamically added nodes?

operator/README.md
pohly: Same question applies: can some rogue deployment trick PMEM-CSI into approving a CSR that it shouldn't approve?

operator/README.md
avalluri: Yes, a valid question.

My idea was that we should make this certificate approval configurable so that admins can choose their own approval mechanisms. But how to is a bit unclear myself. I couldn't decide between 

    a) providing a command-line argument 
    b) missing RBAC(CSR and CSR Approval) rules. 

In the first case, we could provide an option something like `--run-csr-controller` where the operator manifest _must_ have needed RBAC rules otherwise we fail to start the operator. 

In the second case, the operator tries to start csr controller, and if it fails with lack of permissions then just ignore the error continue with no csr controller. In this case, the admin could drop CSR related rules from the provided default operator manifest.

So I just assumed the second option, though not implemented :)

operator/README.md
avalluri: At least till now, I couldn't figure out something to do special for a newly added node. That should be handled by `daemonset` controller. Unless we think of implementing `auto labeling` new nodes feature.

operator/README.md
avalluri: I doubt if some can create CSR object and assign the ownership to a running deployment object. Though I am not sure.

operator/README.md
pohly: > In the second case, the operator tries to start csr controller, and if it fails with lack of permissions then just ignore the error continue with no csr controller.

Ignore silently or log as an error? If it's flagged as an error, then there handling the same in both cases.

If it's ignored silently, then accidentally missing RBAC rules would be hard to debug. Better always log this as an error.

A `--run-csr-controller` option might make sense if we can figure out a use cases for it. I wouldn't bother initially and always run it.

operator/README.md
pohly: Can you explain to me what you mean with ownership? I probably lack some background here.

operator/README.md
pohly: I'm not talking about doing something automatic *on* a node. Configuring PMEM and labeling the node is still something that the admin needs to do.

I'm talking about the situation where we need to do something *for* a node:
- PMEM-CSI operator is running.
- Admin creates a deployment.
- Deployment is running.
- Admin enables PMEM on a node and sets the label that is used by the deployment.

As it's designed right now, PMEM-CSI node driver will start to run on that node, but cannot communicate with the registry because it lacks the necessary certificates.

What we need is a control loop that reacts to label changes on nodes and then re-recreates the content of the `pmem-node-controller` secret.

operator/README.md
avalluri: The [deployment controller adds](https://github.com/intel/pmem-csi/pull/525/commits/0fecdcdc055664608ab9f7fa16ae75a57c74d4e0#diff-ab6d9094a1503b34d88bc9ba1066c814R161-R172)  a [ower reference](https://kubernetes.github.io/cluster-registry/reference/build/index.html#ownerreference-v1) to all the objects it creates for a deployment. Certificate controller checks if a CSR is owned by a Deployment before approving it.

operator/README.md
avalluri: > As it's designed right now, PMEM-CSI node driver will start to run on that node, but cannot communicate with the registry because it lacks the necessary certificates.

We are not generating certificates per node(unlike the current [certificate generation scirpt](https://github.com/intel/pmem-csi/blob/devel/test/setup-ca-kubernetes.sh)), instead, all node divers share the same secret/certificate. So we might not need any extra step for setting up certificates for a newly added node, as the secret/certificate is already there and the Pod template of the daemonset has the reference to the existing secret.

pohly: Oh, so it's a change in the security design. I wasn't aware of that. So it's a bit less secure than before (a rogue node can now impersonate any other node). Might still be good enough.

operator/README.md
pohly: Anyone can set arbitrary owner references in the objects that they are creating. How would the apiserver decide who's allowed to do that?

So this check doesn't provide any additional security. What may work is to restrict signing to CSRs in the namespaces managed by the operator. Then we can argue that an adversary first would have to gain the privilege to create CSRs in that namespace, something that a non-admin users shouldn't have. Once an adversary has that privilege, he probably can also meddle with the other entities.

operator/README.md
pohly: Hmm, I might have misremembered here. I thought we had different certificates per node, but it seems that we don't. Perhaps that's what I had in OIM and PMEM-CSI always was a bit simpler - sorry for the confusion. :man_shrugging:

pohly: Let's take a step back. Why are we creating different certificates for each node?

I believe the intention was to ensure that only a node controller which has the right certificate can register for the node. But that doesn't seem to be implemented (there's no certificate check in https://github.com/intel/pmem-csi/blob/599a94e3fca15384c385f674355cd177f49602f4/pkg/registryserver/registryserver.go#L93-L100) and therefore it now works for the operator where all nodes present the same certificate.

@avalluri: does that make sense? Why was strict certificate checking not implemented?

In OIM, I had implemented such additional checks, for example: https://github.com/intel/oim/blob/de83caab34dd9df2565ac2dcf7e1c164e79a3f2b/pkg/oim-registry/registry.go#L174-L184

avalluri: > I believe the intention was to ensure that only a node controller which has the right certificate can register for the node.

That was not the intention of the implementation though the script generating per node certificates.

> @avalluri: Why was strict certificate checking not implemented?

I felt like it's ok to not tiding the certificate to a node, instead the certificate _CommonName_ should match. Do you think we should have a more strict check?

pohly: > That was not the intention of the implementation though the script generating per node certificates.

Then the script should have also generated only one certificate for all nodes. The way it is now is just confusing.

> I felt like it's ok to not tiding the certificate to a node, instead the certificate CommonName should match. Do you think we should have a more strict check?

I'm not sure. Probably we don't need it: if one node gets compromised, it can already disrupt the operation of PMEM-CSI also without impersonating other nodes.

So let's keep it simple, but that means that `setup-ca-kubernetes.sh` should also get simplified to match the actual implementation.

operator/README.md
avalluri: I changed the CSR approver, now it considers that:
  - If the CSR is owned by a valid Deployment, and that deployment is in _Pending_ state
  - the CSR subject is one of the valid common names expected by PMEM-CSI driver.

pkg/pmem-csi-operator/controller/add_deployment.go
avalluri: > Nit: this can be a single `append()`.

With the [new change](https://github.com/intel/pmem-csi/pull/525/commits/79e3841e8b68b0c8194257ed9d3866f6593327a2#diff-a6eb5181b8c90da9284577844583a624), this Nit is not valid anymore ;-)

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+    - [Deploy PMEM-CSI operator](#deploy-pmem-csi-operator)
+    - [Deployment CRD API](#deployment-crd-api)
+  - [Example Deployment](#example-deployment)
+
+<!-- /TOC -->
+# PMEM-CSI-Operator
+
+## About
+
+Kubernetes operator for [PMEM-CSI](https://github.com/intel/pmem-csi) driver deployment. Using this operator one can easily deploy and manage PMEM-CSI driver in a Kubernetes cluster.
```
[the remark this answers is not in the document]
avalluri: Done.

operator/README.md
```diff
@@ -0,0 +1,173 @@
...
+- [PMEM-CSI Operator](#pmem-csi-for-kubernetes)
+  - [About](#about)
+  - [Usage](#usage)
+    - [Pre-requisites](#pre-requisites)
+    - [Deploy PMEM-CSI operator](#deploy-pmem-csi-operator)
+    - [Deployment CRD API](#deployment-crd-api)
+  - [Example Deployment](#example-deployment)
+
+<!-- /TOC -->
+# PMEM-CSI-Operator
```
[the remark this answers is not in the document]
avalluri: Done

operator/build/Dockerfile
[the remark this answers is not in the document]
avalluri: Removed the operator-sdk generated docker file.

operator/README.md
```diff
@@ -0,0 +1,179 @@
...
+the operator creates required `CertificateSigningRequest` Kubernetes objects
+and keeps the driver deployment in `pending` state till those CSRs get approved by 
+an authorized user or component. Once the CSRs are approved and certificates 
+are ready, then the deployment process proceeds.
+
+By default the operator runs a dedicated controller for auto-approving these CSR
+requests. This controller uses Kubernetes Authority (CA) provided by the Kubernetes
+cluster to get those certificates. The controller watches for new CSRs and approves 
+them by ensuring that those requests are generated by a valid deployment and the 
+request subject matches to a valid _CommonName(CN)_ expected by PMEM-CSI driver.
```
pohly: This explains *what* is done, but not *why*. Why are we using mutual TLS (first sentence in this section) and why is that goal still reached when using this approach explained here?

The way it's explained here it makes it sound like our security model is intact because the auto-approver is careful about what it signs. But after thinking a bit about this, I believe the opposite is true: because our approver is careful to not sign unrelated CSRs, the security of *other* components in the system is preserved when enabling it.

We can't protect ourselves against a malicious or careless approver. If they approve a CSR for "our" names although it wasn't the operator which created the CSR, then valid certificates will be created that allow other entities to communicate with PMEM-CSI. But that's okay: the cluster admin has to take care that no-one can approve CSRs without due diligence.

Does that make sense? If yes, can you update the text here?

avalluri: > This explains _what_ is done, but not _why_. Why are we using mutual TLS (first sentence in this section) and why is that goal still reached when using this approach explained here?

Do you mean this section should also explain 'why are we using mutual TLS' and/or 'why we choose this approving model'?

avalluri: @pohly, Does this sounds any better:

```
By default, the operator runs a dedicated controller for auto-approving these
CSR requests without any manual intervention. This controller uses the
Certificate Authority (CA) provided by the Kubernetes cluster to get those
certificates. The controller watches for new CSRs and approves them by ensuring
that those requests are generated by a valid deployment and the request subject
matches a valid _CommonName(CN)_ expected by PMEM-CSI driver. This way we ensure
that the controller is careful enough in not approving any CSRs in the system
other than the ones requested by PMEM-CSI deployment(s).
```

pohly: This proposal looks good. Let me add to it a bit.

> Do you mean this section should also explain 'why are we using mutual TLS'

As this section mentions it, yes. Something like:

> All [PMEM-CSI driver communication is protected by mutual TLS](../README.md#security) to ensure that malicious or compromised entities in the cluster cannot interfere with its operation.
> ...
> This way we ensure
that the controller is careful enough in not approving any CSRs in the system
other than the ones requested by PMEM-CSI deployment(s). 
> The assumption is that other controllers and/or the cluster admins are equally careful to not sign CSRs for the special names used by PMEM-CSI. Under that assumption, only PMEM-CSI is going to have valid certificates for those names and thus its internal communication is protected.

avalluri: Ok, Thanks. I pushed these changes.

pohly: I'm having second thoughts about this:
> The way it's explained here it makes it sound like our security model is intact because the auto-approver is careful about what it signs. But after thinking a bit about this, I believe the opposite is true: because our approver is careful to not sign unrelated CSRs, the security of other components in the system is preserved when enabling it.

This is not the full truth. In addition to protecting others, we *also* need to protect ourselves from an attacker which makes a CSR look like it was created by us.

Does the current solution achieve that reliably? That hinges on how it determines what "a valid deployment" is. Can you elaborate on that?

avalluri: > Does the current solution achieve that reliably? That hinges on how it determines what "a valid deployment" is. Can you elaborate on that?

By checking the if the owner-reference of the CSR is a ' `Deployment('pmem-csi.intel.com/v1alph)` and if that deployment is really in `pending` state that means waiting for csr approval.

pohly: That still doesn't answer why an attacker can't create such a CSR.

There are multiple aspects to this:
- If the attacker is able to create such a deployment, then he's by definition allowed to set up a deployment and thus has access to the resulting secrets as far as PMEM-CSI is concerned. Cluster admins thus must control who gets this permission.
- If the attacks is unable to create such a deployment, he might still try to create a CSR while some other deployment is pending. This is a race, but we shouldn't rely on timing for security. What saves us here is that ownership is only possible for CSRs in the same namespace. Therefore an attacker must not get permission to create CSRs in the namespace of a deployment.

We need to document this as guidance for admins and explain in the discussion of "why this is secure" that we are making these assumptions.

pohly: > and if that deployment is really in pending state that means waiting for csr approval.

That isn't relevant, is it? It doesn't add any additional defense against an attacker.

avalluri: > * What saves us here is that ownership is only possible for CSRs in the same namespace. Therefore an attacker must not get permission to create CSRs in the namespace of a deployment.

CSRs are cluster level resources, not namespace scoped :-(

pohly: > CSRs are cluster level resources, not namespace scoped :-(

And the PMEM-CSI deployment then also is not namespaced? I thought it was because it's not mentioned that it isn't and the example output for `kubectl describe deployment.pmem-csi.intel.com pmem-deployment` shows a `namespace: default`. Or is that always shows, regardless whether the CSR is namespaced?

But apparently they are not, because otherwise you shouldn't be able to set the deployment as owner of the non-namespaced CSR. https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents says "2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners."

Let's take a step back. Why are we using CSRs at all? I can only think of one reason, which is "then someone might theoretically write a custom approver". We have no idea whether that is even possible in a secure way. What about just dropping this approach entirely for now until we actually have someone asking for it, and instead do CA creation and certificate signing in the operator itself?

What was the rationale for using signing certificates? They were introduced in https://github.com/intel/pmem-csi/commit/5b6432c6da44a339de3e58f14760b1f60d067564#diff-ea37666df77a82a46982984778217075 but that commit message only says what is in the commit, not why it is done this way.

avalluri: > And the PMEM-CSI deployment then also is not namespaced? I thought it was because it's not mentioned that it isn't and the example output for `kubectl describe deployment.pmem-csi.intel.com pmem-deployment` shows a `namespace: default`. Or is that always shows, regardless whether the CSR is namespaced?
>

No,  PMEM-CSI deployments are namespace scoped. [This](https://github.com/intel/pmem-csi/pull/525/commits/4f21a630624dbc2053b8effb0d356d7ef8031591) change adds support for deploying in a different namespace other than the operator's namespace.

> But apparently they are not, because otherwise you shouldn't be able to set the deployment as owner of the non-namespaced CSR. https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents says "2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners."

Hmm, This is bit unclear to me, where this restriction is imposed. But, somehow it worked for me with no issues. I was setting owner reference of CSR(in fact other cluster-scoped objects created by deployment - ClusterRole, ClusterRoleBinding) with deployment information. Initially, i was using for cleaning-up of these objects whenever deployment gets deleted.

avalluri: > Let's take a step back. Why are we using CSRs at all?

The idea behind using CSRs are for utilizing the Kubernetes's certificates subsystem and CA.
Kubernetes service account CA certificate is available to all Pods so that we need to create our own CA and deploy CA certificate to Pod.
 
 > I can only think of one reason, which is "then someone might theoretically write a custom approver". 

This is the reason for dividing CSR creation and approval as separate components.

>  We have no idea whether that is even possible in a secure way. 

If this is the case, then we can avoid the certificate approver controller and the operator itself creates an approved CSR, instead of waiting for someone to approve?.

pohly: > Hmm, This is bit unclear to me, where this restriction is imposed.

You can only set a `Name` for the owner, not `Namespace`: https://godoc.org/github.com/ericchiang/k8s/apis/meta/v1#OwnerReference

You can set a UUID. If that works across namespaces, then the Kubernetes documentation should be updated to reflect that.

> But, somehow it worked for me with no issues. I was setting owner reference of CSR(in fact other cluster-scoped objects created by deployment - ClusterRole, ClusterRoleBinding) with deployment information.

With UUID or Name? Probably the former, otherwise you would have noticed the lack of the `Namespace` field. Have you tested that this works in a non-default namespace?

> Kubernetes service account CA certificate is available to all Pods so that we need to create our own CA and deploy CA certificate to Pod.

Which we can do. We can simply put it into the same secret as the client certificate.

> If this is the case, then we can avoid the certificate approver controller and the operator itself creates an approved CSR, instead of waiting for someone to approve?.

Correct.

pohly: > With UUID or Name? Probably the former, otherwise you would have noticed the lack of the Namespace field. Have you tested that this works in a non-default namespace?

In case it wasn't obvious: with "that this works" I meant "garbage collection deletes orphaned objects", not just "apiserver allows to set ownership".

avalluri: > In case it wasn't obvious: with "that this works" I meant "garbage collection deletes orphaned objects", not just "apiserver allows to set ownership".

Yes, delete deploy resulted in the cleaning of all its sub-resources owned including cluster-level objects.

avalluri: > Have you tested that this works in a non-default namespace?

Yes!

pohly: It worked only by accident: https://github.com/kubernetes/kubernetes/issues/65200

pohly: Do we want the PMEM-CSI deployment to be namespaced? It's a privileged object, only admins should be allowed to create it. Given that it creates some non-namespaced objects, I believe it makes more sense to make it non-namespaced.

avalluri: > It worked only by accident: [kubernetes/kubernetes#65200](https://github.com/kubernetes/kubernetes/issues/65200)

:-)

avalluri: > Do we want the PMEM-CSI deployment to be namespaced? It's a privileged object, only admins should be allowed to create it. Given that it creates some non-namespaced objects, I believe it makes more sense to make it non-namespaced.

Ok. I have no strict opinion on this. In fact, those cluster-level objects need not to betide to a deployment. We could pre-create them and share them with all deployment objects.

If you believe it should be cluster scoped, I could implement those changes.

pohly: > Those cluster-level objects need not to betide to a deployment. We could pre-create them and share them with all deployment objects.

Do we really have objects which can be reused by different deployments? Which ones? If we have them and can't create them per deployment, then that sounds like a problem for cleanup (would need to track number of deployments that still need them).

> If you believe it should be cluster scoped, I could implement those changes.

Ownership relationships are useful and we can't do them when it's namespaced, so yes, I think this should be changed.

pohly: > >   It worked only by accident: kubernetes/kubernetes#65200
>
> :-)

Lesson learned: don't use experiments to figure out how an API works, follow the documentation...

avalluri:
@pohly Updated the PR with:
 - generating self-signed CA certificates for deployment
 - made deployment a cluster-scoped resource.

I feel like it's ready for further review.

Jenkinsfile
```diff
@@ -525,6 +525,10 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                                          TEST_E2E_SKIP=\$(if [ \"${env.CHANGE_ID}\" ] && [ \"${env.CHANGE_ID}\" != null ]; then echo \\\\[Slow\\\\]; fi) \
                            ' \
            "
+    } catch (exc) {
+        echo "Handling exception, get pod state and kubelet logs:"
+        sh "_work/${env.CLUSTER}/ssh.0 kubectl get pods --all-namespaces -o wide"
+        sh "for cmd in `ls _work/${env.CLUSTER}/ssh.*`; do \$cmd sudo journalctl -u kubelet; done"
```
pohly: @avalluri this doesn't belong into this PR.

Can you update the "operator" target branch to "origin/devel" and then rebase your PR?

operator/README.md
```diff
@@ -0,0 +1,175 @@
...
+A running Kubernetes cluster with one or more nodes installed with PMEM hardware(NVDIMM). The persistent memory on these nodes must be [pre-provisioned](../README.md#persistent-memory-pre-provisioning).
+
+### Deploy PMEM-CSI operator
+
+<!-- The assumptions:
+  a) pmem-csi-operator binary is part of released intel/pmem-csi-driver image and no additional steps required to build the images
+-->
+
+```sh
+$ kubectl create -f https://raw.githubusercontent.com/intel/pmem-csi/operator/operator/deploy/operator.yaml
```
pohly: Better use the URL before the redirect, then it is simpler to get the non-raw version by stripping out "/raw/":

```
https://github.com/intel/pmem-csi/raw/operator/operator/deploy/operator.yaml
```

Please also add a TODO about replacing the first "operator" with "devel" after merging into "devel" and a second TODO about patching this URL in release branches.

operator/README.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+### Deployment CRD API
+
+Current PMEM-CSI Deployment object supports below API:
+
+#### Deployment
+
```
pohly: Document here that `Deployment` is not namespaced.

operator/README.md
```diff
@@ -0,0 +1,175 @@
...
+| registryCert | string | Encoded tls certificate signed by a certificate authority used for driver's controller registry server | generated by operator self-signed CA |
+| nodeControllerCert | string | Encoded tls certificate signed by a certificate authority used for driver's node controllers | generated by operator self-signed CA |
+| registryKey | string | Encoded RSA private key used for signing by `registryCert` | generated by the operator |
+| nodeControllerKey | string | Encoded RSA private key used for signing by `nodeControllerCert` | generated by the operator |
+| caCert | string | Certificate of the CA by which the `registryCert` and `controllerCert` are signed | self-signed certificate generated by the operator |
+| nodeSelector | string map | [Labels to use for selecting Nodes](../README.md#run-pmem-csi-on-kubernetes) on which PMEM-CSI driver should run. | `{ "storage": "pmem" }`|
+
+<sup>1</sup> Image versions are depend on the Kubernetes cluster version.
```
pohly: s/are depend/depend/

It's worth explaining here that the operator figures that our itself for the default values whereas users have to handle this themselves when overriding the values.

operator/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+The possible `phase` values and their meaning are as below:
+
+| value | meaning |
+|---|---|---|
+| Initializing | The required sub-resources of the `Deployment` are being initialized |
+| Running | All the required sub-resources of the `Deployment` are created. This does not mean that the deployed driver is fully ready to use, there might still possibility of failures. Reasons for such failures can be determined by looking into individual pod status |
+| Failed | For some reason the state of the `Deployment` failed and cannot be progressed<sup>2</sup> |
+
+<sup>2</sup> Failure reason is supposed to be carried by one of additional `DeploymentStatus` filed, but not implemented yet.
```
pohly: s/filed/field/

operator/README.md
```diff
@@ -0,0 +1,175 @@
...
+### Driver certificates
+
+All [PMEM-CSI driver communication is protected by mutual
+TLS](../README.md#security) to ensure that malicious or compromised entities in
+the cluster cannot interfere with its operation. So the driver deployment needs
+to be provided with those certificates and corresponding private keys using appropriate fields in deployment specification. These encoded certificates and private keys are made available to driver pods via Kubernetes [secrets](https://kubernetes.io/docs/concepts/configuration/secret/) by the operator.
+
+If the private keys and/or certificates data is missing in a deployment, the operator generates them using a self-signed CA.
+
+**NOTE:** A production deployment is not supposed to depend on the operator's self-signed CA instead must provide the certificates generated from a trusted certificate authority.
```
pohly: s/deployment is not/deployment that is not/

operator/deploy/operator.yaml
```diff
@@ -0,0 +1,153 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
```
pohly: Do we need this `monitoring.coreos.com`?

operator/deploy/operator.yaml
```diff
@@ -0,0 +1,153 @@
+apiVersion: v1
```
pohly: Please add a TODO here about moving this to `deploy` and enabling kustomize.

operator/deploy/operator.yaml
```diff
@@ -0,0 +1,153 @@
...
+kind: Namespace
+metadata:
+  name: pmem-csi
+---
```
pohly: Do we want to use this as default or should "default" be the default, with selecting a namespace done via kustomize?

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -0,0 +1,427 @@
...
+	// DeviceModeLVM represents 'lvm' device manager
+	DeviceModeLVM DeviceMode = "lvm"
+	// DeviceModeDirect represents 'direct' device manager
+	DeviceModeDirect DeviceMode = "direct"
+)
+
+// DeploymentSpec defines the desired state of Deployment
+type DeploymentSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
```
pohly: Can you replace this with instructions for how to do that in PMEM-CSI?

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -0,0 +1,427 @@
...
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	//DriverName represents the name of the PMEM-CSI driver to be deployed
```
pohly: nit: missing space

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -0,0 +1,427 @@
...
+		ProvisionerImage:    "provisionerImage",
+		NodeRegistrarImage:  "nodeRegistrarImage",
+		ControllerResources: "controllerResources",
+		NodeResources:       "nodeResources",
+		NodeSelector:        "nodeSelector",
+	}[c]
+}
+
+// EnsureDefaults make sure that the deployment object has all defaults set properly
+func (d *Deployment) EnsureDefaults() {
```
pohly: Is this hand-written or generated?

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	RunSpecs(t, "PMEM Operator API test suite")
+}
+
+var _ = Describe("Operator", func() {
+
+	BeforeEach(func() {
+	})
+
+	Context("API", func() {
```
pohly: Is this where we can verify that invalid values are rejected? If yes, please add such tests.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: This method is written by us _not_ generated by sdk.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
avalluri: The validation part is not handled inside our code, it is done by Kubernetes(api-server?) as per CRD json schema we provide. I will check if we can add such tests.

operator/deploy/operator.yaml
avalluri: I chose a non-default namespace just to illustrate how we could limit the operator and all driver resources to a specific namespace. Once we add customize support for this we can make this configurable. If you prefer to keep 'default' namespace I would change.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
pohly: `deployment_types.go` doesn't seem to enough information information for that kind of validation. For example:
```
// DeviceMode to use to manage PMEM devices. One of lvm, direct
DeviceMode DeviceMode `json:"deviceMode,omitempty"`
```

This "one of lvm, direct" is a restriction that we should enforce via validation, if possible.
`LogLevel` is the (only?) other field where validation seems possible at creation time. 

Everything else is something that the operator must validate because the semantic is too complex. But that also needs tests.

operator/deploy/operator.yaml
pohly: It's okay for now, we'll probably remember to update this when adding kustomize support.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
pohly: I think it's okay to add a TODO item about "input validation" for now, i.e. solve this after merging the PR into the "operator" branch (but before merging into "devel").

operator/deploy/operator.yaml
avalluri: May be not, will remove.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
avalluri: > For example:
> 
> ```
> // DeviceMode to use to manage PMEM devices. One of lvm, direct
> DeviceMode DeviceMode `json:"deviceMode,omitempty"`
> ```
> 
> This "one of lvm, direct" is a restriction that we should enforce via validation,

This supposed to be enforced by the schema, as we defined it [as enum]( https://github.com/intel/pmem-csi/pull/525/files/81e8d693c9dd88c8802e582461a1f546111d6484#diff-6dbccdd31f0d920af3f26b2bbb144b09R303-R310)

avalluri: Added a TODO for this.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: Fixed.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: Adde the `make` instruction.

operator/deploy/operator.yaml
avalluri: Added TODOs about kustomization, and location of this yaml file.

operator/deploy/operator.yaml
avalluri: Removed this.

operator/README.md
avalluri: Done.

operator/README.md
avalluri: Fixed.

operator/README.md
avalluri: Added the suggested explanation about default versions.

operator/README.md
avalluri: Done. Added a sentence about the resource.

operator/README.md
avalluri: Done.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
pohly: > This supposed to be enforced by the schema, as we defined it as [enum](https://github.com/intel/pmem-csi/pull/525/files/81e8d693c9dd88c8802e582461a1f546111d6484#diff-6dbccdd31f0d920af3f26b2bbb144b09R303-R310).

That link is now invalid. Did you mean the `type DeviceMode string` definition? I doubt that this has that effect, because the Go type system cannot list all valid strings. `DeviceMode("foobar")` is valid in Go.

Let's not speculate, let's test it! (motto of the day, eh, week/month/year/career ...) :grin:

pohly: > That link is now invalid.

One tip regarding that: what I do for long-running PRs with multiple links and rebases is to tag the current branch before rebasing and pushing that tag. That way, links that point to commits remain valid.

pohly: > I doubt that this has that effect, because the Go type system cannot list all valid strings. DeviceMode("foobar") is valid in Go.

I was wrong, it does have an effect:
```
$ kubectl apply -f operator/examples/pmem-csi.intel.com_v1alpha1_deployment_cr.yaml 
The Deployment "example-deployment" is invalid: spec.deviceMode: Unsupported value: "foobar": supported values: "lvm", "direct"
```

Excellent. We still need an automated test, though.

operator/README.md
pohly: It is not possible to select the namespace in which the operator creates its objects, is it? Instead it seems to use the namespace in which the operator itself is deployed. That may be good enough, but then should documented.

I thought we had discussed adding a "Namespace" field in the spec, but perhaps I misremember.

pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go
avalluri: > > That link is now invalid.
> 
> One tip regarding that: what I do for long-running PRs with multiple links and rebases is to tag the current branch before rebasing and pushing that tag. That way, links that point to commits remain valid.

Thanks @pohly for the useful tip.

avalluri:
> I've now also tried this in practice by deploying the operator. Some observations...
> 
> ```
> $ kubectl get deployment.pmem-csi.intel.com 
> NAME                 AGE
> example-deployment   115s
> ```
> 
> Can we get additional columns listed here? The "Phase" would be useful.

Yes, That would be useful at least as part of wide(`-o wide`) listing. But I need to dig this how to do it.

> 
> ```
> $ kubectl describe deployment.pmem-csi.intel.com 
> ...
> Events:   <none>
> ```
> 
> Would it be useful to post some events, like major status changes or information on where to find the generated objects?
> 
Ok, I agree will do this.

avalluri:
> ```
> $ kubectl get csidrivers
> No resources found
> ```
> 
> That's while the driver is running. We need a CSIDriver object because that's where we enable the skipping of attach.
> 

It's my bad I was thinking that installing CSIDriver is for pre-1.16 clusters. I will fix it. Thanks for finding this.

avalluri:
> > ```
> > $ kubectl get csidrivers
> > No resources found
> > ```
> > 
> > 
> > That's while the driver is running. We need a CSIDriver object because that's where we enable the skipping of attach.
> 
> It's my bad I was thinking that installing CSIDriver is for pre-1.16 clusters. I will fix it. Thanks for finding this.

Pushed a fix for installing missing CSIDriver object.

pkg/pmem-csi-operator/utils/utils.go
```diff
@@ -47,6 +72,17 @@ func GetKubernetesVersion() (*version.Info, error) {
 	return cs.Discovery().ServerVersion()
 }
 
+// CombinedVersion returns single version number for given major and minor versions
+// The assumption is that the minor version is single dit numbers
```
pohly: Besides the spelling ("dit"), your probably meant "at most a *double* digit number". One of your example even uses more than a single digit...

This smells like a hack. Instead of mapping a semantic version into an uint64, can you introduce a type with helper functions for version comparisons?

pkg/pmem-csi-operator/utils/utils.go
```diff
@@ -34,7 +36,30 @@ const (
 	defaultOperatorNamespace = metav1.NamespaceSystem
 )
 
-func GetKubernetesVersion() (*version.Info, error) {
+var (
+	k8sVersion  uint64
+	versionLock sync.Mutex
+)
+
+// GetKubernetesVersion retruns kubernetes server version
```
pohly: "returns Kubernetes"

pkg/pmem-csi-operator/utils/utils.go
```diff
@@ -34,7 +36,30 @@ const (
 	defaultOperatorNamespace = metav1.NamespaceSystem
 )
 
-func GetKubernetesVersion() (*version.Info, error) {
+var (
+	k8sVersion  uint64
+	versionLock sync.Mutex
```
pohly: Package-level variables are generally considered bad style (https://dave.cheney.net/2020/02/23/the-zen-of-go, "Avoid package level state"). Is that really necessary here? It's going to be simpler to cache the Kubernetes version where it is needed multiple times.

pkg/pmem-csi-operator/utils/utils.go
avalluri: > "utils" is not a good name. Ideally, this package should not exist at all.

Initially, this was started to host some utility functions like `GetKubernetesVersion()` `GetNamespace()` which are expected to be used inside the operator. I will check if we can avoid this after moving version and tls specifics to separate functions.

avalluri:
@pohly I updated the PR with:below changes:
 - removing the `utils` package by moving them to `k8sutils` package
 - made `version` type into it's own package with unit tests
 - made all `tls` related functions and types into a separate package with tests
 - updated deployment tests to check for `csidriver` object
 - Rebased aginst latest devel branch

Can you please have a look if it can be merged.

avalluri:
updated the commits to remove vendor folder changes.

pkg/pmem-csi-operator/pmem-tls/tls.go
```diff
@@ -0,0 +1,186 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pmemtls
```
pohly: Why is there "pmem" in the package name? The package has nothing to do with PMEM...

The package also isn't specific to the operator. Why not simply "pkg/tls"?

operator/operator.make
```diff
@@ -0,0 +1,16 @@
+OPERATOR_SDK_VERSION=v0.12.0
+
+# download operator-sdk binary
+_work/bin/operator-sdk-$(OPERATOR_SDK_VERSION):
+	mkdir -p _work/bin/ 2> /dev/null
```
pohly: What is redirecting stderr for? If there is output, don't we want to see it?

pohly:
> updated the commits to remove vendor folder changes.

I misunderstood that as "rebased on top of latest vendor changes in master". As it stands, the "operator" branch conflicts with "devel". Shall we fix that via a merge commit or with rebasing?

I'm leaning towards rebasing.

avalluri:
> > updated the commits to remove vendor folder changes.
> 
> I misunderstood that as "rebased on top of latest vendor changes in master". As it stands, the "operator" branch conflicts with "devel". Shall we fix that via a merge commit or with rebasing?
> 
> I'm leaning towards rebasing.

Actually i pushed latest(your change of deleted vendor foler) devel to operator and then rebased my branch. Might be i messed-up something. I will try to rebase now.

pohly:
Never mind, I accidentally checked out an old "operator" branch.

PR #528, opened 2020-01-30
Logs cleanup, document logs use in DEVELOPMENT

### LINKED ISSUE #497: Document (optionally: make closer to k8s convention) use of log levels
Discussion under #495 brought up a point we do not document use of log levels.
We could add note in README (or better DEVELOPMENT ?) about our style, which seems to be off-by-one from guidelines at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
i.e. we use (Info levels) 3 for production level and 4,5 for extended/development.
I recall the history was, when we cleaned our logging levels, we had most of statements at level 3 already so we wanted to keep those and we used 4 and 5 as more verbose levels.
We do not use info levels 0,1,2 currently at all.
At some point, do we wish to change all levels so that we conform better to convention in above mentioned page?

okartau:
Address log-related issues listed under #497
Started with that:
Do not use Infof() logger if there are no formatted content.

Let's add more under same PR here.
Resolves  #497

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,6 +20,7 @@ Table of Contents
     - [Specific arguments to pmem-vgm](#specific-arguments-to-pmem-vgm)
     - [Specific arguments to pmem-csi-driver](#specific-arguments-to-pmem-csi-driver)
     - [Environment variables](#environment-variables)
+    - [Logging](#logging)
 - [Notes about switching device mode](#notes-about-switching-device-mode)
     - [Going from LVM device mode to direct device mode](#going-from-lvm-device-mode-to-direct-device-mode)
     - [Going from direct device mode to LVM device mode](#going-from-direct-device-mode-to-lvm-device-mode)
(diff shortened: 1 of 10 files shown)


okartau:
I added WIP in title to indicate we can add more changes here.
I added a commit fixing misspelled `direcotry` in few places.
I rebased onto Fedora-31 tester branch to give more CI coverage to F-31 changes where we study the creation of VMs. F-31 changes should be transparent for changes of this PR so no harm is these are in the testing base.

okartau:
pushed a new commit adding description of log levels in DEVELOPMENT.md.
- Is it OK to have this as sub-chapter under API ? or should we have it as toplevel chapter somewhere?
- This is just quick draft, you are welcome to improve the wording
- grep of klog uses reveals we have log stmts without level, using just klog.Info, Watning, Error, should we attempt to unify, for example here in this PR (right now I added last statement in descriptio about those, but this mix of styles does not leave impression of solid approach)

okartau:
pushed new version with few more commits, added verbosity level version to few remaining Info stmts, thus addressing 3rd bullet of previous comment

DEVELOPMENT.md
```diff
@@ -258,6 +259,16 @@ Environment variables
 TEST_WORK is used by registry server unit-test code to specify path to certificates in test system. 
 Note, THIS IS NOT USED IN PRODUCTION
 
+Logging
+-------
+
+The klog.Info statements are used via the verbosity checker using the following levels:
```
grahamwhaley: Are these levels derived from the k8s klog logging conventions doc at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md ?
Looking at the two, I'm not sure they align? If they did, then it would probably makes sense to add a reference here.

okartau: Our levels developed separately and are about "off by one" related to guidelines pointed by this link, that's actually documented in issue this PR is addressing.
We discussed that off-line and current understanding is, we do not want to shift all V-levels right now in pmem-csi code.

okartau: I was thinking about adding reference, but then the obvious question will be "hey why are you off by one?", and right now the idea is we may want to make some other changes to logging, so it's not worth it to change all the Info code lines (plus all deployment files which set the verbosity level).

PR #530, opened 2020-02-07
scheduler components: extender + webhook

pohly:
The two new components should be compatible with all supported
Kubernetes versions, just the installation may be a bit difficult. For
kubeadm and the QEMU cluster a solution has been found, so the new
components are part of deployment and automatic testing even though
they are not enabled by default in the normal deployments.
    
As an interim solution they improve the situation with pod scheduling,
but a solution built into Kubernetes would be better.

The PR also includes some related fixes and the code from https://github.com/intel/pmem-csi/pull/525/commits/7168553f646508307bd954e4341a697599f75a11


pohly:
Tests are passing now. Please have a look. Start with the README.md change, it explains the concepts and usage.

README.md
```diff
@@ -416,6 +414,88 @@ Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The v
...
+
+Having to add `pmem-csi.intel.com/scheduler` manually is not
+user-friendly. To simplify this, PMEM-CSI provides a [mutating
+admission
+webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
+which intercepts the creation of all pods. If that pod uses volumes
+provided by PMEM-CSI, the webhook transparently adds the extended
+resource request.
+
+The webhook can only do that a) if the persistent volume claim (PVC)
```
okartau: Here the structure of stmt is "do that A if and B"
while better would be "do that if A and B"

README.md
```diff
@@ -416,6 +414,88 @@ Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The v
...
+
+The webhook deals with this uncertainty by allowing the creation of
+the pod without adding the extended resource when it lacks the
+necessary information. The alternative would be to reject the pod, but
+that is a change of behavior of the cluster that may affect also pods
+that don't use PMEM-CSI at all.
+
+Because of this limitation, it is more reliable to manually add the
+resource section.
+
```
okartau: The above chapter about webhook: What is the final point I get? 
We have manual config, and to make it more friendly, we have webhook, which is not reliable? Why would I choose unreliable solution over user-unfriendly but reliable solution?
Is webhook feature present only to help with this not-user-friendly issue, but the cost is lower reliability?

pohly: It's a tradeoff that users have to make. We can offer both, but we can't decide for them which approach they want to use. Neither one nor the other is perfect (one is user-friendly, the other reliable).

pohly: I kept thinking about this. The unreliability stems from using an informer, which by design may be not up-to-date yet when it is queried. I think it is possible to add fallbacks to the slower direct get calls when the informer doesn't have an item, and then this can be reduced to "it'll not work when PVC/StorageClass are created after pod", which is something that users can avoid.

pohly:
@okartau : I pushed one commit which should make the webhook more reliable.

pohly:
instead of using the cluster root CA, let's simply use self-signed certificates. I'll submit a separate PR for that, then rebase this PR.

pohly:
TODO:
- extend docs/images/communication/pmem-csi-communication-diagram.png
- extend API section in DEVELOPMENT.md

README.md
pohly: Resolved by removing the timing condition.

pohly:
> -  extend docs/images/communication/pmem-csi-communication-diagram.png
> -  extend API section in DEVELOPMENT.md

Both done. @okartau, @avalluri : please have another look.

I would like to get this merged before https://github.com/intel/pmem-csi/pull/525 because the latter still needs to update the architecture diagram.

avalluri:
> > * extend docs/images/communication/pmem-csi-communication-diagram.png
> > * extend API section in DEVELOPMENT.md
> 
> Both done. @okartau, @avalluri : please have another look.
> 
> I would like to get this merged before #525 because the latter still needs to update the architecture diagram.

@pohly If you believe it's ready to merge, can you please remove WIP from PR subject.

README.md
```diff
@@ -414,6 +413,85 @@ Volume requests embedded in Pod spec are provisioned as ephemeral volumes. The v
...
+Having to add `pmem-csi.intel.com/scheduler` manually is not
+user-friendly. To simplify this, PMEM-CSI provides a [mutating
+admission
+webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
+which intercepts the creation of all pods. If that pod uses volumes
+provided by PMEM-CSI, the webhook transparently adds the extended
+resource request.
+
+The webhook can only do that if the persistent volume claim (PVC) and
+its storage class have been created already. This is normally not
```
avalluri: Doe this webook also mutates the Pods which are using neither late-binding or ephemeral volumes but just claims with immediate binding?

pohly: It mutates all pods using PMEM-CSI. It intentionally checks as little as possible, because between the situation might theoretically still change between pod creation and pod scheduling.

The scheduler extender then looks at volumes and storage classes more thoroughly (https://github.com/intel/pmem-csi/blob/d1c38b6a1c4e9b96f7dd2bc8905675974f9df022/pkg/scheduler/scheduler.go#L223-L227).

avalluri: > It mutates all pods using PMEM-CSI.

I  wouldn't feel like we should touch the Pods using normal persistent volumes, those should go via the normal scheduling process. 

> It intentionally checks as little as possible,

If I am not wrong there are not many extra checks needed except checking the binding mode of the storage class linked to the volume claim, as anyway need the storage class at this point to check if it's provisioned by us.

>  because the situation might theoretically still change between pod creation and pod scheduling.

I won't expect if there would be any change in binding mode between those phases.

pohly: Okay, I made change in a new commit.

pohly:
@avalluri I fixed the rebase error, can you please have another look?

pkg/pmem-csi-driver/main.go
```diff
@@ -51,6 +57,24 @@ func Main() int {
 
 	klog.V(3).Info("Version: ", version)
 
+	var client *kubernetes.Clientset
+	if *schedulerListenAddr != "" {
+		if DriverMode(*mode) != Controller {
+			pmemcommon.ExitError("scheduler listening", errors.New("only supported in the controller"))
+			return 1
+		}
+		config, err := rest.InClusterConfig()
```
avalluri: Can you please get this Kubernetes specific code out of this package.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -258,6 +278,62 @@ func (pmemd *pmemDriver) registerNodeController() error {
...
+// logs errors and cancels the context when it runs into a problem,
+// either during the startup phase (blocking) or later at runtime (in
+// a go routine).
+func (pmemd *pmemDriver) startScheduler(ctx context.Context, cancel func(), rs *registryserver.RegistryServer) error {
+	if pmemd.cfg.schedulerListen == "" {
+		return nil
+	}
+
+	resyncPeriod := 1 * time.Hour
+	factory := informers.NewSharedInformerFactory(pmemd.cfg.client, resyncPeriod)
```
avalluri: Please limit all this Kubernetes specific code to the newly introduced scheduler package. Apparently this `startScheduler` function should be placed inside `scheduler`. So I would expect just something like:

```
.....
 if pmemd.cfg.schedulerListen != "" {
   err := scheduler.Start(ctx, cancel, rs, pmemd.cfg.schedulerListen)
   ....
}
```

pohly: I won't be that simple because this `scheduler.Start` would need more parameters than just `rs` and `pmemd.cfg.schedulerListen`. The API will also become more complex because of the need to manage Go routines.

As Dave wrote in https://dave.cheney.net/2020/02/23/the-zen-of-go, "Leave concurrency to the caller". That's what I am doing here.

pkg/pmem-csi-driver/main.go
avalluri: > How to get connected to Kubernetes is specific to the command. For example, the command might have specific command line options for specifying the configuration and which cluster to use.

Holding the configuration option is ok in this package, but I am concerned about the creating the clientset using that configuration.

pohly: Does 998d940f address this?

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: Another way to describe the current approach is "the scheduler package contains the business logic, the pmem-csi-driver package the setup code". Both Goroutine handling and creation of client and informer instances fall under "setup code".

avalluri: @pohly my point is about using the Kubernetes specifics into core(pmem-csi-driver) package.

> Another way to describe the current approach is "the scheduler package contains the business logic, the pmem-csi-driver package the setup code". Both Goroutine handling and creation of client and informer instances fall under "setup code".

I agree for some level. But I feel like `NewSchedule()` only needs _optional_ `client`(other than DriverName, RegistryServer) from which it could set up required informers itself. I called `client` optional because it only requires for testing this package. In reality, it should depend on incluster-config to setting-up the client.  So that we could avoid all client-go API calls from here.

pohly: This may be true right now, but may not be true forever: if we add another component which needs informers, then we'll want to set them up once here and then hand them around. So conceptually this is still the right place to make these calls.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -258,6 +278,62 @@ func (pmemd *pmemDriver) registerNodeController() error {
...
+	klog.V(5).Infof("synchronized caches: %+v", cacheSyncResult)
+	for t, v := range cacheSyncResult {
+		if !v {
+			cancel()
+			return fmt.Errorf("failed to sync informer for type %v", t)
+		}
+	}
+
+	go func() {
+		err := server.ListenAndServeTLS(pmemd.cfg.CertFile, pmemd.cfg.KeyFile)
```
pohly: The elephant in the room here is TLS again: this will run an HTTP server which allows more ciphers than we want...

I've added the new TLS check (PR #545) in this PR. That check should trigger.

pohly: I force-pushed a version which uses the same `LoadServerTLS` config as for grpc servers.

pohly:
@avalluri: rebased and retested. Okay to merge?

avalluri:
> @avalluri: rebased and retested. Okay to merge?

@pohly, my only concern is usage Kubernetes specifics stuff into "pmem-csi-driver" package. It would be great if we could avoid that. Other than that I am ok to merge.

pohly:
> my only concern is usage Kubernetes specifics stuff into "pmem-csi-driver" package

Let's agree to disagree on that and merge, okay?

avalluri:
> > my only concern is usage Kubernetes specifics stuff into "pmem-csi-driver" package
> 
> Let's agree to disagree on that and merge, okay?

After rethinking about this, I feel like we can make the scheduler out of the driver completely and run it as a sidecar container to the controller. It can communicate to the driver on unix socket and make use of CSI's `GetCapacity` with topology.

So, It's worth giving it a try before merging.

pohly:
> After rethinking about this, I feel like we can make the scheduler out of the driver completely and run it as a sidecar container to the controller. It can communicate to the driver on unix socket and make use of CSI's GetCapacity with topology.

That's a more complex and thus error-prone setup. Why should we do that when the same code can also run directly in the same binary?

pohly:
> So, It's worth giving it a try before merging

Try what? Of course this is doable. But that doesn't mean that it is going to be nicer.

avalluri:
> That's a more complex and thus error-prone setup.

I don't understand why it's an error-prone setup. Sidecars are a more proven setup in Kubernetes.

avalluri:
> > So, It's worth giving it a try before merging
> 
> Try what? Of course this is doable. But that doesn't mean that it is going to be nicer.

Seems like my suggestion is not so attractive, please go ahead with the current code.

Issue #529, opened 2020-02-06
low pass rate of pmem-csi-periodic CI job


okartau:
pmem-csi-periodic has failed most of recent runs, 1/5 OK.
The failure seems to come in one (but different) of "production" type stages, when timeout of 45 min is reached.
That is in contrast with PR-triggered jobs which mostly work, so either there is something different happening in production tester, or perhaps (as we re-use test cluster) the state of cluster somehow degrades when more is performed? (as PR-based jobs runs just few sessions on same cluster) and reports OK).

okartau:
I looked into logs of failing part of https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi-periodic/458/
which starts at 22:44:19

I cant see immediate error spelled out (can someone?), but is cluster in good shape when testing starts?
Few test are SKIPPED, then gotests is started,  tar/untar is performed.
```
Running in PMEM-CSI pod pmem-csi-node-b4g8
```
driver-registrar and pmem-driver pods reach RUNNING state at 22:45:04.
Then, there is first log after startup from driver.
But after that, there are only "CONNECTING....Reconnected" logs from pmem-csi driver,
plus output from top and df, until timeout kills the stage 45 min later.
Why is testing not running?

pohly:
It is suspicious that `pmem-csi-node-mjwdq` terminates at around the same time as the go testing starts. Why does it do that? Was the node getting overloaded (but why?) and Kubernetes started to remove pods?

A new node pod is getting created and that's where testing is about to start, but somehow executing this command never completes:
```
 [1mSTEP[0m: Running in PMEM-CSI pod pmem-csi-node-b4g8l
 Feb  5 22:45:00.131: INFO: ExecWithOptions {Command:[/bin/sh -c tar -xf - &&if _work/test.test -h 2>&1 | grep -q ginkgo; then TEST_WORK=_work _work/test.test -test.v -ginkgo.v; else TEST_WORK=_work _work/test.test -test.v; fi] Namespace:default PodName:pmem-csi-node-b4g8l ContainerName:pmem-driver Stdin:0xc000010ef8 CaptureStdout:true CaptureStderr:true PreserveWhitespace:false}
```

We don't have enough information about the failure. We need a copy of the kubelet log. It's probably too large to dump all the time, so we should capture it only after a timeout.

pohly:
Here's and idea how that additional log gathering may work:
```
[...]

PR #534, opened 2020-02-18
Jenkinsfile: collect pods state and kubelet logs in caught exception

okartau:


pohly:
Have you tested that this works?

Jenkinsfile
```diff
@@ -525,6 +525,12 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                                          TEST_E2E_SKIP=\$(if [ \"${env.CHANGE_ID}\" ] && [ \"${env.CHANGE_ID}\" != null ]; then echo \\\\[Slow\\\\]; fi) \
                            ' \
            "
+    } catch (exc) {
+        echo "Handling exception, trying to get state and logs:"
+        sh '''
+            _work/${env.CLUSTER}/ssh.0 kubectl get pods --all-namespaces -o wide
+            _work/${env.CLUSTER}/ssh.0 sudo journalctl -u kubelet
```
pohly: We want the kubelet log for all nodes in the cluster, not just the first one.

okartau:
> Have you tested that this works?

good point! As we mostly get good runs in PR jobs, having change here in current form does not show much useful information.
I added same in another Jenkins where I try to debug this pipeline, and it appears variable ${env.CLUSTER} is not set inside catch-block, although this is not intuitive as it appears to be set in try-block. Seems we have to hard-code few values.

okartau:
or is it so that triple-single-quote what I used, does not let me pass variable there, and I should use double quote.

pohly:
> good point! As we mostly get good runs in PR jobs, having change here in current form does not show much useful information.

What I proposed in https://github.com/intel/pmem-csi/issues/529#issuecomment-586876944 was to artificially make the test run time out by replacing the actual test with a sleep. Combine that with shortening the timeout for the step and you get relatively fast feedback whether this works (still need to build and deploy the driver).

> or is it so that triple-single-quote what I used, does not let me pass variable there, and I should use double quote.

Yes, that's probably it.

Jenkinsfile
```diff
@@ -525,6 +525,17 @@ void TestInVM(deviceMode, deploymentMode, distro, distroVersion, kubernetesVersi
                                          TEST_E2E_SKIP=\$(if [ \"${env.CHANGE_ID}\" ] && [ \"${env.CHANGE_ID}\" != null ]; then echo \\\\[Slow\\\\]; fi) \
                            ' \
            "
+    } catch (exc) {
+        echo "Handling exception, trying to get state and logs:"
+        sh "_work/${env.CLUSTER}/ssh.0 kubectl get pods --all-namespaces -o wide"
+        echo "====== KUBELET LOG MASTER ======"
+        sh "_work/${env.CLUSTER}/ssh.0 sudo journalctl -u kubelet"
+        echo "====== KUBELET LOG WORKER-1 ======"
```
pohly: Please don't hard-code the number of nodes. You can iterate over `_work/${env.CLUSTER}/ssh.*` instead.

okartau: removed hard coding, now runs using all ssh.* wrappers

okartau:
Now there is version that gets activated on error (including timeout of stage), showing pods list and kubelet logs using all existing ssh.X.
Anything else we want to add?
The forcibly shortened timeout triggers abortion now. Just sleep is not enough as kube system is set up in first part of same stage.

pohly:
Looks good to me (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-534/6/execution/node/113/log/?consoleFull). Let's remove the timeout change and merge this.

PR #535, opened 2020-02-19
TLS cleanup

pohly:
There's no reason for using the cluster root certificate and for each
node using its own key (the registry is not checking that). We also
don't need to rename the content of the secrets inside the pod specs.
    
This makes the setup code easier and avoids the (false!) impression
that the deployment needs different keys for each node. The
description of our TLS handling also becomes simpler.

The `setup-ca.sh` script itself is made a bit more readable.


pohly:
> you agree that the certificate generation scripts would be deprecated once we move to the operator.

Depends on whether we want to only support installation via the operator. As long as we still support manually customizing and deploying yaml files, we also need a way to generate the secrets even when it's just as example. Let's decide that once we have more experience with using the operator.

PR #537, opened 2020-02-26
pmem-csi-driver: disable insecure ciphers

avalluri:
By default 3DES ciphers are enabled by 'crypto/tls' package which are vulnerable
to SWEET32 attack.  This change disables insecure ciphers from supported
ciphersuite list.  Currently blocked ciphers are 3DES and RC4.

These is no direct way offered by 'crpyto/tls' package to disable a single
cipher from tls.Config. So we do this by preparing per client connection config
using tls.Conig.GetConfigForClient. This function gets called while connection
handshaking where client sends ClientHelloInfo with the cipher suites it
supported. Where we filter out the insecure ciphers from that list and set to
newly created tls.Config.

Here is the `nmap` output run aginst node controller server:
```
$ nmap --script +ssl-enum-ciphers -p 10001 172.17.0.4

Starting Nmap 7.60 ( https://nmap.org ) at 2020-02-26 12:37 EET
Nmap scan report for 172.17.0.4
Host is up (0.00033s latency).

PORT      STATE SERVICE
10001/tcp open  scp-config
| ssl-enum-ciphers:
|   TLSv1.2:
|     ciphers:
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA (rsa 2048) - A
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (rsa 2048) - A
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (rsa 2048) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA (rsa 2048) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (rsa 2048) - A
|       TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_128_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_AES_128_CBC_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_128_GCM_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_CBC_SHA (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_GCM_SHA384 (rsa 2048) - A
|     compressors:
|       NULL
|     cipher preference: client
|_  least strength: A

Nmap done: 1 IP address (1 host up) scanned in 0.42 seconds
```

### DIFF
--- a/pkg/pmem-grpc/grpc.go
+++ b/pkg/pmem-grpc/grpc.go
@@ -3,6 +3,7 @@ package pmemgrpc
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


pkg/pmem-grpc/grpc.go
```diff
@@ -85,26 +85,48 @@ func LoadServerTLS(caFile, certFile, keyFile, peerName string) (*tls.Config, err
...
-		ClientCAs:     certPool,
-		ClientAuth:    tls.RequireAndVerifyClientCert,
-		VerifyPeerCertificate: func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
-			// Common name check when accepting a connection from a client.
-			if peerName == "" {
-				// All names allowed.
-				return nil
+		GetConfigForClient: func(info *tls.ClientHelloInfo) (*tls.Config, error) {
+			if info == nil {
+				return nil, fmt.Errorf("nil client info passed")
```
pohly: Nit: this isn't a format string, better use `errors.New`.

pkg/pmem-grpc/grpc.go
```diff
@@ -85,26 +85,48 @@ func LoadServerTLS(caFile, certFile, keyFile, peerName string) (*tls.Config, err
...
+				CipherSuites:  ciphers,
+				VerifyPeerCertificate: func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+					// Common name check when accepting a connection from a client.
+					if peerName == "" {
+						// All names allowed.
+						return nil
+					}
+					if len(verifiedChains) == 0 ||
+						len(verifiedChains[0]) == 0 {
+						return fmt.Errorf("no valid certificate")
```
pohly: Nit: `errors.New`

pkg/pmem-grpc/grpc.go
```diff
@@ -85,26 +85,48 @@ func LoadServerTLS(caFile, certFile, keyFile, peerName string) (*tls.Config, err
...
+				return nil, fmt.Errorf("nil client info passed")
 			}
-			if len(verifiedChains) == 0 ||
-				len(verifiedChains[0]) == 0 {
-				return fmt.Errorf("no valid certificate")
+			klog.Infof("GetConfigForClient: servername(%q)", info.ServerName)
+			ciphers := []uint16{}
+			for _, c := range info.CipherSuites {
+				// filter out all insecure ciphers from client offered list
+				if c == tls.TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA ||
```
pohly: This looks better with a switch:
```
   switch c {
       case tls.TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,
            tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
            tls.TLS_ECDHE_RSA_WITH_RC4_128_SHA,
            tls.TLS_RSA_WITH_RC4_128_SHA:
            continue
     }
````

pkg/pmem-grpc/grpc.go
avalluri: Fixed.

pkg/pmem-grpc/grpc.go
avalluri: Fixed.

pohly:
I looked at this topic again in the context of securing the HTTPS server from PR #530 and doing it with a whitelist (i.e. only use selected ciphers) instead of blacklisting known-bad ones seems rather straightforward: https://github.com/golang/go/issues/11047

@avalluri: was there a particular reason why you decided to use blacklisting?

avalluri:
> I looked at this topic again in the context of securing the HTTPS server from PR #530 and doing it with a whitelist (i.e. only use selected ciphers) instead of blacklisting known-bad ones seems rather straightforward: golang/go#11047
> @avalluri: was there a particular reason why you decided to use blacklisting?

Yes, setting only supported ciphersuite list is stright forward, It is useful if one has that specific list, but our case is to block only known weak ciphers. There is no way to fetch all supported ciphers from where we could filter-out blocked list. Unless we know that full supported list, we might new ciphers would be added in later releases.

pohly:
> but our case is to block only known weak ciphers

Note quite - the ask is "don't use weak ciphers", which can also be achieved by selecting only strong ones (whitelist). We are not asked to "*only* block known weak ciphers".

But I agree, maintaining that whitelist is extra work and it may be better to let the Go community do that for us.

PR #545, opened 2020-03-06
check TLS settings

pohly:
The recent tightening of TLS ciphers came without automated testing and turned out to be surprisingly hard to test manually due to bugs in nmap.

This PR simplifies the deployment a bit and then adds full port scanning (not just currently known ports!) and checking of the TLS settings.

### DIFF
--- a/test/e2e/tls/tls.go
+++ b/test/e2e/tls/tls.go
@@ -1,0 +1,185 @@
+/*
+Copyright 2019 Intel Corporation.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package tls
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+
+	pmempod "github.com/intel/pmem-csi/test/e2e/pod"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+)
+
+var _ = Describe("TLS", func() {
+	f := framework.NewDefaultFramework("tls")
+
+	// All of the following pod names, namespaces and ports match
+	// those in the current deployment files.
+
+	var nodePod *v1.Pod
+	BeforeEach(func() {
+		// Find one node driver pod.
+		label := labels.SelectorFromSet(labels.Set(map[string]string{"app": "pmem-csi-node"}))
+		pods, err := f.ClientSet.CoreV1().Pods("default").List(metav1.ListOptions{LabelSelector: label.String()})
+		framework.ExpectNoError(err, "list PMEM-CSI node pods")
+		Expect(pods.Items).NotTo(BeEmpty(), "have PMEM-CSI node pods")
+		nodePod = &pods.Items[0]
+	})
+
+	Context("controller", func() {
+		It("is secure", func() {
+			checkTLS(f, "pmem-csi-controller-0.pmem-csi-controller.default")
+		})
+	})
+	Context("node", func() {
+		It("is secure", func() {
+			checkTLS(f, nodePod.Status.PodIP)
+		})
+	})
+})
+
+func checkTLS(f *framework.Framework, server string) {
+	containerName := "nmap"
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      containerName,
+			Namespace: f.Namespace.Name,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    "nmap",
+					Image:   os.Getenv("PMEM_CSI_IMAGE"),
+					Command: []string{"sleep", "1000000"},
+				},
+			},
+			RestartPolicy: v1.RestartPolicyNever,
+		},
+	}
+
+	By(fmt.Sprintf("Creating pod %s", pod.Name))
+	ns := f.Namespace.Name
+	podClient := f.PodClientNS(ns)
+	createdPod := podClient.Create(pod)
+	defer func() {
+		By("delete the pod")
+		podClient.DeleteSync(createdPod.Name, &metav1.DeleteOptions{}, framework.DefaultPodDeletionTimeout)
+	}()
+	podErr := e2epod.WaitForPodRunningInNamespace(f.ClientSet, createdPod)
+	framework.ExpectNoError(podErr, "running pod")
+
+	By("scanning ports")
+	// We have to patch nmap because of https://github.com/nmap/nmap/issues/1187#issuecomment-587031079.
+	output, _ := pmempod.RunInPod(f, os.Getenv("REPO_ROOT")+"/test/e2e/tls", []string{"nmap-ssl-enum-ciphers.patch"},
+		strings.Join([]string{
+			fmt.Sprintf("https_proxy=%s swupd bundle-add nmap patch >&2", os.Getenv("HTTPS_PROXY")),
+			"patch /usr/share/nmap/scripts/ssl-enum-ciphers.nse <nmap-ssl-enum-ciphers.patch",
+			fmt.Sprintf("nmap --script +ssl-enum-ciphers --open -Pn %s 2>&1", server),
+		},
+			" && "),
+		ns, pod.Name, containerName)
+
+	// Now analyze all ports and the ciphers found for them.
+	// The output will be something like this:
+	//   Nmap scan report for pmem-csi-controller-0.pmem-csi-controller.default (10.44.0.1)
+	//   Host is up (0.00013s latency).
+	//   rDNS record for 10.44.0.1: pmem-csi-controller-0.pmem-csi-controller.default.svc.cluster.local
+	//   Not shown: 997 closed ports
+	//   PORT      STATE SERVICE
+	//   8000/tcp  open  http-alt
+	//   | ssl-enum-ciphers:
+	//   |   TLSv1.0:
+	//   |     ciphers:
+	//   |       TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA (ecdh_x25519) - A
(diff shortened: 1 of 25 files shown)


pohly:
@avalluri let's see what result we get in the CI.

I still don't understand why you get the following ciphers listed:
```
TLS_RSA_WITH_AES_128_CBC_SHA256 (rsa 2048) - A
TLS_RSA_WITH_AES_128_GCM_SHA256 (rsa 2048) - A
TLS_RSA_WITH_AES_256_GCM_SHA384 (rsa 2048) - A
```

Also, why did you *not* get the insecure `TLS_ECDHE_ECDSA_WITH_RC4_128_SHA` listed when testing manually?

test/e2e/tls/tls.go
```diff
@@ -0,0 +1,185 @@
...
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      containerName,
+			Namespace: f.Namespace.Name,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    "nmap",
+					Image:   os.Getenv("PMEM_CSI_IMAGE"),
```
avalluri: @pohly So, this expect pmem-csi image to have `nmap` installed?

test/e2e/tls/tls.go
```diff
@@ -0,0 +1,185 @@
...
+		label := labels.SelectorFromSet(labels.Set(map[string]string{"app": "pmem-csi-node"}))
+		pods, err := f.ClientSet.CoreV1().Pods("default").List(metav1.ListOptions{LabelSelector: label.String()})
+		framework.ExpectNoError(err, "list PMEM-CSI node pods")
+		Expect(pods.Items).NotTo(BeEmpty(), "have PMEM-CSI node pods")
+		nodePod = &pods.Items[0]
+	})
+
+	Context("controller", func() {
+		It("is secure", func() {
+			checkTLS(f, "pmem-csi-controller-0.pmem-csi-controller.default")
```
avalluri: Could we avoid this hardcoding name, as it depends on the driver's namespace.

test/e2e/tls/tls.go
avalluri: Ok, Got it. The test installs the bundle here: https://github.com/intel/pmem-csi/pull/545/commits/a2ee870293a82b94efd3ef55bfcbc85af9b697df#diff-d5ba72216cc5edc8c882d8d767bbad46R90

test/e2e/tls/tls.go
pohly: I could look for a running PMEM-CSI driver, but that leads to a bigger question: which one should it use if there is more than one?

How do I test that the code that does the lookup work?

I prefer to solve that together with actually installing the different in different ways, including installing it in different namespaces. I want to work on that this week.

pohly:
> let's see what result we get in the CI.

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-545/5/execution/node/97/log/?consoleFull

```
Nmap scan report for 10.32.0.2
 Host is up (0.013s latency).
 Not shown: 999 closed ports
 PORT      STATE SERVICE
 10001/tcp open  scp-config
 | ssl-enum-ciphers: 
 |   TLSv1.2: 
 |     ciphers: 
 |       TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
 |     compressors: 
 |       NULL
 |     cipher preference: client
 |_  least strength: A
 MAC Address: 2E:34:D0:E1:B6:E5 (Unknown)
 
 Nmap done: 1 IP address (1 host up) scanned in 0.63 seconds
...
Nmap scan report for pmem-csi-controller-0.pmem-csi-controller.default (10.32.0.3)
 Host is up (0.0043s latency).
 rDNS record for 10.32.0.3: pmem-csi-controller-0.pmem-csi-controller.default.svc.cluster.local
 Not shown: 998 closed ports
 PORT      STATE SERVICE
 10000/tcp open  snet-sensor-mgmt
 | ssl-enum-ciphers: 
 |   TLSv1.2: 
 |     ciphers: 
 |       TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
 |       TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
 |     compressors: 
 |       NULL
 |     cipher preference: client
 |_  least strength: A
 10002/tcp open  documentum
 MAC Address: C2:8E:25:2F:BF:29 (Unknown)
 
 Nmap done: 1 IP address (1 host up) scanned in 0.76 seconds
```

This is identical to what I got locally. The question why @avalluri got RSA reported is still open - is the result reported by nmap here incomplete or is RSA really no longer enabled? Should it be?

test/e2e/tls/tls.go
avalluri: > I could look for a running PMEM-CSI driver, but that leads to a bigger question: which one should it use if there is more than one?

Could you consider using the controller's service ip.

avalluri: @pohly otherwise, we can deal with this part of operator changes as it also changes the name of the controller service.

pohly:
Let's clarify the difference in scan results first before merging.

avalluri:
> Let's clarify the difference in scan results first before merging.

I could figure out the difference, The fedora cluster I been working on was based on fedora:30,  where the RSA ciphers reported(surprisingly, the same result even if I run the `nmap` inside fedora:31 container). But the issue is not present on the latest fedora cluster which is using 31.

pohly:
> I could figure out the difference, The fedora cluster I been working on was based on fedora:30, where the RSA ciphers reported(surprisingly, the same result even if I run the nmap inside fedora:31 container).

That is rather unsettling and doesn't really explain why there is a difference. What we run our containers on should not affect the TCP communication between clients (like nmap) and servers (PMEM-CSI). Can you double-check by bringing up a cluster based on fedora 30 (= check out old source code, start cluster) and 31 (= check out current master, start cluster with different `CLUSTER` env var) in parallel and run the same unit test against both? If they behave differently, we need to understand why.

pohly:
Another, more likely explanation is that your Fedora 30 cluster also used the old approach for CA certificate creation, the one with the cluster root CA. I don't know whether the certificates affect the choice of ciphers, but they might.

This can be tested like this:
- bring up Fedora 30 and PMEM-CSI with old certificates
- run test
- update to current code, delete certificates in `_work`, run `make start` again => this will update the PMEM-CSI deployment
- run test again

avalluri:
Opps, Sorry for misleading. It was not the cluster that makes the difference. @pohly your guess was right.  The way we generate the server/client certificates makes the difference. On my earlier cluster, i was deploying the driver using the operator which generates certificates using RSA keys, whereas the `make start` uses "ecdsa".  When I started testing on the new cluster using make start/test_e2e, the first difference attracted me was the cluster version.  

Now i tested this with two driver deployments, one with `make start` and one with the operator. When i run `nmap` agist the driver deployed by the operator shows RSA ciphers:

```
 Nmap scan report for example-deployment-controller-0.example-deployment.pmem-csi.svc.cluster.local (10.42.0.2)
Host is up (0.0014s latency).

PORT      STATE SERVICE
10000/tcp open  snet-sensor-mgmt
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
|       TLS_RSA_WITH_AES_128_CBC_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_128_GCM_SHA256 (rsa 2048) - A
|       TLS_RSA_WITH_AES_256_GCM_SHA384 (rsa 2048) - A
|     compressors: 
|       NULL
|     cipher preference: client
|_  least strength: A
MAC Address: E6:E1:82:9C:8F:36 (Unknown)

Nmap scan report for pmem-csi-controller-0.pmem-csi-controller.default.svc.cluster.local (10.32.0.2)
Host is up (0.0014s latency).

PORT      STATE SERVICE
10000/tcp open  snet-sensor-mgmt
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_ECDHE_ECDSA_WITH_CHACHA20
_POLY1305_SHA256 (ecdh_x25519) - A
|     compressors: 
|       NULL
|     cipher preference: client
|_  least strength: A
MAC Address: 02:F4:50:31:B7:72 (Unknown)
 ```

pohly:
This makes the test a bit sensitive to the deployment method. I'll add a note to this PR which highlights that. As soon as we start testing against different deployments, we need to fix this somehow. The easiest solution might be to extend the example certs so that they also includes keys for RSA.

pohly:
@avalluri: does 9066ebe9 look okay?

It's good that we found this, it'll safe time later.

avalluri:
> @avalluri: does [9066ebe](https://github.com/intel/pmem-csi/commit/9066ebe9606b024fc0a81dfe7acce8e909811f48) look okay?

Yes, looks good. Let's merge this once after this CI test run.

PR #546, opened 2020-03-09
Dockerfile: remove ldconfig workaround

pohly:
The workaround was needed probably because symlink changes were not
recorded properly by Docker. When updating Clear Linux from 32480 to
version 32550 that workaround no longer was enough because
/usr/lib64/ld-linux-x86-64.so.2 itself become broken, making it
impossible to run any non-static command.

The right fix is to reconfigure Docker and overlay2 so that it uses
"Native Overlay Diff: true":

$ docker info
...
 Storage Driver: overlay2
  Backing Filesystem: <unknown>
  Supports d_type: true
  Native Overlay Diff: true

This can be done by using a kernel with CONFIG_OVERLAY_FS_REDIRECT_DIR
disabled or with a module parameter:

sudo modprobe -r overlay && sudo modprobe overlay redirect_dir=off
sudo systemctl restart docker


pohly:
@okartau this worked fine in the CI. Okay to merge?

PR #547, opened 2020-03-16
Small Jenkinsfile improvements

okartau:
Docker registry is used via hard-coded uuid value that is specific to
CI jenkins system, thus limiting Jenkinsfile to be functional
in this Jenkins system only where this UUID is defined.
We can make Jenkinsfile more generic by using a variable.
This variable needs to be provided by Jenkins instance.


okartau:
note that first CI run is expected to fail, as I did not add that variable in Jenkins yet. One step at a time...

okartau:
I added DOCKER_REGISTRY in Jenkins and triggered re-test, started running, good.

okartau:
I added another commit adding -i option to top. The idea is that -i should show meaningful lines, and we can stop cutting hard-coded 20 lines, which will mostly  show many irrelevant lines.

okartau:
Taking away WIP, ready for merging

PR #548, opened 2020-03-16
E2E deployment

pohly:
This currently enhances how e2e.test waits for a functional PMEM-CSI installation, without changing how PMEM-CSI gets installed by `make start`. Prometheus metrics are enabled and used for that, which may be a useful feature by itself. Right now it's not documented, which will be changed in a separate PR after considering more carefully which metrics data would be useful.

The longer-term goal is to also move the actual installation into the e2e.test binary. This could be done via this PR or a separate one.


pkg/pmem-csi-driver/main.go
```diff
@@ -12,39 +12,45 @@ import (
...
+	flag.StringVar(&config.StateBasePath, "statePath", "", "Directory path where to persist the state of the driver running on a node, defaults to /var/lib/<drivername>")
 
 	/* scheduler options */
-	schedulerListenAddr = flag.String("schedulerListen", "", "listen address (like :8000) for scheduler extender and mutating webhook, disabled by default")
+	flag.StringVar(&config.schedulerListen, "schedulerListen", "", "listen address (like :8000) for scheduler extender and mutating webhook, disabled by default")
 
-	version = "unknown" // Set version during build time
-)
+	/* metrics options */
+	flag.StringVar(&config.metricsListen, "metricsListen", "", "listen address (like :8001) for prometheus metrics endpoint, disabled by default")
```
pohly: I'm undecided whether this really should be a separate port. It might be simpler to have a single HTTPS server for unknown clients which serves both scheduler and metrics endpoints.

Thoughts?

pohly: I think I prefer to keep it separate. I asked in Kubernetes-CSI about the confidentiality of the metrics data and the statement there was that Google's own security team did not consider the data exposed by the sidecars confidential, i.e. exposing our own metrics data to anyone also should be fine. However, if we ever decided to require authentication of the client, then it is easier to do with a separate port because then we can do client checks already at the TLS level.

pohly:
Skipping tests in PR jobs is not quite working yet. Despite `-ginkgo.skip`, tests get executed. Will investigate in more detail...

pohly:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-548/21/testReport/fedora-1_16.lvm-testing/sanity/ does not contain sanity tests?

pohly:
> Skipping tests in PR jobs is not quite working yet. Despite -ginkgo.skip, tests get executed. Will investigate in more detail...

The string that gets matched against starts with "PMEM E2E suite [Top Level]", not with the top-level "Describe" strings, and therefore anchoring the regex with `^` prevented matching.

> does not contain sanity tests?

I broke that while trying to fix registration of our custom sanity tests. Should work now.

pohly:
This has now passed full testing (i.e. not just pre-merge): https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/e2e-deployment/2/

Test results and runtimes can be seen here: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-548/32/testReport/

For comparison, here's another PR: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-554/1/testReport/

I believe this is now working as intended and thus ready for review.

test/e2e/storage/sanity.go
```diff
@@ -496,10 +492,12 @@ var _ = Describe("sanity", func() {
 			)
 
 			BeforeEach(func() {
+				// Worker nodes with PMEM.
 				nodes = make(map[string]nodeClient)
-				for i, addr := range workerSocatAddresses {
+				for i := 1; i < cluster.NumNodes(); i++ {
```
okartau: if we start with one and check for less than NumNodes, is there a risk we don't reach last one?

pohly: We skip the first one (index == 0), which is intentional here because we just want the "worker nodes with PMEM".

okartau: ok, just wanted to be sure this is intentional

PR #549, opened 2020-03-16
vendoring and dependencies

pohly:
Vendoring is of questionable value, let's remove it. We can also reduce the amount of checking that we do to direct dependencies.


test/test.make
```diff
@@ -48,15 +48,13 @@ test_runtime_deps: check-go-version-$(GO_BINARY)
 
 RUNTIME_DEPS =
 
-# We use "go list" because it is readily available. A good replacement
-# would be godeps. We list dependencies recursively, not just the
-# direct dependencies.
-# Filter out the go standard runtime packages from dependecies
-RUNTIME_DEPS += diff <($(GO) list -f '{{join .Deps "\n"}}' ./cmd/pmem-csi-driver/ | grep -v ^github.com/intel/pmem-csi | sort -u) \
+# List direct imports of our commands, ignoring the go standard runtime packages.
+RUNTIME_DEPS += diff <(env "GO=$(GO)" hack/list-direct-imports.sh $(IMPORT_PATH) ./cmd/... | grep -v ^github.com/intel/pmem-csi | sort -u) \
```
avalluri: nit, script is already doing `sort -u`, we can remove this additional sorting here.

pohly: That the output is sorted isn't documented and I wanted to reserve the right to perhaps change that in the future, therefore I kept the sorting here. It makes it obvious here that we deal with predictable output, too.

avalluri: > That the output is sorted isn't documented and I wanted to reserve the right to perhaps change that in the future, therefore I kept the sorting here. It makes it obvious here that we deal with predictable output, too.

Ok, never mind then. It just took my attention of double sorting.

avalluri:
@pohly Does this change also supposed to remove vendor folder itself from the repo right?

pohly:
> Does this change also supposed to remove vendor folder itself from the repo right?

It was supposed to, but somewhere along the way (rebasing?) I lost that change. Let me fix that in a separate PR.

PR #553, opened 2020-03-23
deploy: fix 1.14 and 1.15, webhook was missing

pohly:
PR #530 broke deployment on 1.14 and 1.15 because it made all
components of a deployment mandatory but didn't add the "webhook"
link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1005,9 +1005,10 @@ kubectl label ns kube-system pmem-csi.intel.com/webhook=ignore
 ```
 
 This special label is configured in [the provided web hook
-definition](./deploy/kustomize/webhook/webhook.yaml). It can also be
-used to let individual pods bypass the webhook by adding that label.
-The CA gets configured explicitly, which is supported for webhooks.
+definition](./deploy/kustomize/webhook/webhook.yaml). On Kubernetes >=
+1.15, it can also be used to let individual pods bypass the webhook by
+adding that label. The CA gets configured explicitly, which is
+supported for webhooks.
 
 ``` sh
 mkdir my-webhook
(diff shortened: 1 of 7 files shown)


pohly:
This is intentionally a branch in the main repo to get full testing:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/fix-1.14/

pohly:
Support for `objectSelector` was only added in 1.15:
- 1.14: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#webhook-v1beta1-admissionregistration-k8s-io
- 1.15: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#mutatingwebhook-v1beta1-admissionregistration-k8s-io

I'll add a base configuration of the webhook which doesn't use it, which means that on 1.14 one can only exclude PMEM-CSI itself by installing it in a namespace that is excluded, like kube-system. For all other releases I'll patch in the `objectSelector`, so nothing will change there.

pohly:
> which means that on 1.14 one can only exclude PMEM-CSI itself by installing it in a namespace that is excluded

This implies that at least on 1.14, we either need to install it in kube-system or exclude the "default" namespace. I'm leaning towards the later because it is simpler and keeps the deployment similar to other releases - unless we want to use a non-default namespace also there?

pohly:
A test run with all tests (i.e. not just pre-merge) is currently on-going: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/fix-1.14-II/

pohly:
I had to create a second branch for that, somehow Jenkins started to ignore the existing one.

PR #557, opened 2020-03-27
Made headers consistent throughout all readmes in repo.

intelkevinputnam:


test/test-config.d/README.md
```diff
@@ -1,3 +1,5 @@
+# Congifure test-config.d
```
pohly: s/Congifure/Configure/

intelkevinputnam: Ha! Good catch.

intelkevinputnam:
@pohly When you have a chance, take a look at https://intelkevinputnam.github.io/pmem-csi/ 

You can see the source here: https://github.com/intelkevinputnam/pmem-csi/tree/kp-rtd-sphinx-reorg

As you can see the read the docs site exists in parallel with (and is dependent on) the existing markdown documentation, but makes all of the readmes a bit more accessible. I created new docs for the Design, Automated Testing, and Application examples. 

I also added a rule to the Makefile (html) that builds the docs for local viewing and created a GitHub Workflow to deploy the docs automatically to GitHub pages.

pohly:
> When you have a chance, take a look at https://intelkevinputnam.github.io/pmem-csi/

That looks very promising. Can you do a PR where we discuss those changes?

PR #561, opened 2020-03-31
Set k8s 1.18 as base for 1st CI stage, drop 1.14 deploy and test support

okartau:
When a popular distro, like Ubuntu or Fedora, installs Kubernetes now, it gets 1.18. I verified already that simple deployment completes (at least on Ubuntu 19.04) just by symlinking 1.18 entry to 1.16 under deploy/.
Lets see with this trial how CI tries 1.18 (temporarily changed 1.15 test pass to 1.18).


okartau:
We don't plan to add testing on 1.18 as new stage, but we rotate, keeping 3 versions? That would mean we remove 1.14 Fedora stage, right?
Note that this would leave Fedora stages 1.18, 1.16, 1.15 i.e. there is hole in 1.17 (which is currently tested on Clear). Thats for long-pass (pmem-csi-periodoc).
PR-tester currently tests on Fedora with 1.16, and Clear with 1.17, do we want to change that with addition of Fedora with 1.18?

Jenkinsfile
```diff
@@ -248,32 +248,32 @@ pipeline {
...
             options {
                 timeout(time: 180, unit: "MINUTES")
             }
             steps {
-                TestInVM("fedora", "", "1.15", "")
+                TestInVM("fedora", "", "1.16", "")
             }
         }
 
-        stage('testing 1.14') {
```
pohly, line 270: We cannot just stop testing on 1.14 entirely without reporting that in the "Supported Kubernetes versions" section of the README.md. Let's do that in this PR, with this comment:
"Lacks support for ephemeral inline volumes. Not supported officially by upstream anymore."

okartau: There are few other places 1.14 is mentioned in README, these need to be replaced with 1.15 I assume (and should verify these examples still work).
Also, once we touch "Supported Kubernetes Versions" table: this table currently goes up to 1.16, do we want to add 1.17 and 1.18 there, and do we need to add any "support status" about those versions?

okartau: About "Supported table", we even have 1.13 mentioned there, should we remove?

pohly: > do we want to add 1.17 and 1.18 there

Yes.

> do we need to add any "support status" about those versions

No, because there's nothing special to report.

> we even have 1.13 mentioned there, should we remove?

Let's keep it, to avoid questions along the lines of "do you support 1.13?" when we don't say anything about it.

okartau:
What about deploy/kustomize/kubernetes-1.14-* ?
And symlinks (directories `direct` and `lvm` and multiple yaml files) in deploy kubernetes-1.15 that point to 1.14 area? 
(the latter likely to get re-generated after changing former, right)

pohly:
> What about deploy/kustomize/kubernetes-1.14-* ?

Just remove everything related to Kubernetes 1.13 and 1.14. This is part of simplifying our source code. Otherwise we keep accumulating cruft even though we no longer support old Kubernetes releases.

We should have done that already earlier, when removing support for 1.13.

pohly:
Please squash commits. 6ec980f and c7323ab remove the patch files, but aren't self-contained because those patch files must have become unused already earlier.

Issue #564, opened 2020-04-02
log line "NodePublishVolume failed at attempt NNNNN" repeated many times


okartau:
I detected a pattern contributing to #563
In big log of #561 run-1, There are about 1.5 million lines, and about 370 thousand are
"NodePublishVolume failed at attempt NNNNN" where NNNNN rolls up to high numbers, max=209153
During one sampled second 14:40:06, there are 23551 lines about worker-1-volume-61:
```
~ $ grep "NodePublishVolume failed at attempt" log-pr561 |grep 14:40:06 |grep worker-1-volume-61 |wc
  23551  730081 5911301
```

But that volume is not the only one, at least one more different Volume has similar ongoing at same time and clocks 22817 attempts:
```
~ $ grep "NodePublishVolume failed at attempt" log-pr561 |grep 14:40:06 |grep worker-9-volume-62 |wc
  22817  707327 5727067
```
Means, the retry rate is quite high, like 23 kHz per volume, and 46 kHz overall during given second.

okartau:
What is the error returned?
In many cases there is "all SubConns are in TransientFailure, latest connection error: connection closed"
In these cases the retry count rolls up to few thousands, like 5 or 6 thousands, and lasts for short time.

But there is also a case where retry count gets up to 208k and handling response has:
```
[2020-03-31T14:44:55.414Z] pmem-csi-node-b7bmq/pmem-driver: I0331 14:44:55.024141       1 nodeserver.go:126] NodePublishVolume request: VolID:worker-b3750ab271e8b5acd2ce36e30a4f76f8fb34373769c9bf2ea67\
bf837 targetpath:/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-target.BCkKfJ/worker-1/target sourcepath:/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-staging.aNnjMC/worker-1 re\
adonly:false mount flags:[] fsType: context:map["name":"worker-1-volume-61-289E24D3-8D2FC8EA"]
[2020-03-31T14:44:55.414Z] pmem-csi-node-b7bmq/pmem-driver: I0331 14:44:55.024179       1 pmem-exec.go:12] Executing: file -bsL /dev/ndbus0region0fsdax/worker-b3750ab271e8b5acd2ce36e30a4f76f8fb3437376\
9c9bf2ea67bf837
[...]

Issue #563, opened 2020-04-02
Jenkins job logs are big


okartau:
After some commit around March-26, Jenkins job console logs size increased significantly.
I think the trigger was changed e2e testing.
pmem-csi-periodic, where logs used to be 22...25 MB, now has sizes 180...280 MB,
so we see about 10-fold increase.
```
-rw-r--r-- 1 ubuntu ubuntu  24M Mar 24 22:10 558/log
-rw-r--r-- 1 ubuntu ubuntu  28M Mar 25 10:19 559/log
-rw-r--r-- 1 ubuntu ubuntu 245M Mar 26 03:30 560/log
-rw-r--r-- 1 ubuntu ubuntu 237M Mar 26 15:31 561/log
-rw-r--r-- 1 ubuntu ubuntu 229M Mar 27 03:35 562/log
-rw-r--r-- 1 ubuntu ubuntu 228M Mar 27 15:46 563/log
-rw-r--r-- 1 ubuntu ubuntu 276M Mar 28 03:34 564/log
-rw-r--r-- 1 ubuntu ubuntu 187M Mar 28 15:35 565/log
-rw-r--r-- 1 ubuntu ubuntu 213M Mar 29 03:33 566/log
-rw-r--r-- 1 ubuntu ubuntu 187M Mar 29 15:16 567/log
-rw-r--r-- 1 ubuntu ubuntu  21M Mar 29 23:20 568/log
-rw-r--r-- 1 ubuntu ubuntu 228M Mar 30 15:16 569/log
-rw-r--r-- 1 ubuntu ubuntu 277M Mar 31 03:40 570/log
-rw-r--r-- 1 ubuntu ubuntu 219M Mar 31 15:15 571/log
-rw-r--r-- 1 ubuntu ubuntu 184M Apr  1 03:31 572/log
-rw-r--r-- 1 ubuntu ubuntu 211M Apr  1 15:35 573/log
-rw-r--r-- 1 ubuntu ubuntu 192M Apr  2 03:09 574/log
```

I tried to look into some 200+ MB logs of pmem-csi-periodic, but I dont see easy reason there, just lots of testing output. Did we increase amount of tests with recent e2e change ?

In PR-s the picture is not so clear.
Overall, a PR job should produce much shorter log than periodic, as not all stages run. That makes it especially interesting what happened in #561 

In one end #562 run shows relatively moderate 39 MB log file.

[...]

PR #565, opened 2020-04-02
test output

pohly:
This should reduce the amount of "failed at attempt" log messages.

Fixes: #564

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -74,4 +74,4 @@ replace (
 
 // Temporary fork. Can be removed once https://github.com/kubernetes/kubernetes/pull/85540
 // is merged and we update to a version >= 1.18.
-replace k8s.io/kubernetes => github.com/pohly/kubernetes v1.10.0-alpha.3.0.20191122094604-c7faaae98a14
+replace k8s.io/kubernetes => github.com/pohly/kubernetes v1.10.0-alpha.3.0.20200403121116-2ca8baa1a05a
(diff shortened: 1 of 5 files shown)


okartau:
This run produced 40M log which Jenkins again has trouble even showing. But I found repeated older log blocks in log, see recent comment in #563

pohly:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-565/1/:
> Apr  2 16:18:03.626: INFO: worker-4-volume-20: DeleteVolume failed at attempt 1: rpc error: code = DeadlineExceeded desc = context deadline exceeded

The default context timeout of 5 minutes per test was too low for the direct testing stress test. Previously that timeout was ignored, now it is considered. That explains the failure, but not the log output size.

The more interesting part is:
> I0402 16:25:22.667933  107082 dns_resolver.go:196] ccResolverWrapper: sending new addresses to cc: [{172.17.0.4:30208 0  <nil>}]

This first appears right after removing the first driver deployment. It comes from dangling gRPC connections to the nodes which now no longer exist. I thought I had solved this with https://github.com/intel/pmem-csi/blob/da7c9991944d55dd349385656497f7cd69b8429c/test/e2e/deploy/deploy.go#L50-L53 but perhaps not quite. Will check.

pohly:
> I thought I had solved this

I had missed the connections established for the original sanity tests. Might be fixed now.

pohly:
I think I sorted everything out, and https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-565/5 only had 12,040 KB of total log data.

@okartau ready to merge?

okartau:
Major improvement!
I haven't counted lines, but by just looking to log, it's obvious we don't have much repetitions, as jenkins timestamps mostly match pod log timestamps, i.e. there is mainly no repeating of older lines.

PR #559, opened 2020-03-30
Operator E2E tests

avalluri:
This is the initial set of e2e tests for the operator. The operator tests are kept
 in it's own package. The tests setup and tear down the operator for every test
 it runs.
    
 TODOs:
     - Add more test cases like, disruptive tests to ensure that the operator could
       recover state after a sudden reboot
     - testing vs production(?)

PR #566, opened 2020-04-02
test: always test with a fresh cluster

pohly:
After removing the "reuse cluster" logic in
https://github.com/intel/pmem-csi/pull/548, the same cluster was used
throughout all steps, so instead of testing on the desired OS, the
tests just ran again on the initial one.

A potentially running cluster is intentionally torn down before
testing because that way we can avoid the tear down after the last
step. We don't need to do it then because the whole VM will go down.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -434,6 +434,7 @@ void TestInVM(distro, distroVersion, kubernetesVersion, skipIfPR) {
                            loggers=; \
                            atexit () { set -x; kill \$loggers; kill \$( ps --no-header -o %p ); }; \
                            trap atexit EXIT; \
+                           make stop && \
                            make start && \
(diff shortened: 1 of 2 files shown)


avalluri:
I took this change to #559, lets see how it effects..

pohly:
> I took this change to #559, lets see how it effects..

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-559/16 was succesful. Can we merge this PR now?

avalluri:
> > I took this change to #559, lets see how it effects..
> 
> https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-559/16 was succesful. Can we merge this PR now?

Great. Yes, will merge this change.

PR #567, opened 2020-04-02
pmem-state: do not allow tampering of state file data

avalluri:
Driver's state should not be altered outside of the driver. So added an
integrity check to state data. StateManager computes the hash for the
data being saved and stores the generated hash along with the data.

This change is not backward compatible and breaks loading of older state
files which have no hash information.

### DIFF
--- a/pkg/pmem-state/pmem-state.go
+++ b/pkg/pmem-state/pmem-state.go
@@ -6,21 +6,26 @@ SPDX-License-Identifier: Apache-2.0
 package pmemstate
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
+	"reflect"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


pohly:
What is that supposed to protect against? A malicious admin which tries to break his own machine?
Bit flips that weren't detected by the filesystem?

pkg/pmem-state/pmem-state.go
```diff
@@ -73,7 +87,13 @@ func (fs *fileState) Create(id string, data interface{}) error {
 		return errors.Wrapf(err, "file-state: failed to create metadata storage file %s", file)
 	}
 
-	if err := json.NewEncoder(fp).Encode(data); err != nil {
+	hdata := hashedData{}
+	hdata.Data = clone(data)
```
pohly, line 91: Why do you need to clone?

avalluri:
This seems is not the right solution for attacking the tampering sate issue. I will close this PR

Issue #569, opened 2020-04-05
"stores state across reboots" failure in pmem-csi-periodic job 581


okartau:
Starting on Apr-2nd, pass rate of pmem-csi-periodic went down again.
In job 581 there is failure where volumes state after reboot did not match:

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi-periodic/581/testReport/junit/fedora-1_15.direct-testing/sanity/PMEM_CSI_stores_state_across_reboots_for_single_volume/

I compared Expected list to "to consist" (needs re-arrange of blocks as these are shown in different order, plus re-arrange of name vs size order in some places, and diff in ignore-whitespace style).
Here is the actual diff:
Expected:
```
VolumeId: "state--da663aa8543a457c07dfc0f0513aa9a660e5222f38c5898d5993508b",
VolumeContext: {
    "name": "state-volume-58EDBE6F-6B898A05",
    "size": "115343360",
},
```

To consist:
```
VolumeId: "state--da663aa8543a457c07dfc0f0513aa9a660e5222f38c5898d5993508b",
VolumeContext: {
    "name": "state-volume-58EDBE6F-6B898A05",
},
```

Seems "to consist" side lacks the "size" attribute in VolumeContext.
Why this started to fail recently?
Related to recent e2e testing changes?

There is another same-titled failure in lvm test, I haven't looked yet into details there.

okartau:
Another side-track to think here, why do we have entries in state that may have size attribute in map, or may not have. Is this known and planned, or does it point to another potential/real issue.

pohly:
> Can that possibly happen because we re-use nodeVolume here for different state entries, without clearing it in between?

Your probably nailed it. The question remains why this occurs now.

`GetAll` has to clear the variable first before reading into it with `json.Decode` because `Decode` adds values without removing existing ones, but `GetAll` doesn't clear: https://github.com/intel/pmem-csi/blob/d03cf74be569b333e274c2aaee5f732c0b49d45d/pkg/pmem-state/pmem-state.go#L110-L137

@avalluri: is there a reason why state manager was written with `interface{}` as type in the API? If the purpose was to allow storing different structs, then that doesn't work because `GetAll` assumes that all entries fit into the same struct.

Making it generic also doesn't help us, we only need it for nodeVolume. Unless you have a very simple way of clearing that variable in a generic way, then let's just get rid of `interface{}`.

okartau:
[...]

PR #573, opened 2020-04-07
pmem-csi-driver: change semantics of StateManager.GetAll()

avalluri:
Current sm.GetAll() is error prone if the caller does not reset the dataPtr read for each element, this was visible in issue #569.
    
With this change GetAll() only returns all ids of the elements found, then caller has to call Get() on each record he interested in.

Fixes: #569

### DIFF
--- a/pkg/pmem-csi-driver/controllerserver-node.go
+++ b/pkg/pmem-csi-driver/controllerserver-node.go
@@ -31,20 +31,6 @@ type nodeVolume struct {
 	Params map[string]string `json:"parameters"`
 }
 
-func (nv *nodeVolume) Copy() *nodeVolume {
-	nvCopy := &nodeVolume{
-		ID:     nv.ID,
-		Size:   nv.Size,
-		Params: map[string]string{},
-	}
-
-	for k, v := range nv.Params {
-		nvCopy.Params[k] = v
-	}
-
-	return nvCopy
-}
-
 type nodeControllerServer struct {
 	*DefaultControllerServer
 	nodeID      string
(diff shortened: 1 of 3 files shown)


pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -31,7 +31,7 @@ type nodeVolume struct {
 	Params map[string]string `json:"parameters"`
 }
 
-func (nv *nodeVolume) Copy() *nodeVolume {
+func (nv nodeVolume) Copy() *nodeVolume {
```
pohly, line 34: Isn't Go now forced to create a copy of the `nodeVolume` instance prior to calling `Copy`? Why not keep the function as it was?

Note that I am not worried about the performance overhead. Go even recommends to pass parameters by value instead of pointer and optimizes for that. I just don't understand the purpose of this change.

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -82,12 +82,13 @@ func NewNodeControllerServer(nodeID string, dm pmdmanager.PmemDeviceManager, sm
...
 		cleanupList := []string{}
-		v := &nodeVolume{}
-		err = sm.GetAll(v, func(id string) bool {
+		v := nodeVolume{}
+		err = sm.GetAll(&v, func(id string) bool {
 			// See if the device data stored at StateManager is still valid
 			for _, devInfo := range devices {
 				if devInfo.VolumeId == id {
 					ncs.pmemVolumes[id] = v.Copy()
+					v = nodeVolume{}
```
pohly, line 91: And what if we don't end up in this location? Won't that skip the resetting of `v`?

This whole API for `GetAll` is becoming very awkward. This doesn't look like the right solution.

It's not possible to keep the state manager completely generic. Let's change the `interface{}` that it works with into a custom interface with a `New` method.

Then `GetAll` can use `dataPtr.New()` to create a clean instance, fill that with `json.Decode`, and pass that to the callback.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: Yes, you might be true. The change might break the intention, and unnecessary. The reason behind the change was previously we had a pointer to nodeVolume while calling Copy(), whereas now we have the value of it.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: > And what if we don't end up in this location? Won't that skip the resetting of v?

Yes, this is valid case should be fixed.

avalluri: > This whole API for GetAll is becoming very awkward. This doesn't look like the right solution.


I pushed new revision of the patch, where I changed the sematics of GetAll(), now it returns only available ids,  then caller would use a Get() on those ids. This also reduces unnessasary(stale state entries) file-reads in this case.

okartau:
as testing time is not short, I can inform now (although this information seems gotten out of date already) that the test run with first version of this PR just completed and problem was fixed, i.e. the size was not propagating from different volume entry.

pkg/pmem-state/pmem-state.go
```diff
@@ -107,33 +101,26 @@ func (fs *fileState) Get(id string, dataPtr interface{}) error {
...
 }
 
-// GetAll retrieves metadata of all volumes found in fileState.location directory.
-// reads all the .json files in fileState.location directory and decodes the filedata
-func (fs *fileState) GetAll(dataPtr interface{}, f GetAllFunc) error {
+// GetAll retrieves the names of all .json files found in fileState.location directory
+func (fs *fileState) GetAll() ([]string, error) {
 	fs.stateDirLock.Lock()
 	files, err := ioutil.ReadDir(fs.location)
 	fs.stateDirLock.Unlock()
```
pohly, line 108: This could now have used the idiomatic
```
fs.stateDirLock.Lock()
defer fs.stateDirLock.Unlock()
```

But don't bother changing it.

okartau:
> Let's wait for a successful test run and then merge.

please note that CI PR run does seems not triggering the problem, so clean CI PR run does not show how useful this change is.
In all cases I monitored, both in official CI jenkins pmem-csi-periodic, and in my own local instance running same job, the 1.16 stage is clean and failure comes in 1.15, which we do not run in PR jobs.
That means, 1.15 test sequence is different from 1.16, getting into state with multiple leftover volumes, thus triggering the problem.

I will start a job in my local system with this patch applied.

PR #574, opened 2020-04-08
Reduce logs, cover edge cases in /sys mount check

okartau:
Looking at logs shows few possibilities to reduce redundant or not relevant logs.
While changing /sys detection logging, I discovered this function may return OK in some edge cases where search for mount point or mount option does not match, which is not what we want, so there is one commit improving that part.

### DIFF
--- a/test/e2e/deploy/deploy.go
+++ b/test/e2e/deploy/deploy.go
@@ -420,6 +420,7 @@ func EnsureDeployment(deploymentName string) *Deployment {
 		cmd.Dir = os.Getenv("REPO_ROOT")
 		cmd.Env = os.Environ()
 		cmd.Env = append(cmd.Env,
+			"TEST_DEPLOYMENT_QUIET=quiet",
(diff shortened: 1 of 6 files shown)


test/e2e/deploy/deploy.go
```diff
@@ -420,6 +420,7 @@ func EnsureDeployment(deploymentName string) *Deployment {
 		cmd.Dir = os.Getenv("REPO_ROOT")
 		cmd.Env = os.Environ()
 		cmd.Env = append(cmd.Env,
+			"TEST_DEPLOYMENT_QUIET="+string("quiet"),
```
pohly, line 423: This looks odd. Why the string cast, and why not simply "TEST_DEPLOYMENT_QUIET=quiet"?

okartau: thanks, good point, I had over-complicated this by copying existing line. Changed it to
"TEST_DEPLOYMENT_QUIET=quiet" and pushed.

PR #575, opened 2020-04-08
test: unify handling of driver and operator deployment

pohly:
This removes the code duplication introduced by
b8bfa5a71ab0ddbda0cc1df35b3105eab6a188ba which copied much of
deploy.go into operator.go.

Besides making the overall code base smaller, this is also a step
towards deploying the actual driver through the operator and then
running tests against that driver installation: because the deploy
code will now install either the operator or the driver, we don't get
conflicts when the operator installs the real driver. However, for the
current tests the non-existent image continues to be used because it
probably makes the tests run a bit faster.

Another side effect of the unification is that the operator only gets
deployed once per test suite run.


pohly:
@avalluri please have a look. Tests have passed.

test/e2e/operator/deployment_api.go
```diff
@@ -36,13 +37,17 @@ var (
...
 )
 
-var _ = DescribeForAll("pmem-csi", func(o *Operator, f *framework.Framework) {
+var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
+	// Run these tests for all bare operator deployments, i.e.
+	// those which did not already install the driver.
+	return d.HasOperator && !d.HasDriver
+}, func(d *deploy.Deployment) {
+
+	f := framework.NewDefaultFramework("operator")
```
avalluri, line 46: We can reuse the framework initialized for deployment I guess.

test/start-operator.sh
```diff
@@ -30,6 +30,8 @@ EOF
   ${SSH} "cat >'$tmpdir/kustomization.yaml'" <<EOF
 resources:
 - operator.yaml
+commonLabels:
+  pmem-csi.intel.com/deployment: ${TEST_OPERATOR_DEPLOYMENT:-operator}
```
avalluri, line 34: can you please add a comment about this new `TEST_OPERATOR_DEPLOYMENT` test environment variable how and when should be used.

pohly:
> But in the case of testing driver deployed via operator requires both the combinations.

That then can be "operator-[lvm/direct]-[testing/production]".

> And what in my mind was that in this case we deploy single operator for all driver tests suite(regardless of lvm, direct, testing or production).

We could tweak the `EnsureDeployment` so that it keeps the operator running when switching between operator-* deployments. There are some details to take care of though, like the deployment label of the operator. I see this as an optimization and I am not sure whether it is needed. Deploying the operator is faster than deploying the driver.

If we want to test that a single operator can deploy drivers in different modes, then we should write a dedicated test for it instead of testing it indirectly.

> And the other thing what i feel missing is that the operator used to use custom namespace for running the tests.

So we need a test for deploying the operator in custom namespace.

test/e2e/deploy/deploy.go
```diff
@@ -333,33 +375,86 @@ func (d Deployment) DeploymentMode() string {
...
+	driver, err := findDriver(c)
 	if err != nil {
 		return nil, err
 	}
+	operator, err := findOperator(c)
+	if err != nil {
+		return nil, err
+	}
+	if operator != nil && driver != nil && operator.Name != driver.Name {
+		return nil, fmt.Errorf("found two different deployments: %s and %s", operator.Name != driver.Name)
```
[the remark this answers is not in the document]
pohly: > Isn't it a valid case of running driver tests deployed via operator?

Yes, but even in that case the operator and driver are supposed to be from the same deployment.

I'll fix the parameters.

test/e2e/operator/deployment_api.go
pohly: The deployment framework doesn't have a namespace created. We therefore need a new one here.

test/start-operator.sh
pohly: You didn't document `TEST_OPERATOR_NAMESPACE` either and that made sense to me because both variables are obvious. But I can explain both, of course.

pohly: Done.

test/e2e/deploy/deploy.go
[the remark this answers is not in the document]
pohly: Fixed.

test/e2e/deploy/deploy.go
```diff
@@ -309,12 +342,21 @@ func RemovePMEMDriver(c *Cluster, deploymentName string) error {
...
 	// Name string that all objects from the same deployment must
 	// have in the DeploymentLabel.
 	Name string
 
+	// HasDriver is true if the driver itself is running. The
+	// driver is reacting to the usual pmem-csi.intel.com driver
+	// name.
+	HasDriver bool
+
+	// HasDriver is true if the operator is running.
```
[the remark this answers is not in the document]
pohly: Fixed.

pohly:
> That then can be "operator-[lvm/direct]-[testing/production]".

I've added that, so far only with a simple "does the driver really run" test. That is currently work in progress and doesn't pass:
- direct mode: fails to come up (https://github.com/intel/pmem-csi/issues/581)
- direct and LVM mode: the deployment lacks the metrics service

pohly:
I believe this ready now. However, testing on Clear Linux failed, perhaps because the test is flaky (https://github.com/intel/pmem-csi/issues/585). I'm currently retesting.

@avalluri: please take another look

pohly:
> However, testing on Clear Linux failed, perhaps because the test is flaky 

This was probably caused by not deleting secrets immediately when tearing down the previous deployment. The operator then proceeded (without an error!) although the existing secrets were in the process of being deleted by the apiserver's garbage collection.

pohly:
I've extended `RemoveObjects` to also delete secrets.

avalluri:
> I've extended RemoveObjects to also delete secrets.

@pohly I fail to understand why you needed to this deltion manually? this should be done automatically when the deployment gets deleted.

pohly:
> I fail to understand why you needed to this deltion manually? this should be done automatically when the deployment gets deleted.

Because `DeleteObjects` is supposed to delete everything itself and check that it really is gone (not just in "terminating" state). If the operator is broken for whatever reason, `DeleteObjects` will fix the cluster and allow the next test to continue.

pohly:
> this should be done automatically when the deployment gets deleted.

Yes, but as we've (probably) seen in the failed test, even in that case we need to wait for deletion. `RemoveObjects` does that and expedites deletion by trigger it itself.

avalluri:
> this should be done automatically when the deployment gets deleted.
> 
> Yes, but as we've (probably) seen in the failed test, even in that case we need to wait for deletion. RemoveObjects does that and expedites deletion by trigger it itself.

If deleting of a Deployment object is not cleanig the sub-resources it crated, then its a real failure which we cannot hide by manually deleting them. To make sure that every operator tests starts in a calean environment without effecting these type of issues, running them in diffrente namespace is the right soulution in my opinion.

pohly:
> If deleting of a Deployment object is not cleanig the sub-resources it crated, then its a real failure which we cannot hide by manually deleting them.

This particular feature needs an explicit test, instead of relying on other tests to discover a regression accidentally.

> To make sure that every operator tests starts in a calean environment without effecting these type of issues, running them in diffrente namespace is the right soulution in my opinion.

We can't rely on namespaces because we have non-namespaced objects.

pohly:
@avalluri can we merge this once the current test run (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-575/10) completes?

pohly:
It has completed now, once without explicit secret removal (which shows that it was indeed a race condition) and once with (which should be more reliable).

I'd like to get this merged so that we can write more tests based on this infrastructure.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -158,6 +158,10 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, exist
 			d.Spec.PMEMPercentage = existing.Spec.PMEMPercentage
 			updateDeployment = true
 			err = fmt.Errorf("changing %q of a running deployment %q is not allowed", c, d.Name)
+		case api.Labels:
+			d.Spec.Labels = existing.Spec.Labels
+			updateDeployment = true
+			err = fmt.Errorf("changing %q of a running deployment %q is not allowed", c, d.Name)
```
avalluri, line 165: @pohly We could use `Deployment.ObjectMeta.Labels` instead of adding it to DeploymentSpec. It's more convenient to the user, treating like other Kubernetes objects. And, Is there any specific reason behind choosing `labels` not editable? It feels to me like a useful feature to make it editable.

pohly: > We could use Deployment.ObjectMeta.Labels instead of adding it to DeploymentSpec. It's more convenient to the user, treating like other Kubernetes objects.

I'm not aware of any other object where attributes of created objects are copied from somewhere other than the "Spec" field. Do you have an example?

> And, Is there any specific reason behind choosing labels not editable? It feels to me like a useful feature to make it editable.

Just a feeling that this would be potentially dangerous and/or confusing because there will be time periods where some objects have the old labels (like running pods) and some the new labels (the daemonset).

Can you give an example under which circumstances an admin might want to mutate labels?

The label selector in a service is also immutable. This doesn't affect us because our service object doesn't use the custom labels, I'm just mentioning it as an example where others have decided that supporting label changes isn't a good idea.

If you want to discuss this further, can we make this an open issue and in the meantime merge this PR?

avalluri: > I'm not aware of any other object where attributes of created objects are copied from somewhere other than the "Spec" field. Do you have an example?

Sorry, If I couldn't get exactly. But my point was we need not introduce a new spec field for labels, as every Kubernetes objects already have ObjectMeta. Similar to how we use `Deployment.Name` we could use `Deployment.Labels`.

avalluri: > Just a feeling that this would be potentially dangerous and/or confusing because there will be time periods where some objects have the old labels (like running pods) and some the new labels (the daemonset).

Ok, I do not have a strong feeling here. I just felt like to match the behaviour of other Kubernetes objects.

avalluri:
@pohly I have no other concerns on this PR than to use  `metadata.labels` in place of `spec.labels`.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: > Sorry, If I couldn't get exactly. But my point was we need not introduce a new spec field for labels, as every Kubernetes objects already have ObjectMeta.

Yes, but it's not normally used for derived objects. All of the app controllers allow separate labels for the main object and for the objects that get created for it.

> Similar to how we use Deployment.Name

Which is also a bit unusual...

pohly:
> I have no other concerns on this PR than to use metadata.labels in place of spec.labels.

Have I addressed that? Can we merge?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > Sorry, If I couldn't get exactly. But my point was we need not introduce a new spec field for labels, as every Kubernetes objects already have ObjectMeta.
> 
> Yes, but it's not normally used for derived objects. All of the app controllers allow separate labels for the main object and for the objects that get created for it.

@pohly Just for reference, can you please give some example API which uses separate labels other than metadata.labels. Kubernetes mentions labels & annotations as [common features](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#common-features) for CRs.

pohly: That page just says that CRs automatically have labels. It does not say that objects created by an operator inherit the labels.

Example APIs:
- statefulset: labels of pods come from spec.template.metadata.labels, not from metadata.labels; labels of PVCs come from spec.template.metadata.labels
- deployment: labels of the replicaset and pods come from spec.template.metadata.labels, not from metadata.labels

Do you have an example of an API where the labels are copied from the CR's metadata.labels? Which one do we want to follow, core Kubernetes behavior or some add-on?

pohly: > Example APIs: ...

My point is that these APIs show that objects that describe other objects do so in their "spec" field, not with their own meta data. My gut feeling is that this is a useful, more flexible design.

I know that we don't follow that with the namespace, but that's for technical reasons. The Kubernetes deployment also uses the same namespace even if something else is specified in the template; I haven't tested statefulset, but it probably behaves the same.

I know that we reuse metadata.name as base for created objects. I see that as a shortcut; we might want to add a optional spec.Name field for the sake of consistency.

avalluri: > That page just says that CRs automatically have labels. It does not say that objects created by an operator inherit the labels.
> Example APIs:
> 
> statefulset: labels of pods come from spec.template.metadata.labels, not from metadata.labels; labels of PVCs come from spec.template.metadata.labels
> deployment: labels of the replicaset and pods come from spec.template.metadata.labels, not from metadata.labels
> 

Thanks, @pohly, for pointing examples. Yes, my thinking about labels, in this case, is wrong. I agree let's keep them in spec.

avalluri:
> I have no other concerns on this PR than to use metadata.labels in place of spec.labels.
> 
> Have I addressed that? Can we merge?

Yes, it looks good to me. Let's merge this change. so that I will rebase #584.

PR #584, opened 2020-04-11
more operator fixes

avalluri:
This PR contains below fixes:
 - validating deployment certificates
 - validating deployment driver mode
 - using operator image as the default driver image 
 - do not delete CRD on exit if any active deployments
 - fix /sys mounting in direct mode
 - handle reconcile on restart operator


pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -103,6 +106,17 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
...
+		if !foundInCache {
+			// Possible that operator restarted, so we can't assume that
+			// everything upto date. Look at running deployment if any changes
+			// needed
+			oldDeployment, err := d.buildFromPreDeployed(r)
+			if err != nil {
+				return true, err
+			}
+
+			changes = d.Compare(oldDeployment)
```
pohly, line 118: I suspect there is one situation that isn't handled by just looking at the deployment spec.

Consider the case where we upgrade the operator and that the new operator will change the driver deployment somehow, for example by adding a new `--foobar` parameter for one of the sidecars. This difference between the running driver deployment and the one that the new operator would create is not detected, is it?

What should happen is that the operator should update the objects that it creates such that they match exactly what it would create from scratch. The simplest (?) solution would be to just do that unconditionally each time the operator starts. The apps operator can then check whether it really needs to restart pods.

This logic hinges on both operator versions creating exactly the same objects. At one point me might add some new object (for example, a service) that the old operator doesn't know about. In case of a downgrade, it then would leave that service object unchanged although it's not supposed to be part of the downgraded driver deployment.

I don't have a good solution for this because there is no generic "list all objects with certain labels" API in Kubernetes. One has to do that for all relevant object types, and "relevant" is unknown for a downgrade (the older operator would need to know what will be added in the future). However, we can implement this for an upgrade because then we know all types that may have been created in previous versions.

avalluri: > I suspect there is one situation that isn't handled by just looking at the deployment spec.
> Consider the case where we upgrade the operator and that the new operator will change the driver deployment somehow, for example by adding a new --foobar parameter for one of the sidecars. This difference between the running driver deployment and the one that the new operator would create is not detected, is it?

Wouldn't be better to separate normal operator restart from version upgrade? say for version upgrades we could deal by adding an annotation to Deployment object, so on recocile we compare versions and decide if needs to re-deploy a Deployment.

avalluri: On re-deployment we might loose find and reject if any conflicting deployment changes. Say, driverMode/pmemPercentage changed when the operator is not running.

pohly: So for invariant attributes we must still check the running driver deployment and update the deployment object accordingly. But then we can simply update all generated objects.

pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
```diff
@@ -469,6 +486,87 @@ var _ = Describe("Operator", func() {
 			Expect(s.Data[corev1.TLSPrivateKeyKey]).Should(Equal(encodedKey), "mismatched private key")
 		})
 
+		It("shall use provided privatekeys and certificates", func() {
```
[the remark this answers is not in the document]
avalluri: Done

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -264,6 +275,22 @@ func (d *PmemCSIDriver) initDeploymentSecrests(r *ReconcileDeployment) error {
 	return nil
 }
 
+func validateCertificate(encodedCert []byte, certType, commonName string) error {
```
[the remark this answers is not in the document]
avalluri: Added the check to verify that the provided keys/certificates are valid by initiating a client(with nod key/certificate) connection to the server(with registry key/certificate) get succeeds.

pkg/pmem-csi-operator/pmem-tls/tls.go
```diff
@@ -131,53 +132,65 @@ func (ca *CA) EncodedCertificate() []byte {
...
 
-// GenerateCertificate returns a new certificate signed for public key of given private key.
-func (ca *CA) GenerateCertificate(cn string, key *rsa.PrivateKey) (*x509.Certificate, error) {
+// GenerateCertificate returns a new certificate signed for given public key.
+func (ca *CA) GenerateCertificate(cn string, key crypto.PublicKey) (*x509.Certificate, error) {
+	return ca.generateCertificate(cn, ca.cert.NotBefore, time.Now().Add(time.Hour*24*365), key)
+}
+
+// GenerateCertificateWithDuration returns a new certificate signed for given public key.
+// The duration of this certificate is with in the given notBore and notAfter bounds.
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/pmem-tls/tls.go
```diff
@@ -131,53 +132,65 @@ func (ca *CA) EncodedCertificate() []byte {
...
+// Intented use this API is only by tests
```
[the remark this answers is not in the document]
avalluri: Fixed.

operator/README.md
```diff
@@ -72,7 +72,11 @@ $ kubectl create -f https://github.com/intel/pmem-csi/raw/operator/operator/depl
 | nodeSelector | string map | [Labels to use for selecting Nodes](../README.md#run-pmem-csi-on-kubernetes) on which PMEM-CSI driver should run. | `{ "storage": "pmem" }`|
 | pmemPercentage | integer | Percentage of PMEM space to be used by the driver on each node. This is only valid for a driver deployed in `lvm` mode. This field cannot be changed of a running deployment. | 100 |
 
-<sup>1</sup> Image versions depend on the Kubernetes cluster version. The operator figures out itself the appropriate image version(s). Whereas users have to handle choosing the right version(s) themselves when overriding the values.
+<sup>1</sup> To use the same container image as default driver image the operator pod must set with below environment variables with appropriate values:
+ - POD_NAME: Name(´metadata.name`) of the operator pod
```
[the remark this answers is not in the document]
avalluri: Fixed.

operator/README.md
```diff
@@ -72,7 +72,11 @@ $ kubectl create -f https://github.com/intel/pmem-csi/raw/operator/operator/depl
...
+ - OPERATOR_NAME: Name of the operator container
```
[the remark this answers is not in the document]
avalluri: Added statements about the optional and operator namespace.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -128,7 +133,14 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}()
 
-	if err := deployment.EnsureDefaults(); err != nil {
+	driverImage, err := r.ContainerImage()
+	if err != nil {
+		klog.Warningf("failed to find container image: %v", err)
+	} else {
+		klog.Infof("Using driver image: %s", driverImage)
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -194,3 +206,37 @@ func (r *ReconcileDeployment) Update(obj runtime.Object) error {
...
+		// Operator container name shall be provided by deployment
+		operatorName := os.Getenv(operatorNameEnv)
+		if operatorName == "" {
+			operatorName = "pmem-csi-operator"
+		}
+		for _, c := range pod.Spec.Containers {
+			if c.Name == operatorName {
+				r.containerImage = c.Image
+			}
+		}
```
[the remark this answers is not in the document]
avalluri: Covered this case by returning appropriate error.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -194,3 +206,37 @@ func (r *ReconcileDeployment) Update(obj runtime.Object) error {
...
+	} else {
+		return "", fmt.Errorf("failed find operator name. Please set %q environment variable", podNameEnv)
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/main.go
```diff
@@ -113,6 +114,24 @@ func Main() int {
...
+
+	activeList := []string{}
+	for _, d := range list.Items {
+		if d.DeletionTimestamp == nil {
+			activeList = append(activeList, d.Name)
+		}
+	}
+
+	if len(activeList) != 0 {
+		klog.Infof("There are active pmem-csi deployments(%v), hence not deleting the CRD.", activeList)
```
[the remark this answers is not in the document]
avalluri: Fixed.

test/e2e/operator/deployment_api.go
```diff
@@ -464,6 +464,40 @@ var _ = DescribeForAll("pmem-csi", func(o *Operator, f *framework.Framework) {
...
+
+			defer deleteDeployment(f, deployment.Name)
+			validateDriverDeployment(f, o, deployment)
+
+			c, err := deploy.NewCluster(f.ClientSet)
+			Expect(err).ShouldNot(HaveOccurred(), "create new cluster")
+
+			By("Deleting operator")
+			// Stop the operator
+			EnsureOperatorRemoved(c, o)
```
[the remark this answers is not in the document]
avalluri: Did the changes to delete only operator Deployment object without deleting the roles/role bindings. This avoids the case of the CRD exists just because of lack of permissions to delete.

test/e2e/operator/deployment_api.go
```diff
@@ -464,6 +464,40 @@ var _ = DescribeForAll("pmem-csi", func(o *Operator, f *framework.Framework) {
...
+
+			By("validating driver post operator delete")
+			validateDriverDeployment(f, o, deployment)
```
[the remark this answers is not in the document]
avalluri: Modified the validation to be consistent for 1m. The Deployment should be deleted by `defer` at the end of test

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -103,6 +106,17 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 		// Deployment successfull, so no more reconcile needed for this deployment
 		return false, nil
 	case api.DeploymentPhaseRunning:
+		if !foundInCache {
+			// Possible that operator restarted, so we can't assume that
+			// everything upto date. Look at running deployment if any changes
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -103,6 +106,17 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
...
+			// needed
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -103,6 +106,17 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 		// Deployment successfull, so no more reconcile needed for this deployment
 		return false, nil
 	case api.DeploymentPhaseRunning:
+		if !foundInCache {
+			// Possible that operator restarted, so we can't assume that
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Made the changes as suggested, so now deployment reconcile on operator restart refreshes the objects and checks if any incompatible changes to be rejected.

avalluri:
@pohly I believe I addressed all the points you mentioned. Can you please have a look, If it's ok to merge.

test/e2e/operator/deployment_api.go
```diff
@@ -721,3 +756,35 @@ func findOperatorDeployment(f *framework.Framework) (*appsv1.Deployment, error)
...
+	dep, err := findOperatorDeployment(f)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	framework.Logf("Deleting the operator '%s/%s' running", dep.Namespace, dep.Name)
+	if err = f.ClientSet.AppsV1().Deployments(dep.Namespace).Delete(dep.Name, nil); err != nil {
```
pohly, line 771: It'll be simpler to just repeat the `Delete` call until you get a `NotFound` error. As an added bonus, the code becomes more robust against temporary delete failures.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -106,7 +109,18 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 		// Deployment successfull, so no more reconcile needed for this deployment
 		return false, nil
 	case api.DeploymentPhaseRunning:
-		requeue, err := d.reconcileDeploymentChanges(r, oldDeployment, changes)
+		if !foundInCache {
+			// Possibly that operator restarted, so we can't assume that
+			// everything is upto date. Check running deployment to determine
```
pohly, line 114: "up-to-date"

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -116,7 +130,10 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 	return true, nil
 }
 
-func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, existing *api.Deployment, changes map[api.DeploymentChange]struct{}) (bool, error) {
+// reconcileDeploymentChanges examines the changes and updates the appropriate objects.
+// Reverts deployment if and incompatible found compareted to existing deployment, such as deviceMode, driverName, pmemSpace.
+// In case of foundInCache is false, all the objects gets refreshed/updated.
+func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, existing *api.Deployment, changes map[api.DeploymentChange]struct{}, foundInCache bool) (bool, error) {
```
pohly, line 136: Please create an issue about another gap: `reconcileDeploymentChanges` must also delete obsolete sub-resources.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -179,18 +203,29 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, exist
...
 	}
 
-	if updateController {
-		klog.Infof("Updating controller driver for deployment %q", d.Name)
-		if e := r.Update(d.getControllerStatefulSet()); e != nil {
-			requeue = true
-			err = e
+	// Force update all deployment objects
+	if updateAll {
+		klog.Infof("Updating all objects for deployment %q", d.Name)
```
pohly, line 208: Please add a test for this case. A thorough test ensures that all objects are slightly different than they should be before the operator starts and afterwards pass the `validateCSIDriver` function.

That function itself is not as thorough as it should be (only checks some fields, but not everything); you can ignore that for now, I am working on it.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -179,18 +203,29 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, exist
...
+		for _, obj := range d.getDeploymentObjects(r) {
+			if e := r.Update(obj); e != nil {
```
pohly, line 210: This doesn't update the secrets. Let's merge https://github.com/intel/pmem-csi/pull/586 and the `getDeploymentObjects` truly returns *all* objects.

pohly: Another thought: what happens if the object doesn't exist because the operator we update from didn't create it?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -1065,3 +1100,133 @@ func joinMaps(left, right map[string]string) map[string]string {
 	}
 	return result
 }
+
+// buildFromPreDeployed builds a new Deployment object by fetching details
+// from existing deplyment sub-resources.
+func (d *PmemCSIDriver) buildFromPreDeployed(r *ReconcileDeployment) (*api.Deployment, error) {
```
pohly, line 1106: Same here: if an object cannot be found, we shouldn't treat that as an error because otherwise the driver deployment gets stuck forever.

Instead, the operator must recover by using the desired deployment parameters for those objects which don't exist, then later create them.

Needs a test...

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Secrets update is a bit tricky, With the current code, for a deployment with no keys/certificates set, It creates a new set of secrets for every call.

Strictly speaking of this function, it only deals with the objects we created in the Initializing phase, so this will not/need not update the secrets.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -128,7 +133,14 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}()
 
-	if err := deployment.EnsureDefaults(); err != nil {
+	driverImage, err := r.ContainerImage()
+	if err != nil {
+		klog.Warningf("Failed to find the operator image: %v", err)
```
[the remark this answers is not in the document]
avalluri: If we want to treat this as error then, we should move this call to the operator initialization phase, instead of reconciling loop. Ok, i will do the needed changes.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > Another thought: what happens if the object doesn't exist because the operator we update 
> from didn't create it?

To handle such cases we could change the semantics of Update()/Create() such that it creates if not exists else it updates.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: I don't follow. We need to update all objects only if we found an existing deployment in an unknown state. Why would that then get triggered for every call of the function?

avalluri: We end up in this function, only when the deployment is in "Running" phase. So for that deployment secretes must have already created and needs no update. And moreover, secrets creation is not idempotent in case of a deployment using operator generated keys/certificates.

test/e2e/deploy/deploy.go
```diff
@@ -136,7 +136,7 @@ func WaitForPMEMDriver(c *Cluster, namespace string) {
 		}
 		label := buildMetric.Label[0]
 		if *label.Name != "version" {
-			return fmt.Errorf("expected build_info to contain a version label, got: %s", label.Name)
+			return fmt.Errorf("expected build_info to contain a version label, got: %s", *label.Name)
```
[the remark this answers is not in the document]
avalluri: This code was not written by me, I was just trying to fix a warning raised by linter. :-)

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > Same here: if an object cannot be found, we shouldn't treat that as an error because otherwise the driver deployment gets stuck forever.
> Instead, the operator must recover by using the desired deployment parameters for those objects which don't exist, then later create them.

I believe this is how the current implementation of this function treats, it doesn't return any error for an object that was not found. Can you please be specific which part of the code does not follow this.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: > We end up in this function, only when the deployment is in "Running" phase.

Which is a problem. We need to update stale objects regardless of the phase.

Let's do this:
- review and merge https://github.com/intel/pmem-csi/pull/586 because it simplifies where we create objects
- remove the  "operator: reload state of a running deployment if not found in cache" from this PR because the rest should be okay
- write E2E test cases for reconciling on startup, then check whether the new code passes those tests

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -193,24 +193,24 @@ func (r *ReconcileDeployment) Get(obj runtime.Object) error {
...
+	metaObj, _ := meta.Accessor(obj)
+	err := r.Get(objCopy)
 	if err == nil {
-		// Already found an active object
-		return nil
+		// FIXME(avalluri): Check the ownership of the object(?) to make sure
+		// that was created by the same deployment
+		klog.Infof("Updating: %q of type %q ", metaObj.GetName(), obj.GetObjectKind().GroupVersionKind())
+		// Update existing active object,
+		return r.client.Update(context.TODO(), obj)
```
pohly, line 206: I doubt that the apiserver will accept the update because the resourceVersion won't match.

Write an E2E test for this and we'll know...

pkg/pmem-csi-operator/pmem-tls/tls.go
```diff
@@ -131,53 +132,65 @@ func (ca *CA) EncodedCertificate() []byte {
 	return EncodeCert(ca.cert)
 }
 
-// GenerateCertificate returns a new certificate signed for public key of given private key.
-func (ca *CA) GenerateCertificate(cn string, key *rsa.PrivateKey) (*x509.Certificate, error) {
+// GenerateCertificate returns a new certificate signed for given public key.
+func (ca *CA) GenerateCertificate(cn string, key crypto.PublicKey) (*x509.Certificate, error) {
+	return ca.generateCertificate(cn, ca.cert.NotBefore, time.Now().Add(time.Hour*24*365), key)
```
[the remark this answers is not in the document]
avalluri: Done. https://github.com/intel/pmem-csi/issues/594

pkg/pmem-grpc/grpc.go
```diff
@@ -25,12 +25,15 @@ func unixDialer(addr string, timeout time.Duration) (net.Conn, error) {
...
-func Connect(endpoint string, tlsConfig *tls.Config) (*grpc.ClientConn, error) {
+func Connect(endpoint string, tlsConfig *tls.Config, dialOptions ...grpc.DialOption) (*grpc.ClientConn, error) {
 	proto, address, err := parseEndpoint(endpoint)
 	if err != nil {
 		return nil, err
 	}
-	dialOptions := []grpc.DialOption{grpc.WithBackoffMaxDelay(time.Second)}
+	if dialOptions == nil {
+		dialOptions = []grpc.DialOption{}
+	}
```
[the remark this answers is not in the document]
avalluri: Done

pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
```diff
@@ -469,6 +485,87 @@ var _ = Describe("Operator", func() {
 			Expect(s.Data[corev1.TLSPrivateKeyKey]).Should(Equal(encodedKey), "mismatched private key")
 		})
 
+		It("shall use provided private keys and certificates", func() {
+			ca, err := pmemtls.NewCA(nil, nil)
+			Expect(err).Should(BeNil(), "faield to instantiate CA")
+			// Generate private keys and certificates
+			// CASE 1 - all valid data
```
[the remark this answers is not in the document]
avalluri: Done.

pkg/pmem-grpc/grpc.go
```diff
@@ -70,14 +72,33 @@ func NewServer(endpoint string, tlsConfig *tls.Config) (*grpc.Server, net.Listen
 		return nil, nil, err
 	}
 
-	opts := []grpc.ServerOption{grpc.UnaryInterceptor(pmemcommon.LogGRPCServer)}
+	if opts == nil {
+		opts = []grpc.ServerOption{}
+	}
```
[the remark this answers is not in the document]

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
avalluri: Done, As per now new implementation, the operator exists if fail to get self-image.

operator/README.md
```diff
@@ -63,9 +63,9 @@ The current API for PMEM-CSI `Deployment` resources is:
...
-| provisionerImage | string | [CSI provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) docker image name | latest [external provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) stable release image<sup>1</sup> |
-| registrarImage | string | [CSI node driver registrar](https://github.com/kubernetes-csi/node-driver-registrar) docker image name | latest [node driver registrar](https://kubernetes-csi.github.io/docs/node-driver-registrar.html) stable release image<sup>1</sup> |
+| image | string | PMEM-CSI docker image name used for the deployment | the same image as the operator<sup>1</sup> |
+| provisionerImage | string | [CSI provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) docker image name | latest [external provisioner](https://kubernetes-csi.github.io/docs/external-provisioner.html) stable release image<sup>2</sup> |
+| registrarImage | string | [CSI node driver registrar](https://github.com/kubernetes-csi/node-driver-registrar) docker image name | latest [node driver registrar](https://kubernetes-csi.github.io/docs/node-driver-registrar.html) stable release image<sup>2</sup> |
```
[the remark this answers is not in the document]
avalluri: Done.

test/e2e/operator/deployment_api.go
avalluri: Done. Instead of deleting deployment, making its replication count to zero and waiting till the operator pod gets deleted in a repeated loop.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: I believe I fixed really this time :-)

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Added a test that:
  * stats a deployment with defaults
  * removes operator pod
  * updates all deployment fields
  * restarts pod
  * wait till the deployment timestamp gets updated and ensures that all new values are applied to its sub-objects.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: Yes, true. I fixed this by choosing relevant metadata from the object returned by the API server to our new object before update.

avalluri:
I suspect the test failure is related to #593. @pohly Can you please check if it's the real issue or something I messed up in my tests.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Done. https://github.com/intel/pmem-csi/issues/595

pohly:
> operator: do not requeue reconcile on driver name clash 

That commit argues that reconciliation can be stopped until the user updates the deployment spec. But what if the user resolves the conflict by removing the other driver instance? Then the deployment becomes usable without updating it.

pohly:
> operator: fix test failures

Which failure is that fixing? Did some other commit change the behavior? Then this commit should be squashed into that other one.

pohly:
> I suspect the test failure is related to #593.

Did you change something? The latest test run was successful.

avalluri:
All of the commits except(https://github.com/intel/pmem-csi/pull/584/commits/146cde6b7bedafa27c46cfcd093bce83ea6807e9) are merged through other PRs, so I would close this PR.

PR #586, opened 2020-04-16
operator: simplify object creation

pohly:
Spreading object creation across phases is unnecessarily complex and would have to be changed anyway when introducing the "is driver running" check.

While working on this I noticed that controller test errors were attributed to the helper functions instead of the actual call site where the assertion was made, so I changed that, too.

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -70,34 +70,15 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 			return true, fmt.Errorf("driver name %q is already taken by deployment %q",
 				dep.Spec.DriverName, dep.Name)
 		}
-		r.deployments[d.Name] = d.Deployment
-
-		if err := d.initDeploymentSecrests(r); err != nil {
-			d.Status.Phase = api.DeploymentPhaseFailed
-			return true, err
-		}
-		d.Status.Phase = api.DeploymentPhaseInitializing
-	case api.DeploymentPhaseInitializing:
-		// There might be change in deployment name
-		// Reject if any incompatible name change
-		if _, ok := changes[api.DriverName]; ok {
-			if dep := d.checkIfNameClash(r); dep != nil {
-				// Revert Deployment object
-				d.Spec.DriverName = oldDeployment.Spec.DriverName
-				if err := r.Update(d.Deployment); err != nil {
-					return true, err
-				}
-				return true, fmt.Errorf("driver name %q is already taken by deployment %q",
-					oldDeployment.Spec.DriverName, dep.Name)
-			}
-		}
-
 		// Update local cache so that we can catch if any change in deployment
 		r.deployments[d.Name] = d.Deployment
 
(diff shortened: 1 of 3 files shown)


pohly:
@avalluri this passed testing, can we merge?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -194,8 +175,21 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, exist
...
+	for _, obj := range objects {
+		if err := r.Create(obj); err != nil {
+			return err
+		}
+	}
+	return nil
+}
 
+// getDeploymentObjects returns all objects that are part of a driver deployment.
+func (d *PmemCSIDriver) getDeploymentObjects(r *ReconcileDeployment) ([]runtime.Object, error) {
```
avalluri, line 192: @pohly By adding secrets creation to `getDeploymentObjects`, calling this method more than once creates new keys/certificates for every call in case the deployment chooses to use defaults certificates.

pohly: Is that a bad thing? It would ensure that those certificates are always fresh.

We haven't sorted out how to handle periodic updates of those certificates and the current code also doesn't have your "populate deployment from objects" code yet. Let's figure out how to avoid updating the secrets too often later, okay?

avalluri: > Is that a bad thing? It would ensure that those certificates are always fresh.
> We haven't sorted out how to handle periodic updates of those certificates and the current code also doesn't have your "populate deployment from objects" code yet. Let's figure out how to avoid updating the secrets too often later, okay?

That depends on how often we call this method, At least I could think of on every restart of operator pod. Updating secrets also lead to the restart of all driver pods.

pohly: > That depends on how often we call this method, At least I could think of on every restart of operator pod.

Which isn't very often. And even if it was called more often, why would that be a reason to be concerned?

> Updating secrets also lead to the restart of all driver pods.

No, it doesn't: https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically

It's still not perfect because the certificates may get rolled out on different nodes by kubelet at different times, which leads to a time window where controller and node cannot communicate. When designing the certificate update feature we need to take that into account, for example by keeping old and new CA root as trusted for a certain period of time.

For now I think it is okay that the system eventually recovers.

avalluri: > Which isn't very often. And even if it was called more often, why would that be a reason to be concerned?
> 
Ok, If you feel it's not an issue.
 
> No, it doesn't: https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically
> It's still not perfect because the certificates may get rolled out on different nodes by kubelet at different times, which leads to a time window where controller and node cannot communicate. When designing the certificate update feature we need to take that into account, for example by keeping old and new CA root as trusted for a certain period of time.
> For now I think it is okay that the system eventually recovers.

Interesting, Thanks for sharing.

PR #588, opened 2020-04-16
update to Kubernetes 1.18.1

pohly:
This ensures that we have all the latest fixes and enhancements.


pohly:
One additional benefit is that our runtime binaries no longer depend on `k8s.io/kubernetes`. This is architecturally cleaner.

okartau:
what about trying this commit in combination of #561 where we run on Fedora with 1.18 deployed_

test/e2e/e2e.go
```diff
@@ -74,7 +75,7 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
 			framework.Failf("Error deleting orphaned namespaces: %v", err)
 		}
 		framework.Logf("Waiting for deletion of the following namespaces: %v", deleted)
-		if err := framework.WaitForNamespacesDeleted(c, deleted, framework.NamespaceCleanupTimeout); err != nil {
+		if err := framework.WaitForNamespacesDeleted(c, deleted, 15*time.Minute); err != nil {
```
okartau, line 78: Move wrong way, from configured value to hard-coded value, why is that?

test/e2e/storage/scheduler/scheduler.go
```diff
@@ -151,18 +152,17 @@ func (l local) testSchedulerInPod(
...
+	e2epod.SetNodeSelection(&pod.Spec, config.ClientNodeSelection)
 
 	By(fmt.Sprintf("Creating pod %s", pod.Name))
 	ns := f.Namespace.Name
 	podClient := f.PodClientNS(ns)
 	createdPod := podClient.Create(pod)
 	defer func() {
 		By("delete the pod")
-		podClient.DeleteSync(createdPod.Name, &metav1.DeleteOptions{}, framework.DefaultPodDeletionTimeout)
+		podClient.DeleteSync(createdPod.Name, metav1.DeleteOptions{}, framework.DefaultPodDeletionTimeout)
```
okartau, line 165: what is reason here, is it fix to problem, or something else?

test/e2e/tls/tls.go
```diff
@@ -79,7 +80,7 @@ func checkTLS(f *framework.Framework, server string) {
 	createdPod := podClient.Create(pod)
 	defer func() {
 		By("delete the pod")
-		podClient.DeleteSync(createdPod.Name, &metav1.DeleteOptions{}, framework.DefaultPodDeletionTimeout)
+		podClient.DeleteSync(createdPod.Name, metav1.DeleteOptions{}, framework.DefaultPodDeletionTimeout)
```
okartau, line 83: what is reason here, is it fix to problem, or something else? change is other way as similar change above.

test/e2e/e2e.go
pohly: `NamespaceCleanupTimeout` is no longer getting exported.

test/e2e/tls/tls.go
pohly: An API change in client-go.

pohly:
> what about trying this commit in combination of #561 where we run on Fedora with 1.18 deployed

The current test configurations should give us enough confidence to merge this already now. There's no API change between client and apiserver; all the changes are in the client API.

PR #600, opened 2020-04-19
Repair deployments

pohly:
This is useful for PR #584.


pohly:
https://github.com/intel/pmem-csi/pull/605 no longer needed this because it restarted the operator itself.

I still think this is a useful enhancement. @avalluri, what do you think?

avalluri:
> I still think this is a useful enhancement. @avalluri, what do you think?

Ok, let's take it.

PR #597, opened 2020-04-19
operator: Validating deployment Certificates and Driver mode

avalluri:
Validating provided certificates and device modes are valid.

PR #601, opened 2020-04-19
test: relax TLS error check even further

pohly:
Yet another error text appeared in CI testing ("rpc error: code =
Unavailable desc = write unix @->pmem-registry.sock: write: broken
pipe").

Instead of checking for specific error texts, all "rpc error: code =
Unavailable" errors are now accepted.


avalluri:
I took this change into #597, which was affected by this issue. Let's wait until the result of #597.

pohly:
> I took this change into #597, which was affected by this issue. Let's wait until the result of #597.

A successful test run of #597 will not provide us any additional information because the failure is random. If #597 now succeeds, it's either because the patch fixed it or we have been lucky...

Issue #555, opened 2020-03-26
Clear Linux: node registration stuck


pohly:
I just ran local E2E testing on latest Clear Linux (cluster and container image) and observed that bringing up PMEM-CSI got stuck because one out of three nodes failed to register itself:
```
I0326 10:21:35.853359       1 pmem-csi-driver.go:482] Failed to register: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.104.207.128:10000: i/o timeout", retrying after 10 seconds...
I0326 10:21:45.853504       1 pmem-csi-driver.go:477] Registering controller...
```

The other nodes were successful with that same address (after one failed attempt):
```
I0326 10:12:16.594333       1 pmem-csi-driver.go:482] Failed to register: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 10.104.207.128:10000: connect: connection refused", retrying after 10 seconds...
I0326 10:12:26.594723       1 pmem-csi-driver.go:477] Registering controller...
I0326 10:12:26.600413       1 grpc.go:134] VerifyPeerCertificate: CN=pmem-registry
I0326 10:12:26.602616       1 tracing.go:19] GRPC call: /csi.v1.Controller/ListVolumes
I0326 10:12:26.602635       1 tracing.go:20] GRPC request: {}
I0326 10:12:26.605626       1 controllerserver-node.go:337] ListVolumes
I0326 10:12:26.605640       1 tracing.go:25] GRPC response: {}
I0326 10:12:26.608769       1 pmem-csi-driver.go:488] Registration success
```

Networking really was broken, curl couldn't connect either. The network plugin on Clear is flannel.

okartau:
Now that pmem-csi-periodic reaches Clear stage after recent state-save fix, it can be observed that set of tests often fails, and the reason seems to be similar, one node fails to join  the cluster. The root cause is same as here?

pohly:
@okartau : in your observations, is it limited to Clear or does it also occur on Fedora?

Depending on that, perhaps changing the CNI plugin to the one which works more reliably may help? I tried to use Weaveworks also on Clear, but couldn't get it to work and gave up at that time.

okartau:
> is it limited to Clear or does it also occur on Fedora?

That is limited to Clear.
I can try "weave on clear" on a local instance where it's easier to debug.

[...]

PR #603, opened 2020-04-20
test: avoid running controller and node drivers on same node

pohly:
We haven't found a good solution for the networking issues on Clear
Linux (https://github.com/intel/pmem-csi/issues/555) that plague our
periodic builds.

As a workaround, ensuring that the controller runs on the master node
and thus never on the same node as a node driver avoids the
problematic case.

### DIFF
--- a/test/setup-deployment.sh
+++ b/test/setup-deployment.sh
@@ -113,12 +113,34 @@ patchesJson6902:
       version: v1
       kind: StatefulSet
       name: pmem-csi-controller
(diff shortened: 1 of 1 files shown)


pohly:
Before we merge, let's check that the full branch build also succeeds: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/clear-linux-network-workaround/

pohly:
Superseded by PR #604.

PR #626, opened 2020-05-04
idempotent ephemeral volumes

pohly:
Several fixes and tests for ephemeral inline volumes.

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -469,6 +469,11 @@ func (ns *nodeServer) createEphemeralDevice(ctx context.Context, req *csi.NodePu
 		return nil, status.Error(codes.InvalidArgument, "ephemeral inline volume parameters: "+err.Error())
 	}
 
+	// If the caller has use the heuristic for detecting ephemeral volumes, the flag won't
+	// be set. Fix that here.
+	ephemeral := parameters.PersistencyEphemeral
+	p.Persistency = &ephemeral
+
 	// Create new device, using the same code that the normal CreateVolume also uses,
 	// so internally this volume will be tracked like persistent volumes.
 	volumeID, _, err := ns.cs.createVolumeInternal(ctx, p, req.VolumeId,
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -273,8 +273,40 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 		vol = ns.cs.getVolumeByName(req.VolumeId)
 	}
 
+	// The CSI spec 1.2 requires that the SP returns NOT_FOUND
+	// when the volume is not known.  This is problematic for
+	// idempotent calls, in particular for ephemeral volumes,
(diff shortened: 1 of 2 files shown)


pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -274,7 +274,10 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	}
 
 	if vol == nil {
-		return nil, status.Errorf(codes.NotFound, "no volume found with volume id %q", req.VolumeId)
+		// May be an invalid call (wrong persistent volume) or an idempotent call for an ephemeral volume.
+		// We cannot be sure, so let's assume that because we don't have such a volume, it must
+		// have been unpublished and not return an error.
+		return &csi.NodeUnpublishVolumeResponse{}, nil
```
okartau, line 280: I understand logic of explained case, but we are still supposed to conform to CSI spec which is at
https://github.com/container-storage-interface/spec/blob/master/spec.md ?
This change creates a spec violation, as by spec CSI driver MUST return NOT_FOUND to NodeUnpublish call if it does not know about VolumeID. The spec does not cover ephemeral volumes or any special cases, that's just MUST.
If we change this code as here, a simple test sending  NodeUnpublish with non-existing VolumeID, will expose a spec violation. I also understand that if current code creates forever-loop, there is problem somewhere. The spec says that "Recovery: Caller MUST verify that the volume_id is correct and that the volume is accessible and has not been deleted before retrying with exponential back off." In there they tried to make sure that bad things should not happen, But ephemeral volume which changes volume lifecycle significantly, seems to change the game.

pohly: > The spec says that "Recovery: Caller MUST verify that the volume_id is correct and that the volume is accessible and has not been deleted before retrying with exponential back off."

I don't see what bad things could happen if we don't follow this. It only applies to a broken CO and a situation that should never happen, whereas idempotent removal of ephemeral volumes is a real issue.

It would have been nicer to avoid this, but that's out of our control.

pohly: I've filed https://github.com/kubernetes/kubernetes/issues/90752 against Kubernetes, perhaps this can be fixed in kubelet. But we should still merge this because it's uncertain if all Kubernetes versions will get that fix.

okartau: Thanks, I also think there is CSI spec violation on CO side if we see forever-loop. If we have to create that change in pmem-csi, lets do it, but we should add comment in both code and doc that (and why) we intentionally violate CSI spec, because if anyone compares that code and spec, there is chance of getting "spec conformance fix proposal" because we will have a clear bug in code against spec. Adding spec violations to work around other bugs/violations is slippery road that generally just leads to bigger amount of overall mess, After we add some, the next changes may build up on such "seems working well" state. We should not forget that this code will pass csi-test only because of missing test case for that part. But once someone (hypothetically) creates such test case then how we organize this? By excluding tests?

pohly: I've updated the code with more comments and another fix: if the code got interrupted between unmounting and deleting the volume, it returned early without deleting the volume.

PR #633, opened 2020-05-07
examples: memcached with and without warm restart

pohly:
Memcached is a good example because images are readily available, PMEM
can be used with and without persistency, and plain yaml files are
sufficient for deployment.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -387,6 +387,8 @@ String RunInBuilder() {
 */
 String SourceRepo() {
     // Content of CHANGE_FORK varies, see https://issues.jenkins-ci.org/browse/JENKINS-58450.
+    (! env.CHANGE_FORK) ?
+        "github.com/intel/pmem-csi" :
     env.CHANGE_FORK.matches('.*/.*') ?
         env.CHANGE_FORK :
         env.CHANGE_FORK + '/pmem-csi'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -206,7 +206,7 @@ pipeline {
 
         stage('docsite') {
             steps {
-                sh "${RunInBuilder()} ${env.BUILD_CONTAINER} make vhtml"
+                sh "${RunInBuilder()} ${env.BUILD_CONTAINER} env GITHUB_SHA=${GIT_COMMIT} GITHUB_REPOSITORY=${SourceRepo()} make vhtml"
                 publishHTML([allowMissing: false, alwaysLinkToLastBuild: false, keepAll: false, reportDir: '_output/html', reportFiles: 'index.html', reportName: 'Doc Site', reportTitles: ''])
             }
         }
@@ -380,6 +380,16 @@ String RunInBuilder() {
     --user `id -u`:`stat --format %g /var/run/docker.sock` \
     --workdir ${WORKSPACE} \
     "
+}
+
+/*
+ Returns <owner>/<repo> from which the code was built.
+*/
+String SourceRepo() {
+    // Content of CHANGE_FORK varies, see https://issues.jenkins-ci.org/browse/JENKINS-58450.
+    env.CHANGE_FORK.matches('.*/.*') ?
+        env.CHANGE_FORK :
+        env.CHANGE_FORK + '/pmem-csi'
 }
 
 void TestInVM(distro, distroVersion, kubernetesVersion, skipIfPR) {
--- a/examples/memcached.md
+++ b/examples/memcached.md
@@ -48,10 +48,10 @@ name and other parameters in the deployment can be modified with
 how one can change the namespace, volume size or add additional
 command line parameters:
 
-```sh
-mkdir -p my-memcached-deployment
+```console
+$ mkdir -p my-memcached-deployment
 
-cat >my-memcached-deployment/kustomization.yaml <<EOF
+$ cat >my-memcached-deployment/kustomization.yaml <<EOF
 namespace: demo
 bases:
  - github.com/intel/pmem-csi/deploy/kustomize/memcached/ephemeral
@@ -59,7 +59,7 @@ patchesStrategicMerge:
 - update-deployment.yaml
 EOF
 
-cat >my-memcached-deployment/update-deployment.yaml <<EOF
+$ cat >my-memcached-deployment/update-deployment.yaml <<EOF
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -81,24 +81,17 @@ EOF
 ```
 
 These commands then create the namespace and the deployment:
-```sh
-kubectl create ns demo
-kubectl create --kustomize my-memcached-deployment
-```
-
-The expected output is:
-```
+```console
+$ kubectl create ns demo
 namespace/demo created
+$ kubectl create --kustomize my-memcached-deployment
 service/pmem-memcached created
 deployment.apps/pmem-memcached created
 ```
 
 Eventually, there will be one pod with memcached running:
-```sh
-kubectl wait --for=condition=Ready -n demo pods -l app.kubernetes.io/name=pmem-memcached
-```
-
-```
+```console
+$ kubectl wait --for=condition=Ready -n demo pods -l app.kubernetes.io/name=pmem-memcached
 pod/pmem-memcached-96f6b5986-hdx6n condition met
 ```
 
@@ -108,21 +101,15 @@ use [`kubectl
 port-forward`](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/)
 to access the service. Run this command in one shell and keep it
 running there:
-```sh
-kubectl port-forward -n demo service/pmem-memcached 11211
-```
-
-```
+```console
+$ kubectl port-forward -n demo service/pmem-memcached 11211
 Forwarding from 127.0.0.1:11211 -> 11211
 Forwarding from [::1]:11211 -> 11211
 ```
 
 In another shell we can now use `telnet` to connect to memcached:
-```sh
-telnet localhost 11211
-```
-
-```
+```console
+$ telnet localhost 11211
 Trying ::1...
 Connected to localhost.
 Escape character is '^]'.
@@ -162,21 +149,14 @@ Connection closed by foreign host.
 
 The following command verifies the data was stored in a persistent
 memory data volume:
-```sh
-pod=$(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name})
-kubectl exec -n demo $pod grep 'I am PMEM.' /data/memcached-memory-file
-```
-
-```
+```console
+$ kubectl exec -n demo $(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name}) grep 'I am PMEM.' /data/memcached-memory-file
 Binary file /data/memcached-memory-file matches
 ```
 
 To clean up, terminate the `kubectl port-forward` command and delete the memcached deployment with:
-```sh
-kubectl delete --kustomize my-memcached-deployment
-```
-
-```
+```console
+$ kubectl delete --kustomize my-memcached-deployment
 service "pmem-memcached" deleted
 deployment.apps "pmem-memcached" deleted
 ```
@@ -212,11 +192,8 @@ This example can also be kustomized. It uses the [`pmem-csi-sc-ext4`
 storage class](/deploy/common/pmem-storageclass-ext4.yaml). Here we
 just use the defaults, in particular the default namespace:
 
-```sh
-kubectl apply --kustomize github.com/intel/pmem-csi/deploy/kustomize/memcached/persistent
-```
-
-```
+```console
+$ kubectl apply --kustomize github.com/intel/pmem-csi/deploy/kustomize/memcached/persistent
 service/pmem-memcached created
 statefulset.apps/pmem-memcached created
 ```
@@ -224,41 +201,29 @@ statefulset.apps/pmem-memcached created
 We can verify that memcached really does a warm restart by storing
 some data, removing the instance and then starting it again.
 
-```sh
-kubectl wait --for=condition=Ready pods -l app.kubernetes.io/name=pmem-memcached
-```
-
-Because we use a stateful set, the pod name is deterministic:
-```
+```console
+$ kubectl wait --for=condition=Ready pods -l app.kubernetes.io/name=pmem-memcached
 pod/pmem-memcached-0 condition met
 ```
 
-There is also a corresponding persistent volume:
-```sh
-kubectl get pvc
-```
+Because we use a stateful set, the pod name is deterministic.
 
-```
+There is also a corresponding persistent volume:
+```console
+$ kubectl get pvc
 NAME                                     STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS       AGE
 memcached-data-volume-pmem-memcached-0   Bound    pvc-bb2cde11-6aa2-46da-8521-9bc35c08426d   200Mi      RWO            pmem-csi-sc-ext4   5m45s
 ```
 
 First set a key using the same approach as before:
-
-```sh
-kubectl port-forward service/pmem-memcached 11211
-```
-
-```
+```console
+$ kubectl port-forward service/pmem-memcached 11211
 Forwarding from 127.0.0.1:11211 -> 11211
 Forwarding from [::1]:11211 -> 11211
 ```
 
-```sh
-telnet localhost 11211
-```
-
-```
+```console
+$ telnet localhost 11211
 Trying ::1...
 Connected to localhost.
 Escape character is '^]'.
@@ -281,35 +246,17 @@ Then scale down the number of memcached instances down to zero, then
 restart it. To avoid race conditions, it is important to wait for
 Kubernetes to catch up:
 
-```sh
-kubectl scale --replicas=0 statefulset/pmem-memcached
-```
-
-```
+```console
+$ kubectl scale --replicas=0 statefulset/pmem-memcached
 statefulset.apps/pmem-memcached scaled
-```
 
-```sh
-kubectl wait --for delete pod/pmem-memcached-0
-```
-
-```
+$ kubectl wait --for delete pod/pmem-memcached-0
 Error from server (NotFound): pods "pmem-memcached-0" not found
-```
 
-```sh
-kubectl scale --replicas=1 statefulset/pmem-memcached
-```
-
-```
+$ kubectl scale --replicas=1 statefulset/pmem-memcached
 statefulset.apps/pmem-memcached scaled
-```
 
-```sh
-kubectl wait --for=condition=Ready pods -l app.kubernetes.io/name=pmem-memcached
-```
-
-```
+$ kubectl wait --for=condition=Ready pods -l app.kubernetes.io/name=pmem-memcached
 pod/pmem-memcached-0 condition met
 ```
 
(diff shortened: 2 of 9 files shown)


examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+[`kustomize`](https://github.com/kubernetes-sigs/kustomize). Here's
+how one can change the namespace, volume size or add additional
+command line parameters:
+
+```
+$ mkdir -p my-memcached-deployment
+$ cat >my-memcached-deployment/kustomization.yaml <<EOF
+namespace: demo
+bases:
+ - ../deploy/kustomize/memcached/ephemeral
```
pohly, line 45: The implicit assumption here is that users check out the repo when following this example. Alternatively, once we have merged this we could use a `http://` URL that points into the repo, ideally with a specific commit hash.

In particular when reading this on our new docsite, a link into github *in the text* would be useful.

I think I'll try to add that separately.

examples/readme.rst
```diff
@@ -4,12 +4,16 @@ Application examples
 `Redis-pmem operator <redis-operator.md>`__
    Deploy a Redis cluster through the redis-operator using QEMU-emulated persistent memory devices
 
+`memcached with PMEM <memcached.md>__
```
pohly, line 7: Missing backtick. Will fix...

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+```
+$ kubectl wait --for=condition=Ready -n demo pods -l app.kubernetes.io/name=pmem-memcached
+pod/pmem-memcached-5f57bcfd58-qjz74 condition met
+$ pod=$(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name})
+$ kubectl port-forward -n demo $pod 11211 &
+$ telnet 127.0.0.1 11211
+Trying 127.0.0.1...
+Connected to 127.0.0.1.
+
+set demo_key 0 500000 10
```
eadamsintel, line 83: For someone not that familiar with memcached I had a hard time following which is a command and which is output. I talked to one of our content engineers for advice and the recommendation for readability would be to separate into different code blocks. Here is how he suggested doing it based on other docs he has done when commands and outputs are intermixed without a prompt.

```sh
set demo_key 0 500000 10
I am PMEM.
```
```sh
[Output]

STORED
```
```sh
get demo_key
```
```sh
[Output]

VALUE demo_key 0 10
I am PMEM.
END
```
```sh
quit
```

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+        - name: MEMCACHED_ARGS
+          value: --conn-limit=500
+      volumes:
+      - name: data-volume
+        csi:
+          volumeAttributes:
+            size: 500Mi
+EOF
+$ kubectl create ns demo
+$ kubectl create --kustomize my-memcached-deployment
```
eadamsintel, line 69: This might be redundant but I always find it helpful to have the commands to verify that something was created correctly.   Consider adding a `kubectl describe service -n demo` and `kubectl describe deployment -n demo` set of commands.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+Eventually, there will be one pod with memcached running which we can
+connect to with port forwarding through kubectl:
```
eadamsintel, line 73: Same comment as above. I find it helpful to know how to verify that something setup correctly. The next wait command ensures the pod is running, but I personally prefer something like `kubectl get pods -n demo`

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+Handling connection for 11211
+set demo_key 0 500000 10
+I am PMEM.
+STORED
+quit
+Connection closed by foreign host.
+$ kubectl scale --replicas=0 statefulset/pmem-memcached
+statefulset.apps/pmem-memcached scaled
+$ kubectl wait --for delete pod/$pod
+Error from server (NotFound): pods "pmem-memcached-0" not found
```
eadamsintel, line 175: Is there a race condition to worry about here?  My pods did not take that long to terminate so I don't understand the need for the wait commands.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+$ kill $kubectl
```
eadamsintel, line 176: I had opened up two terminal windows to try this out after I did the telnet so I could see some of the telnet commands. On the second window this command did not work. However, if I used `pkill -f kubectl*` that did work in either shell.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+service "pmem-memcached" deleted
+statefulset.apps "pmem-memcached" deleted
+$ kubectl delete pvc -l app.kubernetes.io/name=pmem-memcached
+persistentvolumeclaim "memcached-data-volume-pmem-memcached-0" deleted
+```
+
+We can verify that memcached really does a warm restart by storing some
+data, removing the instance and then starting it again.
+```
+$ kubectl apply -f deploy/kustomize/memcached/persistent/memcached-persistent.yaml
```
eadamsintel, line 153: I'd really consider breaking up this long code block into separate more manageable code blocks.  I found this a bit difficult to follow especially when telnet commands and output are intermixed.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+$ kubectl create --kustomize my-memcached-deployment
+```
+
+Eventually, there will be one pod with memcached running which we can
+connect to with port forwarding through kubectl:
+```
+$ kubectl wait --for=condition=Ready -n demo pods -l app.kubernetes.io/name=pmem-memcached
+pod/pmem-memcached-5f57bcfd58-qjz74 condition met
+$ pod=$(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name})
+$ kubectl port-forward -n demo $pod 11211 &
```
eadamsintel, line 78: When I did the above on Clear Linux, I had to hit enter twice to properly see my prompt. Otherwise, after the $ I had  `$ Forwarding from 127.0.0.1:11211 -> 11211` which looked like a command I was about to run.  Hitting enter brought me back to a normal looking prompt.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+I am PMEM.
+STORED
+get demo_key
+VALUE demo_key 0 10
+I am PMEM.
+END
+quit
+```
+
+This data really was stored in the data volume:
```
eadamsintel, line 93: The way this is written you act surprised that it works.  :)  Perhaps say "The following command verifies the data was stored in a persistent memory data volume"

Is there another way to verify that a persistent memory device actually was used? Maybe that dax check? If not that is fine. This is my first time using memcached so it it might be more obvious to an experienced user that a persistent memory volume is used.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+For this to work, the PMEM-CSI driver must have been deployed with
+`pmem-csi.intel.com` as driver name. No storage classes are
+needed. That name and other parameters in the deployment can be
+modified with
+[`kustomize`](https://github.com/kubernetes-sigs/kustomize). Here's
+how one can change the namespace, volume size or add additional
+command line parameters:
+
+```
+$ mkdir -p my-memcached-deployment
```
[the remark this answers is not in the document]
pohly: I'd rather change this so that the example works without having to check out code. The current ambiguity comes from me not quite knowing how to do that properly and thus ignoring the problem :grimacing: 

It can be done by replacing all file references with URLs. There are just two issues with that:
- the instructions will only work after merging (only a problem for reviewers, can be worked around)
- ideally, we want the versioned documentation for a release to refer to the versioned files, because otherwise we may end up with https://intel.github.io/pmem-csi/0.7/README.html pointing to files that no longer exist on the master branch (no solution yet)

It also would be nice if the URLs were actual URLs in HTML rendering, but that conflicts with putting them into code blocks.

examples/memcached.md
```diff
@@ -0,0 +1,197 @@
...
+$ kubectl create ns demo
+$ kubectl create --kustomize my-memcached-deployment
+```
+
+Eventually, there will be one pod with memcached running which we can
+connect to with port forwarding through kubectl:
+```
+$ kubectl wait --for=condition=Ready -n demo pods -l app.kubernetes.io/name=pmem-memcached
+pod/pmem-memcached-5f57bcfd58-qjz74 condition met
+$ pod=$(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name})
```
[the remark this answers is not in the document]
pohly: Command and output in separate blocks is indeed better. It's also what the Kubernetes documentation uses. The downside is that the overall text becomes longer but I suppose the extra clarity is worth it. I'll give it a try.

examples/memcached.md
pohly: Yes, there's a race without it. The stateful set might get scaled down and back up before the controller reacts to the change, in which case the original pod keeps running.

examples/memcached.md
pohly: `$kubectl` is only set in the window where `kubectl` was sent into the background. Let me try to make this simpler by running the port forwarding in a separate window, without putting it into the background.

examples/memcached.md
pohly: > The way this is written you act surprised that it works.

"Relieved" is close, but I get what you mean. :grin: The dax check doesn't help here, because memcached doesn't need dax semantic.

pohly:
@eadamsintel please have another look, ideally using the docsite page for the new example:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-633/Doc_20Site/examples/memcached.html

The URLs are valid and all commands can be executed as shown without having to check out any code.

Ignore that the page currently refers to my repo: that's intentional (the main repo doesn't have the files...) and will not be like that after merging (fingers crossed - I'll have to check that because this PR introduces that URL mangling).

examples/memcached.md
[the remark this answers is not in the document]
eadamsintel: I got the following script from Graham to help reviewers validate unmerged code. I even used it to test these files out. You do a `git clone` of the repo, change into the directory, and then run the script with the pull number and it checks out the files for you.  In this case I would `./pullpr.sh 633`

```
$ cat pullpr.sh
#!/bin/bash
# Do a git clone of the repository in question
# Change to root directory of repo
# Run ./pullpr.sh 445   where 445 is the PR number

if [ $# -ne 1 ]; then
        echo "Require one argument (PR number)"
        exit 1
fi

pr=${1}
pr_branch=PR_${pr}

echo "Fetching PR ${pr} to branch ${pr_branch}"
git fetch origin "refs/pull/$pr/head:$pr_branch"
git checkout $pr_branch
git branch
```

examples/memcached.md
```diff
@@ -0,0 +1,369 @@
...
+
+In this mode, the PMEM-CSI driver is referenced through its name
+(usually `pmem-csi.intel.com`). No storage classes are needed. That
+name and other parameters in the deployment can be modified with
+[`kustomize`](https://github.com/kubernetes-sigs/kustomize). Here's
+how one can change the namespace, volume size or add additional
+command line parameters:
+
+```sh
+mkdir -p my-memcached-deployment
```
eadamsintel, line 52: This is easier to read now.

examples/memcached.md
```diff
@@ -0,0 +1,369 @@
...
+      - name: data-volume
+        csi:
+          volumeAttributes:
+            size: 500Mi
+EOF
+```
+
+These commands then create the namespace and the deployment:
+```sh
+kubectl create ns demo
```
eadamsintel, line 85: I think for all the commands prepending with $ still makes sense.    `$ kubectl create ns demo`

examples/memcached.md
```diff
@@ -0,0 +1,369 @@
...
+```
+
+How to access a service from remote varies between cluster
+installations and how they handle incoming connections. Therefore we
+use [`kubectl
+port-forward`](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/)
+to access the service. Run this command in one shell and keep it
+running there:
+```sh
+kubectl port-forward -n demo service/pmem-memcached 11211
```
eadamsintel, line 112: I think combining the output with the command makes sense in this case if you prepend the command with $ so users know it is a command. It is the telnet session where you don't have the $ anymore where I think separating and indicating one is output makes sense. 

```
$ kubectl port-forward -n demo service/pmem-memcached 11211
Forwarding from 127.0.0.1:11211 -> 11211
Forwarding from [::1]:11211 -> 11211
```

examples/memcached.md
```diff
@@ -0,0 +1,369 @@
...
+
+```
+VALUE demo_key 0 10
+I am PMEM.
+END
+```
+
+To disconnect, use:
+```
+quit
```
eadamsintel, line 156: This is easier to follow now where one is a command and the other is output.

examples/memcached.md
```diff
@@ -0,0 +1,369 @@
...
+
+```
+Connection closed by foreign host.
+```
+
+The following command verifies the data was stored in a persistent
+memory data volume:
+```sh
+pod=$(kubectl get -n demo pods -l app.kubernetes.io/name=pmem-memcached -o jsonpath={..metadata.name})
+kubectl exec -n demo $pod grep 'I am PMEM.' /data/memcached-memory-file
```
eadamsintel, line 167: As I said before I think with these commands you can combine the output with the command as just put $ in front of the command.

examples/memcached.md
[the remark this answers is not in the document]
pohly: > I got the following script from Graham to help reviewers validate unmerged code.

There's also `hub checkout <PR URL>`. But as the goal is to make this example and the deployment files usable without people checking out code, it's better to also test the PR that way.

examples/memcached.md
pohly: The problem with that is that it then becomes harder to cut-and-paste the command. As it is now (without the $), I can double-click on the command block to select all of it and the directly paste. That even works for multiple commands in a single block.

Ideally, the markup on the docsite should make it clearer which blocks contain code and which contain output. @intelkevinputnam: any suggestions how we can improve that?

See https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-633/Doc_20Site/examples/memcached.html

In Kubernetes, this looks like this:
https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#creating-redis-deployment-and-service

intelkevinputnam: We addressed this problem with block labels in the [Clear Linux docs](https://docs.01.org/clearlinux/latest/guides/clear/mixer.html). I can do something similar with this. For example, I could put a "sh" label on input, and output could be labeled "output".

eadamsintel: You would be surprised how many times this discussion has come up in the docs meeting we have.  With GitHub Markdown you are limited in what you can do.  Almost all GitHub repo instructions I come across including kata all have the $ in front of the command.  It does make it harder to copy,  but it makes it more readable.  I've come to accept the $ because most people are doing instructions for the first or maybe second time I think they would prefer to do one command at a time.  I have my own personal way to remove all $ from a GitHub page that I am viewing since I expect it.

@intelkevinputnam At https://stackoverflow.com/questions/20303826/highlight-bash-shell-code-in-markdown I saw that "console" or "shell" can be used with code blocks and it supposedly colors the prompt. I've not tried it yet but that might make it even easier to read and follow.

intelkevinputnam: Using "ShellSession" in a fenced code block in markdown produces an interesting result when including $, however it doesn't fix the copy/paste problem.

What the markdown looks like.
````` md
``` ShellSession
$ make clean
Removing everything under '_build'...
$ make html
Running Sphinx v1.6.7
```
`````

What the output looks like: 
``` ShellSession
$ make clean
Removing everything under '_build'...
$ make html
Running Sphinx v1.6.7
```

It automatically colors the output differently from the input, keying off the inclusion of the $. Not sure how Sphinx/Read the Docs would handle it, though. I'll check.

eadamsintel: @intelkevinputnam I did some more digging and both   https://github.com/cri-o/cri-o and https://github.com/kubernetes/kubernetes skip using $ but https://github.com/containerd/containerd did use a $ to indicate bash code.  There is no consistency here so really it is whatever we want to do.

intelkevinputnam: I'd prefer to implement a solution that works equally well with the GitHub parser and Sphinx, unfortunately they use different lexers and the aliases for Shell sessions aren't the same. Markdown uses "ShellSession" and Sphinx uses "shell-session". So far, my preferred solution is to separate shell input and console output into different blocks and label them. I can split the difference fairly easily by adding a "$" to each line of a certain kind of code block via CSS that won't be picked up with select/copy.

intelkevinputnam: I take it back. Both GitHub markdown and Sphinx support the use of the alias "console" for a shell session to generate similar output with respect to using the $ character.

pohly: So what's the conclusion?

pohly: I have switched to using `console` and `$` because it is more consistent with the rest of the PMEM-CSI documentation (however, we still need to add `console` there).

IMHO, the increased readability is worth the loss of easy cut-and-paste:
- https://github.com/intel/pmem-csi/blob/a8041e3f3f5e355a688c2940d03869d273aaac97/examples/memcached.md
- https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-633/Doc_20Site/examples/memcached.html

eadamsintel: I skimmed through both of these and it does look a lot nicer to read.

intelkevinputnam: Looks good. I'm looking at an extension my colleague in Clear Linux docs found: sphinx_copybutton. We can add this extension and configure it with this:

copybutton_prompt_text = "$ "

This way the copy button that is appended to the code block will only pick up the commands and strip the "$" character.

intelkevinputnam: FYI, I just tested it on the current version of the Memcached page. Works like a charm. Block looks like this:
image
Clicking the copy button (top right) produces this:

mkdir -p my-memcached-deployment

cat >my-memcached-deployment/kustomization.yaml <<EOF

cat >my-memcached-deployment/update-deployment.yaml <<EOF

pohly: Unfortunately, `cat >my-memcached-deployment/kustomization.yaml <<EOF` is incomplete: the lines up to and including `EOF` also need to be copied.

Let's merge this PR as it is and continue investigating this copy button.

intelkevinputnam: Ha, ha. Of course. I did a little tinkering and came up with a way to make this work.  Just requires a few lines of code in conf.py and some javascript. If you want to keep all of the lines and still strip the "$", we can use the "ShellSession" lexer alias. This way nothing changes in pure markdown, and we can use "console" when we use it for alternating input/output.

Jenkinsfile
```diff
@@ -382,6 +382,16 @@ String RunInBuilder() {
     "
 }
 
+/*
+ Returns <owner>/<repo> from which the code was built.
+*/
+String SourceRepo() {
+    // Content of CHANGE_FORK varies, see https://issues.jenkins-ci.org/browse/JENKINS-58450.
+    env.CHANGE_FORK.matches('.*/.*') ?
```
pohly, line 390: `env.CHANGE_FORK` is nil when building *after* merging. Will fix.

examples/memcached.md
pohly: I merged the memcached PR.

@intelkevinputnam: can you follow up with a PR that adds the copy-and-paste support? Do the other examples also need better annotations like `console`?

intelkevinputnam: Copy/Paste support with the extension adds it to all code blocks, so nothing else required. I'll put together some more explanation in the PR.

PR #636, opened 2020-05-08
deploy: fix kustomize

pohly:
Commit f4b7d552d0aedb615cbed35fa9d29b5ce0cbe43f removed the conversion
of deploy/kustomize input files into the 1.14 output files without
introducing the creation of the 1.15 output files. As a result,
changing for example deploy/kustomize/driver/pmem-csi.yaml and then
running "make kustomize" no longer had an effect.

deploy/kubernetes-1.15/pmem-csi.yaml was added in
a2ea982e6ac0e7e1bde5ef2df88c8ae346672101, probably by mistake. It's
not a generated file and shouldn't be there.


pohly:
@okartau tests passed.

PR #637, opened 2020-05-11
pmem-csi-driver: load only volumes creates by the current device manager

### LINKED ISSUE #632: switching device mode
One thing that I noticed recently: we use `/var/lib/<csi driver name>` for persistent information about volumes. This probably breaks badly when an admin reinstalls the driver in a different mode while keeping the driver name the same.

Instead of telling admins "don't do that", can we make the driver more resilient against such a situation?

A good first step would be to add the device mode to the volume information, instead of assuming that the volume is for the current mode.

Then we can either ignore the "wrong" volumes or perhaps support *both* modes in the driver at the same time, at least for existing volumes? Instead of putting "device mode" into the deployment parameters, it would become a per-volume parameter.

We still need to figure out when to create volume groups, though.

Some work was done for this, but some work remains:
- When changing the device mode while a pod is using a volume, deleting the pod no longer works because the volume operations during the pod removal will fail. To recover, an admin has to revert back to the original device mode.

avalluri:
To support switching device manager save name of the device manager in
state file and consider only volumes that were created by current volume
manager for loading.

FIXES #632


pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -26,9 +26,10 @@ import (
 )
 
 type nodeVolume struct {
-	ID     string            `json:"id"`
-	Size   int64             `json:"size"`
-	Params map[string]string `json:"parameters"`
+	ID         string            `json:"id"`
+	DeviceMode string            `json:"deviceMode"`
+	Size       int64             `json:"size"`
+	Params     map[string]string `json:"parameters"`
```
pohly, line 32: I would prefer a solution where "deviceMode" is a new parameter. We then don't allow setting it anywhere except in the `NodeVolumeOrigin` for now.

Should we ever decide to make it a parameter that can be set also in a storage class, then we don't need any special handling for it.

avalluri: > Should we ever decide to make it a parameter that can be set also in a storage class, then we don't need any special handling for it.

Making the driver's internal device mode feature a storage parameter sounds me not the right solution unless we plan to support creating volumes with chosen mode by admin/user.

avalluri:
> We also need an E2E test for this. Implementing one as an operator test is relatively straight-forward:

Currently, this PR is against devel, shall I resend this to the operator branch with a E2E test?

pkg/pmem-csi-driver/controllerserver-node.go
pohly: > Making the driver's internal device mode feature

That's an implementation detail of our current release. It might change.

> a storage parameter

Conceptually, it could be both a parameter for the volume ("give me a volume in direct mode") as well as a deployment option ("I want to run PMEM-CSI in direct mode"). Here we are dealing with volumes, so treating it as a parameter seems fine to me.

> unless we plan to support creating volumes with chosen mode by admin/user

We might, which is exactly the point I was trying to make.

pohly:
> Currently, this PR is against devel, shall I resend this to the operator branch with a E2E test?

No. Let's merge the operator branch into devel first.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: > Conceptually, it could be both a parameter for the volume ("give me a volume in direct mode")

I couldn't get why we need to expose this detail to the end-user. If the driver deployed in "lvm" mode and the user asks to create a volume in "direct" mode? or vice-versa. What is the benefit he gets?

Still, If this is what we are targetting, I feel like there is no point in choosing the device mode while deploying the driver.  So in my view, it should be either storage parameter(if we support provisioning of volumes in both modes) or deployment API option but not both.

pohly: > I couldn't get why we need to expose this detail to the end-user.

If we as developers can't decide which mode is better, perhaps the admin cannot decide it either? It might depend on the application. Just a thought.

> I feel like there is no point in choosing the device mode while deploying the driver. So in my view, it should be either storage parameter(if we support provisioning of volumes in both modes) or deployment API option but not both.

Because we haven't decided about that, we now should pick an implementation that works for both.

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -74,6 +74,21 @@ func NewNodeControllerServer(nodeID string, dm pmdmanager.PmemDeviceManager, sm
...
+			// retrieve volume info
+			vol := &nodeVolume{}
+			if err := sm.Get(id, vol); err != nil {
+				klog.Warningf("Failed to retrieve volume info for id %q from state: %v", id, err)
+				continue
+			}
+
+			v, err := parameters.Parse(parameters.NodeVolumeOrigin, vol.Params)
+			if err != nil {
+				klog.Warningf("Failed to parse volume parameters for volume %q: %v", id, err)
```
pohly, line 86: It should also ignore volumes where it cannot parse the parameters, i.e. there's a "continue" (or something similar, see other comment) missing here. This will become relevant in the following scenario:
- create a volume with new parameters using a newer PMEM-CSI release
- downgrade to an older release which doesn't know about that parameter

The older release should not act upon that volume because the parameter might be crucial for correctly working with it.

BTW, this is a technical reason why extending parameters is better than a new field: the JSON parser will ignore unknown fields in JSON and then there is no way for this code here to detect that it doesn't have all information about the volume.

pkg/pmem-csi-driver/controllerserver-node.go
```diff
@@ -74,6 +74,21 @@ func NewNodeControllerServer(nodeID string, dm pmdmanager.PmemDeviceManager, sm
...
+			}
+			if v.GetDeviceMode() != dm.GetName() {
+				continue
```
pohly, line 89: I wonder whether we should really ignore the volume. For `NodePublishVolume` that's kind of okay (results in "not found" error, right?), albeit a bit confusing (the volume exists, it's just not usable at the moment).

But what about a repeated `CreateVolume` call (can happen when our initial response didn't get through to Kubernetes)? Is that still idempotent or will we end up with one volume created in direct mode and another in LVM mode for the same Kubernetes PVC?

pkg/pmem-csi-driver/parameters/parameters.go
```diff
@@ -114,6 +116,7 @@ type Volume struct {
 	Persistency *Persistency
 	Size        *int64
 	VolumeID    *string
+	DeviceMode  *string
```
pohly, line 119: Let's make this type-safe:
- merge the operator branch
- use `* api.DeviceMode` as type

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: > But what about a repeated `CreateVolume` call (can happen when our initial response didn't get through to Kubernetes)? Is that still idempotent or will we end up with one volume created in direct mode and another in LVM mode for the same Kubernetes PVC?

I believe, If we want to cover theses cases we should fully support volumes in multiple modes, not just that driver switching modes.

pkg/pmem-csi-driver/parameters/parameters.go
avalluri: I was keeping this intentionally as a string, because we get this value from `DeviceManager.GetName()` which returns a string.

pkg/pmem-csi-driver/controllerserver-node.go
pohly: If that's what it takes to prevent bad things from happening when admins make mistakes, then yes, we'll have to do that.

pkg/pmem-csi-driver/parameters/parameters.go
pohly: That also should be changed.

pkg/pmem-csi-driver/controllerserver-node.go
avalluri: > If that's what it takes to prevent bad things from happening when admins make mistakes, then yes, we'll have to do that.

This is a big design change, PMEM-CSI driver was not designed to support volumes with multiple device modes at a time.  And, I doubt how far the current code helps to switch modes. Say, if the driver deployed in `LVM` mode and then redeploying with `ndctl` mode is really useful as the space used for "lvm" could not be unitized by driver anymore. 

The intention of this change is: not to remove the volumes that were created in the previous mode from the state. That gives the admin to revert to the previous mode and still could access those volumes.

pkg/pmem-csi-driver/parameters/parameters.go
avalluri: Done.

avalluri:
> We also need an E2E test for this

@pohly I added the E2E test could you please review if it's good enough to merge.

test/e2e/operator/deployment_api.go
```diff
@@ -223,6 +227,175 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+				Provisioner:       driverName,
+				ReclaimPolicy:     &reclaim,
+				VolumeBindingMode: &immediate,
+			}
+
+			Eventually(func() error {
+				_, err := f.ClientSet.StorageV1().StorageClasses().Create(context.Background(), sc, metav1.CreateOptions{})
+				deploy.LogError(err, "create storage class error: %v, will retry...", err)
+				return err
+			}, "3m", "1s").Should(BeNil(), "create storage class %q", sc.Name)
```
pohly, line 272: For errors it is better to use `ShouldNot(HaveOccurred())` instead of `BeNil`. It's more specific and I think the output is better when it fails.

This also applies to the rest of the function.

test/e2e/operator/deployment_api.go
```diff
@@ -225,6 +229,136 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+				app := &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "switch-mode-app",
+						Namespace: corev1.NamespaceDefault,
+					},
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:            "test-driver",
+								Image:           "intel/pmem-csi-driver-test:canary",
```
pohly, line 252: This will literally pull that image, right?

Better use `os.Getenv("PMEM_CSI_IMAGE")`, that's the one that is already available.

test/e2e/operator/deployment_api.go
```diff
@@ -352,7 +486,137 @@ func startOperator(c *deploy.Cluster, d *deploy.Deployment) {
...
+	deployment := deploy.GetDeploymentCR(f, depName)
+	deployment.Spec.DeviceMode = mode
+	deployment = deploy.UpdateDeploymentCR(f, deployment)
+
+	waitForDriverReady(c, f, depName)
+
+	return deployment
+}
+
+// waitForDriverReady checks if all node driver pods are in Ready state
```
pohly, line 505: Can you use `WaitForPMEMDriver` instead (https://github.com/intel/pmem-csi/blob/7e2f199db11376f44efcdbb7245b4b9a7f81353e/test/e2e/deploy/deploy.go#L80)?

This looks like a less complete version of that function.

If `WaitForPMEMDriver` needs some additional parameters to be suitable, then just add them.

test/e2e/operator/deployment_api.go
```diff
@@ -352,7 +486,137 @@ func startOperator(c *deploy.Cluster, d *deploy.Deployment) {
...
+func waitForDriverReady(c *deploy.Cluster, f *framework.Framework, depName string) {
+	nWorkerNodes := int32(c.NumNodes()) - 1
+
+	// Ensure all the worker driver is ready
+	EventuallyWithOffset(1, func() int32 {
+		ds := c.WaitForDaemonSet(depName + "-node")
```
pohly, line 511: This isn't in `WaitForPMEMDriver`, but I also doubt that it works reliably: if you check immediately after changing the deployment, then it could be that either the PMEM-CSI or the DaemonSet controller have not reacted to the change and thus the DaemonSet is ready because nothing has changed yet.

test/e2e/operator/deployment_api.go
```diff
@@ -225,6 +229,136 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+
+						// Wait till a volume get provisioned for this claim
+						err := e2epv.WaitForPersistentVolumeClaimPhase(v1.ClaimBound, f.ClientSet, pvc.Namespace, pvc.Name, framework.Poll, framework.ClaimProvisionTimeout)
+						Expect(err).NotTo(HaveOccurred(), "Persistent volume claim bound failure")
+
+						pvc, err = f.ClientSet.CoreV1().PersistentVolumeClaims(pvc.Namespace).Get(context.Background(), pvc.Name, metav1.GetOptions{})
+						Expect(err).NotTo(HaveOccurred(), "failed to get updated volume claim: %q", pvc.Name)
+						framework.Logf("PVC '%s', Volume Ref: %s", pvc.Name, pvc.Spec.VolumeName)
+
+						// Switch to direct device mode
```
pohly, line 347: Comment is out-dated.

test/e2e/operator/deployment_api.go
```diff
@@ -225,6 +229,136 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
 		})
 	})
 
+	Context("switch device mode", func() {
+		postSwitchFuncs := map[string]func(from, to api.DeviceMode, depName string, pvc *corev1.PersistentVolumeClaim){
+			"delete volume": func(from, to api.DeviceMode, depName string, pvc *corev1.PersistentVolumeClaim) {
+				/* Try deleting Volume should fail */
```
pohly, line 235: Should *not* fail?

test/e2e/operator/deployment_api.go
avalluri: > If `WaitForPMEMDriver` needs some additional parameters to be suitable, then just add them.

Does the existing function(in general metrics logic) work even for the deployment update that resulted in the restart of the only deployment set?

pohly: It doesn't wait for a restart (that's what you need to deal with here), but once you have verified that the old controller pod is gone, `WaitForPMEMDriver` should work.

test/e2e/operator/deployment_api.go
avalluri: Fixed

test/e2e/operator/deployment_api.go
avalluri: Pushed a commit (https://github.com/intel/pmem-csi/pull/637/commits/114f71ef2fadbb4cf3757eccf58da65fe8ea50bd) to unregister a node on exit. So that, metrics from controller get updated on the daemon set controller restarts the node pods. This should make `WaitForPMEMDriver` work in case of driver updates only resulted in restart node pods.

test/e2e/operator/deployment_api.go
avalluri: Changed the code to wait till daemonset get updated before calling `WaitForPMEMDriver`.

test/e2e/operator/deployment_api.go
avalluri: Fixed.

test/e2e/operator/deployment_api.go
```diff
@@ -347,7 +482,147 @@ func startOperator(c *deploy.Cluster, d *deploy.Deployment) {
...
+
+	// Wait till the daemon set get updated
+	Eventually(func() bool {
+		ds, err := c.GetDaemonSet(dsName, corev1.NamespaceDefault)
+		deploy.LogError(err, "Failed to fetch daemon set: %v", err)
+		if err != nil {
+			return false
+		}
+		return ds.ObjectMeta.ResourceVersion != dsVersion
+	}, "3m", "1s").Should(BeTrue(), "Update daemonset")
```
pohly, line 516: This still has the race between "daemon set updated" and "pod restarted".

Better identify the controller pod with `GetAppInstance`, switch the mode, then wait for that pod to be gone with https://github.com/kubernetes/kubernetes/blob/ca7a906ab793c13f3def2d7ff0357ab88bfecf92/test/e2e/framework/pod/wait.go#L380

pohly: This didn't take into account that changing the driver mode will not restart the controller, right?

But the race still exists. The fix just becomes more complicated, because you will then have to identify all node pods and wait for those to disappear.

test/e2e/deploy/deploy.go
```diff
@@ -581,7 +581,7 @@ func EnsureDeployment(deploymentName string) *Deployment {
 				framework.Logf("reusing existing %s PMEM-CSI components", deployment.Name)
 				// Do some sanity checks on the running deployment before the test.
 				if deployment.HasDriver {
-					WaitForPMEMDriver(c, deployment.Namespace)
+					WaitForPMEMDriver(c, deployment.Name, deployment.Namespace)
```
pohly, line 584: `deployment.Name` is incorrect here, the name is "pmem-csi".

test/e2e/deploy/deploy.go
```diff
@@ -628,7 +628,7 @@ func EnsureDeployment(deploymentName string) *Deployment {
 			// We check for a running driver the same way at the moment, by directly
 			// looking at the driver state. Long-term we want the operator to do that
 			// checking itself.
-			WaitForPMEMDriver(c, deployment.Namespace)
+			WaitForPMEMDriver(c, deployment.Name, deployment.Namespace)
```
pohly, line 631: And also here.

test/e2e/deploy/deploy.go
pohly: I also had to make changes to `WaitForPMEMDriver` (see  https://github.com/intel/pmem-csi/pull/667/commits/588e887141fdf48c8770528f6b9b4f9557267d95). To avoid conflicts, I cherry-picked your change and fixed it in https://github.com/intel/pmem-csi/pull/667/commits/9e23107ffec1d3590a40bb2e16b758436b2d1f56.

If you include that commit in your branch, we may be able to merge both branches without conflicts.

avalluri: I picked your commit.

pohly:
Looks like there is now a volume leak somewhere. I haven't looked at it in detail.

avalluri:
> Looks like there is now a volume leak somewhere. I haven't looked at it in detail.

Yes, seems like the reason was missing /sys mounting in LVM mode:
```
[2020-06-04T16:57:03.214Z] direct-to-lmv-delete-volume-node-wwn9v/pmem-driver@pmem..ker3: I0604 16:57:02.776058       1 pmd-ndctl.go:37] NewPmemDeviceManagerNdctl: sysfs mount options:[ro nosuid nodev noexec relatime]
[2020-06-04T16:57:03.214Z] direct-to-lmv-delete-volume-node-wwn9v/pmem-driver@pmem..ker3: W0604 16:57:02.779025       1 controllerserver-node.go:94] Failed to initialize device manager for state volume 'pvc-a7-5f6129c2390cc82f7a5c24161395821d27e699dea963511b36c52a1a'(volume mode: 'direct'): FATAL: /sys mounted read-only, can not operate
```

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -317,6 +317,13 @@ func (pmemd *pmemDriver) Run() error {
 		// Here we want to shut down cleanly, i.e. let running
 		// gRPC calls complete.
 		klog.V(3).Infof("Caught signal %s, terminating.", sig)
+		if pmemd.cfg.Mode == Node {
+			klog.V(3).Info("Unregistering node...")
+			if err := pmemd.unregisterNodeController(); err != nil {
```
pohly, line 297: Can it happen that this attempt to unregister slows down the shutdown when the controller happens to exit first?

pohly: Yes, it does...

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi/branches/version-skew-2/runs/1/nodes/97/steps/201/log/?start=0

```
[2020-09-08T22:10:21.367Z] pmem-csi-controller-0/pmem-driver@pmem..ster: I0908 22:10:21.209248       1 pmem-csi-driver.go:296] Caught signal terminated, terminating.
[2020-09-08T22:10:21.367Z] pmem-csi-controller-0/pmem-driver@pmem..ster: I0908 22:10:21.215828       1 server.go:55] Server on 'unix:///csi/csi-controller.sock' stopped
[2020-09-08T22:10:21.367Z] pmem-csi-node-zrcfj/pmem-driver@pmem..ker2: I0908 22:10:21.218459       1 pmem-csi-driver.go:468] Connection state: CONNECTING
[2020-09-08T22:10:21.367Z] pmem-csi-node-zrcfj/pmem-driver@pmem..ker2: I0908 22:10:21.223549       1 pmem-csi-driver.go:468] Connection state: TRANSIENT_FAILURE
[2020-09-08T22:10:21.367Z] pmem-csi-node-mvvzx/pmem-driver@pmem..ker1: I0908 22:10:21.225583       1 pmem-csi-driver.go:468] Connection state: CONNECTING
[2020-09-08T22:10:21.367Z] pmem-csi-controller-0/pmem-driver@pmem..ster: I0908 22:10:21.231423       1 server.go:55] Server on 'tcp://0.0.0.0:10000' stopped
[2020-09-08T22:10:21.367Z] pmem-csi-controller-0/pmem-driver@pmem..ster: PASS
...
[2020-09-08T22:10:37.050Z] Sep  8 22:10:37.018: INFO: waiting for deletion of pmem-csi-node-jp5kd (v1.Pod, 855cb2bf-ebfc-4fb6-abce-2a9991f05191)
[2020-09-08T22:10:37.050Z] Sep  8 22:10:37.018: INFO: waiting for deletion of pmem-csi-node-mvvzx (v1.Pod, 7254a27f-eccd-4079-a1cf-9c5707f1c98d)
[2020-09-08T22:10:37.050Z] Sep  8 22:10:37.018: INFO: waiting for deletion of pmem-csi-node-zrcfj (v1.Pod, 4462c334-fae4-41a0-99ab-54634440cef2)
...
[2020-09-08T22:10:41.715Z] pmem-csi-node-jp5kd/pmem-driver@pmem..ker3: E0908 22:10:41.395483       1 pmem-csi-driver.go:300] Failed to unregister node: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.105.63.129:10000: i/o timeout"
[2020-09-08T22:10:41.715Z] pmem-csi-node-jp5kd/pmem-driver@pmem..ker3: failed to run driver: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.105.63.129:10000: i/o timeout"
[2020-09-08T22:10:41.715Z] pmem-csi-node-jp5kd/pmem-driver@pmem..ker3: PASS
```

Issue #638, opened 2020-05-12
docsite: make warnings failures


pohly:
I mistyped paths:
- first I used `../deploy/...` in `examples` (relative path instead of absolute path),
- then I misspelled a path (customize instead of kustomize): 

```
/mnt/workspace/pmem-csi_PR-633/examples/memcached.md:44: WARNING: None:any reference target not found: /deploy/customize/memcached/ephemeral
```

It's easy to miss these warnings. Can we turn them into errors which case "make vhtml" to fail?

@intelkevinputnam

intelkevinputnam:
I use these switches to the sphinx build:  -W --keep-going

This will error on warnings, but it will also continue to build. That way we can catch all of them in one go if there are multiple warnings.

PR #640, opened 2020-05-13
docs: print all warnings, abort if any were found

pohly:
When building the docsite, we want to know when new warnings show up,
like malformed file references or new files that aren't included in
the docsite.

Fixes: #638


pohly:
@avalluri: okay to merge?

PR #642, opened 2020-05-15
Run operator and controller-driver as non-root user

avalluri:
No root user privileges are required to run the operator binary.
So we better run as normal user.


avalluri:
> As discussed, please re-open as PR against "devel" and also do a similar change for the controller. For the controller, the change needs to go into our yaml files and the operator source code.

Made the needed changes to run the controller as non-root user.  Now by default, the driver container uses the non-root user, and we choose the root user only for node containers.

deploy/kustomize/testing/controller-coverage-patch.yaml
```diff
@@ -14,3 +14,24 @@
     hostPath:
       path: /var/lib/pmem-csi-coverage
       type: DirectoryOrCreate
+# driver in controller mode runs as non-root user
+# which cannot have enough permissions to access
+# coverage-dir(/var/lib/pmem-csi-coverage)
+# Change the ownership of this folder to "pmem-csi:pmem-csi"
+# so that driver could write coverage reports.
+# NOTE: The assumption is that the user,group exist on the VM host with id 1000
```
pohly, line 22: Why this assumption? Files can be owned by numeric IDs that don't have entries in /etc/passwd or /etc/groups.

avalluri: > Why this assumption? Files can be owned by numeric IDs that don't have entries in /etc/passwd or /etc/groups.

Ok, this I do not know. I will remove the note.

PR #648, opened 2020-05-18
Step up to ndctl v68

okartau:


pohly:
A comment in the commit message about why this is necessary would have been good. Let's merge anyway.

PR #649, opened 2020-05-19
Ginkgo timeout

pohly:
We have had several cases recently where testing ran into the Jenkins timeout for the testing step. Usually that's because some tests ran much slower than usual and probably timed out. But because Jenkins aborts the e2e.test program, we don't get the JUnit test file and thus have to resort to log reading to figure out which tests ran resp. failed.

By setting a Ginkgo timeout that is a bit smaller than the Jenkins timeout, we should be able to avoid this.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -248,10 +248,12 @@ pipeline {
 
         stage('testing 1.18') {
             options {
-                timeout(time: 240, unit: "MINUTES")
+                timeout(time: 20, unit: "MINUTES")
             }
             steps {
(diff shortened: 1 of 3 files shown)


pohly:
This is work-in-progress because I artificially injected errors into tests and used lower timeouts. When this works, I'll fix the PR.

pohly:
Doesn't work. Besides not passing `GINKGO_TEST_E2E` properly in the PR, it also wouldn't have worked: "ginkgo -timeout" is implemented as passing "-test.timeout" to e2e.test, and that then just kills the process *without* writing the JUnit file.

PR #657, opened 2020-05-27
Jenkins parallel

pohly:
Let's see whether we can parallelize the long-running test steps...

If exporting images and restoring the right environment on the other workers is fast enough, this should be worthwhile.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,28 +73,8 @@ pipeline {
             }
 
             steps {
-                sh 'docker version'
-                sh 'git version'
-                sh 'free || true'
-                sh 'command -v top >/dev/null 2>&1 || \
-                    if command -v apt-get >/dev/null 2>&1; then \
-                        sudo apt-get install procps; \
-                    else \
-                        sudo dnf -y install procps; \
-                    fi'
-                sh 'head -n 30 /proc/cpuinfo; echo ...; tail -n 30 /proc/cpuinfo'
-                sh "git remote set-url origin git@github.com:intel/pmem-csi.git"
-                sh "git config user.name 'Intel Kubernetes CI/CD Bot'"
-                sh "git config user.email 'k8s-bot@intel.com'"
+                SetupHost()
 
-                // Create a tmpfs for use as backing store for a large file that will be passed
-                // into QEMU for storing the etcd database.
-                sh "mkdir -p '${env.TEST_ETCD_TMPFS}'"
-                sh "sudo mount -osize=${env.TEST_ETCD_VOLUME_SIZE} -t tmpfs none '${env.TEST_ETCD_TMPFS}'"
-                sh "sudo truncate --size=${env.TEST_ETCD_VOLUME_SIZE} '${env.TEST_ETCD_VOLUME}'"
-
-                // known_hosts entry created and verified as described in https://serverfault.com/questions/856194/securely-add-a-host-e-g-github-to-the-ssh-known-hosts-file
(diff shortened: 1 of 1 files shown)


pohly:
Looking good so far. https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-657/13/pipeline/76 really ran on two different VMs in parallel. 1.18 failed because of the quay.io outage, which incidentally confirms that test failures do show up for those additional VMs, too: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-657/13/

pohly:
A full test run just took only 3h45: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/jenkins-parallel/1/pipeline/86

I suggest we merge it.

@avalluri, @okartau: what do you think?

avalluri:
> @avalluri, @okartau: what do you think?

Parallel builds always good. My Jenkins script knowledge is so limited, so I couldn't review the changes. If @okartau agrees let's merge this.

okartau:
There is commented out "make test" with TODO marking, do you plan to merge it like this?

okartau:
Parallel run is really good for job duration. But I notice that logs get now interleaved, at least when looking at job log in regular Jenkins job display. This makes logs processing (which has not been easy  already so far) even more difficult as it is not easy to see from which VM the log comes from.
I wonder how Jenkins keeps those logs internally, as when I look to job at BlueOcean side, there seems to be way to look at one VM only, by selecting bubble of that job. But on traditional side (not blue ocean, there is just one job log, so where the interleaving really happens?

okartau:
> I suggest we merge it.

Do you propose merging regardless of most recent PR job failure? (i.e. is failure planned)?

pohly:
> Do you propose merging regardless of most recent PR job failure?

I did, but I changed my mind.

> (i.e. is failure planned)?

Not planned, but unrelated ("imagefile ext4 [It] 512Mi").

However, that failure and all other Clear Linux tests are not reported on the final JUnit report: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-657/14/testReport/

I thought that had worked earlier, but I probably didn't check carefully enough then. I need to fix this first.

pohly:
> There is commented out "make test" with TODO marking, do you plan to
> merge it like this?

No, that needs to be undone. It was just there to speed up testing of
the PR.

pohly:
> This makes logs processing (which has not been easy already so far) even more difficult as it is not easy to see from which VM the log comes from.

We shouldn't need to read the entire log that often. Ideally, the per-test log (as in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/view/change-requests/job/PR-657/13/testReport/junit/fedora-1_18.direct-testing/E2E/_Testpattern__inline_ephemeral_CSI_volume__ephemeral_should_create_read_only_inline_ephemeral_volume/) contains enough information.

If it doesn't, then the log for each step is still per-VM.

The steps are currently not shown here:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-657/14/

But I've seen them before also for that UI; seems to be a glitch. When it happens, use the Ocean Blue UI:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-657/14/pipeline/90

And then unfold the failed script, which has a "complete log" button leading to:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi/branches/PR-657/runs/14/nodes/90/steps/162/log/?start=0

pohly:
> However, that failure and all other Clear Linux tests are not reported on the final JUnit report: 

A much simpler solution than the previous `post` step is to call `junit` directly after processing results of one test step. That works even when junit then gets called multiple times on different workers: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/PR-657/19/testReport/

In that run, testing was artificially limited to image files. The current code should no longer have that and also runs `make test` again, i.e. once it passes testing, it should be ready for merging:
- PR testing: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-657/20/pipeline/
- full testing: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/jenkins-parallel/2/pipeline

BTW, testing in build 19 took 16 min for Fedora 1.18 and 25 min for Clear Linux. Assuming that the test itself ran equally fast, the difference is the time it took for bringing up the worker VM.

pohly:
The PR test failed again with "back-off pulling image "quay.io/k8scsi/csi-provisioner:v1.6.0".

Shall we merge it or try again?

okartau:
> Shall we merge it or try again?

if it only fails because of img load issue, (and we know code is working) then retry does not give new information. I think we can merge

PR #673, opened 2020-06-05
pmem-csi-driver: avoid caching pre-mounted volume information

### LINKED ISSUE #502: volInfo: mutex locking, persist across restarts
Looking at https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/nodeserver.go#L54 and how ns.volInfo is used I have two observations:

- It gets accessed in read/write mode from potentially parallel gRPC calls for different volumes. I think mutex locking for it is missing.
- It doesn't get stored on disk. I suspect that means that https://github.com/intel/pmem-csi/blob/fbaa784d853c063236d039dcd73aa356135b3cc4/pkg/pmem-csi-driver/nodeserver.go#L222-L233 will not work correctly after a node service restart or node reboot.

avalluri:
Caching volume information of pre-published volumes could be error prone
on node driver restart. Instead, we could rebuild this information by
looking into existing mounts.

FIXES #502

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -11,7 +11,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 
@@ -57,7 +56,6 @@ type nodeServer struct {
 	cs       *nodeControllerServer
 	// Driver deployed to provision only ephemeral volumes(only for Kubernetes v1.15)
 	mounter mount.Interface
-	volInfo map[string]volumeInfo
 
 	// A directory for additional mount points.
 	mountDirectory string
@@ -79,7 +77,6 @@ func NewNodeServer(cs *nodeControllerServer, mountDirectory string) *nodeServer 
 		},
 		cs:             cs,
 		mounter:        mount.New(""),
-		volInfo:        map[string]volumeInfo{},
 		mountDirectory: mountDirectory,
 	}
 }
@@ -196,11 +193,6 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		mountFlags = append(mountFlags, "ro")
 	}
 
-	// After mountFlags additions are complete: sort mountFlags slice and join to
-	// become a single string, as this is what we use to compare against existing
-	sort.Strings(mountFlags)
-	joinedMountFlags := strings.Join(mountFlags[:], ",")
-
 	rawBlock := false
 	switch req.VolumeCapability.GetAccessType().(type) {
 	case *csi.VolumeCapability_Block:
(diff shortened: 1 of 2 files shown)


pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -224,22 +216,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			//    VolumeCapability/fsType (if present in request) must match used fsType.
 			// 3) Readonly MUST match
 			// If there is mismatch of any of above, we return ALREADY_EXISTS error.
-			existingFsType, err := determineFilesystemType(device.Path)
+			mpList, err := ns.mounter.List()
 			if err != nil {
-				return nil, err
+				return nil, status.Errorf(codes.Internal, "Failed to fetch existing mount details for %q: %v", targetPath, err)
```
pohly, line 221: The error is not specific to the target path, is it? If yes, then the error message will be misleading - better say "Failed to fetch existing mount details while checking %q".

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -224,22 +216,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+				if mpList[i].Path == targetPath {
+					klog.V(5).Infof("NodePublishVolume[%s]: Existing mountFlags:'%s', fsTyp: '%s'", req.VolumeId, mpList[i].Opts, mpList[i].Type)
+					if (fsType == "" || mpList[i].Type == fsType) && findMountFlags(mountFlags, mpList[i].Opts) {
+						klog.V(5).Infof("NodePublishVolume[%s]: paremeters match existing, return OK", req.VolumeId)
+						return &csi.NodePublishVolumeResponse{}, nil
+					}
+					break
+				}
 			}
+			klog.V(5).Infof("NodePublishVolume[%s]: paremeters do not match existing, return ALREADY_EXISTS", req.VolumeId)
```
pohly, line 233: "parameters" and "existing volume"

test/e2e/storage/sanity.go
```diff
@@ -312,6 +312,41 @@ var _ = deploy.DescribeForSome("sanity", func(d *deploy.Deployment) bool {
 			v.remove(vol, name)
 		})
 
+		It("could able to handle publish volume calls after a node driver restart", func() {
```
okartau, line 315: What about:
"can publish volume after a node driver restart"
or, if _handle_ and _call_ are important:
"can handle PublishVolume calls after a node driver restart"

avalluri: > What about:
> "can publish volume after a node driver restart"

Sounds good, I made this change.

pkg/pmem-csi-driver/nodeserver.go
avalluri: Done

pkg/pmem-csi-driver/nodeserver.go
avalluri: > The error is not specific to the target path, is it?

The check is for the target path.

 > If so, then the error message will be misleading - better say "Failed to fetch existing mount details while checking %q".

Rephrased the error text.

pohly: > The check is for the target path.

Yes, but the error is not *about* the error path. Usually, errors follow the format "object: operation" and deviating from that would imply some connection that isn't there. Anyway, thanks for changing it.

Issue #664, opened 2020-06-04
asciinema video


pohly:
Eric Adams did a nice asciinema video based on the upcoming 0.7.0 relase: https://asciinema.org/a/4M5PSwbYkaXs0dPHYUIVbakqB

IMHO it would be nice to have that embedded into the starting page: https://intel.github.io/pmem-csi/latest/README.html

@intelkevinputnam: what do you think? Do you know how to embed it instead of merely linking to it?

pohly:
"howto" is explained in https://asciinema.org/docs/embedding.

PR #676, opened 2020-06-09
docs: asciinema demo with thumnail preview

pohly:
This will show a static image that can be clicked on to play the
animated version. We could embed the player
script (https://asciinema.org/docs/embedding) but this may be
overkill.

Fixes: https://github.com/intel/pmem-csi/issues/664

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,12 @@ officially by upstream anymore.
 <sup>2</sup> Lacks support for ephemeral inline volumes.
 Not supported officially by upstream anymore.
 
+## Demo
+
+Click the image to watch the animated demo on asciinema.org:
+
+[![asciicast](https://asciinema.org/a/4M5PSwbYkaXs0dPHYUIVbakqB.png)](https://asciinema.org/a/4M5PSwbYkaXs0dPHYUIVbakqB)
+
 ## Content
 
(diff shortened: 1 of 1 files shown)


pohly:
@avalluri Is it obvious enough that the demo can be started by clicking?

See https://github.com/intel/pmem-csi/blob/994685254173cd85fbcb1413705890ec857a37b8/README.md#demo.

avalluri:
> @avalluri Is it obvious enough that the demo can be started by clicking?
> 
> See https://github.com/intel/pmem-csi/blob/994685254173cd85fbcb1413705890ec857a37b8/README.md#demo.

Yes, I had a look at that. As you mentioned, embedding the video script overkills, then I feel It's ok to start the video by cling the screenshot.

avalluri:
> First, add asciinema-player.js, asciinema-player.css and the .cast file with your recording to your site's assets.

So we need to have those three files to have an embedded video?, I think it's ok to have them in our source. What do you think.

pohly:
> So we need to have those three files to have an embedded video?

No, the JS gets loaded from asciinema.org. I don't know how to get the necessary HTML tags into the rendered output, and it wouldn't work on GitHub.

I added some clarifying text instead.

Issue #679, opened 2020-06-13
check that release 0.7.x is getting updated


pohly:
Should be done by https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi-release/

pohly:
The job ran (instead of the one for release-0.6?!): https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi-release/detail/pmem-csi-release/142/pipeline/77

However, it failed because parallel testing did not anticipate the tag name change from v0.7.0 to v0.7.1. Needs to be fixed.

PR #680, opened 2020-06-15
Fix release 0.7

pohly:
Fixes: #679


pohly:
This should not break PR testing (which will see here). It *should* fix also the release-0.7 branch, but I have no way of testing that other then merging this PR, merging into release-0.7 and then re-running the release job.

PR #683, opened 2020-06-17
{Docker,Make}file: Support to build operator and driver specific images

avalluri:
PMEM-CSI driver depends on quite many packages that are not relevant for running the operator. So distributing the operator part of the driver image is not so appropriate. Instead, we could generate an operator-specific image with no additional dependencies.
    
 We generate below images:
    - pmem-csi-operator: Contains only the operator binary
    - pmem-csi-operator-test: Contains operator binary with debug symbols
    - pmem-csi-driver: Contains driver binaries and dependencies
    - pmem-csi-driver-test: Contains driver binaries with debug symbols +
    test binaries


pohly:
Please explain "why" this is better. Presumably it is because images become smaller. Can you quantify that?

avalluri:
> Please explain "why" this is better. Presumably it is because images become smaller. Can you quantify that?

Image sizes based on `BUILD_ID=33380`:
 Operator: 220MB
 Driver: 619MB

Size is one factor and conceptually the operator is different from the driver that deserves a specific image.

pohly:
> Image sizes based on BUILD_ID=33380:

How did you measure, and how does that compare to previous images?

> that deserves a specific image.

"deserves" is subjective. If it has no tangible advantages other than "it's cleaner", then that may not be enough to justify the extra complications.

For example, your patch is not complete yet. You also need to update the Jenkinsfile and who knows what in the image publishing CI - it may or may not work if we publish new images into internal repository.

PR #684, opened 2020-06-17
pmem-csi-driver: use appropriate klog APIs for "Failed" messages

bart0sh:
Used klog.WarningF and klog.Errorf instead of klog.V(N).Info to output
"Failed..." messages.

This should ensure that the "Failed" messages will not be skipped
from the output.

### DIFF
--- a/pkg/pmem-csi-driver/pmem-csi-driver.go
+++ b/pkg/pmem-csi-driver/pmem-csi-driver.go
@@ -295,7 +295,8 @@ func (pmemd *pmemDriver) Run() error {
 		if pmemd.cfg.Mode == Node {
 			klog.V(3).Info("Unregistering node...")
 			if err := pmemd.unregisterNodeController(); err != nil {
-				klog.V(4).Infof("Failed to node unregister: %v", err)
+				klog.Errorf("Failed to unregister node: %v", err)
+				return err
 			}
 		}
 
@@ -319,7 +320,7 @@ func (pmemd *pmemDriver) registerNodeController() error {
 		if err == nil {
 			break
 		}
-		klog.V(4).Infof("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
+		klog.Warningf("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
 		time.Sleep(retryTimeout)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -295,7 +295,7 @@ func (pmemd *pmemDriver) Run() error {
 		if pmemd.cfg.Mode == Node {
 			klog.V(3).Info("Unregistering node...")
 			if err := pmemd.unregisterNodeController(); err != nil {
-				klog.V(4).Infof("Failed to node unregister: %v", err)
+				klog.V(3).Infof("Failed to unregister node: %v", err)
```
pohly, line 298: @avalluri why were these info messages and not `klog.Errorf` messages?

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -319,7 +319,7 @@ func (pmemd *pmemDriver) registerNodeController() error {
 		if err == nil {
 			break
 		}
-		klog.V(4).Infof("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
+		klog.V(3).Infof("Failed to connect registry server: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
```
pohly, line 322: Here `klog.Warning` seems more appropriate.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -478,7 +478,7 @@ func register(ctx context.Context, conn *grpc.ClientConn, req *registry.Register
 			if s, ok := status.FromError(err); ok && s.Code() == codes.InvalidArgument {
 				return fmt.Errorf("Registration failed: %s", s.Message())
 			}
-			klog.V(5).Infof("Failed to register: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
+			klog.V(3).Infof("Failed to register: %s, retrying after %v seconds...", err.Error(), retryTimeout.Seconds())
```
pohly, line 481: And also here.

bart0sh: updated

pkg/pmem-csi-driver/pmem-csi-driver.go
bart0sh: updated

pkg/pmem-csi-driver/pmem-csi-driver.go
bart0sh: replaced with klog.Errorf

avalluri: > @avalluri why were these info messages and not `klog.Errorf` messages?

No specific reason I could recall. It could be left-over of a copy-paste.

Issue #671, opened 2020-06-05
switch Fedora from docker to containerd


pohly:
We let Kubernetes pick "docker" as container runtime when testing on Fedora. It would be more useful to use "containerd" directly, because then we can test Kata Containers also on Fedora. Right now, we have to use Clear Linux for that because that uses crio. Kata Containers works with that, but not docker.

pohly:
Related: https://groups.google.com/d/msgid/kubernetes-sig-testing/CANw6fcG2N7L6K_c-DBjHMXyTD%2BbVJyR7QpKiufGbvXDhUz5reg%40mail.gmail.com

PR #686, opened 2020-06-18
Fedora containerd

pohly:
Using containerd directly should be simpler and allows us to test Kata
Containers also with Fedora.

Fixes: #671


pohly:
@avalluri shall we switch?

PR #688, opened 2020-06-22
Remove init-containers in LVM mode

### LINKED ISSUE #532: merge pmem-ns-init and pmem-vgm into pmem-csi-driver
The information shown in those setup stages (these run in LVM mode only) can be helpful for debugging issues and monitoring storage devices state.

avalluri:
As we are using 'shared/Bidirectional' mounts for LVM driver also, we
could create the needed physical devices(namespaces, volume groups) in
the same driver session. This invalidates the usage of init-containers.

Few advantages of merging the init-containers functionality to driver code:
 - smaller container image size
 - allows us to implement features that are common to both modes(ex:
 pmemPercentage)

FIXES #532


docs/install.md
```diff
@@ -341,10 +341,10 @@ for `kubectl kustomize`. For example:
          path: lvm-parameters-patch.yaml
      EOF
      $ cat >my-pmem-csi-deployment/lvm-parameters-patch.yaml <<EOF
-     # pmem-ns-init is in the init container #0. Append arguments at the end.
+     # pmem-driver is in the init container #0. Append arguments at the end.
```
pohly, line 344: It's "the container #0", not "the *init* container #0".

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -287,3 +295,86 @@ func getVolumeGroups(groups []string) ([]vgInfo, error) {
 
 	return vgs, nil
 }
+
+func createNS(r *ndctl.Region, uselimit uint) error {
+	const align uint64 = 1024 * 1024 * 1024
+	realalign := align * r.InterleaveWays()
+	// uselimit is the percentage we can use
+	canUse := uint64(uselimit) * r.Size() / 100
```
pohly, line 303: As you noticed here, `uselimit` is a name that cannot be understood without additional explanation. Why not just call the parameter "percentage"?

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -287,3 +295,86 @@ func getVolumeGroups(groups []string) ([]vgInfo, error) {
...
+	klog.V(3).Infof("Create fsdax-namespaces in %v, allowed %d %%, real align %d:\ntotal       : %16d\navail       : %16d\ncan use     : %16d",
+		r.DeviceName(), uselimit, realalign, r.Size(), r.AvailableSize(), canUse)
+	// Subtract sizes of existing active namespaces with currently handled mode and owned by pmem-csi
+	for _, ns := range r.ActiveNamespaces() {
+		klog.V(5).Infof("createNS: Exists: Size %16d Mode:%v Device:%v Name:%v", ns.Size(), ns.Mode(), ns.DeviceName(), ns.Name())
+		if ns.Name() == "pmem-csi" {
+			canUse -= ns.Size()
```
pohly, line 310: Are you absolutely sure that the uint64 cannot underflow?

At some point in the past, the percentage may have been higher and a volume might have been created which is higher than the currently allowed maximum.

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -287,3 +295,86 @@ func getVolumeGroups(groups []string) ([]vgInfo, error) {
 
 	return vgs, nil
 }
+
+func createNS(r *ndctl.Region, uselimit uint) error {
```
pohly, line 299: The logic in this function is complex enough that we should have a unit test which covers different scenarios.

avalluri: The particular code was authored by @okartau. I just moved the existing code from `pmem-ns-init.go` to here.

pohly: Then tests were already missing before. Now would still be a good time to add them.

avalluri: Tests for this part of code is a bit complicated as the code depends on underlined PMEM hardware region properties(size, interleaves, max extent).  Current LVM unit tests are run on a pre-populated logical volume groups.

pkg/pmem-device-manager/pmd-lvm.go
avalluri: Fixed this case.

docs/install.md

pkg/pmem-device-manager/pmd-lvm.go
avalluri: Renamed the variable.

pkg/pmem-device-manager/pmd-lvm.go
pohly: The usual approach for cases like this is to introduce an interface and then test the code with a mock implementation of the interface.

pohly: Okay. Please file an issue to track it.

avalluri: Filed #692

PR #689, opened 2020-06-22
Rename pmem -> csi (variables and fields)

bart0sh:
This is a first step in generalizing current APIs.
Renamed variables and fields of pkg/pmem-csi-driver/ and some of its depdendencies.

Avoided renaming files to keep git history cleaner.

### DIFF
--- a/pkg/pmem-csi-driver/controllerserver-master.go
+++ b/pkg/pmem-csi-driver/controllerserver-master.go
@@ -37,7 +37,7 @@ const (
 	Deleted
 )
 
-type pmemVolume struct {
+type volume struct {
 	// VolumeID published to outside world
 	id string
 	// Name of volume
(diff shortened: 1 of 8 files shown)


pohly:
@bart0sh can you rebase?

I think renaming files is fine and actually desirable if it makes the code simpler.

bart0sh:
Rebased. I'll rename them as a second step. This is a first PR and I didn't want to make a lot of changes in it.

pkg/pmem-csi-driver/controllerserver-master.go
```diff
@@ -4,7 +4,7 @@ Copyright 2017 The Kubernetes Authors.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package pmemcsidriver
+package csidriver
```
pohly, line 7: This is the only change that concerns me: while some of the code in  pkg/pmem-csi-driver is generic, not everything is. Therefore "pmemcsidriver" is the better package name.

Ultimately, we would need a `pkg/csi-driver` with the generic code and `pkg/pmem-csi-driver` with the PMEM-CSI specific code. But that is a larger change where I am not sure whether it is justified.

Can we leave the package name as it is for now?

bart0sh: I though that the idea is to make it generic. Otherwise I'd need to copy most of its code to a new driver.

pohly: In the new driver, do you want to work with a fork of the code or import he code from the PMEM-CSI repo?

My assumption was that we clean up the PMEM-CSI driver code so that it better distinguishes between generic code and PMEM code. That's worthwhile by itself.

But I thought you would still maintain a full copy of the code with additional modifications. In that model, a generic "pkg/csi-driver" is nice, but not required.

bart0sh: My idea was to generalise driver code to the point it can be used(imported?) by other drivers.
However, if you just want a names cleanup it's even better as it's much less work for me :)
I've updated the PR as you suggested.

bart0sh:
@pohly Updated the PR according to review comments. Please review again. Thanks.

avalluri:
> @avalluri : what do you think?

Looks good to me.

PR #697, opened 2020-07-14
parameters: add origin to the messages

bart0sh:
- changed type Origin from int to string
- added origin to the errors and info messages

This should make parameters parsing easier to debug.

### DIFF
--- a/pkg/pmem-csi-driver/parameters/parameters.go
+++ b/pkg/pmem-csi-driver/parameters/parameters.go
@@ -11,12 +11,14 @@ import (
 	"strconv"
 	"strings"
 
+	"k8s.io/klog"
+
 	api "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 type Persistency string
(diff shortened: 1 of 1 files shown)


pkg/pmem-csi-driver/parameters/parameters.go
```diff
@@ -148,7 +151,7 @@ func Parse(origin Origin, stringmap map[string]string) (Volume, error) {
 			}
 		}
 		if !valid {
-			return result, fmt.Errorf("parameter %q invalid in this context", key)
+			return result, fmt.Errorf("%s: parameter %q invalid in this context", origin, key)
```
pohly, line 154: I'm worried that this will lead to incomprehensible error messages for normal users. Suppose they use an invalid parameter in a storage class. Does `CreateVolumeOrigin: parameter "foobar" invalid in this context` really help them?

Changing the type and using them in logs is fine, but I'm not convinced about error messages. "This context" then refers to where the error occurred, not specifically the internal parameter origin.

bart0sh: Agreed. closing.

PR #708, opened 2020-08-19
deploy: make csi.sock work on OpenShift 4.5 + SELinux

pohly:
For some (unknown) reason SELinux got in the way when using
/var/lib/pmem-csi.intel.com/csi.sock. The symptoms were that
node-driver-registrar was unable to connect and thus the CSI driver
never got registered. Disabling SELinux "fixed" this.
    
When using the more traditional
/var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock, it worked with
SELinux enabled, so that is what we are using now.
    
It makes changing the driver name a bit more complicated, though, in
particular when trying to do it via kustomize.


avalluri (approved):
Looks good to me. Let's wait for test results before merging.

pohly:
@avalluri: please take another look. I had to rebase and fix one issue (socat paths not updated, which broke sanity testing).

Originally had one commit that would have been back-portable and another commit which updated it for "devel", but during rebasing I lost that, so now the commit no longer applies cleanly to "release-0.7" - won't be hard to solve, so let's go ahead with just one commit for "devel".

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -250,6 +266,8 @@ spec:
             secretName: pmem-csi-node-secrets
         - name: pmem-state-dir
           hostPath:
+            # This state directory must always be the same, regardless of the CSI driver name,
+            # because we want to pick up old volumes even when the driver name gets changed.
```
avalluri, line 270: @pohly Is this same even for the operator deployments? IIRC operator uses /var/lib/<driver-name> a state directory.

pohly: I replaced the comment with one stating the opposite:

> This state directory must include the CSI driver name because each
>installation with a certain name must have its own, separate state.

PR #712, opened 2020-08-24
deploy: make csi.sock work on OpenShift 4.5 + SELinux

pohly:
For some (unknown) reason SELinux got in the way when using
/var/lib/pmem-csi.intel.com/csi.sock. The symptoms were that
node-driver-registrar was unable to connect and thus the CSI driver
never got registered. Disabling SELinux "fixed" this.

When using the more traditional
/var/lib/kubelet/plugins/pmem-csi.intel.com/csi.sock, it worked with
SELinux enabled, so that is what we are using now.

It makes changing the driver name a bit more complicated, though, in
particular when trying to do it via kustomize.


avalluri (approved):

avalluri:
some test failures:
```
[2020-08-24T08:45:49.262Z]                 deployment_controller_test.go:206: 

[2020-08-24T08:45:49.262Z]                     	Error Trace:	deployment_controller_test.go:206

[2020-08-24T08:45:49.262Z]                     	            				deployment_controller_test.go:419

[2020-08-24T08:45:49.262Z]                     	            				deployment_controller_test.go:452

[2020-08-24T08:45:49.262Z]                     	Error:      	Received unexpected error:

[2020-08-24T08:45:49.262Z]                     	            	deployed driver different from expected deployment:

[2020-08-24T08:45:49.262Z]                     	            	spec content for "pmem-csi-with-values-node" of type "apps/v1, Kind=DaemonSet" in namespace "test-namespace" does not match:

[2020-08-24T08:45:49.262Z]                     	            	   ! spec.template.spec.containers.#0.image expected base-image.7.14, got base-image

[2020-08-24T08:45:49.262Z]                     	            	spec content for "pmem-csi-with-values-controller" of type "apps/v1, Kind=StatefulSet" in namespace "test-namespace" does not match:

[2020-08-24T08:45:49.262Z]                     	            	   ! spec.template.spec.containers.#0.image expected base-image.7.14, got base-image

[2020-08-24T08:45:49.262Z]                     	Test:       	TestDeploymentController/Kubernetes_1.15#01/updating/all_in_deployment_with_specific_values

[2020-08-24T08:45:49.262Z]                     	Messages:   	validate deployment

[2020-08-24T08:45:49.262Z]                 --- FAIL: TestDeploymentController/Kubernetes_1.15#01/updating/all_in_deployment_with_specific_values/while_running (0.02s)

[2020-08-24T08:45:49.262Z]                     testing.go:864: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

pohly:
This puzzled me for a while. The reason for the failure is that "make kustomize" was run after setting a fixed image version in the YAML files, which then broke the test. It works when the builtin YAML files don't have the version.

I'm not sure what the right fix is here. Perhaps always use input files without version when running "make kustomize"? That's then consistent. Alternatively, the code that works with the builtin YAMLs could be changed to ignore the version. @avalluri : any preference?

The short-term fix is to run "make kustomize" on the pre-0.7.0 commit.

avalluri:
> Alternatively, the code that works with the builtin YAMLs could be changed to ignore the version.

This sounds good to me as this part of the code already doing similar things for other fields.

pohly:
> > Alternatively, the code that works with the builtin YAMLs could be changed to ignore the version.
> 
> This sounds good to me as this part of the code already doing similar things for other fields.

It's also trivial to implement (just update the image regex). Okay, let's do it that way.

avalluri (approved):
Looks good to me.

Issue #692, opened 2020-06-25
Write unit test for LVM device manager's `creteNS()` functionality 


avalluri:
As pointed in https://github.com/intel/pmem-csi/pull/688#discussion_r443503303, currently, there are no tests for validating if the LVM device manager uses only requested `pmemPercentage` out of available PMEM space of a region. This is a bit complicated as the testing this requires underlined PMEM region/namespace properties.

It would be good if we could find a way to develop a test case for this. As suggest by @pohly, one way is making [`createNS()`](https://github.com/intel/pmem-csi/pull/688/files#diff-dabd8ee55590c66f44f07a15baac3813R303) work on an interface, so that tests could have a mock implementation.

pohly:
We already had two bugs in that code (fixed by https://github.com/intel/pmem-csi/pull/714):
- did not ignore regions in the wrong mode (not sure how relevant)
- caused restart of driver (did happen each time the driver ran for the first time on a node)

okartau:
I just detected a third bug, fix in #717 
I agree, unit test code would help significantly. One of the bugs was regression caused by omitted code lines in function move where original code was correct.
We cant have really same what comes form HW, but we can use some pre-defined sets of typical value combinations.

pohly:
We do have infrastructure for mocking libndctl now in https://github.com/intel/pmem-csi/tree/devel/pkg/ndctl/fake

That could be used for this.

PR #717, opened 2020-08-28
Dev-manager canUse calc: fix operator order

okartau:
Division as integer and then multiply, sheds off some value
from canUse size, and later when we align down by Align size,
we will lose whole alignment block because of that.
This creates issue where 1st run leaves one realAlign-sized block
unallocated, but 2nd start on same node will find it and create
a new relatively small namespace.

### DIFF
--- a/pkg/pmem-device-manager/pmd-lvm.go
+++ b/pkg/pmem-device-manager/pmd-lvm.go
@@ -308,13 +308,18 @@ const (
 // setupNS checks if a namespace needs to be created in the region and if so, does that.
 func setupNS(r *ndctl.Region, percentage uint) error {
 	align := GB
+	// In doc for "ndctl create-namespace" https://pmem.io/ndctl/ndctl-create-namespace.html
+	// it is stated that:
+	// For pmem namepsaces the size must be a multiple of the interleave-width and the namespace alignment.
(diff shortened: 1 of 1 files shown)


pohly (changes requested):
Not strictly related, but can you use the opportunity to explain what "realalign" is? It's not obvious when reading this function.

The fix itself looks fine. Unit tests would be good, but we are already tracking that (issue #692 ).

pohly:
> Not strictly related, but can you use the opportunity to explain what "realalign" is?

And with that I meant "explain in the source code" - not here in the GitHub issue.

okartau:
testing failed but I believe it is not related to this change:

The Service "pmem-csi-scheduler" is invalid: spec.ports[0].nodePort: Invalid value: 32000: provided port is already allocated
[AfterEach] direct-testing-metrics

I will add another commit with asked "realalign" explanation.

pohly (approved):

pohly:
> The Service "pmem-csi-scheduler" is invalid: spec.ports[0].nodePort: Invalid value: 32000: provided port is already allocated
[AfterEach] direct-testing-metrics

I just hit the same locally. Seems to be timing issue with plenty of upstream issues that report it. I'll probably add a retry loop.

PR #718, opened 2020-09-01
pmem-csi-operator: delete obsolete objects of a deployment

### LINKED ISSUE #595: operator should delete obsolete sub-resources if any
On upgrade, the operator must be able to find and delete obsolete sub-resources if any for a deployment.

avalluri:
Different operator releases might result in obsolete sub-resources that were created for a deployment.  Operator shall detect these obsolete objects if any and delete them in the reconcile loop.
    
There is no direct API to detect all the objects owned by an object. So, we solve this by annotating the Deployment CR with the list of object types & names, and the version of the operator that reconciled. On operator upgrade/degrade we check if any pre-deployed objects are still valid.
    
FIXES #595

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package deployment
 
 import (
+	"context"
 	"crypto/rsa"
 	"crypto/tls"
 	"fmt"
@@ -25,10 +26,15 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/klog"
+	"k8s.io/kubectl/pkg/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -36,6 +42,25 @@ const (
 	controllerMetricsPort = 10010
 	nodeControllerPort    = 10001
 )
+
+// A list of all object types potentially created by the operator,
+// in this or any previous release. In other words, this list may grow,
+// but never shrink, because a newer release needs to delete objects
+// created by an older release.
+// This list also must be kept in sync with the operator RBAC rules.
+var AllObjectTypes = []schema.GroupVersionKind{
+	rbacv1.SchemeGroupVersion.WithKind("RoleList"),
+	rbacv1.SchemeGroupVersion.WithKind("ClusterRoleList"),
+	rbacv1.SchemeGroupVersion.WithKind("RoleBindingList"),
+	rbacv1.SchemeGroupVersion.WithKind("ClusterRoleBindingList"),
+	corev1.SchemeGroupVersion.WithKind("ServiceAccountList"),
+	corev1.SchemeGroupVersion.WithKind("SecretList"),
+	corev1.SchemeGroupVersion.WithKind("ServiceList"),
+	corev1.SchemeGroupVersion.WithKind("ConfigMapList"),
+	appsv1.SchemeGroupVersion.WithKind("DaemonSetList"),
+	appsv1.SchemeGroupVersion.WithKind("StatefulSetList"),
+	storagev1beta1.SchemeGroupVersion.WithKind("CSIDriverList"),
+}
 
 type PmemCSIDriver struct {
 	*api.Deployment
@@ -137,6 +162,15 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
 			return true, err
 		}
 		objects = append(objects, objs...)
+
+		// If not found cache might be result of operator restart
+		// check if this deployment has any objects to be deleted
+		if !foundInCache {
+			if err := d.deleteObsoleteObjects(r, objs); err != nil {
+				klog.Infof("Failed to delete obsolete objects: %v", err)
+				return true, err
+			}
+		}
 	} else {
 		if updateSecrets {
 			objs, err := d.getSecrets()
@@ -203,14 +237,69 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
 	return false, nil
 }
 
-func (d *PmemCSIDriver) deployObjects(r *ReconcileDeployment) error {
-	objects, err := d.getDeploymentObjects()
-	if err != nil {
-		return err
+func objectIsObsolete(objList []apiruntime.Object, toFind unstructured.Unstructured) (bool, error) {
+	for i := range objList {
+		metaObj, err := meta.Accessor(objList[i])
+		if err != nil {
+			return false, err
+		}
+		if metaObj.GetName() == toFind.GetName() &&
+			objList[i].GetObjectKind().GroupVersionKind() == toFind.GetObjectKind().GroupVersionKind() {
+			return false, nil
+		}
 	}
-	for _, obj := range objects {
-		if err := r.Create(obj); err != nil {
+
+	return true, nil
+}
+
+func (d *PmemCSIDriver) isOwnerOf(obj unstructured.Unstructured) bool {
+	for _, owner := range obj.GetOwnerReferences() {
+		if owner.UID == d.GetUID() {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (d *PmemCSIDriver) deleteObsoleteObjects(r *ReconcileDeployment, newObjects []apiruntime.Object) error {
+	for _, gvk := range AllObjectTypes {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(gvk)
+		opts := &client.ListOptions{
+			Namespace: d.namespace,
+		}
+
+		klog.Infof("Fetching '%s' list with options: %v", gvk, opts.Namespace)
+		if err := r.client.List(context.TODO(), list, opts); err != nil {
 			return err
+		}
+
+		for _, obj := range list.Items {
+			if !d.isOwnerOf(obj) {
+				continue
+			}
+			obsolete, err := objectIsObsolete(newObjects, obj)
+			if err != nil {
+				return err
+			}
+			if !obsolete {
+				continue
+			}
+			o, err := scheme.Scheme.New(obj.GetObjectKind().GroupVersionKind())
+			if err != nil {
+				return err
+			}
+			metaObj, err := meta.Accessor(o)
+			if err != nil {
+				return err
+			}
+			metaObj.SetName(obj.GetName())
+			metaObj.SetNamespace(obj.GetNamespace())
+
+			if err := r.Delete(o); err != nil && !errors.IsNotFound(err) {
+				return err
+			}
 		}
 	}
 	return nil
@@ -352,19 +441,6 @@ func validateCertificates(caCert, regKey, regCert, ncKey, ncCert []byte) error {
 	conn.Close()
 
 	return nil
-}
-
-func (d *PmemCSIDriver) getOwnerReference() metav1.OwnerReference {
-	blockOwnerDeletion := true
-	isController := true
-	return metav1.OwnerReference{
-		APIVersion:         d.APIVersion,
-		Kind:               d.Kind,
-		Name:               d.GetName(),
-		UID:                d.GetUID(),
-		BlockOwnerDeletion: &blockOwnerDeletion,
-		Controller:         &isController,
-	}
 }
 
 func (d *PmemCSIDriver) getCSIDriver() *storagev1beta1.CSIDriver {
@@ -1038,7 +1114,7 @@ func (d *PmemCSIDriver) getObjectMeta(name string, isClusterResource bool) metav
 	meta := metav1.ObjectMeta{
 		Name: name,
 		OwnerReferences: []metav1.OwnerReference{
-			d.getOwnerReference(),
+			d.GetOwnerReference(),
 		},
 		Labels: d.Spec.Labels,
 	}
--- a/pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
+++ b/pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
@@ -26,6 +26,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -198,23 +219,152 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+
+	d.ObjectMeta.Annotations[versionAnnotation] = operatorVersion
+	d.ObjectMeta.Annotations[ownedObjectsAnnotation] = string(jsonString)
+	d.needsUpdate = true
+
 	return nil
 }
 
-// getDeploymentObjects returns all objects that are part of a driver deployment.
+// retrieveTaggedObjects retrives the objects tagged for this deployment by the
```
pohly, line 272: "retrieveTaggedObjects retrives the objects tagged" just repeats the same information. When describing a function, at least try to paraphrase or (better) add additional information. In this case, when reading just this function name and comment, it is not clear what "tagged" means.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -198,23 +219,152 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+	}
+
+	for i := range nameGVKs {
+		parts := strings.SplitN(nameGVKs[i], ":", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("failed to parse annototion %q: %s", ownedObjectsAnnotation, nameGVKs[i])
+		}
+		gvk, err := parseGroupVersionKind(parts[1])
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse annototion %q: %v", ownedObjectsAnnotation, err)
```
pohly, line 294: typo: annototion -> annotation

Same above.

The annotation value is already in JSON format. Why don't you use JSON encoding also for the items inside the list? Then this entire code here goes away.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -198,23 +219,152 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+func (d *PmemCSIDriver) getObsoleteObjects(currentOperatorVersion string) ([]apiruntime.Object, error) {
+	obsoleteList := []apiruntime.Object{}
+
+	deploymentVersion, ok := d.ObjectMeta.Annotations[versionAnnotation]
+	if !ok {
+		return obsoleteList, nil
+	}
+	if deploymentVersion == currentOperatorVersion {
+		return obsoleteList, nil
+	}
```
pohly, line 315: Like the other version check this one here seems like an unnecessary optimization.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -198,23 +219,152 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+			obsoleteList = append(obsoleteList, oThen)
+		}
+	}
+
+	return obsoleteList, nil
+}
+
+// getDeploymentOwnedObjects returns the simple list of object with name and its API GVK
+// that are owned(supposed to own) by this deployment.
+func (d *PmemCSIDriver) getDeploymentOwnedObjects() []apiruntime.Object {
```
pohly: Hmm, keeping this list up-to-date will be error-prone. Can't you derive it dynamically by generating the objects and the iterating over them?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -1046,6 +1196,69 @@ func (d *PmemCSIDriver) getObjectMeta(name string, isClusterResource bool) metav
 	return meta
 }
 
+func (d *PmemCSIDriver) objectForGVK(gvk schema.GroupVersionKind, name string) apiruntime.Object {
```
pohly, line 1223: This entire function also disappears when not hand-coding `getDeploymentOwnedObjects`.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -1056,3 +1269,23 @@ func joinMaps(left, right map[string]string) map[string]string {
 	}
 	return result
 }
+
+// parseGroupVersionKind parses the gvk string to from a schema.GroupVersionKind object
+// valid format is: <AIPGroup>/>Version>,Kind=<kind>
+func parseGroupVersionKind(strGVK string) (schema.GroupVersionKind, error) {
```
pohly, line 1304: This function is no longer needed when using JSON encoding for the list items.

pohly (changes requested):
It just occurred to me that we have a conceptual issue with the downgrade scenario and unknown object kinds: if an old operator version didn't create objects of kind "foo", then it also had no RBAC rules that allow it to delete them.

Making the old version aware of objects that it needs to delete then doesn't help when it no longer has permission to do so because the RBAC rules were downgraded together with the operator version. I don't have a good solution for this.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Stored GVK is not a JSON string, its formatted by `GVK.String()`. So we cannot unmarshal that using JSON.

pohly: And who says that we need to use GVK.String? Just define a list of structs which contain a GVK and you can marshal the entire list with json.

avalluri:
@pohly Now I rewrote the code such that instead of annotating the CR with the object list, generating the list dynamically by retrieving objects with a list of hard-coded types.

pkg/pmem-csi-operator/controller/controller.go
```diff
@@ -15,6 +15,8 @@ import (
 
 // ControllerOptions type defintions for options to be passed to reconcile controller
 type ControllerOptions struct {
+	// Version holds Operator/Controller version
+	Version string
```
pohly, line 19: Is the version still needed?

Please squash the two commits.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -37,6 +43,27 @@ const (
...
 )
 
+// A list of all object types potentially created by the
+// operator. It's okay and desirable to list more than actually used
+// at the moment, to catch new objects.
+// NOTE: Add the new object types to this list.
+// But, This list should not shrink even if the later versions of
+// operator does not install the object listed here.
+// map: GVK -> isClusterScoped
+var allObjectTypes = map[schema.GroupVersionKind]bool{
```
pohly, line 53: This is similar the list that we have in test/e2e/operator/validate/validate.go.

Please refactor so that we have only one list.

Also add a comment that we need to keep this in sync with RBAC rules.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -201,16 +237,86 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+	foundList := []NameGVK{}
+	obsoleteList := []NameGVK{}
+
+	for gvk, clusterScoped := range allObjectTypes {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(gvk)
+		opts := client.ListOptions{}
+		if !clusterScoped {
+			opts.Namespace = d.namespace
+		}
```
pohly, line 259: I don't think we need the `clusterScoped` boolean and thus can make our type list simpler. My expectation is that `opts.Namespace` will be ignored when it isn't needed.

pohly (changes requested):

pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
```diff
@@ -69,7 +69,7 @@ func UpdateTests() []UpdateTest {
 			}
 		},
 		"pmemPercentage": func(d *api.Deployment) {
-			d.Spec.PMEMPercentage++
+			d.Spec.PMEMPercentage--
```
pohly, line 72: When does that test failure occur? We don't see it when testing "devel" and it is not obvious why the other commits should cause it.

test/e2e/operator/validate/validate.go
```diff
@@ -114,6 +115,13 @@ func DriverDeployment(client client.Client, k8sver version.Version, namespace st
 						deployment.Spec.NodeControllerCert)...)
 					continue
 				}
+			} else if actual.GetKind() == "ConfigMap" {
+				opVersion, ok := deployment.Annotations[operatordeployment.VersionAnnotation]
+				// Operator in TestVersion creates an additional ConfigMap, which is valid
+				// Refer to version upgrade/degrade" test in deployment_controller_test.go
```
pohly, line 121: degrade -> downgrade

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -25,18 +26,52 @@ import (
...
 )
 
 const (
 	controllerServicePort = 10000
 	controllerMetricsPort = 10010
 	nodeControllerPort    = 10001
+
+	// Annotation used to tag/track the version of the operator that
+	// successfully reconciled a deployment.
+	VersionAnnotation = "pmem-csi.intel.com/operatorVersion"
```
pohly, line 47: This new concept is not clear to me. Why do we need this annotation?

pohly: It only seems to be used for testing. We shouldn't introduce extra, otherwise unused features into the operator API just for testing purposes.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -25,18 +26,52 @@ import (
...
+	TestVersion = "testing"
 )
 
+// A list of all object types potentially created by the
+// operator. It's okay and desirable to list more than actually used
+// at the moment, to catch new objects.
+// NOTE: Add the new object types to this list.
+// But, This list should not shrink even if the later versions of
+// operator does not install the object listed here.
+// This list also should keep in syc with the operator RBAC rules.
```
pohly, line 59: syc -> sync

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -25,18 +26,52 @@ import (
...
+	// successfully reconciled a deployment.
+	VersionAnnotation = "pmem-csi.intel.com/operatorVersion"
+
+	// Used by unit testing
+	TestVersion = "testing"
 )
 
+// A list of all object types potentially created by the
+// operator. It's okay and desirable to list more than actually used
+// at the moment, to catch new objects.
```
pohly, line 55: This comment was applicable only when this was used for testing. It's not desirable to add unused objects anymore because that then would imply that we also add those to the RBAC rules.

Now a more accurate description is:

```
A list of all object types potentially created by the operator, in this or any previous release. In other words, this list may grow, but never shrink, because a newer release needs to delete objects created by an older release.

 This list also must be kept in syc with the operator RBAC rules.
```

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -25,18 +26,52 @@ import (
...
+// NOTE: Add the new object types to this list.
+// But, This list should not shrink even if the later versions of
+// operator does not install the object listed here.
+// This list also should keep in syc with the operator RBAC rules.
+// map: GVK -> isClusterScoped
```
pohly, line 60: Out-dated comment.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -200,24 +244,82 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+
+		klog.Infof("Fetching '%s' list with options: %v", gvk, opts.Namespace)
+		if err := r.client.List(context.TODO(), list, opts); err != nil {
 			return err
 		}
+
+		for _, obj := range list.Items {
+			for _, owner := range obj.GetOwnerReferences() {
+				if owner.UID == d.GetUID() {
+					foundList = append(foundList, NameGVK{
```
pohly, line 276: This collection of found objects in a list is unnecessary. The code becomes simpler (less variables, smaller functions, function names that explain what the code does) when moving loop bodies below into functions (`objectIsObsolete`, `checkExistingObject`). The you just call `checkExistingObject` here.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -236,6 +338,11 @@ func (d *PmemCSIDriver) getDeploymentObjects() ([]apiruntime.Object, error) {
 		d.getNodeDaemonSet(),
 	)
 
+	if operatorVersion == TestVersion {
+		// create a dummy config map
+		objects = append(objects, d.getConfigMap())
+	}
```
pohly, line 344: No test code in the production binary, please.

If you want to test that obsolete objects are removed, then you can let the operator install the driver, shut down the operator, add some extra object *from the test code*, then restart the operator.

pohly (changes requested):

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > If you want to test that obsolete objects are removed, then you can let the operator install the driver, shut down the operator, add some extra object _from the test code_, then restart the operator.

You mean the operator must delete that object created by the testcode? How could the operator know that it should that object crated externally which is not owned by the deployment?

pohly: > How could the operator know that it should that object crated externally which is not owned by the deployment?

You can create that extra object with the deployment as owner. As long as it is of a kind which the operator supports, it'll find it - after all, that is the code that we want to test.

pohly: You seem to be under the impression that only the controller can set the ownership relationship. That is not the case. The owner reference is a field (almost) like any other, with the main RBAC check being that write permission is needed for the object where the field is being set. The additional check is that blocking deletion of other objects by referencing them also requires write permission for that other object (more specifically, its finalizers). None of that is a problem for the test code because it runs with admin rights.

avalluri: Now I rewrote the test as you suggested, it creates an object explicitly, and checks if that gets deleted on a fresh reconcile.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Removed the found list.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Removed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Rephrased the comment.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Removed the version annotation.

pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
pohly: Which E2E test fails?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -37,6 +43,25 @@ const (
 	nodeControllerPort    = 10001
 )
 
+// A list of all object types potentially created by the operator,
+// in this or any previous release. In other words, this list may grow,
+// but never shrink, because a newer release needs to delete objects
+// created by an older release.
+// This list also should keep in sync with the operator RBAC rules.
```
pohly, line 50: s/should keep/must be kept/

Add the same comment in the declaration of the RBAC rules, please.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -203,15 +237,61 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
 		}
+
+		for _, obj := range list.Items {
+			for _, owner := range obj.GetOwnerReferences() {
+				if owner.UID == d.GetUID() {
+					obsolete, err := objectIsObsolete(newObjects, obj)
+					if err != nil {
+						return err
+					}
+					if obsolete {
```
pohly, line 275: You can avoid deep indention by using "continue":

```
if !obsolete {
   continue
}
```

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -203,15 +237,61 @@ func (d *PmemCSIDriver) reconcileDeploymentChanges(r *ReconcileDeployment, chang
...
+			Namespace: d.namespace,
+		}
+
+		klog.Infof("Fetching '%s' list with options: %v", gvk, opts.Namespace)
+		if err := r.client.List(context.TODO(), list, opts); err != nil {
 			return err
 		}
+
+		for _, obj := range list.Items {
+			for _, owner := range obj.GetOwnerReferences() {
```
pohly, line 269: Better also move the "is object owned by deployment" into a function.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -159,16 +159,21 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
...
 		klog.Infof("Updating deployment status....")
 		d.Deployment.Status.LastUpdated = metav1.Now()
 		if statusErr := r.client.Status().Update(context.TODO(), d.Deployment); statusErr != nil {
 			klog.Warningf("failed to update status %q for deployment %q: %v",
 				d.Deployment.Status.Phase, d.Name, statusErr)
 		}
+
+		if err := r.client.Patch(context.TODO(), d.Deployment, patch); err != nil {
+			klog.Warningf("Failed update deployment %q: %s", d.GetName(), err)
+		}
```
pohly, line 176: This doesn't belong into this commit, does it?

pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
```diff
@@ -547,6 +548,62 @@ func TestDeploymentController(t *testing.T) {
 				})
 			}
 		})
+
+		t.Run("delete obsolete objects", func(t *testing.T) {
```
pohly, line 552: We also need a negative test: "not delete unrelated objects".

pohly (changes requested):

pohly:
Given that all three commits are independent, wouldn't it be better to also submit them in separate PRs?

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: It is to fix the issue observed in test failure in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-718/12/tests

```
Sep 9 13:36:13.470: INFO: got expected driver deployment pmem-csi-with-defaults
pmem-csi-operator-84b674657d-4xsj8/pmem-csi-operator@pmem..ker2: W0909 13:36:13.457821 1 deployment_controller.go:172] failed to update status "Running" for deployment "pmem-csi-with-defaults": Operation cannot be fulfilled on deployments.pmem-csi.intel.com "pmem-csi-with-defaults": the object has been modified; please apply your changes to the latest version and try again
```

pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
avalluri: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-718/11/tests/

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Done.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Done.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
pohly: I don't like unsolved mysteries like "test randomly fails". Let's take this fix out of this PR and discuss it separately.

pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
avalluri: > We also need a negative test: "not delete unrelated objects".

Extended the test to this.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
pohly: So the change is needed for the PR, but it still doesn't belong into this commit, right?

Should this hunk here be in the "patch deployment status instead of update"?

Or if it is unrelated to that one, put it into a separate commit with a separate explanation. As it stands now, nothing in the "delete obsolete objects of a deployment" commit message explains why this code here is being changed.

Sorry for being pedantic, but after having let an unrelated commit through in another PR I am now taking a stricter position.

pohly (changes requested):

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: Sorry, this change is not related to the above-mentioned issue of updating status. It was part initial plan of updating deployment with annotations. I reverted this.

pohly (approved):
This now looks clean and consistent. Let's merge after a succesful CI test run.

PR #719, opened 2020-09-01
operator: display additional info on `kubectl get`

### LINKED ISSUE #541: operator: additional columns for "kubectl get"
```
$ kubectl get deployment.pmem-csi.intel.com 
NAME                 AGE
example-deployment   115s
```

Can we get additional columns listed here? The "Phase" would be useful.

avalluri:
Showing additional info on `kubectl get` of deployments is useful for
the user. Currently, we are showing DeviceMode, NodeSelector, Image and
Status fields. One caveat is the default values are not shown as they do
not appear in the deployment spec.

FIXES #541


docs/install.md
```diff
@@ -202,8 +202,8 @@ Once the above deployment installation is successful, we can see all the driver
 pods in `Running` state:
 ``` console
 $ kubectl get deployments.pmem-csi.intel.com
-NAME                 AGE
-pmem-deployment      50s
+NAME                 DEVICEMODE   NODESELECTOR   IMAGE   AGE   STATUS
+pmem-deployment      lvm                                 50s   Running
```
pohly, line 206: Other types list the age as last field:
```
$ kubectl get all --all-namespaces
NAMESPACE     NAME                                                    READY   STATUS    RESTARTS   AGE
kube-system   pod/coredns-66bff467f8-tnshp                            1/1     Running   0          122m
kube-system   pod/coredns-66bff467f8-tpstn                            1/1     Running   0          122m
kube-system   pod/etcd-pmem-csi-pmem-govm-master                      1/1     Running   0          122m
kube-system   pod/kube-apiserver-pmem-csi-pmem-govm-master            1/1     Running   0          122m
kube-system   pod/kube-controller-manager-pmem-csi-pmem-govm-master   1/1     Running   0          122m
kube-system   pod/kube-proxy-5gbns                                    1/1     Running   0          122m
kube-system   pod/kube-proxy-bq86z                                    1/1     Running   0          122m
kube-system   pod/kube-proxy-gf4hj                                    1/1     Running   0          122m
kube-system   pod/kube-proxy-ggtxp                                    1/1     Running   0          122m
kube-system   pod/kube-scheduler-pmem-csi-pmem-govm-master            1/1     Running   0          122m
kube-system   pod/weave-net-5h9sb                                     2/2     Running   1          122m
kube-system   pod/weave-net-l4pkw                                     2/2     Running   1          122m
kube-system   pod/weave-net-rb8tk                                     2/2     Running   0          122m
kube-system   pod/weave-net-rj4zq                                     2/2     Running   0          122m

NAMESPACE     NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)                  AGE
default       service/kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP                  122m
kube-system   service/kube-dns     ClusterIP   10.96.0.10   <none>        53/UDP,53/TCP,9153/TCP   122m

NAMESPACE     NAME                        DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
kube-system   daemonset.apps/kube-proxy   4         4         4       4            4           kubernetes.io/os=linux   122m
kube-system   daemonset.apps/weave-net    4         4         4       4            4           <none>                   122m

NAMESPACE     NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
kube-system   deployment.apps/coredns   2/2     2            2           122m

NAMESPACE     NAME                                 DESIRED   CURRENT   READY   AGE
kube-system   replicaset.apps/coredns-66bff467f8   2         2         2       122m
```

IMHO we should follow that.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -111,6 +111,11 @@ type DeploymentStatus struct {
 // Deployment is the Schema for the deployments API
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=deployments,scope=Namespaced
+// +kubebuilder:printcolumn:name="DeviceMode",type=string,JSONPath=`.spec.deviceMode`
+// +kubebuilder:printcolumn:name="NodeSelector",type=string,JSONPath=`.spec.nodeSelector`
+// +kubebuilder:printcolumn:name="Version",type=string,JSONPath=`.spec.image`
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.phase`
+// +kubebuilder:printcolumn:name="Age",type=data,JSONPath=`.metatdata.creationTimestamp`
```
pohly, line 118: What is the effect of this?

"Version" is mapped to ".spec.image" here, which is not consistent with `AdditionalPrinterColumns` below.

pohly (changes requested):

docs/install.md
avalluri: Ok. Pushed the change.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: It was by mistake, I was initially thinking of showing only the image version and later decided to show the full image. Now I fixed this.

pohly: I still don't understand why we have two different places for defining the "kubectl get" output. How are they related? Which one is used where? Do we need to maintain both manually?

avalluri: One is for making it work when installed the CRD generated using the kube-builder tool and the other is CRD installed from within the operator.

pohly: Why do we have two ways of installing the CRD? When do we use which approach?

If this cannot be avoided, then at least cross-reference the two places with comments so that when someone else makes a change in one place they'll see that there is another that they also need to update.

avalluri: > If this cannot be avoided, then at least cross-reference the two places with comments so that when someone else makes a change in one place they'll see that there is another that they also need to update.

Done.

pohly: > One is the CRD installed by the operator itself, the other is generated CRD installed explicitly when the operator installed via OperatorHub.

This still doesn't explain the "why": why do we have this the redundancy? Which of the two methods can be removed?

We shouldn't have two different ways of doing the same thing if it can be avoided. It's a source for bugs (see initial version of this PR...).

pohly: Installing the CRD from inside the operator implies that it must have RBAC permission to do so. It seems simpler and more secure to me to remove the CRD creation from the operator.

avalluri: > Installing the CRD from inside the operator implies that it must have RBAC permission to do so. It seems simpler and more secure to me to remove the CRD creation from the operator.

Yes, we can drop the CRD creation. Let's do it in a separate PR/issue.

pohly: Can we do that first and then rebase this one here? It makes no sense to introduce something here which will just be removed again.

pohly (approved):

PR #720, opened 2020-09-02
deploy: retry after node port reuse error

pohly:
Sometimes when quickly deleting a deployment and recreating it,
Kubernetes rejects a service with
   nodePort: Invalid value: 32000: provided port is already allocated

Kubernetes 1.19 might be a bit better in this regard, but it's unclear
whether this really addresses the issue in our case and we also need
to work with older releases, so we emulate the behavior of an operator
and retry until it succeeds.

Because the error might really be permanent, we time out after one
minute.

### DIFF
--- a/test/setup-deployment.sh
+++ b/test/setup-deployment.sh
@@ -184,7 +184,25 @@ EOF
 EOF
                 ;;
         esac
-        ${KUBECTL} apply --kustomize "$tmpdir/my-deployment"
+        # When quickly taking down one installation of PMEM-CSI and recreating it, sometimes we get:
+        #   nodePort: Invalid value: 32000: provided port is already allocated
+        #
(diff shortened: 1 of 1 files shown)


test/setup-deployment.sh
```diff
@@ -184,7 +184,25 @@ EOF
 EOF
                 ;;
         esac
-        ${KUBECTL} apply --kustomize "$tmpdir/my-deployment"
+        # When quickly taking down one installation of PMEM-CSI and recreating it, we
+        # sometimes we get: nodePort: Invalid value: 32000: provided port is already allocated
```
avalluri, line 188: Please remove one 'we' in 'we sometimes we get'.

avalluri (approved):
Other than the typo in the code comment, looks good to me.

test/setup-deployment.sh
pohly: Fixed. Can we merge without waiting for another test round?

PR #721, opened 2020-09-02
build: document and distribute copyleft code

pohly:
golang-lru is MPL-2.0 and needs to be distributed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,22 @@ RUN set -x && \
     cp /go/LICENSE /usr/local/share/package-licenses/go.LICENSE && \
     cp LICENSE /usr/local/share/package-licenses/PMEM-CSI.LICENSE
 
+# Some of the licenses might require us to distribute source code.
+# We cannot just point to the upstream repos because those might
+# disappear. We could host a copy at a location under our control,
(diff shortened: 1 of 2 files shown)


pohly:
@mythi FYI. Perhaps you can even review this?

avalluri (approved):

Dockerfile
```diff
@@ -96,6 +96,22 @@ RUN set -x && \
     cp /go/LICENSE /usr/local/share/package-licenses/go.LICENSE && \
     cp LICENSE /usr/local/share/package-licenses/PMEM-CSI.LICENSE
 
+# Some of the licenses might require us to distribute source code.
+# We cannot just point to the upstream repos because those might
+# disappear. We could host a copy at a location under our control,
+# but keeping that in sync with the published container images
+# would be tricky. So what we do instead is copy the (small!)
+# source code which has this requirement into the image.
+RUN set -x && \
```
mythi, line 105: what was the reason to not have it in a separate script, like `hack/copy-module-licenses.sh`?

pohly: No particular reason. It seemed simple enough to do it directly in the Dockerfile.

PR #724, opened 2020-09-02
operator: Initialize deployment CR defaults at right place

avalluri:
Set deployment CR defaults before finding the changes. Otherwise we get
wrong change list.


pohly:
This looks alright, but so did the first attempt (https://github.com/intel/pmem-csi/commit/940eb5d3aa97c56063daecdf9a2a06161513aea1). Can you explain in the code why it has to be the way it is now?

avalluri:
> This looks alright, but so did the first attempt ([940eb5d](https://github.com/intel/pmem-csi/commit/940eb5d3aa97c56063daecdf9a2a06161513aea1)). Can you explain in the code why it has to be the way it is now?

This way we first record the 'NewDeployment' event before it fails if there is any failure in `EnsureDefaults()`.

pohly (approved):
I meant add that to the code, but okay.

PR #731, opened 2020-09-11
operator/test: fix update pmemPercentage test failure

avalluri:
Incrementing the default pmemPerencentage of 100 leads to out of
bounds(101) and tests fails as below:

```
Status: "Failure",
Message: "Deployment.pmem-csi.intel.com \"pmem-csi-with-defaults\" is invalid: .... : validation failure list:\nspec.pmemPercentage in body should be less than or equal to 100",
```

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
+++ b/pkg/pmem-csi-operator/controller/deployment/testcases/testcases.go
@@ -69,7 +69,7 @@ func UpdateTests() []UpdateTest {
 			}
 		},
 		"pmemPercentage": func(d *api.Deployment) {
-			d.Spec.PMEMPercentage++
+			d.Spec.PMEMPercentage--
 		},
 		"labels": func(d *api.Deployment) {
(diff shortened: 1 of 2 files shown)


pohly:
The change itself looks reasonable, but I'd still like to know under which circumstances this becomes a problem.

pohly:
*With* this patch applied there are test failures:

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-731/1/tests

> Message: "Deployment.pmem-csi.intel.com \"pmem-csi-with-defaults\" is invalid: spec.pmemPercentage: Invalid value: 100: spec.pmemPercentage in body should be less than or equal to 100",

avalluri:
Looks like the original failure observed in https://github.com/intel/pmem-csi/pull/718 was due to result of other changes(I guess patch updating the deployment, witch updated the default value) and now this change is not needed any more.

PR #734, opened 2020-09-16
test: Kubernetes 1.19.1 without EndpointSlices

pohly:
Kubernetes 1.19.1 may have some relevant bug fixes. More importantly,
EndpointSlices are now also disabled for it by disabling the feature
that uses them in kube-proxy. That part was missing before, causing
kube-proxy to fail.

Perhaps this will increase stability of Kubernetes 1.19 testing. There
have been random failures recently during sanity testing where
connecting to the node port failed.


pohly:
@avalluri testing passed, so it at least didn't get worse :sweat_smile: 

Please review and merge.

avalluri (approved):

PR #738, opened 2020-09-18
test: update docker + containerd

pohly:
We didn't lock down the containerd version, which caused cluster
startup to fail when containerd was updated to 1.3.7 and the
configuration format changed.

### DIFF
--- a/test/e2e/deploy/deploy.go
+++ b/test/e2e/deploy/deploy.go
@@ -194,6 +194,9 @@ func WaitForPMEMDriver(c *Cluster, name, namespace string) (metricsURL string) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pohly:
@avalluri: this PR didn't pass testing because of the sanity connection issues, but PR #739 which includes it did pass.

Perhaps review PR #739 and we'll merge both at the same time?

pohly:
containerd 1.3.7 has an issue, let's address that here. Do not merge yet...

pohly:
> containerd 1.3.7 has an issue, let's address that here. Do not merge yet...

Added 1e13890f for this. PR can be merged once testing is done.

PR #739, opened 2020-09-18
debug connection issues

pohly:
This builds on top of https://github.com/intel/pmem-csi/pull/738 and adds commits that may address the sanity test connection issues.

### DIFF
--- a/test/e2e/deploy/deploy.go
+++ b/test/e2e/deploy/deploy.go
@@ -194,6 +194,9 @@ func WaitForPMEMDriver(c *Cluster, name, namespace string) (metricsURL string) {
 	}
 }
 
+// https://github.com/containerd/containerd/issues/4068
+var containerdTaskError = regexp.MustCompile(`failed to (start|create) containerd task`)
+
 // CheckPMEMDriver does some sanity checks for a running deployment.
 func CheckPMEMDriver(c *Cluster, deployment *Deployment) {
 	pods, err := c.cs.CoreV1().Pods(deployment.Namespace).List(context.Background(),
@@ -206,7 +209,12 @@ func CheckPMEMDriver(c *Cluster, deployment *Deployment) {
 	for _, pod := range pods.Items {
 		for _, containerStatus := range pod.Status.ContainerStatuses {
 			if containerStatus.RestartCount > 0 {
-				framework.Failf("container %q in pod %q restarted %d times, last state: %+v",
+				print := framework.Failf
+				if containerdTaskError.MatchString(fmt.Sprintf("%v", containerStatus.LastTerminationState)) {
+					// This is a known issue in containerd, only document it.
+					print = framework.Logf
+				}
+				print("container %q in pod %q restarted %d times, last state: %+v",
 					containerStatus.Name,
 					pod.Name,
 					containerStatus.RestartCount,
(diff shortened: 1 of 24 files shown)


test/e2e/deploy/deploy.go
```diff
@@ -758,7 +791,7 @@ func (d *Deployment) GetDriverDeployment() api.Deployment {
 			Labels: map[string]string{
 				deploymentLabel: d.Name,
 			},
-			// TODO: replace pmemcsidriver.DeviceMode with api.DeviceMode everywhere
+			// TODO: replace pmemcsidrive1r.DeviceMode with api.DeviceMode everywhere
```
avalluri, line 814: Can you please remove this unintentional change.

avalluri (approved):

test/e2e/deploy/deploy.go
pohly: Fixed.

pohly:
In https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/version-skew-2/17/testReport/junit/fedora-1_18.lvm-testing/sanity/Testing___1_18___PMEM_CSI_can_publish_volume_after_a_node_driver_restart/ there was another `dial tcp 172.17.0.5:9735: connect: connection refused` error despite including these changes here.

Full log in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/version-skew-2/17/execution/node/138/log/?consoleFull

So while this PR here may help a bit (more obvious error reporting, avoid excessive CI test runtime), it doesn't help when the cluster gets into that state where the socat port cannot be reached.

I'm wondering whether this started to happen because the docker.io/alpine/socat image changed. We don't lock it down to a specific version.

pohly:
> I'm wondering whether this started to happen because the docker.io/alpine/socat image changed. We don't lock it down to a specific version.

No, https://hub.docker.com/r/alpine/socat/tags shows that "latest" hasn't changed for over 8 months.

pohly:
I think I have finally nailed the root cause for the "connection refused" issue - see commit f4a7062.

@avalluri : please review, then we can merge this and thus also PR #738.

test/e2e/deploy/deploy.go
```diff
@@ -90,8 +98,14 @@ func WaitForPMEMDriver(c *Cluster, name, namespace string, testing bool) (metric
...
 	info := time.NewTicker(time.Minute)
 	defer info.Stop()
-	deadline, cancel := context.WithTimeout(context.Background(), framework.TestContext.SystemDaemonsetStartupTimeout)
+	deadline, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
 	defer cancel()
+
+	if waitForPMEMDriverTimedOut {
+		// Abort early.
+		skipper.Skipf("installing PMEM-CSI driver during previous test was too slow")
+	}
```
pohly, line 107: I think this is indeed useful: in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/version-skew-2/19/tests the same driver install error would have happened for all skew tests, so instead of repeating that over and over, testing runs into that issue once and then skips the rest of the testing.

We just have to remember that "only one failed test" no longer means "all other tests were okay".

pohly: FWIW, that issue is known and a result of rebasing on top of the metrics code: now the version skew testing must switch between http and https depending on the version.

avalluri: Yes, Let's keep this. It's useful for quick fail.

avalluri (approved):

Issue #742, opened 2020-09-21
test flake: updating labels in default deployment while running – fedora-1_19.olm-operator.API


pohly:
From https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-738/2/tests:

```
/mnt/workspace/pmem-csi_PR-738/test/e2e/operator/deployment_api.go:461
Sep 21 10:17:28.182: validate driver after update and restart
Unexpected error:
<*errors.errorString | 0xc003051a70>: {
s: "timed out waiting for deployment, last error: deployed driver different from expected deployment:\nlabel foo missing for \"pmem-csi-with-defaults-external-provisioner-cfg\" of type \"rbac.authorization.k8s.io/v1, Kind=Role\" in namespace \"default\"\nlabel foo missing for \"pmem-csi-with-defaults-external-provisioner-runner\" of type \"rbac.authorization.k8s.io/v1, Kind=ClusterRole\" in namespace \"\"\nlabel foo missing for \"pmem-csi-with-defaults-csi-provisioner-role-cfg\" of type \"rbac.authorization.k8s.io/v1, Kind=RoleBinding\" in namespace \"default\"\nlabel foo missing for \"pmem-csi-with-defaults-csi-provisioner-role\" of type \"rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding\" in namespace \"\"\nlabel foo missing for \"pmem-csi-with-defaults-controller\" of type \"/v1, Kind=ServiceAccount\" in namespace \"default\"\nlabel foo missing for \"pmem-csi-with-defaults-controller\" of type \"/v1, Kind=Service\" in namespace \"default\"\nlabel foo missing for \"pmem-csi-with-defaults-metrics\" of type \"/v1, Kind=Service\" in namespace \"default\"\nlabel foo missing for \"pmem-csi-with-defaults\" of type \"storage.k8s.io/v1beta1, Kind=CSIDriver\" in namespace \"\"",
}
timed out waiting for deployment, last error: deployed driver different from expected deployment:
label foo missing for "pmem-csi-with-defaults-external-provisioner-cfg" of type "rbac.authorization.k8s.io/v1, Kind=Role" in namespace "default"
label foo missing for "pmem-csi-with-defaults-external-provisioner-runner" of type "rbac.authorization.k8s.io/v1, Kind=ClusterRole" in namespace ""
[...]

PR #743, opened 2020-09-23
operator: Fix issue in deployment comparison

avalluri:
With the recent commit 761d7cef0b96ea3e2ed559c5e89e7f918b2112b9, the
spec changes made to a deployment are get reverted and not been saved
even in the deployment cache. But when we compare this cached object
against the new deployment (with defaults set) result those default
fields as changes.

This could be fixed by resetting those defaults to cached deployment
before comparison.

Also removed unneeded deep copy of deployment object while passing to
client.MergeFrom().

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -91,7 +91,16 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
 	}
 
 	if foundInCache {
-		changes = d.Compare(oldDeployment)
+		// Cached deployment does not have defaults stored in it.
+		// So, reset those defaults before comparing otherwise
+		// those set defaults `d.Deployment` will be detected as
+		// changes.
+		cached := oldDeployment.DeepCopy()
+		if err := cached.EnsureDefaults(r.containerImage); err != nil {
+			r.evRecorder.Event(d, corev1.EventTypeWarning, api.EventReasonFailed, err.Error())
+			return true, err
+		}
+		changes = d.Compare(cached)
 	}
 
 	klog.Infof("Deployment: %q, state %q, changes %v, in cache %v", d.Name, d.Status.Phase, changes, foundInCache)
--- a/pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
+++ b/pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
@@ -177,10 +177,14 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	defer func() {
 		klog.Infof("Updating deployment status....")
 		d.Deployment.Status.LastUpdated = metav1.Now()
-		if statusErr := r.client.Status().Patch(context.TODO(), d.Deployment, patch); statusErr != nil {
+		// Passing copy of CR to patch as the fake client used in tests
+		// will write back the changes to both status and spec.
+		copy := d.Deployment.DeepCopy()
+		if statusErr := r.client.Status().Patch(context.TODO(), copy, patch); statusErr != nil {
 			klog.Warningf("failed to update status %q for deployment %q: %v",
 				d.Deployment.Status.Phase, d.Name, statusErr)
 		}
+		d.Deployment.Status = copy.Status
 	}()
 
 	requeue, err = d.Reconcile(r)


pohly (commented):
Why were tests failing only sometimes?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -91,7 +91,16 @@ func (d *PmemCSIDriver) Reconcile(r *ReconcileDeployment) (bool, error) {
...
+		// Cached deployment does not have defaults stored in it.
+		// So, reset those defaults before comparing otherwise
+		// those set defaults `d.Deployment` will be detected as
+		// changes.
+		cached := oldDeployment.DeepCopy()
+		if err := cached.EnsureDefaults(r.containerImage); err != nil {
+			r.evRecorder.Event(d, corev1.EventTypeWarning, api.EventReasonFailed, err.Error())
+			return true, err
+		}
+		changes = d.Compare(cached)
```
pohly, line 103: This whole "sometimes a deployment has defaults, sometimes not" is a mess that has already caused two different bugs (first storing the modified CR in the apiserver, now comparing an instance with defaults against one without).

Can we avoid this once and for all? For example, remove `EnsureDefaults` and instead introduce `Get` methods that return the value if set and the default if not?

avalluri: > This whole "sometimes a deployment has defaults, sometimes not" is a mess that has already caused two different bugs (first storing the modified CR in the apiserver, now comparing an instance with defaults against one without).

The first one was not really an issue, That was a work around fix for fake client(used in tests) issue of not supporting status patching. Infact, saving modified CR to API server is useful for the user too for knowing what defaults are used. I think few controllers does this.
> 
> Can we avoid this once and for all? For example, remove `EnsureDefaults` and instead introduce `Get` methods that return the value if set and the default if not?

I afraid, there fields are accessed quite many places in overall the code and quite possible that we might miss calling those Get calls and access directly the field.

pohly (changes requested):
Regardless how we address this, we need a test that reliably triggers the issue (before fixing it) and then no longer fails (after fixing it).

Also, the proposed change caused "make test" to fail:
```
Error:      	Not equal: 

[2020-09-23T12:41:57.615Z]                     	            	expected: "Running"

[2020-09-23T12:41:57.615Z]                     	            	actual  : ""

[2020-09-23T12:41:57.615Z]                     	            	

[2020-09-23T12:41:57.615Z]                     	            	Diff:

[2020-09-23T12:41:57.615Z]                     	            	--- Expected

[2020-09-23T12:41:57.615Z]                     	            	+++ Actual

[2020-09-23T12:41:57.615Z]                     	            	@@ -1,2 +1,2 @@

[2020-09-23T12:41:57.615Z]                     	            	-(v1alpha1.DeploymentPhase) (len=7) "Running"

[2020-09-23T12:41:57.615Z]                     	            	+(v1alpha1.DeploymentPhase) ""
```

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: > Infact, saving modified CR to API server is useful for the user too for knowing what defaults are used. 

No, this is not useful. The defaults are meant to change over time (think updating the operator, which then comes with a new default image and associated object generation rules). Storing them in the apiserver forever locks them to what the defaults where when first creating the CR.

> I afraid, there fields are accessed quite many places in overall the code and quite possible that we might miss calling those Get calls and access directly the field.

Then rename the fields temporarily to find all those locations.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -177,10 +177,14 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	defer func() {
 		klog.Infof("Updating deployment status....")
 		d.Deployment.Status.LastUpdated = metav1.Now()
-		if statusErr := r.client.Status().Patch(context.TODO(), d.Deployment, patch); statusErr != nil {
+		// Passing copy of CR to patch as the fake client used in tests
+		// will write back the changes to both status and spec.
+		copy := d.Deployment.DeepCopy()
+		if statusErr := r.client.Status().Patch(context.TODO(), copy, patch); statusErr != nil {
```
pohly, line 183: This code is not obvious. Why is `d.Deployment.Status` getting modified?

pohly: After re-reading this a few times it became clearer.

```
Beware that r.client.Status() as implemented by the fake client during testing will
patch not just the status, but also the spec. We don't want that and therefore
let it work with a copy of the deployment, then later update just the status.
```

pohly: Is this a fix for a different problem than the other commit?

The commit message describe what the patch does, but not why we want it.

avalluri: `Patch()` is updating the passed `obj` with patch result: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/fake/client.go#L353-L403

avalluri: > Is this a fix for a different problem than the other commit?
> 
The first commit revealed this. The `modified_deployment_under_reconcile` test is failing after the change. This test expects   modified changes are supposed to visible in the next reconcile loop, but operator could not figureout the change because the `status.Pactch()` updated the cached CR, and the next reconcile loop couldn't capture the change.

It was hidden previously because, reconcile used to redeploy both the controller, node driver even if there is no real change in deployment because of comparison always returns all default fields as changes.

avalluri:
> Regardless how we address this, we need a test that reliably triggers the issue (before fixing it) and then no longer fails (after fixing it).

I am not clear how to reliably trigger this from tests. The CR 'change detection' logic is internal to the reconciler and there is no direct way to test if the reconcile:
 - is not redeploying the other objects that the one expected by test, and
 - ensuring that the expected object is redeployed only just because of a particular change.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: > The defaults are meant to change over time (think updating the operator, which then comes with a new default image and associated object generation rules). Storing them in the apiserver forever locks them to what the defaults where when first creating the CR.

Let me add one more thought to this: these defaults are not even part of the API, exactly because they can change without changing the API. That is not the case for "normal" API constants - changing those is an API break.

Therefore it is unfortunate and confusing that the PMEM-CSI defaults are defined in `pkg/apis/pmemcsi/v1alpha1/deployment_types.go`. It would be better to define the constants in the operator, because that is what they are: the values that the operator uses when nothing is explicitly specified in the API.

pohly:
> I am not clear how to reliably trigger this from tests.

Perhaps we get further by addressing one aspect: why was it working sometimes, but not always? The tests were flaky, right?

Under which sequence of events did the test fail? If we can reconstruct that sequence reliably, then we have a reproducer.

avalluri:
> Under which sequence of events did the test fail? If we can reconstruct that sequence reliably, then we have a reproducer.

This is the tough part. For example, in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/version-skew-2/23/tests run the test failure was updating configured 'kubeletDir'. But the logs show that the operator could detect that the change, but still the resulted validation failed, which is hard to interpret the reason:

```
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.375103 1 deployment_controller.go:188] Requeue: false, error: <nil>
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.375140 1 deployment_controller.go:178] Updating deployment status....
Sep 23 11:12:07.486: INFO: got expected driver deployment pmem-csi-with-values
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.481666 1 controller_driver.go:97] Deployment: "pmem-csi-with-values", state "Running", changes map[kubeletDir:{}], in cache true
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.481972 1 controller_driver.go:195] Updating node driver for deployment "pmem-csi-with-values"
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.482476 1 deployment_controller.go:284] Updating 'default/pmem-csi-with-values-node' of type '*v1.DaemonSet'
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.513788 1 deployment_controller.go:188] Requeue: false, error: <nil>
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.513889 1 deployment_controller.go:178] Updating deployment status....
Sep 23 11:12:07.519: INFO: Updated deployment "pmem-csi-with-values"
Sep 23 11:12:07.519: INFO: waiting for expected driver deployment pmem-csi-with-values
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.532554 1 controller_driver.go:97] Deployment: "pmem-csi-with-values", state "Running", changes map[], in cache true
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.535290 1 deployment_controller.go:188] Requeue: false, error: <nil>
pmem-csi-operator-c6f6d6ddd-t9zl8/pmem-csi-operator@pmem..ker3: I0923 11:12:07.535702 1 deployment_controller.go:178] Updating deployment status....
Sep 23 11:17:06.159: FAIL: validate driver after update and restart
```

pohly:
> Under which sequence of events did the test fail?

Was it perhaps so that the tests looked at the state right after the initial reconciliation and before the status update followed by sending back the modified CR to the operator triggers another reconciliation? At that point the objects are okay and the tests pass.

But if the test misses that time window, then the next reconciliation uses the broken copy in the cache and breaks the deployed objects, which then gets flagged by the test.

The fix for the test then is to not just check with `Eventually`, but also with `Consistently` that the state is stable.

pohly (approved):
After discussion on chat I understand this PR better. Key points:
- It avoids unnecessary updates of objects.
- It does not fix issue #742, or if it does, we don't know why.

The change itself is okayish. Some parts of it can't be understood without the comments here in the PR and I would have preferred to get rid of the whole "modify object, use it, change it back" dance, but that can also be handled separately.

The PR also lacks a test, but I have one ready that I can submit separately.

Let's merge...

PR #744, opened 2020-09-24
operator testing

pohly:
This is the regression test for PR #743.

Also includes one drive-by comment fix.


avalluri (approved):
IMHO waiting for 10 seconds to check the deployment consistency for each test might increase `make test` runtime.

pohly:
> IMHO waiting for 10 seconds to check the deployment consistency for each test might increase make test runtime.

These 10 seconds are not applied when using `make test`. They are used when doing E2E testing, which already is kind of slow so the additional overhead is (relatively speaking) smaller.

Issue #611, opened 2020-04-22
operator API v1alpha2: revise Status


pohly:
We should align with https://github.com/kubernetes/enhancements/pull/1624.

pohly:
@avalluri please describe what you intend to do regarding the status field in the CRD.

avalluri:
@pohly, here is the draft of the new status type roughly look:

```
// DeploymentConditionType type for representing a deployment status condition
type DeploymentConditionType string

const (
	// CertsVerified means the provided deployment secrets are verified and valid for usage
	CertsVerified DeploymentConditionType = "CertsVerified"
	// CertsReady means secrests/certificates required for running the PMEM-CSI driver
	// are ready and the deployment could progress further
	CertsReady DeploymentConditionType = "CertsReady"
	// DriverDeployed means that the node driver deamonset got created
	DriverDeployed DeploymentConditionType = "DriverDeployed"
)

// +k8s:deepcopy-gen=true
type DeploymentCondition struct {
	// Type of condition.
	Type DeploymentConditionType `json:"type"`
	// Status of the condition, one of True, False, Unknown.
	Status corev1.ConditionStatus `json:"status"`
	// Message human readable text that explain why this condition is in this state
	// +optional
	Message string `json:"reason,omitempty"`
	// Last time the condition was probed.
	// +optional
	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
}

type DriverType string

const (
	ControllerDriver DriverType = "ControllerDriver"
	NodeDriver       DriverType = "NodeDriver"
)

// +k8s:deepcopy-gen=true
type DriverStatus struct {
	// Type represents type of the driver: controller or node
	Type DriverType `json:"type"`
	// Status represents the driver status : Ready, NotReady
	Status string `json:"status"`
	// Reason represents the human readable text that explains why the
	// driver is in this state.
	Reason string `json:"message"`
}

// +k8s:deepcopy-gen=true
// DeploymentStatus defines the observed state of Deployment
type DeploymentStatus struct {
	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
	// Important: Run "make operator-generate-k8s" to regenerate code after modifying this file

	// Phase indicates the state of the deployment
	Phase DeploymentPhase `json:"phase,omitempty"`
	// Conditions
	Conditions []DeploymentCondition `json:"conditions,omitempty"`
	Components []DriverStatus        `json:"components,omitempty"`
[...]

PR #749, opened 2020-09-27
operator: revise deployment status API and operator reconcile loop

avalluri:
Revised the Deployment.Status to accommodate the deployment state
conditions and driver state. Currently Deployment has 3 conditions named
CertsVerified, CertsReady, and DriverDeployed. It also records summary
of controller and node driver state, .i.e, no. of nodes the driver is
running.

In order to record real time status of the driver current had to rewrite
the current reconcile loop. The existing reconcile loop was keen on the
deployment CR changes and redeploy *only* the sub-objects that requires
to redeploy. Instead the new reconcile logic *refresh* all the objects
and CR status to keep the state consistent. The refresh chooses to
merge patching the objects to avoid all unnecessary updates.

There are two reconcile entry points:
- CR reconcile loop: refreshes all the sub-objects and CR status
- sub-object event handler: redeploy only the deleted/changed resource
and updates CR status if required.

This also includes other code cleanups that come across.

TODOs:
 - [x] E2E tests for validating if the operator restores the state of a
 broken deployment - i.e, recovering deleted/modified sub-objects.
 - [x] Investigate the "unexpected object update" failures.
 - [x] Update documentation bits to show new status fields.

FIXES: #611


pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -109,13 +110,76 @@ type DeploymentSpec struct {
...
+	}
+	return ""
+}
+
+// +k8s:deepcopy-gen=true
+type DriverStatus struct {
+	// Type represents type of the driver: controller or node
+	Type string `json:"type"`
+	// Status represents the driver status : Ready, NotReady
+	Status string `json:"status"`
```
pohly, line 163: Please document here what "ready" means for the driver status.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -109,13 +110,76 @@ type DeploymentSpec struct {
...
+		return "controller"
+	case NodeDriver:
+		return "node"
+	}
+	return ""
+}
+
+// +k8s:deepcopy-gen=true
+type DriverStatus struct {
+	// Type represents type of the driver: controller or node
```
pohly, line 160: Do we call this "type of the driver" elsewhere?

pohly: When we talk about "driver deployment", I tend to think of it as the combination of all of its parts. "Driver component" seems more appropriate here.

pohly (changes requested):
I've not looked at the actual code yet. Let's get the current tests to work first. For that, please rebase.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: Do you mean to rename`Type` to `DriverComponent`?

pohly: Yes.

avalluri: Renamed the field. The status looks as below:

```
  Driver Components:
    Component:     Controller
    Last Updated:  2020-10-08T07:45:13Z
    Reason:        1 instance(s) of controller driver is running successfully
    Status:        Ready
    Component:     Node
    Last Updated:  2020-10-08T07:45:11Z
    Reason:        All 3 node driver pod(s) running successfully
    Status:        Ready
```

avalluri:
Split the change into 2 different commits; Revised reconcile loop and Revised status API.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -47,25 +53,128 @@ func Add(mgr manager.Manager, opts pmemcontroller.ControllerOptions) error {
...
+		&rbacv1.ClusterRoleBinding{},
+		&storagev1beta1.CSIDriver{},
+		&appsv1.DaemonSet{},
+		&rbacv1.Role{},
+		&rbacv1.RoleBinding{},
+		&corev1.Secret{},
+		&corev1.Service{},
+		&corev1.ServiceAccount{},
+		&appsv1.StatefulSet{},
+	}
```
pohly, line 166: Don't we already have such a list? If the redundancy cannot be avoided, then at least add comments in both places about keeping the lists in sync.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -303,6 +423,62 @@ func (r *ReconcileDeployment) Delete(obj runtime.Object) error {
 	return r.client.Delete(context.TODO(), obj)
 }
 
+// PatchDeploymentStatus patches the give given deployment CR status
```
pohly, line 428: "give given"

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
```diff
@@ -423,6 +495,30 @@ func (d *Deployment) GetOwnerReference() metav1.OwnerReference {
 	}
 }
 
+// HaveCertificatesConfigured checks if the configured deployment
+// certificate fields are valid. Returns true if valid else appropriate
+// error.
+func (d *Deployment) HaveCertificatesConfigured() (bool, error) {
```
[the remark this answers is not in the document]
avalluri: (false, nil) is a valid case, meaning no certificates provided assuming that the operator provisions the self-signed.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -47,25 +53,128 @@ func Add(mgr manager.Manager, opts pmemcontroller.ControllerOptions) error {
...
+				// Objects owned by it are automatically garbage collected.
+				r.deleteDeployment(e.MetaOld.GetName())
+				return false
+			}
+			if old.GetGeneration() == new.GetGeneration() {
+				// No changes registered
+				return false
+			}
+			// We are intersted in only spec changes, not CR status changes
+			changes := old.Compare(new)
```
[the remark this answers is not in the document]
avalluri: Below are points that led me to keep compare:
 - This makes it easy to understand by looking at logs what exactly changed that leads to "deployment reconcile"
 - "Update" gets invoked on each time status got updated, which we clearly can avoid reconciling. Otherwise, the logs look messy with quite many "redeploy...." statements.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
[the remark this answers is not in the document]
pohly: That's not what the comment says ("Returns true if valid else appropriate error."), so you need to fix the comment.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -47,25 +53,128 @@ func Add(mgr manager.Manager, opts pmemcontroller.ControllerOptions) error {
...
+			if e.MetaNew.GetDeletionTimestamp() != nil {
+				// We can handle this in delete handler
+				return false
+			}
+			klog.Infof("UPDATED: %T/%s", e.ObjectOld, e.MetaOld.GetName())
+			return eventFunc(e.MetaOld, e.ObjectOld)
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			// Do not handle sub-object create events
+			return false
```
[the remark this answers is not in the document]
avalluri: There is nothing we should do on object creation. obviously, those events are supposed to be the result of a previous reconcile loop.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
pohly: So basically we maintain code that we need to test for correctness just for logging.

Can we use some automated diff and print the result of that instead of manually writing the code? There's https://pkg.go.dev/github.com/google/go-cmp/cmp specifically for that purpose or we can generate a patch and print that (probably less readable, but avoids the extra dependency).

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
pohly: Theoretically someone might create objects that are owned by the CR which are not needed. We do remove those on operator startup, so it is worth calling out here *in the comment* that we intentionally don't do that because a) it's a rather obscure situation and b) we want to avoid doing work for the usual case where the object was created by us and thus don't need changes.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
avalluri: Add the description.

pkg/apis/pmemcsi/v1alpha1/deployment_types.go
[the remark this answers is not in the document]
avalluri: Rephrased the comment to match the implementation.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
avalluri: Added the comment, can you please check if that is sufficient.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: This list is different from the `AllObjectTypes` list defined in `controller_driver.go`. As suggested added a note in the comment.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -67,194 +67,490 @@ var AllObjectTypes = []schema.GroupVersionKind{
...
+	ss := &appsv1.StatefulSet{
+		TypeMeta:   metav1.TypeMeta{Kind: "StatefulSet", APIVersion: "apps/v1"},
+		ObjectMeta: d.getObjectMeta(d.ControllerDriverName(), false),
+	}
+	if err := d.getSubObject(r, ss); err != nil {
+		return nil, err
+	}
+	op := NewObjectPatch(ss, ss.DeepCopy())
+	d.getControllerStatefulSet(ss)
+	if err := d.updateSubObject(r, op); err != nil {
```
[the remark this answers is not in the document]
avalluri: Rewrote the code such that all redeploy* functions call a `redeploy()` with `RedeployObject` by setting the appropriate callback handlers. But one exception is `redeploySecrets` which is different from the pattern. It checks if the object is new after the fetching and before the update.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -303,6 +604,8 @@ func (d *PmemCSIDriver) deleteObsoleteObjects(r *ReconcileDeployment, newObjects
 			if !obsolete {
 				continue
 			}
+			klog.Infof("Deleting %q of type '%s' is obsolete object", obj.GetName(), obj.GetObjectKind().GroupVersionKind())
```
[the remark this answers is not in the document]
avalluri: Fixed it.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
avalluri: Replaced `Compare()` with  `Patch.Data()`, which does the json patching for computing the changes,.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -47,25 +53,128 @@ func Add(mgr manager.Manager, opts pmemcontroller.ControllerOptions) error {
...
+		},
+	}
+
+	// All possible object types for a Deployment CR
+	// that would requires recdeploying on change
```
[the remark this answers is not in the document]
avalluri: Fixed.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -67,194 +67,504 @@ var AllObjectTypes = []schema.GroupVersionKind{
...
+	return op.Apply(r.client, d.Spec.Labels)
+}
+
+type redeployObject struct {
+	object     func() apiruntime.Object
+	modify     func(apiruntime.Object)
+	postUpdate func(apiruntime.Object)
+}
+
+func (d *PmemCSIDriver) redeploy(r *ReconcileDeployment, ro redeployObject) (apiruntime.Object, error) {
```
pohly, line 264: It would be useful to explain each step in this function to help someone who is not familiar with the code understand what's going on. Don't just paraphrase the function names, but try to think of additional information that may be useful.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
pohly: You keep skipping the space before an opening bracket, but let's ignore that for now.

I'm not sure about the "that anyway gets removed in the next reconcile loop" argument - what if there is no such loop iteration for quite a while? The whole point of this callback function is to trigger such a reconciliation as soon as the extra object gets created.

The argument needs to be more along the lines of "this shouldn't happen in practice and checking whether it has happened is just going to cause extra overhead in the normal case of objects that were created by the operator".

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > It would be useful to explain each step in this function to help someone who is not familiar with the code understand what's going on.

Added the function description, hope that helps.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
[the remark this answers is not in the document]
avalluri: Okay, rephrased the comment by including the suggested text.

pohly (changes requested):

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: > I'm not convinced that we need a second list.

The AllObjectTypes in `controller_driver.go` might also contain objects that were obsolete now, whereas this watch list is supposed to have only active objects requires reconciling.

avalluri: > See [avalluri#3](https://github.com/avalluri/pmem-CSI/pull/3)

Splitting `AllObjectTypes` list into current + obsolete seems works fine. I took the patch. Thanks.

pohly (changes requested):
`make test` failed in the CI.

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-749/12/pipeline

```
[2020-10-28T08:27:44.157Z] •! Panic [0.006 seconds]
[2020-10-28T08:27:44.157Z] Operator
[2020-10-28T08:27:44.157Z] /mnt/workspace/pmem-csi_PR-749/pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go:31
[2020-10-28T08:27:44.157Z]   API
[2020-10-28T08:27:44.157Z]   /mnt/workspace/pmem-csi_PR-749/pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go:36
[2020-10-28T08:27:44.157Z]     should have valid json schema [It]
[2020-10-28T08:27:44.157Z]     /mnt/workspace/pmem-csi_PR-749/pkg/apis/pmemcsi/v1alpha1/deployment_types_test.go:112
[2020-10-28T08:27:44.157Z] 
[2020-10-28T08:27:44.157Z]     Test Panicked
[2020-10-28T08:27:44.157Z]     runtime error: invalid memory address or nil pointer dereference
[2020-10-28T08:27:44.157Z]     /go/src/runtime/panic.go:212
```

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -117,14 +247,16 @@ func NewReconcileDeployment(client client.Client, opts pmemcontroller.Controller
...
-		reconcileHooks: map[ReconcileHook]struct{}{},
+		client:           client,
+		evBroadcaster:    evBroadcaster,
+		evRecorder:       evRecorder,
+		k8sVersion:       opts.K8sVersion,
+		namespace:        opts.Namespace,
+		containerImage:   opts.DriverImage,
+		deployments:      map[string]*pmemcsiv1alpha1.Deployment{},
+		deploymentsMutex: sync.Mutex{},
+		reconcileMutex:   sync.Mutex{},
```
pohly, line 237: You can remove the initialization of `deploymentsMutex` and `reconcileMutex`, it's not needed.

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -85,8 +213,10 @@ type ReconcileDeployment struct {
 	// container image used for deploying the operator
 	containerImage string
 	// known deployments
-	deployments    map[string]*pmemcsiv1alpha1.Deployment
-	reconcileHooks map[ReconcileHook]struct{}
+	deployments      map[string]*pmemcsiv1alpha1.Deployment
+	deploymentsMutex sync.Mutex
+	reconcileMutex   sync.Mutex
```
pohly, line 197: Please add comments explaining what is protected by these mutexes and why that is necessary. `deploymentsMutex` seems relatively straightforward (protect concurrent access to `deployments`) but the logic around locking `reconcileMutex` isn't.

docs/install.md
```diff
@@ -1182,21 +1219,47 @@ active volumes.
 
 #### DeploymentStatus
 
-A PMEM-CSI Deployment's `status` field is a `DeploymentStatus` object, which has
-a `phase` field. The phase of a Deployment is high-level summary of where the
-Deployment is in it's lifecycle.
+A PMEM-CSI Deployment's `status` field is a `DeploymentStatus` object, which
+carries the detailed state of the driver deployment. It comprises of [deployment
```
pohly, line 1223: "It is comprised of"

docs/install.md
```diff
@@ -1182,21 +1219,47 @@ active volumes.
...
 | Running | The operator has determined that the driver is usable<sup>1</sup>.  |
-| Failed | For some reason the state of the `Deployment` failed and cannot be progressed<sup>2</sup>. |
+| Failed | For some reason the state of the `Deployment` failed and cannot be progressed. The failure reason is placed in the `DeploymentStatus.Reason` field. |
 
 <sup>1</sup> This check has not been implemented yet. Instead, the deployment goes straight to `Running` after creating sub-resources.
-<sup>2</sup> Failure reason is supposed to be carried by one of additional `DeploymentStatus` field, but not implemented yet.
+
+#### Deployment Conditions
+
+PMEM-CSI `DeploymentStatus` has an array of `conditions` through witch the
```
pohly, line 1240: s/witch/which/

docs/install.md
```diff
@@ -1182,21 +1219,47 @@ active volumes.
...
+PMEM-CSI Deployment has or has not passed. Below are the possible condition
+types and their meanings:
+
+| Condition type | Meaning |
+|---|---|
+| CertsReady | Driver certificates/secrets are available. |
+| CertsVerified | Verified that the provided certificates are valid. |
+| DriverDeployed | All the componentes required for the PMEM-CSI deployment has been deployed. |
```
pohly, line 1248: s/has/have/

docs/install.md
```diff
@@ -1182,21 +1219,47 @@ active volumes.
...
+
+#### Driver component status
+
+PMEM-CSI `DeploymentStatus` has an array of `components` of type `DriverStatus`
+in which the operator records the brief driver(Controller and Node) components
```
pohly, line 1253: Space before opening bracket. Do we need the brackets here at all?

Why are controller and node capitalized?

I think "the brief driver components status" might be enough. What the components are is explained below.

docs/install.md
```diff
@@ -276,36 +276,73 @@ pmem-deployment      lvm                                 Running  50s
...
 Events:
-  Type    Reason         Age                From               Message
-  ----    ------         ----               ----               -------
-  Normal  NewDeployment  34s                pmem-csi-operator  Processing new driver deployment
-  Normal  Running        2s (x10 over 26s)  pmem-csi-operator  Driver deployment successful
-
+  Type    Reason         Age   From               Message
+  ----    ------         ----  ----               -------
+  Normal  NewDeployment  58s   pmem-csi-operator  Processing new driver deployment
+  Normal  Running        39s   pmem-csi-operator  Driver deployment successful
```
pohly, line 345: Looks good. However, the "AGE" field in "kubectl get" is now (?) empty:
```
$ kubectl get deployments.pmem-csi.intel.com
NAME                 DEVICEMODE   NODESELECTOR         IMAGE   STATUS    AGE
pmem-csi.intel.com   lvm          {"storage":"pmem"}           Running   
```

Is that related to this PR?

pohly (changes requested):

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
avalluri: Added the field comments.

docs/install.md

docs/install.md

docs/install.md
avalluri: Change the text as suggested.

docs/install.md

docs/install.md
avalluri: There was a typo in the field name. https://github.com/intel/pmem-csi/pull/749/commits/374202455470d6e1f28eb555855ccf27fdc3d7ab fixes this.

pohly: Is this resolved?

pohly: Ah, "operator/crd: Fix typo in additional columns field" was added for it.

deploy/crd/pmem-csi.intel.com_deployments.yaml
```diff
@@ -30,7 +30,7 @@ spec:
     - jsonPath: .status.phase
       name: Status
       type: string
-    - jsonPath: .metatdata.creationTimestamp
+    - jsonPath: .metadata.creationTimestamp
```
pohly, line 33: We need a test for this and the other fields.

pohly (changes requested):

test/e2e/operator/deployment_api.go
```diff
@@ -144,6 +161,30 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+
+			deployment = deploy.CreateDeploymentCR(f, deployment)
+			defer deploy.DeleteDeploymentCR(f, deployment.Name)
+
+			// Run in-cluster kubectl from master node
+			ssh := os.Getenv("REPO_ROOT") + "/_work/" + os.Getenv("CLUSTER") + "/ssh.0"
+			out, err := exec.RunCommand(ssh, "kubectl", "get", "deployments.pmem-csi.intel.com", "--no-headers")
+			Expect(err).ShouldNot(HaveOccurred(), "kubectl get: %v", out)
+			fields := strings.Fields(string(out))
+			Expect(len(fields)).Should(BeEquivalentTo(expectedFiles), "fields mismatch: %v", out)
```
pohly, line 185: The new test isn't passing:
```
/mnt/workspace/pmem-csi_PR-749/test/e2e/operator/deployment_api.go:164
Nov 1 03:50:34.611: fields mismatch: test-get-deployment-fields direct map[storage:unknown-node] unexisting/pmem-csi-driver 1s
Expected
<int>: 5
to be equivalent to
<int>: 6
/mnt/workspace/pmem-csi_PR-749/test/e2e/operator/deployment_api.go:185
```

test/e2e/operator/deployment_api.go
```diff
@@ -144,6 +161,30 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
 			})
 		}
 
+		It("get deployment shall list expected fields", func() {
+			expectedFiles := len([]string{"NAME", "DEVICEMODE", "NODESELECTOR", "IMAGE", "STATUS", "AGE"})
```
pohly, line 165: `expectedFields` instead of `expectedFiles`

test/e2e/operator/deployment_api.go
pohly: `MatchRegexp` (https://github.com/onsi/gomega/blob/v1.10.3/matchers.go#L171) might be a better way to check for the expected output.

avalluri: > The new test isn't passing:

Fixed it now.

avalluri: > `MatchRegexp` (https://github.com/onsi/gomega/blob/v1.10.3/matchers.go#L171) might be a better way to check for the expected output.

Changed the test to use regex.

pohly: This is a bit dangerous: it only works if the strings that you put into the regex have no special meaning. I'm on the edge whether the curly brackets in `{"storage":"unknown-node"}` are matcher literally or might be interpreted as a repetition count. But it seems to work, so I suppose it's okay....

... except that the output with Kubernetes 1.17 is different:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-749/20/tests

```
Expected
<string>: test-get-deployment-fields direct map[storage:unknown-node] unexisting/pmem-csi-driver Running 14s
to match regular expression
<string>: test-get-deployment-fields\s+direct\s+{"storage":"unknown-node"}\s+unexisting/pmem-csi-driver\s+Running\s+[0-9]+(s|m)
/mnt/workspace/pmem-csi_PR-749/test/e2e/operator/deployment_api.go:188
```

avalluri: Made changes to regex to support K8s 1.17 output., this also avoids usage of `{}` in the regex.

pohly:
@avalluri please rebase

avalluri:
> @avalluri please rebase

Rebased.

pohly (approved):
Let's merge this.

One more comment about API versioning: strictly speaking, we are changing from v1alpha1 to v1alpha2 here. But because it's an alpha API, it's okayish and (in practice) easier for everyone to just change some fields.

But as we now intend to move to a beta operator API, the 0.9.0 release should have a v1beta1 deployment API, ideally in addition to the v1alpha1. At some point we need to test how to support more than one API version, so we should better start now...

PR #751, opened 2020-09-28
build: use debian:buster as base images, non-reproducible builds

pohly:
The frequent updates of Clear Linux have been a problem, typically
leading to new PMEM-CSI releases for no good reason other than "there
might have been a bug fix". Debian is expected to update less
frequently. We also no longer update automatically.

The approach taken for that is to use the latest release as base,
whatever that is. In other words, we no longer offer reproducible
releases because there is no way to lock down exactly which files are
taken from the upstream distro. Each release is only built once, so
the binary content is still locked down at build time - we just don't
record in the source code what exactly that binary content is. In
practice that should be good enough.

Compared to Clear Linux, the approach for disabling usage of udev has
changed a bit. use_lvmetad is no longer a valid configuration option.


pohly:
I have already disabled the automated release update job in Jenkins.

Dockerfile
```diff
@@ -1,69 +1,46 @@
...
-RUN swupd update ${SWUPD_UPDATE_ARG} && swupd bundle-add file xfsprogs storage-utils \
-    $(if [ "$BIN_SUFFIX" = "-test" ]; then echo fio; fi) && \
-    rm -rf /var/lib/swupd /var/tmp/swupd
+# xfsprogs, e2fsprogs - formating filesystems
+# lvm2 - volume management
+# ndctl - pulls in the necessary library, useful by itself
+# fio - only included in testing images
+RUN ${APT_GET} update && \
+    ${APT_GET} upgrade -y --no-install-recommends && \
+    ${APT_GET} install -y --no-install-recommends file xfsprogs e2fsprogs lvm2 ndctl \
```
avalluri, line 41: So, now on we no longer building `ndctl` from the source.

avalluri (approved):

okartau:
not many "clear" mentioned any more after change, but there is still (in Jenkinsfile) BUILD_IMAGE composed from `pmem-clear-linux-builder`, but is that intentional, as we have CI registry filled with those images?

Dockerfile
pohly: Correct. It makes our bill of materials and Dockerfile simpler. I don't think we ever really needed the ability to build a newer version than the one offered by the distro, and as ndctl is even more mature now I don't expect that either.

pohly:
> BUILD_IMAGE composed from pmem-clear-linux-builder, but is that intentional

Kind of. By keeping the name we can overwrite the existing images. Not sure how important that is, though.

PR #753, opened 2020-09-28
Go 1.15.2

pohly:
This builds on top of PR #751 to avoid code conflicts and updates to Go 1.15.2 without changing how certificates are handled, i.e. the changed default in the Go runtime is reset to what it was before the update.


deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -112,6 +112,8 @@ spec:
           value: /tmp/termination-log
         - name: PMEM_CSI_DRIVER_NAME
           value: pmem-csi.intel.com
+        - name: GODEBUG
+          value: x509ignoreCN=0
```
avalluri, line 116: The driver deployed via operator also requires this environment change.

pohly: Yes. I thought I'd fire of the PR before calling it a day, but with the early consistency check it didn't get very far... Will update tomorrow...

pohly: Fixed.

pohly:
Updated PR:
- operator itself also needs GODEBUG
- lz4 installed in build container - that was broken in "devel" by the switch to Debian

pohly:
> lz4 installed in build container - that was broken in "devel" by the switch to Debian

That's... hogwash! lz4 is needed on the Jenkins workers. We have an `apt-get install liblz4-tool` in the worker's install script in Jenkins for that. What must have happened in the failed job is that the command failed and the worker came up anyway - a known issue of our CI.

pohly:
@avalluri : this seems to be ready.

avalluri (approved):

PR #755, opened 2020-09-29
go-1.15-2 + certificate generation

avalluri:
Using of certificate CommonName is deprecated in Go-1.15.

Made changes to the self-signed certificate generation and client certificate
verification code so that now they work on DNSNames instead of CommonName.
    
Whereas the certificate generation using `cfssl` tools in`setup-ca.sh`
script already setting the hostname(s) which is good enough for us.
`cfssl` treats as an error if no Subject set, so keep setting the CN.
    
Also updated to the latest cfssl release - v1.4.1

### DIFF
--- a/test/setup-ca.sh
+++ b/test/setup-ca.sh
@@ -15,9 +15,10 @@ if [ $cfssl_found -eq 0 ]; then
 fi
 
 # Generate CA certificates.
-<<EOF cfssl -loglevel=3 gencert -initca - | cfssljson -bare ca
+<<EOF cfssl gencert -initca - | cfssljson -bare ca
 {
     "CN": "$CA",
+    "hosts": [ "$CA" ],
     "key": {
         "algo": "rsa",
         "size": 2048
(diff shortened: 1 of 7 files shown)


avalluri:
This is PR is based on #753. I reverted the "enable legacy certificate support" (https://github.com/intel/pmem-csi/pull/755/commits/487f65bbfccac58e8ad76f322f1e5bddbd86deb1) commit and made changes to self-signed certificate code to generate SAN/hostname certificates. @pohly Can you please check if this is what you wanted to have?

test/setup-ca.sh
```diff
@@ -13,23 +13,24 @@ if [ $cfssl_found -eq 0 ]; then
     echo "cfssl tools not found, Please install cfssl and cfssljson."
     exit 1
 fi
-
+set -x
```
pohly, line 16: This probably needs to be removed, right?

test/setup-ca.sh
```diff
@@ -13,23 +13,24 @@ if [ $cfssl_found -eq 0 ]; then
...
     "CN": "$CA",
+    "hosts": [ "$CA" ],
     "key": {
         "algo": "rsa",
         "size": 2048
     }
 }
 EOF
-
+set +x
```
pohly, line 28: Also here.

test/setup-ca.sh
avalluri: oops. sorry. I will fix it.

avalluri: Done.

pohly (changes requested):
This looks good, except for the added `set +x`.

Can you submit just your one commit? It should work independently of the Go 1.15 change. That way we get some indirect testing for the downgrade scenario ("new" certificates with "old" code).

avalluri:
> This looks good, except for the added `set +x`.
> 
> Can you submit just your one commit? It should work independently of the Go 1.15 change. That way we get some indirect testing for the downgrade scenario ("new" certificates with "old" code).

Ok,  I will submit rebasing on devel.

pohly (approved):

PR #763, opened 2020-10-01
test deployment fixes and cleanup

pohly:
Some minor code cleanup.


pohly:
I'm not happy with commit 9940f96. `Describe` invokes the callback immediately, so this shouldn't make any difference. Let me dive deeper into debugging the issue.

pohly:
I can no longer reproduce the issue and the commit made no sense -> I removed it.

The rest are simple code cleanups.

avalluri (approved):

pohly:
> Describe invokes the callback immediately, so this shouldn't make any difference.

This is the current Ginkgo behavior. The dependency update brings in a Gingko version where that is no longer true, so with that version, the commit is needed. I'll add it back in the other PR...

PR #765, opened 2020-10-02
Dockerfile: include copyleft source code of extra packages

pohly:
Our "apt-get upgrade" + "apt-get install" layer is something where we
pull binaries from Debian and publish on Docker Hub, i.e. we
re-distribute binaries under their respective licenses. To fulfill the
obligation of making the source code available we download the source
code from Debian and include it in the image.

This is easier than maintaining a separate download location for the
source code and the extra size is small enough (10MB for the
production image, 100MB for the test image).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ LABEL description="PMEM CSI Driver"
 RUN ${APT_GET} update && \
     mkdir -p /usr/local/share && \
     bash -c 'set -o pipefail; ${APT_GET} install -y --no-install-recommends file xfsprogs e2fsprogs lvm2 ndctl \
-       $(if [ "$BIN_SUFFIX" = "-test" ]; then echo fio; fi) | tee --append /usr/local/share/package-install.log' && \
+       | tee --append /usr/local/share/package-install.log' && \
     rm -rf /var/cache/*
 
 # Image in which PMEM-CSI binaries get built.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -57,6 +59,57 @@ ARG BIN_SUFFIX
...
+COPY --from=runtime /usr/local/share/package-install.log /usr/local/share/package-install.log
+COPY --from=runtime /usr/share/doc /tmp/runtime-doc
+RUN sed -i -e 's/^deb \(.*\)/deb \1\ndeb-src \1/' /etc/apt/sources.list
+RUN mkdir -p /usr/local/share/package-sources
+RUN cd /usr/local/share/package-sources && \
+    ${APT_GET} update && \
+    grep ^Get: /usr/local/share/package-install.log | cut -d ' ' -f 5,7 | \
+    while read pkg version; do \
+       if ! [ -f /tmp/runtime-doc/$pkg/copyright ]; then \
+           echo "ERROR: missing copyfight file for $pkg"; exit 1; \
```
avalluri, line 96: s/copyfight/copyright/

avalluri (approved):
Though I haven't tried building the image but looks good to me.

Dockerfile
pohly: Oh, nice typo! Will fix.

pohly:
@avalluri please take another look - I added two more commits.

avalluri (approved):

PR #772, opened 2020-10-08
build: better caching of Debian package sources

pohly:
By moving the command which pulls Debian package sources before the
settings for the PMEM-CSI binaries, test and production images can
share the same cached result of that command.

Switching between different PMEM-CSI source code also no longer causes
a re-download.


avalluri (approved):
Looks good to me

Issue #761, opened 2020-10-01
operator: stop creating CRD


pohly:
The approach taken by OLM (= aka installation from Operator Hub) is to create the CRD as part of the installation, i.e. the operator doesn't need to create it and also shouldn't delete it. This is more predictable.

We should do the same when installing the operator via YAML and remove the extra code for managing the CRD.

pohly:
In the "update dependencies" PR (= PR #764 ) I just got a strange error that I cannot reproduce locally:

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi/branches/PR-764/runs/1/nodes/90/steps/142/log/?start=0

```
create deployment error: the server could not find the requested resource, will retry...
```

Tests like "operator-lvm-production driver runs" fail because the CRD is not found?

PR #777, opened 2020-10-09
operator: do not install/uninstall CRD.

avalluri:
As per operator developer [best-practices](https://github.com/operator-framework/community-operators/blob/master/docs/best-practices.md)
the operator shall not self-register CRDs. With this change we stop doing so.

FIXES: #761


deploy/crd/pmem-csi.intel.com_deployments.yaml
```diff
@@ -0,0 +1,183 @@
+
+---
```
pohly, line 2: Where does this file come from? If it is generated, can you add a comment at the top that explains that?

pkg/pmem-csi-operator/main.go
```diff
@@ -19,7 +19,7 @@ import (
 	api "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1alpha1"
 	"github.com/intel/pmem-csi/pkg/k8sutil"
 	"github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller"
-	"github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller/deployment"
+	_ "github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller/deployment"
```
pohly, line 22: What is the anonymous import for? Please explain in the code with a comment.

test/start-operator.sh
```diff
@@ -75,6 +75,10 @@ function deploy_using_olm() {
 }
 
 function deploy_using_yaml() {
+  crd=${REPO_DIRECTORY}/deploy/crd/pmem-csi.intel.com_deployments.yaml
+  echo "Deploying '${crd}'..."
+  cat  ${crd} | ${SSH} kubectl create -f -
```
pohly, line 80: Better use `apply`, as for the operator itself. Sometimes that makes it simpler to re-deploy.

test/test.make
```diff
@@ -134,7 +134,7 @@ RUNTIME_DEPS += LC_ALL=C LANG=C sort -u
 # (https://github.com/intel/pmem-csi/pull/420#discussion_r346850741).
 .PHONY: run_tests
 test: run_tests
-RUN_TESTS = TEST_WORK=$(abspath _work) \
+RUN_TESTS = TEST_WORK=$(abspath _work) REPO_ROOT=`pwd` \
```
pohly, line 137: Did this break the image file tests? https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-777/1/

pohly (changes requested):

deploy/crd/pmem-csi.intel.com_deployments.yaml
avalluri: The file was generated by `controller-gen` and it does not add any comment. Changed the make target to add this comment after generating the crd file.

test/start-operator.sh
avalluri: Change as suggested.

test/test.make
avalluri: This was a bit strange to me. I guess `pkg/imagefile/test` is supposed to be run by e2e test, but not skipped from `make test`. It was not an issue before as those tests check if REPO_ROOT and CLUSTER are set and exit without running if those are missing. Whereas now they get to run as both variables are preset, which results in failure.

Now I added them to the skip list along with device manager tests.

operator/operator.make
```diff
@@ -52,17 +53,19 @@ OPERATOR_OUTPUT_DIR := $(CATALOG_DIR)/$(MAJOR_MINOR_PATCH_VERSION)
...
 # Generate CRD and add kustomization support
 operator-generate-crd: controller-gen
-	@echo "Generating CRD in $(MANIFESTS_DIR)/crd ..."
-	$(CONTROLLER_GEN) crd:trivialVersions=true,crdVersions=v1beta1 paths=./pkg/apis/... output:dir=$(MANIFESTS_DIR)/crd/
-	@echo "resources: [pmem-csi.intel.com_deployments.yaml]" > $(MANIFESTS_DIR)/crd/kustomization.yaml
+	@echo "Generating CRD in $(CRD_DIR) ..."
+	$(CONTROLLER_GEN) crd:trivialVersions=true,crdVersions=v1beta1 paths=./pkg/apis/... output:dir=$(CRD_DIR)/
+	version="$(shell $(CONTROLLER_GEN) --version |cut -f2 -d':')"; \
+	for f in $(shell ls $(CRD_DIR)); do \
+	  sed -i "1s/^/# This file was generated by controller-gen$$version\n/" $(CRD_DIR)/$$f ; \
```
pohly, line 60: please append "via `make operator-generate-crd`".

operator/operator.make
```diff
@@ -52,17 +53,19 @@ OPERATOR_OUTPUT_DIR := $(CATALOG_DIR)/$(MAJOR_MINOR_PATCH_VERSION)
 
 # Generate CRD and add kustomization support
 operator-generate-crd: controller-gen
-	@echo "Generating CRD in $(MANIFESTS_DIR)/crd ..."
-	$(CONTROLLER_GEN) crd:trivialVersions=true,crdVersions=v1beta1 paths=./pkg/apis/... output:dir=$(MANIFESTS_DIR)/crd/
-	@echo "resources: [pmem-csi.intel.com_deployments.yaml]" > $(MANIFESTS_DIR)/crd/kustomization.yaml
+	@echo "Generating CRD in $(CRD_DIR) ..."
+	$(CONTROLLER_GEN) crd:trivialVersions=true,crdVersions=v1beta1 paths=./pkg/apis/... output:dir=$(CRD_DIR)/
+	version="$(shell $(CONTROLLER_GEN) --version |cut -f2 -d':')"; \
+	for f in $(shell ls $(CRD_DIR)); do \
```
pohly, line 59: That looks complicated. Why not just `for f in $(CRD_DIR)/*; do`?

pohly (changes requested):
Some tests were failing: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-777/4/tests

operator/operator.make
avalluri: > That looks complicated. Why not just `for f in $(CRD_DIR)/*; do`?

Yes, I should have done this. Changed it now.

operator/operator.make
avalluri: Done

PR #783, opened 2020-10-12
Dockerfile: remove python3 from image

okartau:
python3 is included only because of dependency in xfsprogs,
there is one script in there which we don't use.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,8 @@ RUN ${APT_GET} update && \
     mkdir -p /usr/local/share && \
     bash -c 'set -o pipefail; ${APT_GET} install -y --no-install-recommends file xfsprogs e2fsprogs lvm2 ndctl \
        | tee --append /usr/local/share/package-install.log' && \
+    dpkg --purge --ignore-depends=xfsprogs python3 python3-minimal libpython3-stdlib libpython3.7-minimal libpython3.7-stdlib python3.7 python3.7-minimal && \
+    sed -i 's/, python3:any//' /var/lib/dpkg/status && \
     rm -rf /var/cache/*
 
 # Image in which PMEM-CSI binaries get built.


Dockerfile
```diff
@@ -171,4 +171,5 @@ ENV LD_LIBRARY_PATH=/usr/lib
 # By default container runs with non-root user
 # Choose root user explicitly only where needed, like - node driver
 RUN useradd --uid 1000 --user-group --shell /bin/bash pmem-csi
+RUN dpkg --purge --ignore-depends=xfsprogs python3 python3-minimal libpython3-stdlib libpython3.7-minimal libpython3.7-stdlib python3.7 python3.7-minimal
```
pohly, line 174: Can you add this to the `RUN` command where packages are installed?

As it stands now, we create a layer where Python is included and later another layer where the files are marked as deleted. But that then just increases the size of our image.

You can check that with `docker history <container name>`.

okartau: yes, can try that.
But seems testing also fails now because TLS test attempts install of nmap, patch and apt complains. We use final production image and modify it for testing. Should we use testing image for such?

okartau: moved dpkg cmd up in Dockerfile. But I think we still hit that TLS test failure.

pohly: > Should we use testing image for such?

And keep Python installed in it? That then doesn't help us with getting rid of Python during our image release process because we'll still have to deal with it for the test image.

This https://serverfault.com/questions/250224/how-do-i-get-apt-get-to-ignore-some-dependencies/506993#506993 answer explains how to hack apt to ignore Python permanently. Alternatively, use `equivs-build` to create a fake python package (https://serverfault.com/a/251091).

okartau: I did not find easy way to accomplish this with apt options, but there is a brute-force way editing dpkg status file directly. I tried this now (raw variant, just single sed without even looking where we edit, but in current state we have exactly one python3 dependency). Have to admit, this starts to turn hackish, so is it worth doing after all...

pohly: Is the approach with equivs-build better? See https://github.com/intel/pmem-csi/pull/784

pohly:
Solved via PR #784 instead.

PR #784, opened 2020-10-13
build: avoid installing Python3

pohly:
Python3 is needed by xfsprogs only for a single script that we don't
need at runtime. Therefore we can avoid installing this rather large
package and its dependency by providing "python3:any" with an empty
stub package.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,11 +15,16 @@ ARG CACHEBUST
 # tools and recommended packages. But this image gets pushed to a registry by the CI as a cache,
 # so it still makes sense to keep this layer small by removing /var/cache.
 RUN ${APT_GET} update && \
-    ${APT_GET} install -y gcc libndctl-dev make git curl iproute2 pkg-config xfsprogs e2fsprogs parted openssh-client python3 python3-venv && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -29,6 +34,8 @@ ARG BIN_SUFFIX
 LABEL maintainers="Intel"
 LABEL description="PMEM CSI Driver"
 
+COPY --from=build python3_100.0_all.deb /var/cache/python3_100.0_all.deb
```
pohly, line 37: This creates a layer, which cannot be avoided easily (https://forums.docker.com/t/copy-a-file-that-doesnt-need-to-be-in-the-final-image/12627/5).

But it's small, so this shouldn't matter.

pohly: 2KB

okartau: I think dummy package approach here is cleaner than that other hack in #783 which requires brutal altering of dpkg state.

pohly: So shall we merge this? I think it's worthwhile.

okartau: yes, I will merge, let's see how the devel branch images will look.

Issue #798, opened 2020-10-30
deploy/crd/pmem-csi.intel.com_deployments.yaml on "devel" branch


pohly:
After `make operator-generate-crd`, `deploy/crd/pmem-csi.intel.com_deployments.yaml` is modified on the current devel branch (= 6f40cebb2e2ea2dbd151caaec12c5a605e484db8):

```
diff --git a/deploy/crd/pmem-csi.intel.com_deployments.yaml b/deploy/crd/pmem-csi.intel.com_deployments.yaml
index ec1b2246..404da56a 100644
--- a/deploy/crd/pmem-csi.intel.com_deployments.yaml
+++ b/deploy/crd/pmem-csi.intel.com_deployments.yaml
@@ -180,56 +180,6 @@ spec:
           status:
             description: DeploymentStatus defines the observed state of Deployment
             properties:
-              conditions:
-                description: Conditions
-                items:
-                  properties:
...
```

This escaped PR testing because we don't have a dedicated "make test" check like the one we have for generated kustomizations. We need to:
- add such a test
- check in the current file

Found because a full branch build fails at the point where it tries to update master:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/devel/273/pipeline
```
[2020-10-29T15:25:07.585Z] + git rev-parse HEAD
[2020-10-29T15:25:07.585Z] + head=6f40cebb2e2ea2dbd151caaec12c5a605e484db8
[2020-10-29T15:25:07.585Z] + git fetch origin master
[2020-10-29T15:25:08.842Z] From github.com:intel/pmem-csi
[2020-10-29T15:25:08.842Z]  * branch              master     -> FETCH_HEAD
[2020-10-29T15:25:08.842Z] + git checkout FETCH_HEAD
[2020-10-29T15:25:08.842Z] error: Your local changes to the following files would be overwritten by checkout:
[2020-10-29T15:25:08.842Z] 	deploy/crd/pmem-csi.intel.com_deployments.yaml
[2020-10-29T15:25:08.842Z] Please commit your changes or stash them before you switch branches.
[2020-10-29T15:25:08.842Z] Aborting
```

PR #800, opened 2020-11-02
operator: fix and test code generation

pohly:
An invalid deploy/crd/pmem-csi.intel.com_deployments.yaml file was
checked in as part of commit b4d17343a8b97785f0aab17b4cd25a9561fb4312
and we didn't noticed until after merging the PR because we didn't
have a test for it and the broken file only caused problems later when
trying to update the master branch (working directory not clean).

Adding the test also showed that "operator-generate-bundle" was
broken (used version in the wrong format).

Fixes: https://github.com/intel/pmem-csi/issues/798


pohly:
@okartau : with Amarnath out this week, can you perhaps have a look whether this looks reasonable?

okartau (approved):

PR #806, opened 2020-11-05
pmem-csi-driver: fix rescheduling after CreateVolume failure

pohly:
If CreateVolume fails for a volume with late binding, the Kubernetes
scheduler is supposed to be told that it needs to reschedule the
pod. This didn't work with PMEM-CSI because it returned an "internal"
error code instead of the "resource exhausted" one that the
external-provisioner knows about.

Instead of having different well-defined errors at different levels of
the PMEM-CSI software stack, a common error package is used. These
errors get wrapped when it is useful to add more information. We still
add the gRPC error code, though. That could be avoided by making
the internal errors based on gRPC but it is uncertain whether we want
to go that far.

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2020 Intel Coporation.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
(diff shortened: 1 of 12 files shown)


avalluri (approved):
Looks good.

PR #808, opened 2020-11-06
set and test with default FS type

pohly:
The updated test reveals that ownership changing fails when using a storage class without explicit FS type. This gets fixed by adding the `--default-fstype` parameter to external-provisioner (https://github.com/kubernetes-csi/external-provisioner/issues/328#issuecomment-714801581).


avalluri (approved):
Looks good.

Issue #685, opened 2020-06-17
test flake: release-0.7, stress test


pohly:
In this log https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi-release/runs/149/nodes/103/steps/187/log/?start=0 the first stress test succeeds, but in the second one, the last output for worker-1 is:
```
[2020-06-17T11:55:00.983Z] Jun 17 11:55:00.964: INFO: ExecWithOptions {Command:[mkdir /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pmem-sanity-staging.MLhbDg/worker-1] Namespace:default PodName:pmem-csi-node-testing-qr4w9 ContainerName:socat Stdin:<nil> CaptureStdout:true CaptureStderr:true PreserveWhitespace:false}
[2020-06-17T11:55:00.983Z] Jun 17 11:55:00.964: INFO: >>> kubeConfig: /mnt/workspace/pmem-csi-release/_work/govm/kube.config
[2020-06-17T11:55:01.339Z] [1mSTEP[0m: Exec Output: 
```

What should come next is `worker-1-volume-X: creating a single node writer volume`, but worker-1 never seems to reach that. It's not clear where it can possibly get stuck.

pohly:
I think the conclusion was that pod startup on Clear Linux didn't work as reliably as necessary, at least at that time:
https://github.com/intel/pmem-csi/blob/07c650a793fdc484a6f62d1f14524096d7af8a4d/test/e2e/storage/csi_volumes.go#L109-L112

Let's see if it's more stable now.

pohly:
I tried enabling the stress tests again and the Jenkins jobs fell over with "no space left on device" *on the host*.

This needs further work, let's postpone to 0.9.

pohly:
Started a full test run of the "stress-test-enable" branch with a 128GB VM disk: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/view/pmem-csi/job/pmem-csi/job/stress-test-enable/1/

PR #809, opened 2020-11-06
test: enable stress test again

pohly:
We no longer test on the platform (Clear Linux) which had problems
passing the test, so now it may be possible to use it again.

Another issue was that the increased usage of the PMEM data files on the build host caused it to run out of filesystem space. This was fixed by increasing the VM's disk.

Fixes: https://github.com/intel/pmem-csi/issues/685


pohly:
The test itself finishes in roughly three minutes, which is okay: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/stress-test-enable/1/tests

avalluri (approved):

PR #811, opened 2020-11-09
enable scale testing of PMEM-CSI

pohly:
Normal clusters in the cloud often don't have PMEM. It's still useful to test scalability of the non-local operations on such clusters.

Deploying on a genuine kubeadm-based cluster must allow the controller part to run on the master node.


avalluri:
The change looks okay to me, but @pohly which part of the change is really enabling the scale testing?

PR requires rebasing.

pohly:
> which part of the change is really enabling the scale testing?

The testing doesn't run in our CI, if that's what you were looking for.

The intended usage is to deploy PMEM-CSI on some cluster manually with the fake device manager instead of LVM or direct.

pohly:
Rebased.

avalluri:
> The intended usage is to deploy PMEM-CSI on some cluster manually with the fake device manager instead of LVM or direct.

Okay, thanks.

Issue #747, opened 2020-09-25
node labelling


pohly:
The example label (`storage=pmem`) should better be replaced with a namespaced label (`pmem-csi.intel.com=true`). It shows that the label is meant to control the placement of the PMEM-CSI driver and avoids name conflicts.

We also should explore, test and document how to use NFD for the labeling part.

pohly:
Fixed in PR #813

PR #813, opened 2020-11-11
use NFD for node labelling

pohly:
By using NFD we ensure that the integration with it works as intended.

TODO:
- [X] documentation
- [X] clean up commits (split out removal of Clear Linux workaround)
- [x] ensure that downgrade tests work - they probably don't, because the node labels prevent running the older version

Fixes: issue #747

### DIFF
--- a/docs/autotest.md
+++ b/docs/autotest.md
@@ -56,8 +56,11 @@ virtual machines.
 The first node is the Kubernetes master without
 persistent memory.
 The other three nodes are worker nodes with one emulated 32GB NVDIMM each.
-After the cluster has been formed, `make start` adds `storage=pmem` label
-to the worker nodes and deploys the PMEM-CSI driver.
+After the cluster has been formed, `make start` installs [NFD](https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html) to label
+the worker nodes. The PMEM-CSI driver can be installed with
+`test/setup-deployment.sh`, but will also be installed as needed by
+the E2E test suite.
+
 Once `make start` completes, the cluster is ready for interactive use via
 `kubectl` inside the virtual machine. Alternatively, you can also
 set `KUBECONFIG` as shown at the end of the `make start` output
--- a/docs/install.md
+++ b/docs/install.md
@@ -209,9 +209,24 @@ on the Kubernetes version.
 
 - **Label the cluster nodes that provide persistent memory device(s)**
 
+PMEM-CSI manages PMEM on those nodes that have a certain label. For
+historic reasons, the default in the YAML files and the operator
+settings is to use a label `storage` with the value `pmem`.
+
+Such a label can be set for each node manually with:
+
 ``` console
 $ kubectl label node <your node> storage=pmem
 ```
+
+Alternatively, the [Node Feature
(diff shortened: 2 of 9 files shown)


pohly:
@avalluri I tested manually that version skew testing still passes. This is ready for review.

@marquiz FYI, we are going to use and recommend NFD for node labeling - this has been long overdue :sweat_smile:

marquiz:
> FYI, we are going to use and recommend NFD for node labeling - this has been long overdue

Good to get some new users onboard 👍 Especially users we're aware of 😄 Please don't hesitate to bring up any issues or wishes for improvement you might have.

docs/autotest.md
```diff
@@ -56,8 +56,11 @@ virtual machines.
 The first node is the Kubernetes master without
 persistent memory.
 The other three nodes are worker nodes with one emulated 32GB NVDIMM each.
-After the cluster has been formed, `make start` adds `storage=pmem` label
-to the worker nodes and deploys the PMEM-CSI driver.
+After the cluster has been formed, `make start` adds NFD to label
```
avalluri, line 59: 'adds NFD' sounds not appropriate to me, what about something line 'starts NFD'? And a link to NFD project seems appropriate here.

docs/install.md
```diff
@@ -209,10 +209,25 @@ on the Kubernetes version.
 
 - **Label the cluster nodes that provide persistent memory device(s)**
 
+The local part of PMEM-CSI runs on all nodes with a certain label. For
```
avalluri, line 212: @pohly What exactly 'local' refers to here?

docs/autotest.md
avalluri: Okay, a later part of the documentation added the link to NFD, so it is fine to omit it here.

docs/install.md
```diff
@@ -448,13 +463,16 @@ verify that the node labels have been configured correctly
...
+The command output must indicate that every node with PMEM has at least two labels:
 ``` console
 pmem-csi.intel.com/node=<NODE-NAME>,storage=pmem
 ```
 
-If **storage=pmem** is missing, label manually as described above. If
-**pmem-csi.intel.com/node** is missing, then double-check that the
+**storage=pmem** is the label that has to be added manually as
+described above.  When using NFD, the node should have the
+`feature.node.kubernetes.io/memory-nv.dax: true` label.
```
avalluri, line 473: Replacing ';' with '=' is most appropriate here.

docs/install.md
pohly: The node server.

Saying the "The node part of PMEM-CSI runs on all nodes with a certain label" felt odd.

docs/autotest.md
pohly: "`make start` starts NFD" stutters.

How about "`make start` installs NFD"?

avalluri: > How about "`make start` installs NFD"?

Sounds good.

docs/install.md
pohly: Replaced with "PMEM-CSI manages PMEM on those nodes that have a certain label."

docs/install.md
pohly: Fixed.

avalluri (approved):

PR #820, opened 2020-11-18
measure resource consumption with Vertical Pod Autoscaler

pohly:
This is useful for replacing the guessed default resource requirements
in the operator.

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -17,6 +17,7 @@
     - [Command line arguments](#command-line-arguments)
     - [Environment variables](#environment-variables)
     - [Logging](#logging)
+- [Performance and resource measurements](#performance-and-resource-measurements)
 - [Switching device mode](#switching-device-mode)
     - [Going from LVM device mode to direct device mode](#going-from-lvm-device-mode-to-direct-device-mode)
     - [Going from direct device mode to LVM device mode](#going-from-direct-device-mode-to-lvm-device-mode)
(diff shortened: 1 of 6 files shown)


pohly:
This was straightforward, but still required a bit of reading of documentation and the TLS issue is a pitfall. Therefore it is worthwhile to record the result in our repo.

docs/DEVELOPMENT.md
```diff
@@ -258,6 +259,40 @@ The klog.Info statements are used via the verbosity checker using the following
...
+
+Resource requirements depend on the workload. To generate some load, run
+```console
+make test_e2e TEST_E2E_FOCUS=lvm-production.*late.binding.*stress.test
+```
+
+Now resource recommendations can be retrieved with:
+
+```console
+kubectl get vpa
```
avalluri, line 293: `kubectl get vpa`  itself does not give any useful information. Would be nice to mention a way to use vpa reccomnedation something like: 
```
$ kubectl get vpa  pmem-csi-node -o jsonpath='{range .status.recommendation.containerRecommendations[*]}{.containerName}{":\n\tRequests: "}{.lowerBound}{"\n\tLimits: "}{.upperBound}{"\n"}{end}'
```

pohly: Right. I think I was using `kubectl describe` and should have put that into the doc.

Let me add that and the command that you showed here.

pohly: Done - let's merge if okay without waiting for tests.

avalluri (approved):

Issue #823, opened 2020-11-23
volume leak during stress test


pohly:
The late binding stress test was recently enabled again and now makes testing unstable, with leaked volumes. 
From https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/rest/organizations/jenkins/pipelines/pmem-csi/branches/disable-0.7-2/runs/8/nodes/87/steps/127/log/?start=0:

```
$ grep -e 

[2020-11-20T12:21:07.176Z] pmem-csi-controller-0/external-provisioner@pmem..ker3: I1120 12:21:07.135249       1 event.go:282] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"latebinding-7248", Name:"pvc-x6rkd", UID:"731cfbca-c8b5-424e-9d9e-f024e204e09e", APIVersion:"v1", ResourceVersion:"33412", FieldPath:""}): type: 'Normal' reason: 'Provisioning' External provisioner is provisioning volume for claim "latebinding-7248/pvc-x6rkd"
[2020-11-20T12:21:07.176Z] pmem-csi-controller-0/external-provisioner@pmem..ker3: I1120 12:21:07.135366       1 controller.go:573] CreateVolumeRequest {Name:pvc-731cfbca-c8b5-424e-9d9e-f024e204e09e CapacityRange:required_bytes:1048576  VolumeCapabilities:[mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > ] Parameters:map[] Secrets:map[] VolumeContentSource:<nil> AccessibilityRequirements:requisite:<segments:<key:"pmem-csi.intel.com/node" value:"pmem-csi-govm-worker2" > > preferred:<segments:<key:"pmem-csi.intel.com/node" value:"pmem-csi-govm-worker2" > >  XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
[2020-11-20T12:21:07.176Z] pmem-csi-controller-0/external-provisioner@pmem..ker3: I1120 12:21:07.136915       1 connection.go:182] GRPC call: /csi.v1.Controller/CreateVolume
[...]

PR #824, opened 2020-11-23
Revert "test: enable stress test again"

pohly:
This reverts commit b9079bedf12ee888b2fae77a90303ce4e244ca57.

The test works and triggers volume leaks, which then fails our test
and branch build jobs (https://github.com/intel/pmem-csi/issues/823).
We have to disable it until we have a fix.


pohly:
@avalluri : please take a look.

avalluri (approved):

PR #826, opened 2020-11-25
operator: per-container resource requests, v1beta API and CRD webhook

avalluri:
The changes in brief:
- Split the deployment API code such that, could be reused for multiple API version.
- Provided per-container resources as part of v1beta1 API.
- Conversion webhook for migrating existing v1alpha1 API to v1beta1. [Not yet supported deploying using OLM]


pohly (commented):
> Conversion webhook for migrating existing v1alpha1 API to v1beta1. [Not yet supported deploying using OLM]

What's the situation when deploying with OLM? Only v1beta1 API available or operator not usable?

avalluri:
> What's the situation when deploying with OLM? Only v1beta1 API available or operator not usable?

Yes, It accepts v1beta1 CRs. For v1alpha1 CR, the default conversion works, i.e, only the `apiVersion` field. This means any v1alpha1 deployment that was deployed with default values will automatically be converted to v1beta1. Only v1alpha1 deployments which explicitly use `{node,controller}Resources` fields will not work.

pohly:
> Yes, It accepts v1beta1 CRs. For v1alpha1 CR, the default conversion works, i.e, only the apiVersion field. This means any v1alpha1 deployment that was deployed with default values will automatically be converted to v1beta1. Only v1alpha1 deployments which explicitly use {node,controller}Resources fields will not work as they are deprecated in the new API.

This needs to be documented in `install.md`.

avalluri:
> > Yes, It accepts v1beta1 CRs. For v1alpha1 CR, the default conversion works, i.e, only the apiVersion field. This means any v1alpha1 deployment that was deployed with default values will automatically be converted to v1beta1. Only v1alpha1 deployments which explicitly use {node,controller}Resources fields will not work as they are deprecated in the new API.
> 
> This needs to be documented in `install.md`.

Current install steps did not cover installing the operator using OLM. I added this to the install steps and also added a note about the lack of webhook support.

docs/install.md
```diff
@@ -1302,6 +1314,45 @@ deployment is in the `Failed` state, then one can look into the event(s) using
 > [`DeploymentSpec`](#deployment-crd-api).
 
 
+## PMEM-CSI operator API conversion webhook
+
+The operator relese v0.9.0 deprecates the `v1alpha1` [deployment API](#pmem-csi-deployment-crd) infavour of
+`v1beta1`. To support auto-migrating of deprecated(v1alpha1) API versions, The
```
pohly, line 1320: Please explain what "auto-migrating" does. Be specific (i.e. provide examples) to illustrate it. This is also the right place to describe what happens without the webhook. We have to assume that the reader is not familiar with CRD concepts.

docs/install.md
```diff
@@ -1302,6 +1314,45 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+operator could be deployed to run a built-in CRD [conversion webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/)
+server. By default it starts an http server at port nubmer 9443 and serving
+traffic on path `/convert`. The conversion service is exposed via a Kubernetes
+service named `pmem-cis-operator-service` in the operator's namespace.
+
+To secure the communication between the Kubernetes APIserver and the conversion
+webhook, the server must be configured with apporopriate private key and a TLS
```
pohly, line 1327: Please run a spell-checker ("apporopriate").

docs/install.md
```diff
@@ -1302,6 +1314,45 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+```
+
+- **Install the PMEM-CSI operator**
+```console
+$ kubectl apply -f https://github.com/intel/pmem-csi/raw/devel/deploy/operator/pmem-csi-operator-webhook.yaml
+```
+
+**NOTE**: Conversion webhook is not yet supported by the operator installed using the
+[Operator Lifecycle Manager](https://operatorhub.io/operator/pmem-csi-operator).
+If the operator is deployed with no conversion webhook then it might
```
pohly, line 1352: "might not" or "will not"? "Might" sounds like it's uncertain.

operator/operator.make
```diff
@@ -1,15 +1,15 @@
...
 	mkdir -p _work/bin/ 2> /dev/null
 #	Building operator-sdk from sources as that needs below fixes:
 #     https://github.com/operator-framework/operator-sdk/pull/3787
 #     https://github.com/operator-framework/operator-sdk/pull/3786
-#   curl -L https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o $(abspath $@)
+#   curl -L https://github.com/operator-framework/operator-sdk/releases/download/v$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o $(abspath $@)
 	tmpdir=`mktemp -d` && \
 	trap 'set -x; rm -rf $$tmpdir' EXIT && \
-	git clone --branch 1.0.0+fixes https://github.com/avalluri/operator-sdk.git $$tmpdir && \
+	git clone --branch $(OPERATOR_SDK_VERSION)+fixes https://github.com/avalluri/operator-sdk.git $$tmpdir && \
```
pohly, line 12: Do we still need additional fixes? Please add links to pending PRs for all of them so that we can check progress.

pkg/apis/pmemcsi/v1beta1/deployment_conversion.go
```diff
@@ -0,0 +1,26 @@
...
+package v1beta1
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+)
+
+var _ conversion.Hub = &Deployment{}
+
+// Hub marks this(v1beta1) type as a conversion hub.
+func (*Deployment) Hub() {}
```
pohly, line 26: Explaining "hub" as "conversion hub" doesn't explain what a "hub" is (circular definition). Please explain what it does.

test/e2e/operator/deployment_api.go
```diff
@@ -775,6 +776,70 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+			deployment := deploy.GetDeploymentCR(f, alphaDep.Name)
+
+			Expect(deployment.Spec.NodeDriverResources).Should(BeEquivalentTo(alphaDep.Spec.NodeResources), "node driver resources")
+			Expect(deployment.Spec.ControllerDriverResources).Should(BeEquivalentTo(alphaDep.Spec.ControllerResources), "controller driver resources")
+
+			defer deploy.DeleteDeploymentCR(f, deployment.Name)
+
+			validateDriver(deployment, true)
+		})
+	})
```
pohly, line 842: According to the conversion functions above, conversion from beta to alpha is supported. This is not tested here, is it?

pohly:
The way I understand, the current design is:
- operator only supports v1beta1 API
- v1alpha must be converted to v1beta either:
  - on-the-fly via webhook or
  - permanently manually by the user and/or admin

Have you considered supporting both v1alpha1 and and v1beta1 in the operator? Is that possible?

docs/install.md
avalluri: Added more explanation about CRD conversion and added suitable examples to the docuentation.

docs/install.md

docs/install.md
avalluri: Rephrased to "will not".

operator/operator.make
avalluri: The pending PR link(s) are available at few lines above in make target.

pkg/apis/pmemcsi/v1beta1/deployment_conversion.go
avalluri: This is copied from the generated code by controller-runtime/operator-sdk. Added a short description:
`// Conversion hub is the version that all other CRD versions get converted into.`

test/e2e/operator/deployment_api.go
avalluri: Though the conversion code is available, the operator does not reconcile `v1alpha1` CRs, and thus no test to cover that.

avalluri:
> Have you considered supporting both v1alpha1 and and v1beta1 in the operator? Is that possible?

I did not consider supporting both(all) versions. It makes the reconciling code bit complicated. Do you see that we should support both API versions without conversion?

operator/operator.make
pohly: Looking at https://github.com/operator-framework/operator-sdk/pull/3786 I see that it was closed as "open for a while without interation". It looks like there was a test failure that didn't get addressed for two months?

Can you do what was suggested in https://github.com/operator-framework/operator-sdk/pull/3786#issuecomment-719659176 and create an issue? Then if you have a working fix, create a PR.

docs/install.md
```diff
@@ -1215,8 +1223,12 @@ The current API for PMEM-CSI `Deployment` resources is:
 | pullPolicy | string | Docker image pull policy. either one of `Always`, `Never`, `IfNotPresent` | `IfNotPresent` |
 | logLevel | integer | PMEM-CSI driver logging level | 3 |
 | deviceMode | string | Device management mode to use. Supports one of `lvm` or `direct` | `lvm`
-| controllerResources | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core) | Describes the compute resource requirements for controller pod |
-| nodeResources | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core) | Describes the compute resource requirements for the pods running on node(s) |
+| controllerResources | [ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#resourcerequirements-v1-core) | Describes the compute resource requirements for controller pod [**_Deprectead_**] |
```
pohly, line 1226: What does "deprecated" mean here? Isn't `controllerResources` gone in v1beta1?

We probably need a more nuanced way of describing the API:
- for `apiVersion`, list v1beta1 and v1alpha1 as valid values
- add a general remark outside of the table that fields are valid in all API revisions unless noted otherwise
- for fields that only exist in some revision(s), list those revisions in the description

docs/install.md
```diff
@@ -1302,6 +1314,140 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
 
+## PMEM-CSI operator API conversion webhook
+
+The operator release v0.9.0 deprecates the `v1alpha1` [deployment API](#pmem-csi-deployment-crd)
+in favor of `v1beta1`. To keep supporting a custom resource that uses the
+`v1alpha1` version must be converted to the latest supported `v1beta1` CRD
+version. This conversion is done by the Kubernetes API server by using the
+conversion strategy chosen in the [deployment CRD sepcification](#pmem-csi-deployment-crd).
+If no conversion strategy is chosen by the CRD, by default the API server
+converts the custom resource `apiVersion` to the latest supported CRD version.
```
pohly, line 1325: Converts when? When a client asks for v1beta1 (i.e. on demand)?

docs/install.md
```diff
@@ -1302,6 +1314,140 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+      cpu: "400m"
+      memory: "150Mi"
+EOF
+
+deployment.pmem-csi.intel.com/pmem-csi.intel.com created
+
+$ kubectl get deployments.pmem-csi.intel.com pmem-csi.intel.com -o jsonpath='{.apiVersion}{"\n"}{.spec}{"\n"}'
+pmem-csi.intel.com/v1beta1
+{"deviceMode":"lvm"}
+```
```
pohly, line 1373: It's worth showing here what happens when a client asks for a v1alpha1 version of that same object.

pkg/apis/pmemcsi/base/defaults.go
```diff
@@ -0,0 +1,85 @@
...
+	defaultRegistrarImageName = "k8s.gcr.io/sig-storage/csi-node-driver-registrar"
+	defaultRegistrarImageTag  = "v1.2.0"
+	// DefaultRegistrarImage default node driver registrar image to use
+	DefaultRegistrarImage = defaultRegistrarImageName + ":" + defaultRegistrarImageTag
+
+	// Below resource requests and limits are derived(with minor adjustments) from
+	// recommendations reported by VirtualPodAutoscaler(LowerBound -> Requests and UpperBound -> Limits)
+
+	// DefaultControllerResourceRequestCPU default CPU resource request used for controller driver container
+	DefaultControllerResourceRequestCPU = "12m" // MilliSeconds
```
pohly, line 41: Please document under DEVELOPMENT.md how these defaults were obtained. I already added instructions for running Vertical Pod Autoscaler; what's missing is a description of the workload and how to run it.

You said that you used a script. In that case we'll need the script so that others can reproduce your experiment. It might be simpler to use the late binding stress test.

pohly (changes requested):
I'm still mostly looking at documentation to understand what I should expect from the code...

The new documentation contains quite a few typos. Can you run a spell checker and fix those? 

I'll probably then propose some language changes in a PR (will be faster than leaving comments).

docs/install.md
avalluri: > What does "deprecated" mean here? Isn't `controllerResources` gone in v1beta1?
> 
Yes, `v1beta1` uses `controllerDriverResources` and `provisionerResources` in place of `controllerResources`.

> We probably need a more nuanced way of describing the API:
> 
> * for `apiVersion`, list v1beta1 and v1alpha1 as valid values
> * add a general remark outside of the table that fields are valid in all API revisions unless noted otherwise
> * for fields that only exist in some revision(s), list those revisions in the description

Made changes as suggested.

docs/install.md
avalluri: > Converts when? When a client asks for v1beta1 (i.e. on demand)?

As pet the [webhook conversion](https://v1-15.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#webhook-conversion) documentation:

```
The API server also supports webhook conversions that call an external service in case a conversion is required. For example when:

- custom resource is requested in a different version than stored version.
- Watch is created in one version but the changed object is stored in another version.
- custom resource PUT request is in a different version than storage version.
```

docs/install.md
avalluri: > It's worth showing here what happens when a client asks for a v1alpha1 version of that same object.

It supposed to return the same object by setting `apiVersion` to `v1alpha1`

pkg/apis/pmemcsi/base/defaults.go
avalluri: Running the stress test is a good enough example, but anyway I also added the alternate to run the script.

docs/install.md
pohly: But with `controllerResources` and `nodeResources` set, so not quite the same. This illustrates what we need to mention earlier about objects being converted on demand instead of being rewritten in the etcd store.

pohly: Long-term that becomes relevant for admins: if we ever *remove* the v1alpha1 API, they need to ensure that none of the stored CRs use that version or they'll become unusable.

avalluri: > But with `controllerResources` and `nodeResources` set, so not quite the same. This illustrates what we need to mention earlier about objects being converted on demand instead of being rewritten in the etcd store.

Only the latest(`v1beta1`)  CR fields get stored and when the client asks for a different revision, then the conversion happens from the stored revision. In this case where no conversion webhook, the `nodeResource`, `controllerResources` fields are unknown to `v1beta1` get lost and hence no change if we ask for `v1beta1` or `v1alpha1`.

test/e2e/operator/deployment_api.go
avalluri: Extended the tests to make sure that the `v1beta1` ->``v1alpha1` conversion works:
 - create `v1alpha1` object
 - retrieve back the `v1alpha1` revision and make sure that both the `spec` fields match.
 - retrieve the `v1beta1` revision for the same object and check that the `status` matches with the `v1alpah1` status.

avalluri:
Seems like the fedora-31 downlink is broken. Is it time to move to the latest release?
```
[2020-12-03T11:09:39.316Z] Downloading https://download.fedoraproject.org/pub/fedora/linux/releases/31/Cloud/x86_64/images/Fedora-Cloud-Base-31-1.9.x86_64.raw.xz image
[2020-12-03T11:09:39.316Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:39.316Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:39.673Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:39.673Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:39.673Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:39.673Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:40.032Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:40.032Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:40.032Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:40.032Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:40.390Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:40.390Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:40.390Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:40.390Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:40.746Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:40.746Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:40.746Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:40.746Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:41.104Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:41.104Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:41.104Z]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
[2020-12-03T11:09:41.104Z]                                  Dload  Upload   Total   Spent    Left  Speed
[2020-12-03T11:09:41.461Z] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
[2020-12-03T11:09:41.461Z] curl: (22) The requested URL returned error: 404 
[2020-12-03T11:09:41.461Z] ERROR: Download failed repeatedly, giving up.
```

hack/stress-driver.sh
```diff
@@ -0,0 +1,27 @@
...
+VOL_COUNT=${VOL_COUNT:-5} # 5 ext4 + 5  xfs = 10 * 4Gi ~= 40Gi
+
+for i in $(seq 1 1 $ROUNDS) ; do
+  echo "Round #$i:"
+  echo "Creating volumes..."
+  for j in  $(seq 1 1 $VOL_COUNT) ; do
+    sed -e "s;\(.*name:\)\(.*\);\1\2-$j;g" < deploy/common/pmem-pvc.yaml | kubectl create -f -
+  done
+  echo "Deleting all pvc..."
+  while [ "$(kubectl get pvc --no-headers | wc -l)" -ne "0" ]; do
```
pohly, line 23: Is this supposed to wait until all volumes are deleted? Then you need to check for `pv` here, not `pvc`. First a PVC gets deleted, then the PV.

docs/install.md
```diff
@@ -1302,6 +1320,144 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+{"controllerResources":{"limits":{"cpu":"400m","memory":"150Mi"}},"deviceMode":"lvm","nodeResources":{"limits":{"cpu":"500m","memory":"250Mi"}}}
+
+```
+
+**NOTE**: Conversion webhook is not yet supported by the operator installed using the
+[Operator Lifecycle Manager](https://operatorhub.io/operator/pmem-csi-operator).
+If the operator is deployed with no conversion webhook then it will not
+support the conversion of deprecated API fields as described above.
+In that case, the user has to manually migrate their existing driver deployments
+to use the latest operator API.
```
pohly, line 1459: Huh? I thought the apiserver will convert internally when no webhook is configured. So why does a user "have to manually migrate their existing driver deployments"?

pohly: Hmm, "in that case" is not clear. It doesn't refer to "deployed with no conversion webhook" (my initial understanding) but rather "conversion of deprecated API fields". Will fix...

operator/operator.make
pohly: Please work with upstream and add a link to that work (probably the new issue) here.

pohly:
Please remove the "Fedora 33" commit.

Tip: if you have to carry the same commit in two branches, then base the larger branch on top of the smaller one. Then if the smaller branch gets merged, the larger one can also be merged without rebasing because it is based on a commit that is in the target branch.

As it stands now, we would have two different "test cluster: Move to fedora-33 cloud image" commits in the "devel" branch if we merged this PR.

hack/stress-driver.sh
pohly: Ping...

test/e2e/operator/deployment_api.go
```diff
@@ -690,7 +722,7 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
 				},
 				"metrics service": func(dep *api.Deployment) apiruntime.Object {
 					obj := &corev1.Service{}
-					key := runtime.ObjectKey{Name: dep.MetricsServiceName(), Namespace: d.Namespace}
+					key := runtime.ObjectKey{Name: base.MetricsServiceName(dep), Namespace: d.Namespace}
```
pohly, line 725: Can this code churn be avoided?

test/e2e/operator/deployment_api.go
```diff
@@ -744,6 +776,84 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+					DeploymentSpec: base.DeploymentSpec{
+						Image: dummyImage,
+					},
+				},
+			}
+		}
+		It("with default values", func() {
+			if d.HasOLM {
+				Skip("Conversion webhook is not supported by OLM")
+			}
```
pohly, line 796: The conversion *webhook* may not be supported, but *conversion* is supported - it just works a bit differently.

In other words, don't skip this and the other tests below for OLM and instead just update the expectations where necessary.

pohly (changes requested):

test/e2e/operator/deployment_api.go
pohly: Yes, this can be avoided. There's no reason for moving these helper functions into the base package.

Partial patch:
```
diff --git a/pkg/apis/pmemcsi/base/base_types.go b/pkg/apis/pmemcsi/base/base_types.go
index 770600bcb..7bd6b2c11 100644
--- a/pkg/apis/pmemcsi/base/base_types.go
+++ b/pkg/apis/pmemcsi/base/base_types.go
@@ -349,12 +349,6 @@ func ControllerServiceName(d metav1.Object) string {
 	return GetHyphenedName(d) + "-controller"
 }
 
-// MetricsServiceName returns the name of the controller metrics
-// Service object used by the deployment
-func MetricsServiceName(d metav1.Object) string {
-	return GetHyphenedName(d) + "-metrics"
-}
-
 // ServiceAccountName returns the name of the ServiceAccount
 // object used by the deployment
 func ServiceAccountName(d metav1.Object) string {
diff --git a/pkg/apis/pmemcsi/v1beta1/deployment_types.go b/pkg/apis/pmemcsi/v1beta1/deployment_types.go
index 8cf85adb3..68c8b6eb9 100644
--- a/pkg/apis/pmemcsi/v1beta1/deployment_types.go
+++ b/pkg/apis/pmemcsi/v1beta1/deployment_types.go
@@ -158,3 +158,10 @@ func (d *Deployment) GetOwnerReference() metav1.OwnerReference {
 func (d *Deployment) HaveCertificatesConfigured() (bool, error) {
 	return d.Spec.HaveCertificatesConfigured()
 }
+
+// MetricsServiceName returns the name of the controller metrics
+// Service object used by the deployment
+func (d *Deployment) MetricsServiceName() string {
+	// TODO: also move GetHyphenedName and all other functions
+	return base.GetHyphenedName(d) + "-metrics"
+}
diff --git a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
index 346a5e3dc..3f13403e2 100644
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -497,7 +497,7 @@ func (d *PmemCSIDriver) redeployMetricsService(r *ReconcileDeployment) (apirunti
 		object: func() apiruntime.Object {
 			return &corev1.Service{
 				TypeMeta:   metav1.TypeMeta{Kind: "Service", APIVersion: "v1"},
-				ObjectMeta: d.getObjectMeta(base.MetricsServiceName(d), false),
+				ObjectMeta: d.getObjectMeta(d.MetricsServiceName(), false),
 			}
 		},
 		modify: func(o apiruntime.Object) error {
@@ -653,7 +653,7 @@ func (d *PmemCSIDriver) HandleEvent(meta metav1.Object, obj apiruntime.Object, r
 		klog.Infof("Redeploying service: %s", meta.GetName())
 		if meta.GetName() == base.ControllerServiceName(d) {
 			_, err = d.redeployControllerService(r)
-		} else if meta.GetName() == base.MetricsServiceName(d) {
+		} else if meta.GetName() == d.MetricsServiceName() {
 			_, err = d.redeployMetricsService(r)
 		}
 		if err != nil {
diff --git a/test/e2e/operator/deployment_api.go b/test/e2e/operator/deployment_api.go
index 297707f63..8b32bcac1 100644
--- a/test/e2e/operator/deployment_api.go
+++ b/test/e2e/operator/deployment_api.go
@@ -620,7 +620,7 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
 				},
 				"metrics service": func(dep *api.Deployment) apiruntime.Object {
 					return &corev1.Service{
-						ObjectMeta: metav1.ObjectMeta{Name: base.MetricsServiceName(dep), Namespace: d.Namespace},
+						ObjectMeta: metav1.ObjectMeta{Name: dep.MetricsServiceName(), Namespace: d.Namespace},
 					}
 				},
 				"provisioner role": func(dep *api.Deployment) apiruntime.Object {
@@ -722,7 +722,7 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
 				},
 				"metrics service": func(dep *api.Deployment) apiruntime.Object {
 					obj := &corev1.Service{}
-					key := runtime.ObjectKey{Name: base.MetricsServiceName(dep), Namespace: d.Namespace}
+					key := runtime.ObjectKey{Name: dep.MetricsServiceName(), Namespace: d.Namespace}
 					EventuallyWithOffset(1, func() error {
 						return client.Get(context.TODO(), key, obj)
 					}, "2m", "1s").ShouldNot(HaveOccurred(), "get metrics service set")
```

test/e2e/operator/deployment_api.go
avalluri: Yes, agree. I will modify the test.

test/e2e/operator/deployment_api.go
avalluri: > Can this code churn be avoided?

My initial changes were as you proposed. But, I felt like code duplication which we could avoid. I mean all(`alpha`, `beta`, and future) versions need not define the same function. So I moved them to the API `base` package.

My definition of base package is the 'common' code that could be shared by all versions without duplicating the code.

pkg/deployments/load.go
```diff
@@ -23,7 +24,7 @@ import (
 )
 
 // LoadObjects reads all objects stored in a pmem-csi.yaml reference file.
-func LoadObjects(kubernetes version.Version, deviceMode api.DeviceMode) ([]unstructured.Unstructured, error) {
+func LoadObjects(kubernetes version.Version, deviceMode base.DeviceMode) ([]unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
avalluri: > We can avoid this code churn by importing `DeviceMode` into the v1beta1 API. If we ever remove `v1alpha` and `base`, then the rest of the code doesn't need to change again

So you wanted to say that DeivceMode is 'v1beta' specific and going to be removed in v1?. My intention is that we ever need to drop the base package. The definitions from this package are shared by all future versions also.

hack/stress-driver.sh
avalluri: Yes, I agree. Will change the check.

test/e2e/operator/deployment_api.go
pohly: > I mean all(alpha, beta, and future) versions need not define the same function.

Why? It's only called for the API revision that the operator code uses, so the function is only needed there. Or am I missing something?

The cleanest solution would be to not define those helper functions in the API package at all. They are not part of the *API*, they just support code in the operator which works with the API object.

pkg/deployments/load.go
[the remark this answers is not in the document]
pohly: > So you wanted to say that DeivceMode is 'v1beta' specific and going to be removed in v1?

No. I'm just saying that `DeviceMode` is part of the `v1beta1` API and we avoid code churn by exposing it as definition in that package. I have no plans to remove anything at this point.

> The definitions from this package are shared by all future versions also.

Only as long as they use exactly the same definitions. If we add another device mode that isn't supported by older PMEM-CSI instances, then we can no longer just update the base package.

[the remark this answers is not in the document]
pohly: > If we add another device mode that isn't supported by older PMEM-CSI instances, then we can no longer just update the base package.

Actually, we already have that situation now. "fake" device mode is allowed in the v1beta1 API, but not in the v1alpha1 API.

deploy/kustomize/operator-webhook/kustomization.yaml
```diff
@@ -0,0 +1,33 @@
...
+- name: SERVICE_NAMESPACE
+  objref:
+    apiVersion: v1
+    kind: Service
+    name: pmem-csi-operator-service
+  fieldref:
+    fieldpath: metadata.namespace
+
+configurations:
+- kustomizeconfig.yaml
```
pohly, line 33: Nit: missing line break

deploy/kustomize/operator-webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,4 @@
+varReference:
+- kind: Certificate
+  group: cert-manager.io
+  path: spec/dnsNames
```
pohly, line 4: Nit: missing line break

deploy/kustomize/patches/crd-conversion-webhook-patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+    strategy: Webhook
+    webhook:
+      conversionReviewVersions: ["v1beta1", "v1alpha1"]
+      clientConfig:
+        #caBundle: Cg==
+        service:
+          name: pmem-csi-operator-service
+          namespace: default # should match with the operator namespace
+          port: 9443
+          path: /convert
```
pohly, line 23: Nit: missing line break

pohly (changes requested):

operator/operator.make
avalluri: Reported the [issue](https://github.com/operator-framework/operator-sdk/issues/4293).

hack/stress-driver.sh
avalluri: Done.

pkg/apis/pmemcsi/v1beta1/deployment_types_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+var _ = Describe("Operator", func() {
+
+	BeforeEach(func() {
+	})
+
+	Context("API", func() {
+		//
+		// TODO: Add input-validation tests for the API spec fields
```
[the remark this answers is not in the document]
avalluri: I feel like this TODO is outdated and no additional such test required. IIRC this to was in the context of validating device mode value., which was covered in a different place. I removed the TODO.

pkg/deployments/load.go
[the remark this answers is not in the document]
avalluri: I pushed the latest revision of PR that removes the additional `base` package. and avoids these changes.

deploy/kustomize/operator-webhook/kustomization.yaml
avalluri: Fixed.

deploy/kustomize/operator-webhook/kustomizeconfig.yaml
avalluri: Fixed.

deploy/kustomize/patches/crd-conversion-webhook-patch.yaml
avalluri: Fixed.

test/e2e/operator/deployment_api.go
avalluri: Done.

deploy/kustomize/crd/pmem-csi.intel.com_deployments.yaml
```diff
@@ -0,0 +1,543 @@
+
+---
```
pohly, line 2: Redundant blank line and item separator?

deploy/kustomize/operator-webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,5 @@
...
+
```
pohly, line 5: nit: extra line at the end.

deploy/kustomize/operator-webhook/webhook-certificate.yaml
```diff
@@ -0,0 +1,25 @@
...
+  dnsNames:
+  # dnsNames must match be: <webhook-service-name><.namespace>.svc[.domain]
+  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
+  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
+  secretName: pmem-csi-operator-service-cert # name referenced in: patches/operator-webhook-patch.yaml
+  issuerRef:
+    kind: Issuer
+    name: self-signed-issuer
+---
+
```
pohly, line 25: nit: extra separator and line at the end

deploy/kustomize/operator-webhook/webhook-service.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: pmem-csi-operator-service # name referenced in webhook-certificate.yaml for dnsNames
+  namespace: default
+spec:
+  ports:
+    - port: 9443
+      targetPort: 9443
+  selector:
+    name: pmem-csi-operator
+
```
pohly, line 12: nit: extra line

deploy/kustomize/patches/crd-conversion-webhook-patch.yaml
```diff
@@ -0,0 +1,24 @@
...
+
```
pohly, line 24: nit: extra line

pkg/pmem-csi-operator/main.go
```diff
@@ -16,13 +16,15 @@ import (
...
 
 	"github.com/intel/pmem-csi/pkg/apis"
-	api "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1beta1"
+	"github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1beta1"
 	"github.com/intel/pmem-csi/pkg/k8sutil"
 	"github.com/intel/pmem-csi/pkg/pmem-csi-operator/controller"
 
 	pmemcommon "github.com/intel/pmem-csi/pkg/pmem-common"
-
 	"github.com/operator-framework/operator-lib/leader"
```
pohly, line 24: nit: I find it useful to separate "our" imports from Go runtime imports (at the top) and third-party imports (below).

In other words, don't delete this blank line.

pohly (changes requested):
Mostly just some nits left now. Looks much better than with the "base" package (shorter diff!) and clean separation between different API versions.

pkg/pmem-csi-operator/main.go
pohly: Or rather, delete line 22 above this.

pohly:
"conversion with explicit values" in "olm.API" is broken (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-826/15/tests).

test/e2e/operator/deployment_api.go
pohly: "conversion with explicit values" fails: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-826/15/tests

pkg/pmem-csi-operator/controller/deployment/deployment_controller.go
```diff
@@ -14,7 +14,7 @@ import (
 	"sync"
 	"time"
 
-	pmemcsiv1alpha1 "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1alpha1"
+	pmemcsiv1beta1 "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1beta1"
```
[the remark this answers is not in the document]
avalluri: Done.

pkg/pmem-csi-operator/controller/deployment/deployment_controller_test.go
```diff
@@ -56,7 +56,7 @@ func getDeployment(d *pmemDeployment) *api.Deployment {
 	dep := &api.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
-			APIVersion: "pmem-csi.intel.com/v1alpha1",
+			APIVersion: "pmem-csi.intel.com/v1beta1",
```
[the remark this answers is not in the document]
avalluri: We have such a variable defined already named `SchemeGroupVersion` in each API version. I used that here and we need not touch in the future.

deploy/kustomize/crd/pmem-csi.intel.com_deployments.yaml
avalluri: In fact, this file should not be committed, I will remove this.

deploy/kustomize/operator-webhook/kustomizeconfig.yaml
avalluri: Fixed.

deploy/kustomize/operator-webhook/webhook-certificate.yaml
avalluri: Fixed.

deploy/kustomize/operator-webhook/webhook-service.yaml
avalluri: Fixed.

deploy/kustomize/patches/crd-conversion-webhook-patch.yaml
avalluri: Fixed.

pkg/pmem-csi-operator/main.go
avalluri: Fixed.

pohly:
Not sure what changed, but now "fedora-1_17.olm.API" is very unhappy (92 test failures in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-826/18/tests).

avalluri:
> Not sure what changed, but now "fedora-1_17.olm.API" is very unhappy (92 test failures in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-826/18/tests).

The error seems not related to this particular change. Somehow installed OLM fails to start the operator: 
```
+ /mnt/workspace/pmem-csi_PR-826/_work/bin/operator-sdk run packagemanifests --version 0.8.0 --namespace default --install-mode OwnNamespace --timeout 3m /tmp/tmp.W3oTimKXBK/olm-catalog
time="2020-12-09T12:40:39Z" level=info msg="Creating pmem-csi-operator registry"
time="2020-12-09T12:40:39Z" level=info msg=" Creating ConfigMap \"default/pmem-csi-operator-registry-manifests-package\""
time="2020-12-09T12:40:39Z" level=info msg=" Creating ConfigMap \"default/pmem-csi-operator-registry-manifests-0-8-0\""
time="2020-12-09T12:40:39Z" level=info msg=" Creating Deployment \"default/pmem-csi-operator-registry-server\""
time="2020-12-09T12:40:39Z" level=info msg=" Creating Service \"default/pmem-csi-operator-registry-server\""
time="2020-12-09T12:40:39Z" level=info msg="Waiting for Deployment \"default/pmem-csi-operator-registry-server\" rollout to complete"
time="2020-12-09T12:40:39Z" level=info msg=" Waiting for Deployment \"default/pmem-csi-operator-registry-server\" to rollout: 0 out of 1 new replicas have been updated"
time="2020-12-09T12:40:40Z" level=info msg=" Waiting for Deployment \"default/pmem-csi-operator-registry-server\" to rollout: 0 of 1 updated replicas are available"
time="2020-12-09T12:40:49Z" level=info msg=" Deployment \"default/pmem-csi-operator-registry-server\" successfully rolled out"
time="2020-12-09T12:40:49Z" level=info msg="Created CatalogSource: pmem-csi-operator-catalog"
time="2020-12-09T12:40:49Z" level=info msg="OperatorGroup \"operator-sdk-og\" created"
time="2020-12-09T12:40:49Z" level=info msg="Created Subscription: pmem-csi-operator-v0-8-0-sub"
time="2020-12-09T12:40:52Z" level=fatal msg="Failed to run packagemanifests: error approving install plan: Operation cannot be fulfilled on installplans.operators.coreos.com \"install-m4tpt\": the object has been modified; please apply your changes to the latest version and try again\n"
```

pohly:
> The error seems not related to this particular change. Somehow installed OLM fails to start the operator:

Can you file an upstream bug for this?

test/e2e/operator/deployment_api.go
```diff
@@ -744,6 +769,84 @@ var _ = deploy.DescribeForSome("API", func(d *deploy.Deployment) bool {
...
+			alphaDep := getAlphaDeployment("alpha-default-values")
+			deploy.CreateAlphaDeploymentCR(f, alphaDep)
+
+			deployment := deploy.GetDeploymentCR(f, alphaDep.Name)
+
+			defer deploy.DeleteDeploymentCR(f, deployment.Name)
+
+			validateDriver(deployment, true)
+
+			// Expect to ruturn alpha object converting from stored beta CR
```
pohly, line 794: I missed this when we merged the PR, but now I am looking at a test failure in this test after fixing the default log level and have some more questions.

Why does it say here "from stored beta CR"? The stored CR is in the v1alpha1 format, isn't it?

avalluri: Stored version is always the latest(`spec.storage: true` in the CRD) version. And when the conversion happens to client requested version from that.

avalluri:
> > The error seems not related to this particular change. Somehow installed OLM fails to start the operator:
> 
> Can you file an upstream bug for this?

I filed the [bug](https://github.com/operator-framework/operator-sdk/issues/4308) and also submitted the [fix](https://github.com/operator-framework/operator-sdk/pull/4309).

test/e2e/operator/deployment_api.go
pohly: > Extended the tests to make sure that the v1beta1 ->``v1alpha1` conversion works:

This isn't really converting something that was created as v1beta1, is it?

avalluri: > > Extended the tests to make sure that the v1beta1 ->``v1alpha1` conversion works:
> 
> This isn't really converting something that was created as v1beta1, is it?

The test creates a `v1alpha1` object, which actually stored as the `v1beta1` by the API server after the conversion. Then the test retrieves back both `v1alpah1` and `v1beta1` versions for the same CR, and checks if both have expected fields.

PR #836, opened 2020-12-18
cluster setup

pohly:
PR #826 added the installation of cert-manager, which turned out to be surprisingly slow for me because downloads of GitHub release artifacts are slow. Caching the YAML file mitigates that. Doesn't seem to affect the CI.

PR #829 changed how the YAML files handle namespaces. This didn't work because secrets couldn't be created. This PR moves the namespace creation into the secret creation because that should happen first. An even cleaner solution would be to add an explicit first step, but then usability is worse.

The PR also contains various other minor changes (code cleanup, language).


avalluri:
> This didn't work because secrets couldn't be created.

@pohly Have you considered moving the secret creation part to post driver deployment and keeping the namespace creation in YAMLS. I just thinking if the user do not use the `setup-ca-kubernetes.sh` script and provides their certificates? Then the namespace creation gets skipped.

pohly:
> Have you considered moving the secret creation part to post driver deployment and keeping the namespace creation in YAMLS.

I have considered that. It will cause error events to be emitted for pods because they cannot start without their secrets, which is not desirable.

> I just thinking if the user do not use the setup-ca-kubernetes.sh script and provides their certificates? Then the namespace creation gets skipped.

If they use a different way of creating secrets in the desired namespace, then they'll know that they have to create that namespace first.

avalluri (approved):

pohly:
merged via PR #837

PR #825, opened 2020-11-24
structured logging

pohly:
This lays the foundation and converts the operator as a smaller piece of code to structured logging.

Both pmem-csi-driver and pmem-csi-operator now have a `-logging-format=json|text` command line parameter.

TODO:
- [x] ~~support choosing the logging format of the operator when deploying through OLM~~ - cannot be implemented at the moment, operator defaults to plain text
- [x] support choosing the logging format of the driver when deploying through the operator
- [x] admin documentation

Issue #733, opened 2020-09-14
volume leak during deployment rollout


pohly:
The controller is designed such that it collects information about volumes from nodes as the nodes register themselves. This implies that the controller cannot know about existing volumes for nodes that haven't registered (yet).

This leads to the following problem:
- DeleteVolume is called for an existing volume that the controller doesn't know about at the moment.
- The controller cannot distinguish between "volume already deleted" (idempotency!) and "need to wait for some node with that volume".
- It assumes that the volume is gone and returns success without doing anything, after a misleading log message about "Volume pvc-bd-adc62b1395a868c243a74ee138e313a19c72211c5fbc0d5f2706e486 not created by this controller".
- external-provisioner removes PV

=> volume leak

This problem was triggered by the new version skew tests which restart the driver while volumes exist, then does some operations (including removal) with them right after the driver deployment comes up again.

pohly:
The whole registration process is also problematic because of https://github.com/intel/pmem-csi/issues/729. It may also be a scalability problem (single instance of the controller), although we don't have any evidence of that because we haven't done scale testing.

I'm currently leaning towards solving this problem by deploying the external-provisioner alongside each node driver instance and removing the central controller entirely. This was originally proposed in https://github.com/kubernetes-csi/external-provisioner/pull/367 but wasn't finished.

What was missing in that PR was support for immediate binding. With immediate binding, all external-provisioner instances need to collaboratively figure out who's the one who should create the volume. I was proposing leadership election for that (https://github.com/kubernetes-csi/external-provisioner/pull/367#discussion_r336886155) but that'll need further thought.

avalluri:
I think the other alternate/quick solution is to persist the controller state(known volume details) in a config map.

pohly:
> I think the other alternate/quick solution is to persist the controller state(known volume details) in a config map.

The controller currently doesn't have access to a config map. Introducing that would introduce a dependency on Kubernetes into the CO-agnostic part of PMEM-CSI.

Even with a config map, getting this right in all cases will be tricky. Consider the naive approach:
- controller gets CreateVolume request
- sends CreateVolume to node
- records new volume in config map
- returns information to external-provisioner

If the controller dies at any point during this flow, the volume may leak. A lot of effort went into external-provisioner to prevent such leaks; we would have to duplicate all of that.

pohly:
This is a valid alternative solution, but I wouldn't call it quick.

pohly:
[...]

Issue #729, opened 2020-09-10
stale node entries in controller registry


pohly:
When a node has registered with the controller and then exists (either during a normal shutdown of that node or because the driver dies unexpectedly), the registry has a stale entry.

We need a strategy for dealing with those stale entries. Right now, the controller will try to connect to the dead nodes in perpetuity (?).

okartau:
Does the controller keep track when/if node is alive and good, and can it easily count connection attempts? The controller could remove stale entries based on some policy, if counting failed attempts/measuring "dead" time is easily doable.

pohly:
I don't think it currently does, but something like that should be easily doable.

pohly:
Resolved via PR #838

Issue #594, opened 2020-04-18
operator shall renew certificates generated on expiry


avalluri:
The operator generated self-signed certificates must be renewed periodically. Otherwise, a deployment that is over one year old will fail.

pohly:
Obsoleted by PR #838

PR #838, opened 2020-12-18
Distributed provisioning

pohly:
This changes the architecture of PMEM-CSI such that the controller part only implements the webhooks and provisioning happens on the nodes.

This PR is based on some pending changes (PR #825, PR #836) and uses the external-provisioner canary image because v2.1.0 has been tagged, but the versioned image is not available yet. Therefore the PR is ready for review and further testing, but merging should wait.

Fixes: https://github.com/intel/pmem-csi/issues/823, https://github.com/intel/pmem-csi/issues/733, https://github.com/intel/pmem-csi/issues/729, https://github.com/intel/pmem-csi/issues/594, https://github.com/intel/pmem-csi/issues/526, https://github.com/intel/pmem-csi/issues/507


pohly:
I've rebased, so only new commits are present now.

I also re-enabled the stress test. It passed in local testing.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -1026,25 +910,14 @@ func (d *pmemCSIDeployment) getControllerStatefulSet(ss *appsv1.StatefulSet) {
...
-						VolumeSource: corev1.VolumeSource{
-							EmptyDir: &corev1.EmptyDirVolumeSource{},
-						},
-					},
-					{
-						Name: "registry-cert",
+						Name: "webhook-cert",
 						VolumeSource: corev1.VolumeSource{
 							Secret: &corev1.SecretVolumeSource{
+								// TODO: replace name, create secret via cert-manager
```
pohly, line 916: This is a bit more complicated. When deploying via OLM, we need OLM to create the certificates. My understanding is that this can be done for "normal" webhooks, just not for conversion webhooks.

For deployment of the operator via YAML files we need to document how to set up those certs.

In both cases, re-configuring the cluster to use the webhooks is something that needs to be done manually.

If the user of the operator doesn't need the scheduler extensions, then it's still okay to start the statefulset unconditionally. The pod simply won't run.

deploy/kubernetes-1.17/direct/pmem-csi.yaml
```diff
@@ -223,42 +303,27 @@ spec:
...
         - /usr/local/bin/pmem-csi-driver
         - -v=3
         - -logging-format=text
-        - -mode=controller
-        - -endpoint=unix:///csi/csi-controller.sock
-        - -registryEndpoint=tcp://0.0.0.0:10000
-        - -nodeid=$(KUBE_NODE_NAME)
+        - -mode=webhooks
+        - -schedulerListen=:8000
+        - -drivername=$(PMEM_CSI_DRIVER_NAME)
```
pohly, line 313: We still need the external-provisioner in the controller as fallback for https://github.com/kubernetes-csi/external-provisioner/issues/544

pohly: ... or a custom controller which removes the selected-node annotation.

pohly: ... or just set `AllowedTopologies` in our storage classes for late binding (https://github.com/kubernetes-csi/external-provisioner/issues/544#issuecomment-748871274).

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
```diff
@@ -0,0 +1,65 @@
...
+patchesJSON6902:
+- target:
+    group: apps
+    version: v1
+    kind: StatefulSet
+    name: pmem-csi-controller
+  patch: |-
+    - op: add
+      path: /spec/template/spec/containers/0/command/-
+      value: -mode=webhooks
```
avalluri, line 16: What does this "webhooks" mode does? Isn't it that valid mods are 'controller' and 'node'?

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
```diff
@@ -0,0 +1,65 @@
...
+      path: /spec/template/spec/containers/0/command/-
+      value: --schedulerListen=:8000
+    - op: add
+      path: /spec/template/spec/containers/0/command/-
+      value: -kube-api-qps=5.0
+    - op: add
+      path: /spec/template/spec/containers/0/command/-
+      value: -kube-api-burst=10
+    - op: remove
+      path: /spec/template/spec/containers/1
```
avalluri, line 27: Why can't we do theses(remove external provisioner from controller) changes to our base driver YAML instead of a kustomize layer.

docs/design.md
```diff
@@ -155,88 +155,35 @@ Kata Container support has to be enabled explicitly via a [storage
 class parameter and Kata Containers must be set up
 appropriately](install.md#kata-containers-support)
 
-## Driver modes
```
avalluri, line 158: The reference is not removed from the table of contents.

docs/design.md
```diff
@@ -155,88 +155,35 @@ Kata Container support has to be enabled explicitly via a [storage
...
+
+Traditionally, Kubernetes expects that a driver deployment has a
+central component, usually implemented with the `external-provisioner`
+and a custom CSI driver component which implements volume creation.
+That central component is hard to implement for a CSI driver that
+creates volumes locally on a node.
+
+PMEM-CSI solves this problem by deploying `external-provisioner`
+alongside each node driver and enabling ["distributed
+provisioning"](https://github.com/kubernetes-csi/external-provisioner/issues/487):
```
avalluri, line 168: Wouldn't it be better to link to the documentation if any  instead of the issue.

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
avalluri: Okay, after going through the design documentation changes, I understood that the new 'webhooks' mode replaces the existing 'controller' mode. But the driver command-line flags in [`pkg/pmem-csi-dirver/main.go`](https://github.com/pohly/pmem-CSI/blob/distributed-provisioning/pkg/pmem-csi-driver/main.go#L38) is not synced with this.

docs/design.md
```diff
@@ -155,88 +155,35 @@ Kata Container support has to be enabled explicitly via a [storage
...
+  Once a node has been selected, provisioning continues as with late
+  binding. This is less efficient and therefore "late binding" is the
+  recommended binding mode.
+
+PMEM-CSI also has a central component which implements the [scheduler
+extender](#scheduler-extender) webhook. That component needs to know
+on which nodes the PMEM-CSI driver is running and how much capacity is
+available there. This information is retrieved by dynamically
+discovering PMEM-CSI pods and connecting to their [metrics
+endpoint](/docs/install.md#metrics-support).
```
avalluri, line 186: This makes 'enabling the metrics support' a hard dependency on 'scheduler extension'. The code should implement this by always enabling node metrics when scheduler extension is enabled.

docs/install.md
```diff
@@ -8,7 +8,7 @@
     - [Install PMEM-CSI driver](#install-pmem-csi-driver)
       - [Install using the operator](#install-using-the-operator)
       - [Install from source](#install-from-source)
-    - [Expose persistent and cache volumes to applications](#expose-persistent-and-cache-volumes-to-applications)
+    - [Expose volumes to applications](#expose-volumes-to-applications)
```
avalluri, line 11: Links in Table of Contents has broken.

pkg/pmem-csi-driver/main.go
```diff
@@ -71,7 +67,7 @@ func Main() int {
 	klog.V(3).Info("Version: ", version)
 
 	if config.schedulerListen != "" {
-		if config.Mode != Controller {
+		if config.Mode != Webhooks {
```
avalluri: I feel like offloading the webhook and scheduler extension part from the CSI driver to a separate binary makes much sense and cleaner than treating it as a mode of the driver.

pohly: What's the benefit? Have you considered the drawbacks?

"Cleaner" is subjective.

docs/design.md
pohly: Yes, now that that it is merged. Will change it.

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
pohly: We can and I thought I had made that change when squashing. I'll check again.... it looks like I simply forgot to remove this file. It shouldn't be needed anymore.

Initially it wasn't done because distributed and centralized provisioning were both supported for a while.

docs/design.md
pohly: Does the node part know whether scheduler extensions are enabled? I don't think so, because those parameters are passed to the controller part.

pohly: Or did you mean some other code than the main.go and pmem-csi-driver.go?

pkg/pmem-csi-driver/main.go
pohly: Is this perhaps about publishing one image that has multiple purposed vs. two different images with different content and different purpose?

I prefer to stick with a single image until someone has some strong argument against it. I doubt that the size of an image that only runs once in a cluster matters that much and publishing a single image is simpler, so I prefer that.

When we have a single image, then a single binary is likely to be smaller overall.

What that binary is called and how it is invoked hardly matters for end-users - they never invoke the binary. Renaming the binary is possible, but makes this PR more complex for (IMHO) little gain.

docs/design.md
avalluri: > Does the node part know whether scheduler extensions are enabled? I don't think so, because those parameters are passed to the controller part.

Yes, your are true, it is not possible to enforce this in the code. Would be good to add a note about "enabling metrics while using scheduler extension" in the documentation.

pkg/pmem-csi-driver/main.go
avalluri: My point was not about different images, more about a different binary. As now there is no central controller  driver component, IMHO the scheduler code is not suitable to be part of the driver code.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
avalluri: > This is a bit more complicated. When deploying via OLM, we need OLM to create the certificates. My understanding is that this can be done for "normal" webhooks, just not for conversion webhooks.

I doubt if OLM can creates certificates for services created by the operator for the CRs, though it provides service certificates for operator itself.

avalluri: @pohly How this scheduler extension and the Pod mutation works in the context of multiple PmemCSIDeployments? I think we should /cannot support multiple deployments.

pkg/pmem-csi-driver/main.go
pohly: We discussed this in the standup meeting.

"pmem-csi-driver" has already before served multiple different purposes and we agreed to not change that in this PR because doing so would make our image larger for not much gain (basically cleaner command line arguments of the binary which isn't that important).

We also raised the question whether "pmem-csi-operator" should be a separate binary. Probably not. But again, that is a different issue.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: For the mutating pod webhook: it must use a resource name which includes the driver name. I thought it already did this, but currently it doesn't (https://github.com/intel/pmem-csi/blob/692818be4679e826fc7a98e77ad0eeb61b40dba5/pkg/scheduler/mutate_pod.go#L31).

For the scheduler extender: it must be triggered by the same resource name (driver name to be added) and then check only pods that belong to that instance (already implemented: https://github.com/intel/pmem-csi/blob/82bcde130b32d006b0da5b11729373a91ae575d1/pkg/scheduler/capacity.go#L103-L106)

pohly: > I doubt if OLM can creates certificates for services created by the operator for the CRs, though it provides service certificates for operator itself.

When we discussed it this morning, my conclusion was that the user needs to supply TLS credentials and that the webhooks need to be disabled if not provided.

pohly: > In both cases, re-configuring the cluster to use the webhooks is something that needs to be done manually.

Have I mentioned how much I hate configuring a scheduler extension? :cry: 

The way we currently do it is by creating the certificates first, then setting up the "kubeadm init" configuration so that those certificates can be used by `kube-scheduler`. This no longer works when creating those certificates *inside* the cluster with a tool like cert-manager.

It is possible to set up the cluster without the extension enabled, then create certificates, and then reconfigure `kube-scheduler`. `kubeadm upgrade` supports that. But this looks complicated. I think I'll stick with the current approach.

pohly:
I've pushed a major update: now the operator supports configuration of the webhooks and distributed provisioning. Some E2E tests passed (in particular the ones for the operator), but for full coverage I let the CI do its work over night.

I'll double-check tomorrow, but I think there is only one minor TODO left (the one about reading the node port from the shell script in deploy.go).

I've not tried to ensure that all tests pass for all commits, so strictly speaking I would have to squash "distributed provisioning", "deploy: consistent object naming, app.kubernetes.io labels" and "operator: distributed provisioning, scheduler extensions". But I think it's easier to review when they are separate.

pohly:
> I'll double-check tomorrow, but I think there is only one minor TODO left (the one about reading the node port from the shell script in deploy.go).

That referred to TODOs inside the patch itself. I still need to address review comments.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -104,9 +98,6 @@ type Config struct {
 	NodeID string
 	//Endpoint exported csi driver endpoint
 	Endpoint string
-	//TestEndpoint adds the controller service to the server listening on Endpoint.
-	//Only needed for testing.
-	TestEndpoint bool
 	//Mode mode fo the driver
 	Mode DriverMode
 	//RegistryEndpoint exported registry server endpoint
```
avalluri, line 104: `RegistryEndpoint` no more needed and could be removed.

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
pohly: Fixed.

deploy/kustomize/kubernetes-distributed-lvm/kustomization.yaml
pohly: Fixed.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -251,24 +189,23 @@ func (csid *csiDriver) Run() error {
...
+		factory := informers.NewSharedInformerFactoryWithOptions(csid.cfg.client, resyncPeriod,
+			informers.WithNamespace(namespace),
+		)
+		podLister := factory.Core().V1().Pods().Lister()
+		c := scheduler.CapacityViaMetrics(namespace, csid.cfg.DriverName, podLister)
+		factory.Start(ctx.Done())
+		if _, err := csid.startScheduler(ctx, cancel, c); err != nil {
+			return err
 		}
 	case Node:
```
avalluri, line 228: We can move identity server creation from https://github.com/intel/pmem-csi/blob/7b7f17953f9a9cb79bb754e4e5bf94d4c9c1076d/pkg/pmem-csi-driver/pmem-csi-driver.go#L169-L172 to here.

docs/design.md
pohly: Fixed.

docs/design.md
pohly: It's documented that the scheduler extension code uses metrics. I think that is enough. We don't describe how to disable metrics support (it's always on in YAML and operator deployments), so the risk that someone accidentally deploys incorrectly is low.

docs/install.md
pohly: Fixed, also in the body.

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: Removed (but not pushed yet).

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: Done. I also removed the unused error return value to make it simpler and consistent.

docs/install.md
```diff
@@ -752,41 +752,27 @@ The PMEM-CSI scheduler extender and admission webhook are provided by
...
+The webhook needs TLS certificates which must be created in
+advance. The YAML files expects them in a secret called
+`pmem-csi-intel-com-controller-secret` and will not work without one.
+The operator is more flexible and creates a driver without the
+scheduler extensions by default. This can be changed by setting the
+`controllerTLSSecret` field in the `PmemCSIDeployment` API.
+
+That secret must contain the following data items:
+- `ca.crt`: root CA certificate
+- `tls.key`: secret key of the webhook
```
avalluri, line 766: Wouldn't be better to mention that the secret/public key is for both webhook and scheduler extender.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -377,6 +424,7 @@ var subObjectHandlers = map[string]redeployObject{
 	},
 	"metrics service": {
 		objType: reflect.TypeOf(&corev1.Service{}),
+		enabled: controllerEnabled,
```
avalluri, line 427: Node metrics service cant be enabled regardless of the controller?

pkg/pmem-csi-driver/pmem-csi-driver.go
pohly: Done.

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -5,6 +5,87 @@ metadata:
   namespace: default
 ---
 apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    pmem-csi.intel.com/deployment: lvm-testing
```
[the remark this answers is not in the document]
pohly: Yes. The usage of the "pmem-csi.intel.com/deployment" to mark different kinds of deployments in deploy.go doesn't change.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
pohly: Node metrics are always enabled. There's nothing in the operator API to control that.

docs/install.md
pohly: The scheduler extender is also a webhook. Both run in the controller.

With that in mind, it's better to say above that "The *controller* needs TLS certificates" - i'll use that instead. The name of the API was already chosen accordingly.

pkg/pmem-csi-driver/pmem-csi-driver.go
```diff
@@ -188,6 +196,16 @@ func (csid *csiDriver) Run() error {
 	)
 	csid.gatherers = append(csid.gatherers, cmm.GetRegistry())
 
+	var client kubernetes.Interface
+	if config.schedulerListen != "" ||
+		csid.cfg.Mode == Node {
```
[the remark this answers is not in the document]
pohly: Indeed, good catch. The initialization of `cmm` can also be moved.

deploy/kustomize/storageclass/pmem-storageclass.yaml
```diff
@@ -3,3 +3,12 @@ kind: StorageClass
...
 provisioner: pmem-csi.intel.com
+# Setting allowedTopology ensures that pods using volumes
+# with late binding only get scheduled to pmem
+allowedTopologies:
+- matchLabelExpressions:
+  # The "pmem-csi.intel.com" part is the driver name.
+  # It must match the "provisioner" field above.
+  - key: pmem-csi.intel.com/driver
+    values:
+    - started
```
pohly, line 14: Looks like this breaks skew testing; I've not looked into it yet, but probably it is related to the older deployment not setting this label.

I wanted to ensure that only nodes where the driver actually runs (or at least ran) is used, vs. nodes where it is supposed to run. The latter is what we would get when using the label that admins need to set for the nodes with PMEM.

Will have to think about this some more...

pohly: The problem is that the 0.8.0 deployment doesn't handle the new topology:
```
pmem-csi-controller-0/pmem-driver@pmem..ker1: I0118 11:38:34.703703       1 tracing.go:19] GRPC call: /csi.v1.Controller/CreateVolume
pmem-csi-controller-0/pmem-driver@pmem..ker1: I0118 11:38:34.703770       1 tracing.go:20] GRPC request: {"accessibility_requirements":{"preferred":[{"segments":{"pmem-csi.intel.com/driver":"started"}}],"requisite":[{"segments":{"pmem-csi.intel.com/driver":"started"}}]},"capacity_range":{"required_bytes":115343360},"name":"pvc-269396e3-8ad6-4b4a-8296-9944cdd3acf0","parameters":{"eraseafter":"true"},"volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}}]}
pmem-csi-controller-0/pmem-driver@pmem..ker1: I0118 11:38:34.704225       1 controllerserver-master.go:181] Controller CreateVolume: Name:pvc-269396e3-8ad6-4b4a-8296-9944cdd3acf0 required_bytes:115343360 limit_bytes:0
pmem-csi-controller-0/pmem-driver@pmem..ker1: I0118 11:38:34.704251       1 controllerserver-master.go:86] Controller CreateVolume: Create VolumeID:pvc-26-328dad453a5df7f8348d57f9f7422e6b084a58303f5400af6f92906a based on name:pvc-269396e3-8ad6-4b4a-8296-9944cdd3acf0
pmem-csi-controller-0/pmem-driver@pmem..ker1: W0118 11:38:34.704273       1 controllerserver-master.go:236] failed to connect to : No node registered with id: 
```

I'm a bit surprised that "pmem-csi.intel.com/node" doesn't show up at all; might be related to how we configure the external-provisioner. Anyway, we can't change our topology labels without breaking compatibility, so I'll switch to using the node selector labels that we use for the PMEM-CSI node pods also for the storage classes.

pohly: Using our node selector label also doesn't work because the external-provisioner only seems to allow node labels that were set by the CSI driver:
```
pmem-csi-intel-com-node-q94vv/external-provisioner@pmem..ker2: I0118 14:46:54.540845       1 event.go:282] Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"skew-4425", Name:"direct-testingv4nl9", UID:"fb94e3a9-62e5-44ec-b48b-33471452142c", APIVersion:"v1", ResourceVersion:"1204", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "pmem-pmem-csi-sc-ext4-skew-4425-unused": error generating accessibility requirements: selected node '"pmem-csi-pmem-govm-worker2"' topology 'map[pmem-csi.intel.com/node:pmem-csi-pmem-govm-worker2]' is not in allowed topologies: [map[feature.node.kubernetes.io/memory-nv.dax:true]]
```

pohly: At this point, running a "deprovisioner" in the PMEM-CSI controller looks like the simplest solution.

pohly: I've implemented that, albeit with the name "(PVC) rescheduler" because that's what it is.

I decided to keep the commits for the previous two attempts because it is something that we might want to refer to. I could also squash.

I used the existing external-provisioner helper lib. It's not a perfect fit because it needs a PV informer without ever doing anything with PVs, but using it was simpler than writing everything from scratch (which probably would have amounted to a lot of copy-and-paste).

I've pushed after some local testing ("make test", the new late binding test, some skew testing). 

@avalluri can you take another look?

pohly: I've filed https://github.com/kubernetes-csi/external-provisioner/issues/548 to discuss why we can't use the node label in `AllowedTopology`. With a modified external-provisioner we might be able to restore that solution, which would be better than the current solution alone because then a bad node selection should happen a lot less. Even if we can enable that again, rescheduling as a second line of defense will still be useful (think "node selected base on labels", "labels change", "provisioning no longer possible" -> "must reschedule").

pohly:
I intentionally left https://github.com/intel/pmem-csi/issues/857 out of the current implementation. It's not going to be a problem for a while and can be addressed later.

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -36,6 +36,15 @@ metadata:
...
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumes
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
```
pohly, line 47: With these changes, the `ClusterRole` becomes a combination of permissions for the scheduler extensions and for the rescheduler.

I suspect putting "webhooks" into the name of the `ClusterRole` was a poor choice from the beginning; "controller" would have been better. I didn't change it in this commit because I wanted to keep it small, but I can do it in a separate cleanup commit.

deploy/kustomize/storageclass/pmem-storageclass.yaml
```diff
@@ -3,12 +3,3 @@ kind: StorageClass
 metadata:
   name: pmem-csi-sc
 provisioner: pmem-csi.intel.com
-# Setting allowedTopology ensures that pods using volumes
-# with late binding only get scheduled to pmem
-allowedTopologies:
```
avalluri: @pohly Why should we remove `allowedTopologies` form the storageclass?

pkg/pmem-csi-driver/main.go
```diff
@@ -48,6 +48,7 @@ func init() {
 
 	/* Controller mode options */
 	flag.StringVar(&config.schedulerListen, "schedulerListen", "", "controller: listen address (like :8000) for scheduler extender and mutating webhook, disabled by default")
+	flag.Var(&config.nodeSelector, "nodeSelector", "controller: reschedule PVCs with a selected node where PMEM-CSI is not meant to run because the node does not have these labels (represented as JSON map)")
```
avalluri, line 51: I am bit confused, where exactly the provided `nodeSelecter`  is used in the code, I could only see that this is used to check if the rescheduler should be started or not by the controller. But I couldn't find anywhere the contents of the provided nodeSeleter is used in the code. In this case why can't it be a simple boolean value. @pohly  Pardon and please correct me if I miss something.

deploy/kustomize/storageclass/pmem-storageclass.yaml
pohly: Because it didn't work - see the commit message in https://github.com/intel/pmem-csi/pull/838/commits/7c4ee8639eb7729a6ea455a1d1dcbca0af26d5de

pkg/pmem-csi-driver/main.go
pohly: It's passed to the rescheduler here: https://github.com/intel/pmem-csi/blob/3a7d2cfc9483ad24c52c83a833865c02e158ea35/pkg/pmem-csi-driver/pmem-csi-driver.go#L219-L223

avalluri: Yes, but it is not used used by rescheduler?

pohly: And it should be used here: https://github.com/intel/pmem-csi/blob/3a7d2cfc9483ad24c52c83a833865c02e158ea35/pkg/pmem-csi-driver/rescheduler.go#L196-L201

... but isn't! Duh. That code is of course bogus. Not sure what I was thinking. I'll fix it.

pkg/pmem-csi-driver/rescheduler.go
```diff
@@ -0,0 +1,215 @@
...
+	if node == nil {
+		// Decide only based on CSINode.
+		reschedule := !driverIsRunning
+		l.V(3).Info("result", "reschedule", reschedule, "driverIsRunning", driverIsRunning)
+		return reschedule, nil
+	}
+
+	driverMightRun := true
+	for key, value := range node.Labels {
+		if node.Labels[key] == value {
```
avalluri, line 198: I think this check is wrong, It is where `pcp.nodeSelecter` supposed to be used?

pkg/pmem-csi-driver/main.go
pohly: I also need to add unit tests.

pkg/pmem-csi-driver/rescheduler.go
pohly: Yes.

deploy/kustomize/storageclass/pmem-storageclass.yaml
avalluri: If I am not wrong That attempt was for using `{"pmem-csi.intel.com/driver":"started"}` as allowed topology. Isn't it?. But it should work for `{storage: "pmem}` as before. And anyways your are passing the same label for controller's `--nodeSelecter` [here](https://github.com/intel/pmem-csi/pull/838/commits/3a7d2cfc9483ad24c52c83a833865c02e158ea35?file-filters%5B%5D=.csv&file-filters%5B%5D=.go&file-filters%5B%5D=.make&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml#diff-c4a858c419d0c78037540b85e6cb044a293f53c45c5a073e68108ed890430ea6R183).  Then why exactly this `--nodeSelector` is needed as we get the same info from from the storage class?

pkg/pmem-csi-driver/rescheduler.go
```diff
@@ -0,0 +1,215 @@
...
+			driverMightRun = false
+		}
+	}
+
+	reschedule := !driverMightRun && !driverIsRunning
```
avalluri, line 203: Isn't it that we should make sure that the selected node has enough PMEM capacity instead of just checking if the driver is running on this node?

pohly: If the driver is running, then it will do that check itself.

Checking capacity here is redundant and racy (we check capacity, fail, removed selected node, while sometime later the driver processes the PVC and creates a volume although it is no longer assigned to the node).

pohly: In other words, the rescheduler only may handle the case where the driver definitely isn't running.

deploy/kustomize/storageclass/pmem-storageclass.yaml
pohly: > If I am not wrong That attempt was for using `{"pmem-csi.intel.com/driver":"started"}` as allowed topology.

That's the first attempt (https://github.com/intel/pmem-csi/pull/838/commits/a03a7004e8288fca31b13a752019399328f37256). The second attempt was `{storage: pmem}` which also didn't work.

> Then why exactly this --nodeSelector is needed as we get the same info from from the storage class?

We can't use `AllowedTopology` at all at the moment, not with the currently released external-provisioner. Even if we can start using `AllowedTopology` with some future external-provisioner, then the label check in kube-scheduler may still go wrong when labels get changed (less likely, but can happen). The rescheduler then handles that case and thus remains useful.

pkg/pmem-csi-driver/rescheduler.go
pohly: I've fixed the commit.

pkg/pmem-csi-driver/main.go
pohly: Done, see updated commit.

avalluri (approved):
Looks good to me

pohly:
There was a test flake (?), filed as https://github.com/intel/pmem-csi/issues/858.

I don't think that it was caused by this PR, so I am going to merge it despite that.

PR #851, opened 2021-01-13
doc: a warning note about delete operator

### LINKED ISSUE #835: operator: uninstall operator, keep driver running
What happens when the operator is installed, deploys a driver, and then gets uninstalled?

With the recent change of adding the Namespace object to `deploy/kustomize/operator`, I suspect that the namespace and thus also the driver will get removed (untested!). This is probably not what we want.

How is that handled when using OLM?

avalluri:
FIXES #835

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -269,6 +269,12 @@ $ kubectl create -f https://github.com/intel/pmem-csi/raw/devel/deploy/operator/
 ```
 The operator gets deployed in a namespace called 'pmem-csi' which gets created by that YAML file.
 
+**WARNING:** This YAML file cannot be used to stop just the operator while
+keeping the PMEM-CSI deployments running. That's because something like
+`kubectl delete -f pmem-csi-operator.yaml` will delete the `pmem-csi`
+namespace which then also causes all PMEM-CSI deployments that might have
+been created in that namespace to be deleted.
+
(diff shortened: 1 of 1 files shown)


pohly:
@avalluri Please rebase.

docs/install.md
```diff
@@ -270,6 +270,8 @@ $ kubectl create -f https://github.com/intel/pmem-csi/raw/devel/deploy/operator/
 ```
 The operator gets deployed in a namespace called 'pmem-csi' which gets created by that YAML file.
 
+**WARNING:** Using this YAML to stop(`kubectl delete -f`) the running operator is not recommended as it deletes the `pmem-csi` namespace object that results in bringing down the PMEM-CSI driver pods if any that were deployed by the operator.
```
pohly, line 273: Additional explanations in round brackets look odd. How about this?

```
**WARNING:** This YAML file cannot be used to stop just the operator while 
keeping the PMEM-CSI deployments running. That's because 
something like `kubectl delete -f pmem-csi-operator.yaml` will delete the
`pmem-csi` namespace which then also causes all PMEM-CSI deployments that
might have been created in that namespace to be deleted.
```

avalluri: Agree, and changed as suggested. Thanks.

pohly (approved):

PR #860, opened 2021-01-21
Driver validation on Kubernetes v1.20

### LINKED ISSUE #853: Kubernetes 1.20
Let's add testing against Kubernetes 1.20.

avalluri:
FIXES #853


test/test-config.sh
```diff
@@ -163,8 +163,8 @@ fi
 # < 1.19) and because there were random connection failures to node ports during sanity
 # testing (Kubernetes 1.19.0)
 : ${TEST_FEATURE_GATES:=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true\
-$(case ${TEST_KUBERNETES_VERSION} in 1.1[0-5]) ;; *) echo ',EndpointSlice=false';; esac)\
-$(case ${TEST_KUBERNETES_VERSION} in 1.19) echo ',CSIStorageCapacity=true,GenericEphemeralVolume=true,EndpointSliceProxying=false';; esac)\
+$(case ${TEST_KUBERNETES_VERSION} in 1.1[0-5]) ;; 1.20) ;; *) echo ',EndpointSlice=false,EndpointSliceProxying=false';; esac)\
+$(case ${TEST_KUBERNETES_VERSION} in 1.19 | 1.20) echo ',CSIStorageCapacity=true,GenericEphemeralVolume=true,EndpointSliceProxying=false';; esac)\
```
pohly, line 167: This looks odd and/or unnecessarily complex.

The first line is responsible for adding `EndpointSlice=false` on those releases which support it - basically anything >= 1.16. The second line then added feature gates for Kubernetes = 1.19.

Why does `EndpointSliceProxying=false` get added in the first line?

How does 1.20 fit into this? Neither the commend above nor the commit message have that.

deploy/kubernetes-1.20/direct/kustomization.yaml
```diff
@@ -0,0 +1 @@
+resources: [ pmem-csi.yaml ]
```
pohly, line 1: The deploy changes don't belong into the first commit, do they?

test/test-config.sh
avalluri: oops, I broke it in rebasing, `EndpointSliceProxying=false` should be removed from the second line. My intention was separate endpoint features and storage features and disabling limit endpoint features to 1.19. I will fix this.

deploy/kubernetes-1.20/direct/kustomization.yaml
avalluri: > The deploy changes don't belong into the first commit, do they?
> 
> Or at least not according to the commit message ("changes to install Kuberntes v1.20 based test cluster") - it's about installing Kubernetes, not PMEM-CSI.

Okay, I split the commit into 2; manifest generation and cluster changes.

test/setup-kubernetes.sh
```diff
@@ -267,7 +265,17 @@ cat ${kubeadm_config_file}
 # 5.0.9-301.fc30.x86_64).
 kubeadm_args="$kubeadm_args --ignore-preflight-errors=SystemVerification"
 
-kubeadm_args_init="$kubeadm_args_init --config=$kubeadm_config_file"
+# Migrate to newer config version incase if we are using outdated config schema.
+# `kubeadm config migrate` vanishing the 'KubeletConfiguration' while doing this.
+# So, append to post-migarted configuraiton.
```
pohly, line 270: @avalluri Do you want to fix spelling mistakes or should we ignore them?

This here and the commit message both have mistakes.

test/setup-kubernetes.sh
```diff
@@ -267,7 +265,17 @@ cat ${kubeadm_config_file}
 # 5.0.9-301.fc30.x86_64).
 kubeadm_args="$kubeadm_args --ignore-preflight-errors=SystemVerification"
 
-kubeadm_args_init="$kubeadm_args_init --config=$kubeadm_config_file"
+# Migrate to newer config version incase if we are using outdated config schema.
+# `kubeadm config migrate` vanishing the 'KubeletConfiguration' while doing this.
```
pohly, line 269: Isn't that a bug? The config was valid, why should something just disappear?

Presumably this is about the `kubelet.config.k8s.io/v1beta1`, right? That's still the latest version in 1.20, so what is out-dated?

What I am looking for is a clear explanation why this workaround is needed and when we can remove it again. Ideally we should also file an issue as reminder.

test/test-config.sh
```diff
@@ -163,8 +163,8 @@ fi
 # < 1.19) and because there were random connection failures to node ports during sanity
 # testing (Kubernetes 1.19.0)
 : ${TEST_FEATURE_GATES:=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true\
-$(case ${TEST_KUBERNETES_VERSION} in 1.1[0-5]) ;; *) echo ',EndpointSlice=false';; esac)\
-$(case ${TEST_KUBERNETES_VERSION} in 1.19) echo ',CSIStorageCapacity=true,GenericEphemeralVolume=true,EndpointSliceProxying=false';; esac)\
+$(case ${TEST_KUBERNETES_VERSION} in 1.1[0-5]) ;; 1.20) ;; *) echo ',EndpointSlice=false,EndpointSliceProxying=false';; esac)\
+$(case ${TEST_KUBERNETES_VERSION} in 1.19 | 1.20) echo ',CSIStorageCapacity=true,GenericEphemeralVolume=true';; esac)\
```
pohly, line 167: There's still no explanation why 1.20 needs to be configured this way. What's wrong with keeping the endpoint feature disabled?

The first line can be shortened to:
```
$(case ${TEST_KUBERNETES_VERSION} in 1.1[6-9]) echo ',EndpointSlice=false,EndpointSliceProxying=false';; esac)\
```

pohly (changes requested):

pohly (changes requested):
Kubernetes 1.20 worked, 1.17 and 1.19 didn't.

test/setup-kubernetes.sh
avalluri: I am fixing them part of fixing <=v1.19 failures

test/setup-kubernetes.sh
avalluri: > Isn't that a bug? The config was valid, why should something just disappear?

Might not be, I am not sure how config migrate is supposed to work with kubelet config or not ,as it is not part of `kubadm.k8s.io` API.

> 
> Presumably this is about the `kubelet.config.k8s.io/v1beta1`, right? That's still the latest version in 1.20, so what is out-dated?

Isn't it v1beta2, according to https://github.com/kubernetes/kubernetes/commit/3a050344d0e6475ec7ead775f3f1db32bc837c83?
> 
> What I am looking for is a clear explanation why this workaround is needed and when we can remove it again. Ideally we should also file an issue as reminder.

I am myself not sure if this is a workaround or not.

test/test-config.sh
avalluri: > There's still no explanation why 1.20 needs to be configured this way. What's wrong with keeping the endpoint feature disabled?

I didn't investigate much, but my observation was, `kube-proxy` was failing to connect to cluster api with some endpoint errors, that was when I tried with reenabling them. And as the issue mentioned in the comment was fixed, so I didn't try much investigating this.

avalluri: To my surprise, the same error message is showing even in v19.1 cluster with the current change, this is what observed while working with v1.20:

```
$ sudo grep feature-gates /etc/kubernetes/manifests/kube-apiserver.yaml 
    - --feature-gates=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true,EndpointSlice=false,EndpointSliceProxying=false,CSIStorageCapacity=true,GenericEphemeralVolume=true
   
$ kubectl logs -n kube-system kube-proxy-kwj6m
I0122 09:20:52.758697       1 node.go:136] Successfully retrieved node IP: 172.17.0.4
I0122 09:20:52.758844       1 server_others.go:111] kube-proxy node IP is an IPv4 address (172.17.0.4), assume IPv4 operation
W0122 09:20:55.569023       1 server_others.go:579] Unknown proxy mode "", assuming iptables proxy
I0122 09:20:55.569301       1 server_others.go:186] Using iptables Proxier.
W0122 09:20:55.569328       1 server_others.go:456] detect-local-mode set to ClusterCIDR, but no cluster CIDR defined
I0122 09:20:55.569335       1 server_others.go:467] detect-local-mode: ClusterCIDR , defaulting to no-op detect-local
I0122 09:20:55.570015       1 server.go:650] Version: v1.19.1
I0122 09:20:55.570906       1 conntrack.go:100] Set sysctl 'net/netfilter/nf_conntrack_max' to 131072
I0122 09:20:55.570957       1 conntrack.go:52] Setting nf_conntrack_max to 131072
I0122 09:20:55.572328       1 conntrack.go:83] Setting conntrack hashsize to 32768
I0122 09:20:55.579243       1 conntrack.go:100] Set sysctl 'net/netfilter/nf_conntrack_tcp_timeout_established' to 86400
I0122 09:20:55.579525       1 conntrack.go:100] Set sysctl 'net/netfilter/nf_conntrack_tcp_timeout_close_wait' to 3600
I0122 09:20:55.579958       1 config.go:315] Starting service config controller
I0122 09:20:55.580084       1 shared_informer.go:240] Waiting for caches to sync for service config
I0122 09:20:55.580343       1 config.go:224] Starting endpoint slice config controller
I0122 09:20:55.580469       1 shared_informer.go:240] Waiting for caches to sync for endpoint slice config
E0122 09:20:55.585394       1 reflector.go:127] k8s.io/client-go/informers/factory.go:134: Failed to watch *v1beta1.EndpointSlice: failed to list *v1beta1.EndpointSlice: endpointslices.discovery.k8s.io is forbidden: User "system:serviceaccount:kube-system:kube-proxy" cannot list resource "endpointslices" in API group "discovery.k8s.io" at the cluster scope
I0122 09:20:55.680448       1 shared_informer.go:247] Caches are synced for service config 
E0122 09:20:56.984422       1 reflector.go:127] k8s.io/client-go/informers/factory.go:134: Failed to watch *v1beta1.EndpointSlice: failed to list *v1beta1.EndpointSlice: endpointslices.discovery.k8s.io is forbidden: User "system:serviceaccount:kube-system:kube-proxy" cannot list resource "endpointslices" in API group "discovery.k8s.io" at the cluster scope
....
```
And coredns and calico is not able to start:
```
$ kubectl  get po -o wide -n kube-system
NAME                                                READY   STATUS              RESTARTS   AGE   IP           NODE                         NOMINATED NODE   READINESS GATES
calico-kube-controllers-744cfdf676-7xxht            0/1     ContainerCreating   0          25m   <none>       pmem-csi-pmem-govm-worker2   <none>           <none>
calico-node-7ftbn                                   0/1     CrashLoopBackOff    7          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
calico-node-bh77f                                   0/1     CrashLoopBackOff    7          21m   172.17.0.4   pmem-csi-pmem-govm-worker2   <none>           <none>
calico-node-qwlx6                                   0/1     Running             9          21m   172.17.0.3   pmem-csi-pmem-govm-worker1   <none>           <none>
calico-node-szq6b                                   0/1     Running             9          21m   172.17.0.5   pmem-csi-pmem-govm-worker3   <none>           <none>
coredns-f9fd979d6-4xcjm                             0/1     ContainerCreating   0          25m   <none>       pmem-csi-pmem-govm-worker1   <none>           <none>
coredns-f9fd979d6-m24q6                             0/1     ContainerCreating   0          25m   <none>       pmem-csi-pmem-govm-worker3   <none>           <none>
etcd-pmem-csi-pmem-govm-master                      1/1     Running             0          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
kube-apiserver-pmem-csi-pmem-govm-master            1/1     Running             0          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
kube-controller-manager-pmem-csi-pmem-govm-master   1/1     Running             2          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
kube-proxy-65vzf                                    1/1     Running             0          21m   172.17.0.3   pmem-csi-pmem-govm-worker1   <none>           <none>
kube-proxy-kwj6m                                    1/1     Running             0          21m   172.17.0.4   pmem-csi-pmem-govm-worker2   <none>           <none>
kube-proxy-ndgg9                                    1/1     Running             0          21m   172.17.0.5   pmem-csi-pmem-govm-worker3   <none>           <none>
kube-proxy-s7tlk                                    1/1     Running             0          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
kube-scheduler-pmem-csi-pmem-govm-master            1/1     Running             2          25m   172.17.0.2   pmem-csi-pmem-govm-master    <none>           <none>
```

test/setup-kubernetes.sh
pohly: > > Isn't that a bug? The config was valid, why should something just disappear?
> 
> Might not be, I am not sure how config migrate is supposed to work with kubelet config or not ,as it is not part of `kubadm.k8s.io` API.

If in doubt, file an upstream bug and clarify it.

> > Presumably this is about the `kubelet.config.k8s.io/v1beta1`, right? That's still the latest version in 1.20, so what is out-dated?
>
> Isn't it v1beta2, according to [kubernetes/kubernetes@3a05034](https://github.com/kubernetes/kubernetes/commit/3a050344d0e6475ec7ead775f3f1db32bc837c83)?

That v1beta2 is for kubeadm. I was talking about `kubelet.config.k8s.io/v1beta1` (= https://pkg.go.dev/k8s.io/kubelet/config) because that is the one that gets the special treatment.

> > What I am looking for is a clear explanation why this workaround is needed and when we can remove it again. Ideally we should also file an issue as reminder.
> 
> I am myself not sure if this is a workaround or not.

What we should do is generate a kubeadm config that works for the `kubeadm init` invocation. We shouldn't need `kubeadm migrate` - that just makes it harder to understand what's going on, in particular when we then restore parts that got lost during it.

Why is that not possible? I haven't tried it and it is not mentioned in the PR, so I don't know.

test/test-config.sh
pohly: I don't follow. Callico doesn't work when the endpoint feature is disabled? Is that always the case or just on 1.20? 

It makes sense to keep it disabled on older Kubernetes (because of the bug) and run with the defaults on Kubernetes 1.20 (because the bug is fixed); I'm just trying to understand whether that is the only reason for this change or there are other reasons.

test/setup-kubernetes.sh
avalluri: > What we should do is generate a kubeadm config that works for the `kubeadm init` invocation

You mean write k8s version specific kubeadm configuration?

test/test-config.sh
avalluri: > I don't follow. Callico doesn't work when the endpoint feature is disabled? Is that always the case or just on 1.20?

It should be calico(or in general CNI plugin) issue, somehow `kube-proxy` it self trying access endpoints even if the feature is disabled, I need to check what was causing this.

pohly: Is `kube-proxy` correctly configured to not use the feature when that happens?

test/setup-kubernetes.sh
avalluri: > We shouldn't need `kubeadm migrate` - that just makes it harder to understand what's going on, in particular when we then restore parts that got lost during it.

You are right, This is causing the issue of misbehaving the kube-proxy, `kubeadm config migrate` only migrates the API configurations which it could validate, that is under APIs `kubeadm.k8s.io` other it just ignores, Because of this the feature-gates we set in the kubeproxy configuration is not passed to kube-proxy and causing the issue. I will revert this.

pohly: Yes, or at least make the creation of that config a bit more flexible with k8s version checks.

I don't know how the configuration for 1.20 has to be different, so it's hard to tell without further information which approach is better.

pohly: I had no problems bringing up 1.20 without this "kubeadm migrate". I removed it with the patch below and "make start" worked.

```
diff --git a/test/setup-kubernetes.sh b/test/setup-kubernetes.sh
index e10dec094..574324258 100755
--- a/test/setup-kubernetes.sh
+++ b/test/setup-kubernetes.sh
@@ -250,6 +250,8 @@ kubernetesVersion: $k8sversion
 cat >${kubeadm_config_file} <<EOF
 $kubeadm_config_init
 ---
+$kubeadm_config_kubelet
+---
 $kubeadm_config_cluster
 ---
 $kubeadm_config_proxy
@@ -265,17 +267,7 @@ cat ${kubeadm_config_file}
 # 5.0.9-301.fc30.x86_64).
 kubeadm_args="$kubeadm_args --ignore-preflight-errors=SystemVerification"
 
-# Migrate to newer config version incase if we are using outdated config schema.
-# `kubeadm config migrate` vanishing the 'KubeletConfiguration' while doing this.
-# So, append to post-migarted configuraiton.
-kubeadm_config_new_file="/tmp/kubeadm-config-new.yaml"
-kubeadm config migrate --old-config $kubeadm_config_file --new-config $kubeadm_config_new_file
-cat >>${kubeadm_config_new_file} <<EOF
----
-$kubeadm_config_kubelet
-EOF
-
-kubeadm_args_init="$kubeadm_args_init --config=$kubeadm_config_new_file"
+kubeadm_args_init="$kubeadm_args_init --config=$kubeadm_config_file"
 sudo kubeadm init $kubeadm_args $kubeadm_args_init || (
     set +e
     # Dump some information that might explain the failure.
```

pohly (approved):
Now it's a clean, minimal solution. Looks good.

avalluri:
> Now it's a clean, minimal solution. Looks good.

I pushed a [new commit](https://github.com/intel/pmem-csi/pull/860/commits/678f2fe762e6fd8a4ae103d48eb684cedf2d3d65) that drops enabling of Beta/GA CSI features, just to avoid unnecessary warnings. Can you pleases check if it's ok.

pohly:
Something is still broken with Kubernetes 1.17 (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-860/5/pipeline):
```
[2021-01-22T17:35:23.610Z] 17:35:21 pmem-csi-govm-master: [kubelet-start] Writing kubelet environment file with flags to file "/var/lib/kubelet/kubeadm-flags.env"
[2021-01-22T17:35:23.610Z] 17:35:21 pmem-csi-govm-master: [kubelet-start] Writing kubelet configuration to file "/var/lib/kubelet/config.yaml"
...
[2021-01-22T17:36:06.239Z] 17:36:04 pmem-csi-govm-master: [kubelet-check] Initial timeout of 40s passed.
[2021-01-22T17:36:06.239Z] 17:36:04 pmem-csi-govm-master: [kubelet-check] It seems like the kubelet isn't running or healthy.
[2021-01-22T17:36:06.239Z] 17:36:04 pmem-csi-govm-master: [kubelet-check] The HTTP call equal to 'curl -sSL http://localhost:10248/healthz' failed with error: Get http://localhost:10248/healthz: dial tcp [::1]:10248: connect: connection refused.
...
[2021-01-22T17:37:21.223Z] 17:37:19 pmem-csi-govm-master: Process: 1133 ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_CONFIG_ARGS $KUBELET_KUBEADM_ARGS $KUBELET_EXTRA_ARGS (code=exited, status=255/EXCEPTWarning: The unit file, source configuration file or drop-ins of kubelet.service changed on disk. Run 'systemctl daemon-reload' to reload units.
[2021-01-22T17:37:21.223Z] 17:37:19 pmem-csi-govm-master: ION)
[2021-01-22T17:37:21.223Z] 17:37:19 pmem-csi-govm-master: 
[2021-01-22T17:37:21.223Z] 17:37:19 pmem-csi-govm-master: Jan 22 17:37:14 pmem-csi-govm-master systemd[1]: kubelet.service: Main process exited, code=exited, status=255/EXCEPTION
[2021-01-22T17:37:21.223Z] 17:37:19 pmem-csi-govm-master: Jan 22 17:37:14 pmem-csi-govm-master systemd[1]: kubelet.service: Failed with result 'exit-code'.
...
[2021-01-22T17:37:21.225Z] 17:37:20 pmem-csi-govm-master: Jan 22 17:34:44 pmem-csi-govm-master kubelet[606]: F0122 17:34:44.169675     606 server.go:198] failed to load Kubelet config file /var/lib/kubelet/config.yaml, error failed to read kubelet config file "/var/lib/kubelet/config.yaml", error: open /var/lib/kubelet/config.yaml: no such file or directory
```

This also happened in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-860/4/pipeline

test/test-config.sh
```diff
@@ -162,9 +162,9 @@ fi
 # EndpointSlice is disabled because of https://github.com/kubernetes/kubernetes/issues/91287 (Kubernetes
 # < 1.19) and because there were random connection failures to node ports during sanity
 # testing (Kubernetes 1.19.0)
-: ${TEST_FEATURE_GATES:=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true\
-$(case ${TEST_KUBERNETES_VERSION} in 1.1[0-5]) ;; *) echo ',EndpointSlice=false';; esac)\
-$(case ${TEST_KUBERNETES_VERSION} in 1.19) echo ',CSIStorageCapacity=true,GenericEphemeralVolume=true,EndpointSliceProxying=false';; esac)\
+: ${TEST_FEATURE_GATES:=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true,\
+$(case ${TEST_KUBERNETES_VERSION} in 1.1[6-9]) echo 'EndpointSlice=false,EndpointSliceProxying=false,';; esac)\
+$(case ${TEST_KUBERNETES_VERSION} in 1.19 | 1.20) echo 'CSIStorageCapacity=true,GenericEphemeralVolume=true,';; esac)\
```
pohly, line 167: There's a subtle change here for 1.17: before, only `EndpointSlice=false` was set. Now `EndpointSliceProxying=false` is also set. But that feature gate was only introduced in 1.18 (https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/), so setting it for 1.17 should trigger an error.

It was disabled before for 1.19 because that's when it became enabled by default.

How about having one `$(case ${TEST_KUBERNETES_VERSION} )` with cases for one or more releases which share the same flags?

pohly:
> I pushed a new commit that drops enabling of Beta/GA CSI features, just to avoid unnecessary warnings.

What are those warnings? Please quote them literally in that commit's message for future reference.

test/test-config.sh
avalluri: Split setting the `EndpointSlice` and `EndpointSliceProxying` in two cases:
```
$(case ${TEST_KUBERNETES_VERSION} in 1.1[6-9]) echo 'EndpointSlice=false,';; esac)\
$(case ${TEST_KUBERNETES_VERSION} in 1.1[8-9]) echo 'EndpointSliceProxying=false,';; esac)\
```
Hopefully, it should fix the 1.17.

avalluri:
> What are those warnings? Please quote them literally in that commit's message for future reference.

Done.

test/test-config.sh
```diff
@@ -162,7 +162,7 @@ fi
 # EndpointSlice is disabled because of https://github.com/kubernetes/kubernetes/issues/91287 (Kubernetes
 # < 1.19) and because there were random connection failures to node ports during sanity
 # testing (Kubernetes 1.19.0)
-: ${TEST_FEATURE_GATES:=CSINodeInfo=true,CSIDriverRegistry=true,CSIBlockVolume=true,CSIInlineVolume=true,\
+: ${TEST_FEATURE_GATES:=\
```
pohly, line 165: Removal of GA features is fine. That's what the warning is about.

I'm less sure about beta features. It is possible that support for them gets removed in a future release. When we explicitly enable one that was removed, we get an unambiguous error when setting up the new Kubernetes release. When we don't enable them, the cluster comes up, but then won't work as expected (i.e. some test failure during E2E testing).

Issue #786, opened 2020-10-14
pmem-csi-intel-com-node pods fails to start in direct mode


krisiasty:
I have OpenShift 4.5 cluster (kubernetes 1.18) with pmem-csi-operator v0.8.0 deployed in default namespace with the following pmem deployment:
```
apiVersion: pmem-csi.intel.com/v1alpha1
kind: Deployment
metadata:
  name: pmem-csi.intel.com
spec:
  deviceMode: direct
  nodeSelector:
    storage: pmem
```
Operator tried to deploy pmem-csi-intel-com-node on each node tagged with storage=pmem label, but the pods fail to start with the following error messages:
```
oc logs pmem-csi-intel-com-node-pwnc4 --all-containers
I1014 20:27:02.303693       1 main.go:110] Version: v1.2.0-0-g6ef000ae
I1014 20:27:02.303763       1 main.go:120] Attempting to open a gRPC connection with: "/csi/csi.sock"
I1014 20:27:02.303779       1 connection.go:151] Connecting to unix:///csi/csi.sock
W1014 20:27:12.304085       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:27:22.304117       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:27:32.304076       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:27:42.304119       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:27:52.304106       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:28:02.304074       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:28:12.304117       1 connection.go:170] Still connecting to unix:///csi/csi.sock
W1014 20:28:22.304013       1 connection.go:170] Still connecting to unix:///csi/csi.sock
I1014 20:27:50.107676       1 main.go:67] Version: v0.8.0-0-gc3728707
I1014 20:27:50.109417       1 mount_linux.go:163] Detected OS without systemd
failed to run driver: FATAL: /sys mounted read-only, can not operate
I1014 20:27:50.204747       1 pmd-ndctl.go:42] NewPmemDeviceManagerNdctl: sysfs mount options:[ro seclabel nosuid nodev noexec relatime]
```
[...]

PR #867, opened 2021-01-26
deploy: work around read-only /sys

pohly:
Depending on the cluster configuration (seen on OpenShift 4.5,
Kubernetes 1.18.3, cri-o 1.18.3), /sys is mounted read-only although a
read/write mount gets requested.

The theory is that some component does something special for /sys, so
hopefully asking for it to be mounted also under /host-sys and then
bind-mounting inside the container will work around that.

We need /host-sys because the special sysfs that normally can be
mounted directly does not exist inside the container. An attempt to do
that simpler approach failed:

pmem-csi-intel-com-node-n992g/pmem-driver@pmem..ker2: Mounting command: mount
pmem-csi-intel-com-node-n992g/pmem-driver@pmem..ker2: Mounting arguments:  -o rw,seclabel,nosuid,nodev,noexec,relatime sysfs /sys
pmem-csi-intel-com-node-n992g/pmem-driver@pmem..ker2: Output: mount: /sys: special device sysfs does not exist.

Fixes: #786


pohly:
I've verified manually that bind-mounting works when `/sys` is requested as a read-only volume. However, I've not been able to verify whether this fixes the issue on OpenShift.

pohly:
@avalluri okay to merge?

avalluri (approved):

PR #868, opened 2021-01-27
operator: set operator's log verbosity level

avalluri:
Operator running in e2e tests was not showing any logs, and the reason is
that there is no log verbosity level set for the operator deployment.
The default level is 0 which suppress all informative logs.

The generated operator manifest is set to log level 3, and the test
script updates it with the TEST_OPERATOR_LOGLEVEL environment value,
which defaults to 5.

### DIFF
--- a/deploy/bindata_generated.go
+++ b/deploy/bindata_generated.go
@@ -103,7 +103,7 @@ func deployKubernetes117DirectPmemCsiYaml() (*asset, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


pohly:
> The default level is 0 which suppress all informative logs.

Shouldn't the default be higher, as in the rest of the PMEM-CSI? Or do we also have 0 as default there and just always increased it during deployment?

avalluri:
> > The default level is 0 which suppress all informative logs.
> 
> Shouldn't the default be higher, as in the rest of the PMEM-CSI? Or do we also have 0 as default there and just always increased it during deployment?

Driver code using the 'klog' for logging and the driver manifests set the log-level to 3 where as, the operator is using 'logger', which defaults to 0 level and the operator manifest was not setting the right log-level. This PR supposed to fix this.

pohly (approved):

PR #869, opened 2021-01-28
Fixes to operator CSV/catalog and example CR

avalluri:
CSV:
- add missing CRD 'displayName','description' fields.
- Deployment -> PmemCSIDeployment naming changes
- mention NFD usage
example CR:
 - set `storage: pmem` as the default node selector.

### DIFF
--- a/deploy/kustomize/olm-catalog/bases/pmem-csi-operator.clusterserviceversion.yaml
+++ b/deploy/kustomize/olm-catalog/bases/pmem-csi-operator.clusterserviceversion.yaml
@@ -43,8 +43,10 @@ spec:
   maturity: alpha
   customresourcedefinitions:
     owned:
-    - kind: Deployment
-      name: deployments.pmem-csi.intel.com
+    - kind: PmemCSIDeployment
+      name: pmemcsideployments.pmem-csi.intel.com
+      displayName: PmemCSIDeployment
+      description: Represents a PMEM-CSI driver deployment in the cluster.
       version: v1beta1
       resources:
       - kind: StatefulSet
@@ -109,23 +111,27 @@ spec:
     $ ipmctl create -goal PersistentMemoryType=AppDirect
     ```
     **Label the cluster nodes that provide persistent memory device(s)**
+
+    You can skip this step if your cluster has [Node Feature Discovery((NFD)](https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html) installed.
     ```
-    $ kubectl label node <your node> storage=pmem
+    $ kubectl label node <your node> feature.node.kubernetes.io/memory-nv.dax="true"
     ```
-    
+
     ## Deploying PMEM-CSI driver
     
     Once after installing the PMEM-CSI operator, you can deploy the PMEM-CSI driver by
     creating an instance of the PMEM-CSI deployment custom resource as shown below:
     ```
     $ kubectl create -f - <<EOF
-    apiVersion: pmem-csi.intel.com/v1alpha1
-    kind: Deployment
+    apiVersion: pmem-csi.intel.com/v1beta1
+    kind: PmemCSIDeployment
     metadata:
       name: pmem-deployment
     spec:
       pmemPercentage: 50
       deviceMode: lvm
+      nodeSelector:
+        feature.node.kubernetes.io/memory-nv.dax: "true"
     EOF
     ```
     For additional configuration options, examples and more information on using the operator,


deploy/common/pmem-csi.intel.com_v1beta1_pmemcsideployment_cr.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   deviceMode: "lvm"
   nodeSelector:
     # When using Node Feature Discovery (NFD):
-    feature.node.kubernetes.io/memory-nv.dax: "true"
+    # feature.node.kubernetes.io/memory-nv.dax: "true"
     # When using manual node labeling with that label:
-    # storage: pmem
+    storage: pmem
```
pohly, line 11: Why this change?

pohly: We shouldn't encourage users to use "storage" as label. It's a bad choice for a label because it isn't namespaced and may conflict with other usage of labels.

We are kind of stuck with it because we promised seamless updates from previous versions, but for someone who is starting out anew (and people using the operator will be in that camp) it makes sense to pick a better label.

The NFD label can also be set manually in case that NFD is not used.

avalluri: > Why this change?

There is a mismatch between the documentation and the example CR, an the example is not usable out of the box in a cluster where no NFD installed. 

> but for someone who is starting out anew (and people using the operator will be in that camp) it makes sense to pick a better label.
> 
> The NFD label can also be set manually in case that NFD is not used.

Then shall we replace the node label name in the OperatorHub documentation as below:

```
**Label the cluster nodes that provide persistent memory device(s)**

$ kubectl label node <your node> feature.node.kubernetes.io/memory-nv.dax="true"

You can skip this step if your cluster is installed with [Node Feature Discovery(NFD)](https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html)

```

pohly: Which documentation is "the documentation"? https://intel.github.io/pmem-csi/latest/docs/install.html#installation-and-setup ?

It introduces the "storage=pmem" label as a "historic" choice. That some other label like the one from NFD is less likely to conflict is not called out explicitly, but could be added.

> Then shall we replace the node label name in the OperatorHub documentation as below:
[...]

Yes, I think that is better.

But how much documentation do we really want to store on OperatorHub? Do we expect that people will be able to use the operator solely based on information there? I'm worried that we'll end up maintaining our documentation in two places.

pohly: "You can skip this step if your cluster is installed with" -> "You can skip this step if your cluster has NFD installed"

avalluri: > But how much documentation do we really want to store on OperatorHub?

Very minimal. It just has software/hardware requirements and nodes setup as part of pre-requisites.

avalluri: > "You can skip this step if your cluster is installed with" -> "You can skip this step if your cluster has NFD installed"

Done.

deploy/kustomize/olm-catalog/bases/pmem-csi-operator.clusterserviceversion.yaml
```diff
@@ -109,23 +111,27 @@ spec:
...
     ```
     **Label the cluster nodes that provide persistent memory device(s)**
+
+    You can skip this step if your cluster has [Node Feature Discovery((NFD)](https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html) installed.
     ```
-    $ kubectl label node <your node> storage=pmem
+    $ kubectl label node <your node> feature.node.kubernetes.io/memory-nv.dax="true"
     ```
-    
+
```
pohly, line 119: We discussed that the setup instructions on operator hub should not repeat content from our main documentation site and instead guide users toward that.

Do you intend to implement that in this PR or separately?

avalluri: > Do you intend to implement that in this PR or separately?

I will prepare a separate PR.

pohly: I'm not sure why we need to merge this PR, but okay.

I opened an issue so that we don't forget: https://github.com/intel/pmem-csi/issues/891

PR #870, opened 2021-01-28
build: replace forked klogr with upstream klog v2.5.0

pohly:
Call depth handling was sorted out upstream, therefore we no longer
need the local fork.


avalluri (approved):
Looks good to me.

Issue #852, opened 2021-01-15
flaky operator test


pohly:
For `go test ./pkg/operator/controller/deployment` I am seeing random errors.
- concurrency issues
- wrong events

The wrong events (so far) only occurred during the E2E version of that test (`make test_e2e TEST_E2E_FOCUS=direct-testing.*gotests.*controller.deployment`):

```
              --- FAIL: TestDeploymentController/Kubernetes_1.17/updating/logFormat_in_default_deployment (2.04s)
                  testinglogger.go:172: INFO reconcile: new instance defaultDriverImage="fake-driver-image"
                  testinglogger.go:172: INFO reconcile: reconcile starting deployment="pmem-csi-with-defaults"
                  testinglogger.go:172: INFO reconcile/reconcile: start deployment="pmem-csi-with-defaults" phase=
                  testinglogger.go:172: INFO reconcile/getSubObject: get deployment="pmem-csi-with-defaults" object="pmem-csi-with-defaults <*v1beta1.CSIDriver>"
                  testinglogger.go:172: INFO reconcile/getSubObject: not found deployment="pmem-csi-with-defaults" pmem-csi-with-defaults <*v1beta1.CSIDriver>=<nil>
                  testinglogger.go:172: INFO reconcile/objectPatch/apply: create deployment="pmem-csi-with-defaults" pmem-csi-with-defaults <*v1beta1.CSIDriver>=<nil>
                  testinglogger.go:172: INFO reconcile/getSubObject: get deployment="pmem-csi-with-defaults" object="test-namespace/pmem-csi-with-defaults-csi-provisioner-role-cfg <*v1.RoleBinding>"
                  testinglogger.go:172: INFO reconcile/getSubObject: not found deployment="pmem-csi-with-defaults" test-namespace/pmem-csi-with-defaults-csi-provisioner-role-cfg <*v1.RoleBinding>=<nil>
                  testinglogger.go:172: INFO reconcile/objectPatch/apply: create deployment="pmem-csi-with-defaults" test-namespace/pmem-csi-with-defaults-csi-provisioner-role-cfg <*v1.RoleBinding>=<nil>
                  testinglogger.go:172: INFO reconcile/getSubObject: get deployment="pmem-csi-with-defaults" object="test-namespace/pmem-csi-with-defaults-node <*v1.DaemonSet>"
[...]

PR #872, opened 2021-02-01
Operator race condition

pohly:
Fixes: https://github.com/intel/pmem-csi/issues/852

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -67,6 +67,35 @@ var currentObjects = []client.Object{
 	&admissionregistrationv1beta1.MutatingWebhookConfiguration{TypeMeta: typeMeta(admissionregistrationv1beta1.SchemeGroupVersion, "MutatingWebhookConfiguration")},
 }
 
+func cloneObject(from client.Object) (client.Object, error) {
+	switch t := from.(type) {
+	case *rbacv1.ClusterRole:
+		return t.DeepCopyObject().(*rbacv1.ClusterRole), nil
+	case *rbacv1.ClusterRoleBinding:
+		return t.DeepCopyObject().(*rbacv1.ClusterRoleBinding), nil
+	case *storagev1beta1.CSIDriver:
+		return t.DeepCopyObject().(*storagev1beta1.CSIDriver), nil
+	case *appsv1.DaemonSet:
+		return t.DeepCopyObject().(*appsv1.DaemonSet), nil
+	case *rbacv1.Role:
+		return t.DeepCopyObject().(*rbacv1.Role), nil
+	case *rbacv1.RoleBinding:
+		return t.DeepCopyObject().(*rbacv1.RoleBinding), nil
+	case *corev1.Secret:
+		return t.DeepCopyObject().(*corev1.Secret), nil
+	case *corev1.Service:
+		return t.DeepCopyObject().(*corev1.Service), nil
+	case *corev1.ServiceAccount:
+		return t.DeepCopyObject().(*corev1.ServiceAccount), nil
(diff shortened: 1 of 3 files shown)


pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -189,6 +218,15 @@ func (op *objectPatch) apply(ctx context.Context, c client.Client) error {
...
 
+	// Patch() will modify op.obj, which is an object that was
+	// generated from our PmemCSIDeployment object and shares some
+	// data structure with it. We don't want those to be modified,
+	// so here we have to do a deep copy first.
+	copy, err := cloneObject(op.obj)
+	if err != nil {
+		return fmt.Errorf("internal error: %v", err)
+	}
+	op.obj = copy
```
pohly, line 229: I'm not happy with the whole `objectPatch` struct here. It's very hard to understand what it does and now `apply` even has side effects on the content of the `objectPatch` struct itself.

It's only used in one place (`redeploy`) and I am tempted to roll it back into that.

@avalluri what do you think?

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -67,6 +67,35 @@ var currentObjects = []client.Object{
...
+	case *corev1.ServiceAccount:
+		return t.DeepCopyObject().(*corev1.ServiceAccount), nil
+	case *appsv1.StatefulSet:
+		return t.DeepCopyObject().(*appsv1.StatefulSet), nil
+	case *admissionregistrationv1beta1.MutatingWebhookConfiguration:
+		return t.DeepCopyObject().(*admissionregistrationv1beta1.MutatingWebhookConfiguration), nil
+	default:
+		return nil, fmt.Errorf("cannot clone client.Object of type %T", from)
+	}
+}
```
pohly, line 97: It would be nicer to avoid this. But `client.Object.DeepCopyObject` returns a `runtime.Object` which cannot be used for `client.Client.Patch` because it is not a `client.Object`.

pkg/pmem-csi-operator/controller/deployment/controller_driver.go
```diff
@@ -289,46 +241,85 @@ type redeployObject struct {
...
+				return nil, fmt.Errorf("internal error: %v", err)
+			}
+			l.V(3).Info("update")
+			if err := r.client.Patch(ctx, copy, patch); err != nil {
+				return nil, fmt.Errorf("patch object: %v", err)
+			}
+		}
+	} else {
+		// For unknown reason client.Create() clearing off the
+		// GVK on obj, so restore it manually.
```
pohly, line 313: I copied this workaround without checking whether it is still needed.

I now removed it and at least `go test ./pkg/pmem-csi-operator/controller/deployment/` still passes.

@avalluri : do you remember further details about this problem? Can you check whether we still need it?

avalluri: > @avalluri : do you remember further details about this problem? Can you check whether we still need it?

I do not remember what was the issue. But seems like `client.Create()` is not clearing the GVK data, so we might not need this workaround anymore.

avalluri (approved):

pohly:
I'm merging without the "operator: remove client.Create workaround" commit because that caused E2E tests to fail for the operator (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-872/3/pipeline). The earlier commits were tested fine except for one unrelated test flake (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-872/2/pipeline).

I tried to reproduce the E2E test failures locally, but so far it just worked for me. Needs further investigation...

PR #874, opened 2021-02-01
test: fix version skew testing on Kubernetes 1.20

pohly:
Skew testing broke when introducing 1.20 because older PMEM-CSI
releases didn't have a deployment directory for it. A symlink to 1.19
works because that deployment is okay also for 1.20.


pohly:
I've verified locally that this worked for one case. I'll also push to the upstream repo to get the branch tested.

avalluri (approved):

PR #875, opened 2021-02-01
test/e2e/operator: use localhost with closed port for dummy image

okartau:
Use of dummy image name only creates many real requests to docker repo
server. The intent here is to avoid connecting real server if
we know we will fail, and also hopefully we can fail quicker.

### DIFF
--- a/test/e2e/operator/deployment_api.go
+++ b/test/e2e/operator/deployment_api.go
@@ -46,9 +46,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// We use intentionally use this non-existing driver image
+// We intentionally use this non-existing container registry for driver image
 // because these tests do not actually need a running driver.
-const dummyImage = "unexisting/pmem-csi-driver"
+const dummyImage = "127.0.0.1:270/unexisting/pmem-csi-driver"
 
 func getDeployment(name string) api.PmemCSIDeployment {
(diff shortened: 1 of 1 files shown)


test/e2e/operator/deployment_api.go
```diff
@@ -46,9 +46,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// We use intentionally use this non-existing driver image
+// We use intentionally use this non-existing service point for driver image
```
pohly, line 49: "service point" - "container registry" probably is a better name.

okartau:
Now the error looks like this:
```
failed to pull and unpack image "127.0.0.1:270/unexisting/pmem-csi-driver:latest": \
failed to resolve reference "127.0.0.1:270/unexisting/pmem-csi-driver:latest": \
failed to do request: Head https://127.0.0.1:270/v2/unexisting/pmem-csi-driver/manifests/latest: \
dial tcp 127.0.0.1:270: connect: connection refused
```
at least the failure is quick and does not connect to remote server

test/e2e/operator/deployment_api.go
okartau: changed this and deleted duplicated "use"

pohly (approved):

Issue #840, opened 2020-12-18
test flake: getting pod output


pohly:
This is from a new test in the structured logging PR (PR #837):

https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-825/9/tests

```
/mnt/workspace/pmem-csi_PR-825/test/e2e/operator/deployment_api.go:306
Dec 18 09:46:30.815: get pod logs for controller-0
Unexpected error:
<*errors.StatusError | 0xc000c485a0>: {
ErrStatus: {
TypeMeta: {Kind: "", APIVersion: ""},
ListMeta: {
SelfLink: "",
ResourceVersion: "",
Continue: "",
RemainingItemCount: nil,
},
Status: "Failure",
Message: "the server rejected our request for an unknown reason (get pods test-deployment-driver-image-controller-0)",
Reason: "BadRequest",
Details: {
Name: "test-deployment-driver-image-controller-0",
Group: "",
Kind: "pods",
UID: "",
Causes: [
{
Type: "UnexpectedServerResponse",
Message: "unknown",
Field: "",
},
],
RetryAfterSeconds: 0,
},
Code: 400,
},
}
the server rejected our request for an unknown reason (get pods test-deployment-driver-image-controller-0)
occurred
/mnt/workspace/pmem-csi_PR-825/test/e2e/operator/deployment_api.go:318
```

PR #876, opened 2021-02-03
test: retry when log output is not available yet

pohly:
"validateDriver" does not verify that the pods are created, so it was
possible to attempt log retrieval before the pod had actually
started. Retrying for all errors fixes that.

Fixes: https://github.com/intel/pmem-csi/issues/840

### DIFF
--- a/test/e2e/pod/logs.go
+++ b/test/e2e/pod/logs.go
@@ -1,0 +1,31 @@
+/*
+Copyright 2021 Intel Corporation.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package pod
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


avalluri (approved):

test/e2e/pod/logs.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Retrieves a container log, with retry in case of errors.
+// It is okay if the pod does not exist yet.
+func Logs(ctx context.Context, client kubernetes.Interface, namespace, pod, container string) (output string, err error) {
+	for {
+		output, err = e2epod.GetPodLogs(client, namespace, pod, container)
+		if err == nil {
+			return
+		}
+		if ctx.Err() == nil {
```
avalluri, line 26: @pohly I missed it while reviewing, isn't it supposed to be `ctx.Err() != nil`?

pohly: Yes. Will fix.

PR #880, opened 2021-02-04
test: fix retry after error

pohly:
Because of == instead of != no retries were attempted.


pohly:
@avalluri Please merge directly.

PR #881, opened 2021-02-04
Overwrite with zero instead of random when shredding

okartau:
shred default behavior is to write random data, but getting random data is slower than writing zeroes. In my small test with 8G file, I saw write with zeroes being almost twice faster than write with random. shred has option to add zero-fill at the end, but with setting number of random writes to zero, there will be only zero-write pass. 
I want to see how this affects overall test duration.
Also, we may discuss is fill with zeroes actually enough for real operation of the plugin, instead of even more secure fill-with-random style.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -629,7 +629,7 @@ the following custom parameters in a storage class:
 
 |key|meaning|optional|values|
 |---|-------|--------|-------------|
-|`eraseAfter`|Clear all data after use and before<br> deleting the volume|Yes|`true` (default),<br> `false`|
+|`eraseAfter`|Clear all data by overwriting with zeroes after use and before<br> deleting the volume|Yes|`true` (default),<br> `false`|
 |`kataContainers`|Prepare volume for use with DAX in Kata Containers.|Yes|`false/0/f/FALSE` (default),<br> `true/1/t/TRUE`|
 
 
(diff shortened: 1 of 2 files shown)


okartau:
This is not very good measurement as different PRs may bring various changes to timing, but timing list of recent jobs:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/

indicate that job for this PR ran in almost 10% shorter time (5h5min) than other PRs close to it (5h30min):

okartau:
I propose we consider changing our erasing methods as in this PR.
The stronger "overwrite with random" method (based on what I have been seen documented) is meant to extra-assure data is even harder to recover from hard drives compared to data overwritten with static pattern. But this claim seems has roots with technology where the media (like hard disk surface) can be post-analyzed with special tools out-of-box. In our PMEM case, data is memory value (no surface), so does it matter at all do we overwrite it with zero or random? I think not, and thus we should select the faster method. Using random data is slower and consumes system random data pool. Proposal: lets shred with zeroes.

avalluri:
> Proposal: lets shred with zeroes.

I agree with this.

pohly (changes requested):
Please update also the documentation.

pohly (changes requested):
Please remove the "experimental" from the commit title.

pohly (approved):

PR #893, opened 2021-02-09
operator: load TLS CA bundle also when handling sub-object changes

pohly:
When handling change or delete events for sub-objects, in particular
MutatingWebhookConfig, a pmemCSIDeployment was created and used
without loading the TLS secret, resulting in a MutatingWebhookConfig
which had an empty CA bundle. That in turn led to:

   failed calling webhook "pod-hook.pmem-csi.intel.com": Post "https://second-pmem-csi-intel-com-scheduler.default.svc:443/pod/mutate?timeout=30s": x509: certificate signed by unknown authority

### DIFF
--- a/pkg/deployments/load.go
+++ b/pkg/deployments/load.go
@@ -8,6 +8,7 @@ package deployments
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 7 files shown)


pohly:
I think this slipped through testing so far because driver deployment validation doesn't check the CA in the webhook configuration. Probably should be addressed.

I noticed it when running `make test_e2e TEST_E2E_FOCUS=lvm-production.*late.*binding.*stress` which happened to hit this issue (by chance?).

pohly:
> I think this slipped through testing so far because driver deployment validation doesn't check the CA in the webhook configuration. Probably should be addressed.

The second commit adds the test. I checked manually that it fails without the other commit.

avalluri (approved):

Issue #878, opened 2021-02-04
test flake: volume with immediate binding not created


pohly:
Seen for
```
Testing / 1.17 / [Testpattern: Dynamic PV (default fs)] volumeIO should write files of various sizes, verify size, validate content [Slow] – fedora-1_17.direct-testing.E2E
```
in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/fix-skew-testing-1.20/5/tests

```
pmem-csi-intel-com-controller-0/pmem-driver@pmem..ker3: I0203 23:55:55.246104       1 rescheduler.go:152] ShouldReschedulePVC: no need to reschedule, no selected node pvc="volumeio-9963/direct-testingg785b"
pmem-csi-intel-com-controller-0/pmem-driver@pmem..ker3: I0203 23:55:55.249054       1 rescheduler.go:152] ShouldReschedulePVC: no need to reschedule, no selected node pvc="volumeio-9963/direct-testingg785b"
pmem-csi-intel-com-node-5vs74/external-provisioner@pmem..ker1: I0203 23:55:55.244811       1 controller.go:1265] should provision: checking node for PVC volumeio-9963/direct-testingg785b with resource version 76099
pmem-csi-intel-com-node-5vs74/external-provisioner@pmem..ker1: I0203 23:55:55.244885       1 controller.go:1267] should provision: done checking node for PVC volumeio-9963/direct-testingg785b with resource version 76099: provision false, err storageclass.storage.k8s.io "pmem-pmem-csi-sc-volumeio-9963" not found
pmem-csi-intel-com-node-5vs74/external-provisioner@pmem..ker1: I0203 23:55:55.244898       1 controller.go:1231] trying to become an owner of PVC volumeio-9963/direct-testingg785b in advance failed, will try again during provisioning: storageclass.storage.k8s.io "pmem-pmem-csi-sc-volumeio-9963" not found
pmem-csi-intel-com-node-5vs74/external-provisioner@pmem..ker1: W0203 23:55:55.244986       1 controller.go:961] Retrying syncing claim "933350b5-579d-487d-82da-b3adb9ffe3c6", failure 0
[...]

PR #898, opened 2021-02-10
Immediate binding

pohly:
This is an attempt to trigger issue #878. Even if the test just passes, having it is worthwhile to prevent regressions.


pohly:
An actual run of the new test is here: https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/immediate-binding/1/tests

As intended, stress tests only ran for LVM.

PR #899, opened 2021-02-13
validating scheduler extender security

avalluri:
E2E tests for validating the security of scheduler extender. It discovered(and thus fixed) in tls config used by scheduler extender with which it allows to connect clients with invalid certificates.

Somehow API server(?) is trying to connect the webhook with missing certificate:
```
Feb 13 22:26:37.751: INFO: Still waiting for PMEM-CSI driver, last error: expected 'Invalid' as error for pod creation, got: Internal error occurred: failed calling webhook "pod-hook.pmem-csi.intel.com": Post "https://pmem-csi-intel-com-scheduler.default.svc:443/pod/mutate?timeout=10s": remote error: tls: bad certificate
pmem-csi-intel-com-controller-0/pmem-driver@pmem..ker3: 2021/02/13 20:26:38 http: TLS handshake error from 192.168.238.0:11057: tls: client didn't provide a certificate
```

So till we fix this, as a workaround , used `VerifyClientCertIfGiven` client authentication method to allow clients to connect the mutation webhook server without providing any client certificate.

### DIFF
--- a/pkg/pmem-grpc/grpc.go
+++ b/pkg/pmem-grpc/grpc.go
@@ -142,12 +142,21 @@ func serverConfig(certPool *x509.CertPool, peerCert *tls.Certificate, peerName s
 				}
 			}
 
-			config := &tls.Config{
+			clientAuth := tls.RequireAndVerifyClientCert
+			if peerName == "" {
+				// NOTE: allow the clients(APIServer) with no certificate
+				// to connect to mutation webhook.
+				// In practice this authentication type should not be used.
+				clientAuth = tls.VerifyClientCertIfGiven
+			}
+
+			return &tls.Config{
 				MinVersion:    tls.VersionTLS12,
 				Renegotiation: tls.RenegotiateNever,
 				Certificates:  []tls.Certificate{*peerCert},
 				ClientCAs:     certPool,
 				CipherSuites:  ciphers,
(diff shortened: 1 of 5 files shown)


pohly:
What is the problem that this PR is trying to solve?

The webhook exposes no confidential data. We therefore don't need to validate the client.

It's nicer if we do, but from the description of the PR it is not clear to me whether it currently works ("So till we fix this, as a workaround"...).

avalluri:
> What is the problem that this PR is trying to solve?

Validating the client tls certificate by the scheduler extender service. i.e., to ensure the SE rejects connections with invalid certificate.

> The webhook exposes no confidential data. We therefore don't need to validate the client.
> 
As a side effect, this also also allows accessing the SE with no client certificate, I assume should SE reject such connections.

> It's nicer if we do, but from the description of the PR it is not clear to me whether it currently works ("So till we fix this, as a workaround"...).

Current behavior with this PR is:
 - Both webhook and SE allows clients with no certificate(Which should be fixed for at least for SE) 
 - If client provide certificate, validate and reject if it's a not valid certificate.

avalluri:
This PR is required if we consider the data(Node(s) details where PMEM installed) exposed by SE is not sensitive.

pohly:
Then let's close it.

PR #900, opened 2021-02-16
Logs as artifacts

okartau:


Jenkinsfile
```diff
@@ -453,13 +453,13 @@ void TestInVM(worker, distro, distroVersion, kubernetesVersion, skipIfPR) {
...
-        (`qemu-system-x86_64: error: failed to set MSR 0x48b to
-        0x1582e00000000`,
-        https://www.mail-archive.com/qemu-devel@nongnu.org/msg665051.html),
+        The default QEMU cpu enables nested virtualization with "-cpu host".
+        However, that fails on some Azure machines:
+        `qemu-system-x86_64: error: failed to set MSR 0x48b to 0x1582e00000000`,
+        https://www.mail-archive.com/qemu-devel@nongnu.org/msg665051.html,
         so for now we disable VMX with -vmx.
 
+        XXX FIX BEFORE MERGE: Do we still need below 2 lines?
```
pohly, line 462: I think you can also remove the "TODO". We are testing in parallel now.

okartau:
In build#2,  with top-loop SLEEP value to set to infinite, kill of exit handler suddenly fails:
```
22:29:20  + kill 17313
22:29:20  /mnt/workspace/pmem-csi_PR-900@tmp/durable-2149b46b/script.sh: 1: kill: No such process
```

somehow, sub-process with sleep loop/infinite, when top was never called, behaves differently than sleep with 60s when top gets called? why is that...
To get around that, we could: 
1) use kill || true, 
2) skip creating the sleep-loop altogether based on top-enabled flag
3) what about we stop recording log PIDs and using exit handler, based on assumption that a VM is created to serve one build only?

Jenkinsfile
okartau: done

okartau:
seems indeed, new logs are now overwriting previous logs, so I will try to make them unique by adding build number in name.
Also, in failure case, we extract journal from workers which can be long, so it makes sense to direct that to separate file as well.
Also, I will add "tee" to extract important strings from stream.

Issue #858, opened 2021-01-20
test flake: LVM namespace disappears


pohly:
From https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/distributed-provisioning/10/tests:

```
Testing / 1.18 / [Testpattern: DynamicPV Filesystem {"FSType":"","Parameters":null}] skew everything [Slow] – fedora-1_18.direct-testing.E2E5m 49s
Stacktrace
/mnt/workspace/mem-csi_distributed-provisioning/test/e2e/deploy/deploy.go:882
Jan 20 17:42:17.943: same volumes before and after the test
Expected
    <map[string][]string | len:3>: {
        "Namespaces on Node 1": [""],
        "Namespaces on Node 2": [
            "mode:fsdax,size:32212254720,uuid:bd22b2b5-5b18-11eb-978a-1ef67ef1e261,",
        ],
        "Namespaces on Node 3": [
            "mode:fsdax,size:32212254720,uuid:bd2406e7-5b18-11eb-8e3a-3606fa9799a4,",
        ],
    }
to equal
    <map[string][]string | len:3>: {
        "Namespaces on Node 2": [
            "mode:fsdax,size:32212254720,uuid:bd22b2b5-5b18-11eb-978a-1ef67ef1e261,",
        ],
        "Namespaces on Node 3": [
            "mode:fsdax,size:32212254720,uuid:bd2406e7-5b18-11eb-8e3a-3606fa9799a4,",
        ],
        "Namespaces on Node 1": [
            "mode:fsdax,size:32212254720,uuid:bd36996a-5b18-11eb-9266-16f5378db05f,",
        ],
    }
/mnt/workspace/mem-csi_distributed-provisioning/test/e2e/deploy/volumeleaks.go:63
```

My interpretation is that the `mode:fsdax,size:32212254720` namespace is the one for LVM. That it really got removed during the test on node 1 seems very unlikely - there's simply node code which does that. I suspect that the ssh command which lists the namespaces produced no output.

pohly:
Not seen anymore.

PR #904, opened 2021-02-22
Volume leaks test: better check for ssh error and retry

okartau:
ssh error was checked, but only to detect end of nodes.
It is also possible ssh fails for some other reason,
so try to detect and log this. Retry once.

### DIFF
--- a/test/e2e/deploy/volumeleaks.go
+++ b/test/e2e/deploy/volumeleaks.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	api "github.com/intel/pmem-csi/pkg/apis/pmemcsi/v1beta1"
 
@@ -44,15 +45,29 @@ func GetHostVolumes(d *Deployment) map[string][]string {
 	// script appears to be "no such file".
 	for worker := 1; ; worker++ {
 		sshcmd := fmt.Sprintf("%s/_work/%s/ssh.%d", os.Getenv("REPO_ROOT"), os.Getenv("CLUSTER"), worker)
-		ssh := exec.Command(sshcmd, cmd)
-		// Intentional Output instead of CombinedOutput to dismiss warnings from stderr.
-		// lvs may emit lvmetad-related WARNING msg which can't be silenced using -q option.
-		out, err := ssh.Output()
-		if err != nil && os.IsNotExist(err) {
+		if _, err := os.Stat(sshcmd); err == nil {
+			i := 0
+			for {
+				ssh := exec.Command(sshcmd, cmd)
+				// Intentional Output instead of CombinedOutput to dismiss warnings from stderr.
+				// lvs may emit lvmetad-related WARNING msg which can't be silenced using -q option.
+				out, err := ssh.Output()
+				if err != nil {
+					if i >= 3 {
+						ginkgo.Fail(fmt.Sprintf("repeated ssh attempts to node %d failed: %v", worker, err.Error()))
+					}
+					ginkgo.By(fmt.Sprintf("ssh attempt No.%d to node %d failed: %v", i, worker, err.Error()))
+					time.Sleep(10 * time.Second)
+				} else {
+					buf := fmt.Sprintf("%s on Node %d", hdr, worker)
+					result[buf] = strings.Split(strings.TrimSpace(string(out)), "\n")
+					break
+				}
+			}
+		} else {
+			// ssh wrapper does not exist: all nodes handled.
 			break
 		}
-		buf := fmt.Sprintf("%s on Node %d", hdr, worker)
-		result[buf] = strings.Split(strings.TrimSpace(string(out)), "\n")
 	}
 	return result
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -461,7 +461,7 @@ void TestInVM(worker, distro, distroVersion, kubernetesVersion, skipIfPR) {
         */
         sh " \
            loggers=; \
-           atexit () { set -x; kill \$loggers ||true; killall sleep; }; \
+           atexit () { set -x; kill \$loggers ||true; killall sleep ||true; }; \
            trap atexit EXIT; \
            mkdir -p build/reports && \
            if ${env.LOGGING_JOURNALCTL}; then sudo journalctl -f; fi & \


okartau:
This addresses situation in #858.
At least failing return value check was covering only part of cases thanks to ANDed condition, we can do better there,
Also, try to log error cause if there is other than "no such file" error.
Retry once, as there is not much else we can do in this situation.

test/e2e/deploy/volumeleaks.go
```diff
@@ -48,8 +48,16 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
 		// lvs may emit lvmetad-related WARNING msg which can't be silenced using -q option.
 		out, err := ssh.Output()
-		if err != nil && os.IsNotExist(err) {
-			break
+		if err != nil {
+			if os.IsNotExist(err) {
+				// ssh wrapper does not exist: all node handled.
+				break
+			}
+			ginkgo.By(fmt.Sprintf("ssh on node %d returned error '%s', retry once", worker, err.Error()))
```
pohly, line 56: Why once? Retrying for a certain time may be more predictable.

test/e2e/deploy/volumeleaks.go
```diff
@@ -48,8 +48,16 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
+			out, err = ssh.Output()
+			if err != nil {
+				ginkgo.By(fmt.Sprintf("ssh on node %d repeatedly returned error '%s', give up on that node", worker, err.Error()))
+			}
```
pohly, line 60: This message is not going to be visible easily when the assertion fails later on. Wouldn't it be better to fail the test here?

okartau: yes, better idea is to fail there as job will fail anyway

test/e2e/deploy/volumeleaks.go
okartau: The logic behind "once" is that, we likely get hit by a random problem, and these usually don't occur multiple in row, so re-trying more than once will not change the result. The problem likely is some resource shortage, like IO slowness, connectivity issue, timeout etc.
Although, I think it would be good idea to add some delay before retry, because if the bad condition gets over in time, we should give some time.
Let's also note that it is very unlikely to hit  the path we try to test here in regular CI run. In CI we can only verify we dont break it more.

okartau:
test run No.2 got hit by another flake #888, for which there is already another PR.

test/e2e/deploy/volumeleaks.go
pohly: If it's simply about probability, then retrying twice is better than retrying once and does not hurt if the assumption is that we never have to try more than once...

test/e2e/deploy/volumeleaks.go
```diff
@@ -48,8 +49,17 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
+		if err != nil {
+			if os.IsNotExist(err) {
+				// ssh wrapper does not exist: all nodes handled.
+				break
+			}
+			ginkgo.By(fmt.Sprintf("ssh on node %d returned error '%s', wait 10s and retry once", worker, err.Error()))
+			time.Sleep(10 * time.Second)
+			out, err = ssh.Output()
+			if err != nil {
+				ginkgo.Fail(fmt.Sprintf("repeated ssh to node %d returned error '%s'", worker, err.Error()))
```
pohly, line 61: The error may contain single quotes. Let's use the usual prefix style for wrapping an error:
```
fmt.Sprintf("repeated ssh to node %d failed again: %v", worker, err.Error())
```

test/e2e/deploy/volumeleaks.go
```diff
@@ -48,8 +49,21 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
 		out, err := ssh.Output()
-		if err != nil && os.IsNotExist(err) {
-			break
+		if err != nil {
+			if os.IsNotExist(err) {
+				// ssh wrapper does not exist: all nodes handled.
+				break
+			}
+			ginkgo.By(fmt.Sprintf("ssh to node %d failed: %v, retry", worker, err.Error()))
+			for retry := 1; retry < 4; retry++ {
```
pohly, line 58: Wouldn't it be better to merge the first attempt into the loop body?

When ssh gets repeated, why does it not need to check for `os.IsNotExist`?

okartau: > When ssh gets repeated, why does it not need to check for `os.IsNotExist`?

IsNotExist is there to detect end of nodes. I assume that set of ssh handlers does not change during running this. We check for IsNotExist above repeat loop, and that means (if assumption above is true) that also in repeated attempts, the ssh handler should be there, therefore we dont check for existence error.

okartau: > Wouldn't it be better to merge the first attempt into the loop body?

makes sense, will avoid repeated code lines. And that would also make the IfNotExist checking more uniform. I will change.

okartau:
the current change has a fatal bug anyway, it lacks repeat of Exec.Command, only repeating check of old exec result. It's possible to miss it entirely in CI as the probability to hit the code-to-be tested is very low. We should use some sort of error injection to actually run this part. Luckily this is so small change we can (hopefully) see it in reviews. Will change that as well...

okartau:
pushed a new version. I took the check for ssh handler existence out of cmd error checking as it was bit dirty approach to run command to find out that command file did not exist. Lets use proper file existence check for that. Now we can write the repeat loop without need to break out from nested loop.

test/e2e/deploy/volumeleaks.go
```diff
@@ -44,15 +46,28 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
+					time.Sleep(10 * time.Second)
+				} else {
+					buf := fmt.Sprintf("%s on Node %d", hdr, worker)
+					result[buf] = strings.Split(strings.TrimSpace(string(out)), "\n")
+					break
+				}
+			}
+			if retry >= 4 {
+				ginkgo.Fail(fmt.Sprintf("repeated ssh attempts to node %d failed", worker))
+			}
```
pohly, line 66: We get no explanation here why it fails. Having to repeat the comparison against 4 here also isn't nice (what if we reduce it above to 3 and forget here?!).

A better retry loop is:
```
i := 0
for {
    ssh := exec.Command(sshcmd, cmd)
    out, err := ssh.Output()
    if err != nil {
        if i >= 4 {
           gingko.Fail(fmt.Sprintf("repeated ssh attempts to node %d failed again: %v", worker, err))
	}
	
         ginkgo.By(fmt.Sprintf("ssh attempt No.%d to node %d failed: %v", retry, worker, err.Error()))
         time.Sleep(10 * time.Second)
   } else {
        ...
        break
   }
   i++
}

File: test/e2e/deploy/volumeleaks.go [hunk 6]
```diff
@@ -44,15 +45,29 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
+		if _, err := os.Stat(sshcmd); err == nil {
+			i := 0
+			for {
+				ssh := exec.Command(sshcmd, cmd)
+				// Intentional Output instead of CombinedOutput to dismiss warnings from stderr.
+				// lvs may emit lvmetad-related WARNING msg which can't be silenced using -q option.
+				out, err := ssh.Output()
+				if err != nil {
+					if i >= 3 {
+						ginkgo.Fail(fmt.Sprintf("repeated ssh attempts to node %d failed", worker))
```
pohly, line 57: The error is still not part of the failure message.

okartau: showing error when all attempts have failed, seems layering violation for me. We show error each time ssh fails. After 3 attempts failed, the problem is not one of the 3 errors, but aggregate state "all attempts failed". It's not guaranteed errors were same, so which one should we show at the end, last one? (right, we dont store others so thats the only one we have). But all errors were already shown. You want to repeat the last one? But why it is special that it needs to be shown twice while it happened once?

pohly: When I see the test failure, I want to know immediately why ssh didn't work. It's faster if I don't have to go hunting for the error somewhere in the log file. Reporting the error directly in the failure message achieves that.

If I then want to double-check that all previous attempts failed the same way, then I can still consult the log file.

okartau: I added error print now in final msg. Actually, my logic about "aggregated" report and repeated log does not hold water 100%, because in current code, we check for repeat limit before reporting error, thus the last report is not repeated.

test/e2e/deploy/volumeleaks.go
```diff
@@ -44,15 +45,29 @@ func GetHostVolumes(d *Deployment) map[string][]string {
...
+			for {
+				ssh := exec.Command(sshcmd, cmd)
+				// Intentional Output instead of CombinedOutput to dismiss warnings from stderr.
+				// lvs may emit lvmetad-related WARNING msg which can't be silenced using -q option.
+				out, err := ssh.Output()
+				if err != nil {
+					if i >= 3 {
+						ginkgo.Fail(fmt.Sprintf("repeated ssh attempts to node %d failed: %v", worker, err.Error()))
+					}
+					ginkgo.By(fmt.Sprintf("ssh attempt No.%d to node %d failed: %v", i, worker, err.Error()))
```
pohly, line 59: `err.Error()` could be replaced with just `err`, but I'm going to merge it as-is.

pohly (approved):

PR #910, opened 2021-03-08
Testing: step up installed k8s package versions on testers.

### LINKED ISSUE #911: typo in Jenkinsfile blocks multiple sleep cmds from starting
In one of my recent changes in Jenkinsfile I made a logging change where I misspelled one word slightly:
`LOGGING_SAMPLING_DELAY=infinite`
As result, multiple sleep commands using that as argument, fail now:
`sleep: invalid time interval ‘infinite’`

The correct spelling for this argument is `infinity`.
There were some cases later about kill cmd not finding processes to kill, which I then "fixed" by adding `||true`.
In the light of recent findings, those were likely caused by infinite-issue.
I could go reverting that `||true` change, but OTOH I believe it does not harm to avoid break on nonzero result of kill cmd.
I can add fix for "infinite" in my recently opened #910 about testing versions update.

okartau:
Minor versions have advanced significantly, lets stick to latest.
The cmd (as comment) for showing latest versions needed update.

also:
Fixes #911

### DIFF
--- a/test/setup-fedora-govm.sh
+++ b/test/setup-fedora-govm.sh
@@ -52,17 +52,16 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cl
 EOF
 
     # For the sake of reproducibility, use fixed versions.
-    # List generated with:
-    # for v in 1.13 1.14 1.15 1.16 1.17 1.18; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo dnf --showduplicates list kubelet | grep " $v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u -n -k 2.6 | tail -n 1)"; done; done
-    # Note that -n and -k 2.6 are needed to handle minor versions > 9, and -k 2.6 works only if k8s major versions have form X.XX
+    # List of latest packages can be generated on Fedora with:
+    # for v in 1.13 1.14 1.15 1.16 1.17 1.18 1.19 1.20; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo dnf --showduplicates list kubelet | grep " $v" | tail -n 1 | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/')"; done; done
     case ${TEST_KUBERNETES_VERSION} in
         1.13) packages+=" kubelet-1.13.12-0 kubeadm-1.13.12-0 kubectl-1.13.12-0";;
         1.14) packages+=" kubelet-1.14.10-0 kubeadm-1.14.10-0 kubectl-1.14.10-0";;
-        1.15) packages+=" kubelet-1.15.11-0 kubeadm-1.15.11-0 kubectl-1.15.11-0";;
-        1.16) packages+=" kubelet-1.16.9-0 kubeadm-1.16.9-0 kubectl-1.16.9-0";;
-        1.17) packages+=" kubelet-1.17.5-0 kubeadm-1.17.5-0 kubectl-1.17.5-0";;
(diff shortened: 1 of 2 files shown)


okartau:
not sure why the commented cmd with -k 2.6 did not work anymore, output format has changed? Anyway, it works with -k 1.6 now. I also changed long list to $(seq 13 20), and updated minor versions.
Sticks out that we repeat same version for 3 packages always, what about using one variable? Current style allows flexibility but do we need/use it? I dont recall we ever needed different versions yet.

test/setup-fedora-govm.sh
```diff
@@ -52,17 +52,17 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cl
 EOF
 
     # For the sake of reproducibility, use fixed versions.
-    # List generated with:
-    # for v in 1.13 1.14 1.15 1.16 1.17 1.18; do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo dnf --showduplicates list kubelet | grep " $v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u -n -k 2.6 | tail -n 1)"; done; done
-    # Note that -n and -k 2.6 are needed to handle minor versions > 9, and -k 2.6 works only if k8s major versions have form X.XX
+    # List of latest packages can be generated on Fedora with:
+    # for v in $(seq 13 20); do for i in kubelet kubeadm kubectl; do echo "$i-$(sudo dnf --showduplicates list kubelet | grep " 1.$v"  | sed -e 's/.* \([0-9]*\.[0-9]*\.[0-9]*[^ ]*\).*/\1/' | sort -u -n -k 1.6 | tail -n 1)"; done; done
```
pohly, line 56: I find the `$(seq 13 20)` less readable. Previously it was obvious at a glance that these are Kubernetes version numbers, now not so much. It's not even that much shorter.

test/setup-fedora-govm.sh
```diff
@@ -52,17 +52,17 @@ gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cl
...
+    # Note that -n and -k 1.6 are needed to handle minor versions > 9, and -k 1.6 works only if k8s major versions have form X.XX
```
pohly, line 57: What are the fields that we are sorting on?

test/setup-fedora-govm.sh
okartau: If that list does not bother anybody, I can put it back, you are right code does not get much smaller. Speaking about list length BTW, do we still want to keep older versions here, even if we start testing with 1.17 ? If we only want older ones (13..16) for historical reference, older versions can as well remain in git history?

test/setup-fedora-govm.sh
okartau: good question! Now that I look at that again, I see we can (?) leave sorting out, if we trust that dnf keeps showing entries in sorted order already. This is what we get before "sort" now:
kubectl-1.18.0-0
1.18.1-0
1.18.2-0
1.18.3-0
1.18.4-0
1.18.4-1
1.18.5-0
1.18.6-0
1.18.8-0
1.18.9-0
1.18.10-0
1.18.12-0
1.18.13-0
1.18.14-0
1.18.15-0
1.18.16-0
---------------
so if we just take last line, it will be what we need.

The history of that example cmd is, before "sort -n -u -k 2.6" that I introduced, there was sort -u, which did not work for minor versions >9 as it always showed 1.18.9 after 1.8.10. I tried to improve it by telling that we need numeric sort, but that had to be applied only at certain sub-field (position 6) to work correctly. It solved the "9 before 1" issue, and it works as in current PR.
But indeed, seems we can simplify by leaving out sort as dnf shows entries in correct versioning order already.
The line without "sort" part currently produces same result as with "sort" part.

pohly (approved):

PR #915, opened 2021-03-15
Jenkinsfile: archive artifacts before junit step

okartau:
Sometimes, junit step does not find reports and this
causes break of execution and some artifacts may get lost.
Let's archive artifacts before junit step.


okartau:
positive-result testing can't show the effect of this change

pohly (approved):

PR #916, opened 2021-03-16
docs: clarify volume binding modes

pohly:
The comment about `allowedTopologies` was incorrect and referred to an
experiment that didn't work out. What we have to use instead are
volume binding modes and scheduler extensions.


avalluri (approved):
Looks good to me.

PR #923, opened 2021-04-08
Operator deployment changes for supporting version skew tests

avalluri:
In order to upgrade/downgrade the operator:
- the `stop-operator.sh` script must support removing the operator without deleting the CRD and namespace.
- the `start-operator.sh` script must support choosing the specific operator image version.

### DIFF
--- a/test/stop-operator.sh
+++ b/test/stop-operator.sh
@@ -14,6 +14,9 @@ CLUSTER_DIRECTORY="${CLUSTER_DIRECTORY:-${REPO_DIRECTORY}/_work/${CLUSTER}}"
 SSH="${CLUSTER_DIRECTORY}/ssh.0"
 KUBECTL="${SSH} kubectl" # Always use the kubectl installed in the cluster.
 
+keep_crd=false
+keep_namespace=false
+
 function delete_olm_operator() {
   set -x
   BINDIR=${REPO_DIRECTORY}/_work/bin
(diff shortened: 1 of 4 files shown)


test/stop-operator.sh
```diff
@@ -45,32 +48,74 @@ function delete_olm_operator() {
...
+  "-h") Usage ;;
+  *) echo "Ignoring unknown argument: $arg"
+     Usage ;;
+  esac
+done
 
 case $deploy_method in
   yaml)
-    delete_operator ;;
+    delete_operator;;
```
pohly, line 115: nit: above you use a space between `Usage` and `;;`. I don't care either way, but it would be nice to be consistent.

test/test-config.sh
```diff
@@ -43,6 +43,10 @@ fi
 # This is needed for "make push-images".
 : ${TEST_BUILD_PMEM_REGISTRY:=localhost:5000}
 
+# The PMEM-CSI driver/operator container image tag use.
+# This is used by version skew tests to set the driver/operator version.
+: ${TEST_PMEM_IMAGE_TAG:=canary}
```
pohly, line 48: How will this work once we have this test in a release-1.0 branch and test against 0.9? We cannot use "canary" as the image in such a branch.

pohly: The `hack/set-version.sh` might have to be extended to replace the "canary" version also in this line.

test/stop-operator.sh
pohly: So you don't want to make it consistent?

avalluri: Oops, I missed it. This change was not intentional. I fixed it now.

test/test-config.sh
avalluri: > The `hack/set-version.sh` might have to be extended to replace the "canary" version also in this line.

Done, Also removed `canary` from the regex in `test/start-operator.sh`.

avalluri:
Let's handle this in #928

PR #883, opened 2021-02-04
test/deployment: stop generating registry and node secrets/certificates

avalluri:

PR #928, opened 2021-04-12
Version skew tests for operator

### LINKED ISSUE #779: operator: up- and downgrade testing
When submitting a new operator version to the Docker Hub, one of the questions is "Have you tested an update to your Operator when deployed via OLM?".

We should add automated up- and downgrade tests.

avalluri:
Added both upgrade and downgrade operator tests. These are similar to driver version skew tests. We test that the running driver deployments using one version keep runs even after upgrading/downgrading the operator.
    
Currently, these tests are limited to running the operator deployed via YAML files not using the OLM. Because there is no way to keep the CRD while deleting the operator using olm.

FIXES #779 

This PR combines the #883 and #923


test/e2e/versionskew/operator.go
```diff
@@ -0,0 +1,227 @@
...
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = deploy.DescribeForSome("versionskew", func(d *deploy.Deployment) bool {
+	// Run these tests only for bare operator deployments, not for the operator
+	// deployed by OLM. Because there is no way how to ask olm to stop the operator
+	// but keep the CRD.
```
pohly, line 31: Isn't that a problem for people using OLM to manage the operator? How are they supposed to do upgrades and downgrades?

test/e2e/versionskew/operator.go
```diff
@@ -0,0 +1,227 @@
...
+					defer evMutex.Unlock()
+					if _, ok := evCaptured[ev.InvolvedObject.UID]; ok {
+						evCaptured[ev.InvolvedObject.UID] = true
+					}
+				}()
+			}
+		}()
+	}
+
+	stopEvenCapture := func() {
```
pohly, line 119: nit: s/stopEvenCapture/stopEventCapture/

test/e2e/versionskew/operator.go
```diff
@@ -0,0 +1,227 @@
...
+		evWatcher.Stop()
+	}
+
+	clearEvents := func() {
+		evMutex.Lock()
+		defer evMutex.Unlock()
+		for k := range evCaptured {
```
pohly, line 126: Isn't it simpler to do `evCaptured = map[types.UID]bool{}`?

test/e2e/versionskew/operator.go
```diff
@@ -0,0 +1,227 @@
...
+
+	captureEvents := func(uuids ...types.UID) {
+		framework.Logf("Checking for events on: %s", uuids)
+
+		// Initialize events map
+		evMutex.Lock()
+		evCaptured = map[types.UID]bool{}
+		for _, uid := range uuids {
+			evCaptured[uid] = false
+		}
```
pohly, line 94: This looks redundant. `evCaptured[uid]` is false when `uid` doesn't have an entry, so explicitly adding it with false as value can be skipped.

test/e2e/versionskew/operator.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	ensureDeploymentReady := func(uids []types.UID, what string) {
+		framework.Logf("Ensuring that the CRs got reconciled")
+
+		Eventually(func() bool {
+			evMutex.Lock()
+			defer evMutex.Unlock()
+			for _, uid := range uids {
+				if found, ok := evCaptured[uid]; !ok || !found {
```
pohly, line 154: It looks like this can be simplified to `if !evCaptured[uid] { return false }`

test/e2e/versionskew/operator.go
avalluri: The operator upgrade is supported by the newer  `operator-sdk` releases for the operators in the 'bundle manifest'(`operator-sdk generate bundle`) format, but not in 'package manifest'. We should anyway move to bundle format as the packagemanifests are becoming obsolete. I am working on this now, will submit a new PR.

But downgrading is not supported by operator-sdk.

test/e2e/versionskew/operator.go
avalluri: Done.

test/e2e/versionskew/operator.go
avalluri: Yes, My intention was to store only the events for given deployment UUIDs, and avoid all the events generated by other objects. But this is anyway addressed in the event condition check. So pre initializing the map is unnecessary. I made the chagnes.

test/e2e/versionskew/operator.go
avalluri: Done.

test/e2e/versionskew/operator.go
avalluri: Yes, Done.

avalluri:
```
    + docker push 172.17.0.1:5000/pmem-csi-bundle:v0.9.0
    The push refers to repository [172.17.0.1:5000/pmem-csi-bundle]
    Get https://172.17.0.1:5000/v2/: http: server gave HTTP response to HTTPS client
    + rm -rf /tmp/tmp.NSpHVGosnU
```

Docker push on our Jenkins host fails as above. The `start-operator.sh` script pushes the generated operator bundle image to the local registry configured via `TEST_LOCAL_REGISTRY'. 

@pohly Do you have any idea where to set the local registry as insecure in our Jenkins host?

pohly:
Don't push to `TEST_LOCAL_REGISTRY`. That's the registry from which the QEMU nodes pull images.

Use `TEST_BUILD_PMEM_REGISTRY` instead. It's "the same registry reachable from the build host".

But I might be misunderstanding what you are pushing and why. I haven't looked at your changes.

avalluri:
> Don't push to `TEST_LOCAL_REGISTRY`. That's the registry from which the QEMU nodes pull images.
> 
> Use `TEST_BUILD_PMEM_REGISTRY` instead. It's "the same registry reachable from the build host".

That was my initial try which also failed. On CI `TEST_BUILD_PMEM_REGISTRY` set to `localhost` and failed to push the bundle image with SSL same error:

```
    + /mnt/workspace/pmem-csi_operator-versionskew/_work/bin/operator-sdk run bundle --namespace default --timeout 5m localhost:5000/pmem-csi-bundle:v0.9.0 --skip-tls
    time="2021-04-23T19:03:02Z" level=fatal msg="Failed to run bundle: pull bundle image: error pulling image localhost:5000/pmem-csi-bundle:v0.9.0: error resolving name : failed to do request: Head \"http://localhost:5000/v2/pmem-csi-bundle/manifests/v0.9.0\": dial tcp 127.0.0.1:5000: connect: connection refused\n"
    + rm -rf /tmp/tmp.pRyOFNiS3X
```
> 
> But I might be misunderstanding what you are pushing and why. I haven't looked at your changes.

The OLM starts the operator by pulling the bundle image. For that, we prepare the bundle image and push it to the local test registry and as the OLM to pull the bundle from that registry.

test/stop-operator.sh
```diff
@@ -45,32 +48,74 @@ function delete_olm_operator() {
...
+resources:
+- pmem-csi-operator.yaml
+$($keep_namespace && echo -n \
+'patchesStrategicMerge:
+- patch.yaml'
+)
+EOF
+
+  if $keep_namespace ; then
+  cat > ${tmpdir}/patch.yaml <<EOF
```
pohly, line 73: nit: indention of "cat" is the same as "if".

test/stop-operator.sh
```diff
@@ -45,32 +48,74 @@ function delete_olm_operator() {
...
   DEPLOY_DIRECTORY="${REPO_DIRECTORY}/deploy/operator"
-  deploy="${DEPLOYMENT_DIRECTORY}/pmem-csi-operator.yaml"
+  deploy="${DEPLOY_DIRECTORY}/pmem-csi-operator.yaml"
+
+  tmpdir=$(mktemp -d)
+  trap "rm -rf $tmpdir" SIGTERM SIGINT EXIT
+
+  cat > ${tmpdir}/pmem-csi-operator.yaml <<EOF
+$(cat "${deploy}")
+EOF
```
pohly, line 61: How is this different from `cp "${deploy}" ${tmpdir}/pmem-csi-operator.yaml`?

pohly: Introducing and using the `deploy` variable only once makes this function harder to follow.

test/e2e/deploy/deploy.go
```diff
@@ -1051,30 +1051,48 @@ func EnsureDeploymentNow(f *framework.Framework, deployment *Deployment) {
...
-		if running.HasOperator {
-			cmd := exec.Command("test/stop-operator.sh")
-			if running.HasOLM {
-				cmd.Args = append(cmd.Args, "-olm")
-			}
-			// Keep both the CRD and namespace in case the required deployment is also the operator
-			// required for version skew tests, otherwise it brings down the existing driver deployment(s)
-			if deployment.HasOperator {
-				cmd.Args = append(cmd.Args, "-keep-crd", "-keep-namespace")
+		// If both running and wanted operator deployments are managed by
```
pohly, line 1054: The commit message has a typo: s/vesion-skew/version-skew/

test/e2e/deploy/deploy.go
```diff
@@ -1051,30 +1051,48 @@ func EnsureDeploymentNow(f *framework.Framework, deployment *Deployment) {
...
+		// OLM, then we do nothing. The ./test/start-operato-sh -olm scirpt
```
pohly, line 1055: s/scirpt/script/

test/e2e/versionskew/operator.go
```diff
@@ -202,10 +216,18 @@ var _ = deploy.DescribeForSome("versionskew", func(d *deploy.Deployment) bool {
 	}
 
 	It("downgrade [Slow]", func() {
+		if d.HasOLM {
+			Skip("OLM yet not support operator downgrade!")
```
pohly, line 220: "OLM does not support operator downgrade yet."

test/e2e/versionskew/versionskew.go
```diff
@@ -39,6 +39,8 @@ import (
 const (
 	// base is the release branch used for version skew testing. Empty if none.
 	base = "0.9"
+	// Expected release version(used by the operator)
+	current = "1.0"
```
pohly, line 43: This assumes that the next version will be 1.0.0, right? How does it work given that this version doesn't exist yet?

I'm giving up on reminding you to add a space in English prose before an opening bracket...

pohly (changes requested):

test/stop-operator.sh
avalluri: Fixed.

test/stop-operator.sh
avalluri: Removed the unneeded variables.

test/e2e/deploy/deploy.go
avalluri: Fxied.

test/e2e/deploy/deploy.go
avalluri: Fixed.

test/e2e/versionskew/operator.go
avalluri: Done.

test/e2e/versionskew/versionskew.go
avalluri: > This assumes that the next version will be 1.0.0, right? How does it work given that this version doesn't exist yet?

Its need not exist. This is the version number used for the 'operator bundle' generated for the devel code, The bundle gets generated by the 'upgrade test'. And it assumes that the operator gets an upgrade from the 'base'(released) version to the 'current'(future) version.

pohly (approved):

Issue #925, opened 2021-04-09
LVM: more resilient output parsing


pohly:
On some machines, tools like `vgs` print errors in addition to the normal error:

```
/dev/sda: open failed: No medium failed
VG ...
ndbus0region0fsdax ...
```

This break volume creation in PMEM-CSI because our output parsing doesn't expect the `/dev/sda` line.

We should more clearly distinguish between stdout and stderr when invoking LVM tools and only parse stdout.

PR #929, opened 2021-04-12
exec: cleanly separate stdout/stderr

pohly:
Commit 40ba3555a43fa2aa8aa1f081af1f2386edda4dd8 intentionally captured
stderr in the output because earlier it was hard to understand why LVM
commands failed. However, we now learned the some error messages can
be safely ignored and not doing so broke LVM mode: when /dev/sda is
unreadable, "vgs" prints a line about it on stderr, but the output on
stdout is fine.

Now we capture both output streams separately, log both, but only
return stdout to the caller.

The TestResult call is unrelated to that fix and will be used for
additional debugging in another PR. It gets added here because it can
use the same test cases and is related.

Fixes: #925

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -32,27 +32,59 @@ func Run(cmd *exec.Cmd) (string, error) {
 	klog.V(4).Infof("Executing: %s", cmd)
 
 	r, w := io.Pipe()
+	r2, w2 := io.Pipe()
 	cmd.Stdout = w
-	cmd.Stderr = w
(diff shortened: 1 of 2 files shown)


pohly:
I tried to trigger the error output from `vgs` also under QEMU, but without success. Adding an empty CD-ROM driver with `-drive if=ide,index=0,media=cdrom` did not lead to an error messages. Therefore we have to rely on the unit tests to become confident that output parsing will now ignore stderr.

Log output is also not tested.

avalluri (approved):
Looks good to me.

PR #931, opened 2021-04-14
operator: avoid unexpected object patching

### LINKED ISSUE #924: operator: unexpected patching of objects
We've observed that the operator patches objects when nothing changed in the object (https://github.com/intel/pmem-csi/issues/884#issuecomment-814352954).

The patch is a lot larger than I expected:
```
pmem-csi-operator-7d8f55f866-pc2ht/pmem-csi-operator@pmem..ker3: I0406 20:08:35.694748       1 controller_driver.go:280] "redeploy: update" object="default/second-pmem-csi-intel-com-node-setup" patch="{\"spec\":{\"revisionHistoryLimit\":null,\"template\":{\"spec\":{\"containers\":[{\"command\":[\"/usr/local/bin/pmem-csi-driver\",\"-v=3\",\"-logging-format=text\",\"-mode=force-convert-raw-namespaces\",\"-nodeSelector={\\\"feature.node.kubernetes.io/memory-nv.dax\\\":\\\"true\\\"}\",\"-nodeid=$(KUBE_NODE_NAME)\"],\"env\":[{\"name\":\"KUBE_NODE_NAME\",\"valueFrom\":{\"fieldRef\":{\"apiVersion\":\"v1\",\"fieldPath\":\"spec.nodeName\"}}},{\"name\":\"TERMINATION_LOG_PATH\",\"value\":\"/tmp/termination-log\"}],\"image\":\"172.17.42.1:5001/pmem-csi-driver:canary\",\"imagePullPolicy\":\"IfNotPresent\",\"name\":\"pmem-driver\",\"resources\":{},\"securityContext\":{\"privileged\":true,\"runAsUser\":0},\"terminationMessagePath\":\"/tmp/termination-log\",\"volumeMounts\":[{\"mountPath\":\"/dev\",\"name\":\"dev-dir\"},{\"mountPath\":\"/sys\",\"name\":\"sys-dir\"},{\"mountPath\":\"/host-sys\",\"name\":\"sys-dir\"}]}],\"dnsPolicy\":null,\"restartPolicy\":null,\"schedulerName\":null,\"securityContext\":null,\"serviceAccount\":null,\"terminationGracePeriodSeconds\":null}},\"updateStrategy\":{\"rollingUpdate\":null,\"type\":null}}}"
```

This is from testing https://github.com/intel/pmem-csi/pull/919 but that PR doesn't modify the operator logic.

One test failure there is about `objects were changed after reaching expected state`, which might be related. https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-919/8/tests

```
/mnt/workspace/pmem-csi_PR-919/test/e2e/operator/driver.go:37
Apr  8 14:04:25.229: validate driver
Unexpected error:
    <*errors.errorString | 0xc0025d0e00>: {
        s: "objects were changed after reaching expected state: deployed driver different from expected deployment:\nobject was modified unnecessarily: \"pmem-csi-intel-com-controller\" of type \"apps/v1, Kind=StatefulSet\" in namespace \"kube-system\"",
    }
    objects were changed after reaching expected state: deployed driver different from expected deployment:
    object was modified unnecessarily: "pmem-csi-intel-com-controller" of type "apps/v1, Kind=StatefulSet" in namespace "kube-system"
occurred
/mnt/workspace/pmem-csi_PR-919/test/e2e/operator/driver.go:54
```

avalluri:
On redeploying the objects operator should not unset the defaults chosen
by the API server, otherwise it results in patch diff and leads to
unnecessary update object call. This is observed both driver
statefulset and daemonset objects.

To avoid this we have two options:
 1. set only specific options instead of total refresh.
 2. set also the default values

We chose option 1 while setting daemonset and statefulset specification
fields, and option 2 for setting container values.

FIXES #924

### DIFF
--- a/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
+++ b/pkg/pmem-csi-operator/controller/deployment/controller_driver.go
@@ -269,6 +269,7 @@ func (d *pmemCSIDeployment) redeploy(ctx context.Context, r *ReconcileDeployment
 		}
 		// Check whether we really need to patch.
 		if string(data) != "{}" && len(data) >= 0 {
+			l.V(5).Info("patch", "diff", string(data))
 			// Patch() will modify the object, which is an object that was
 			// generated from our PmemCSIDeployment object and shares some
 			// data structure with it. We don't want those to be modified,
@@ -1031,6 +1032,17 @@ func (d *pmemCSIDeployment) getControllerStatefulSet(ss *appsv1.StatefulSet) {
 	true := true
 	pmemcsiUser := int64(1000)
 
+	// To make sure that the default values set by the API server
+	// are not unset by the operator we choose to update only specific
+	// we are interested.
+	//
+	// NOTE: Do not ferget to unset the fields that are set conditionally, as below:
+	// if expr{
+	//   ss.Spec.FieldX = some_value
+	// } else {
+	//	ss.Spec.FieldX = unset
+	// }
+
 	if ss.Labels == nil {
 		ss.Labels = map[string]string{}
 	}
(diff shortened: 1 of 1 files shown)


pohly (changes requested):
Can you add a test for this? As it stands now, we have to be very careful about how we update the objects, and that is easy to get wrong because the effect is very subtle when done incorrectly.

Please also add a comment somewhere that *removing* values that were previously set by the operator but now no longer are desired needs special code.

avalluri:
> Can you add a test for this?

Testing this is a bit hard. These defaults are set by the API server, which cannot be tested in unit tests, and in E2E tests it is not possible to determine that the sub-objects updates are coming from the operator. We might introduce an event and emit that while patching an object., which could be used in the test if is expected or not.

> As it stands now, we have to be very careful about how we update the objects, and that is easy to get wrong because the effect is very subtle when done incorrectly.

In fact,  I guess these patch calls would not result in updating the object at the server-side, that is the reason is not caught by driver validation of unexpected update check in e2e tests.

pohly:
> > Can you add a test for this?
> 
> Testing this is a bit hard. These defaults are set by the API server, which cannot be tested in unit tests, and in E2E tests it is not possible to determine that the sub-objects updates are coming from the operator. We might introduce an event and emit that while patching an object., which could be used in the test if is expected or not.

Events are not reliable. There is garbage collection for them. It shouldn't be a big problem in practice for us therefore I am not too worried that the downgrade testing is using them, but it's still better to avoid the dependency if possible.

What about introducing some metric for the operator? That then can be queried reliably - if it can be queried from outside the cluster. In those cases where we currently read metrics, we do so via the socat pod running alongside the driver pod, but that only works in testing deployments, not in production deployments, and not for the operator.

We probably need to copy some code from https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/drivers/proxy/portproxy.go and then will have a `Dial(pod, port)` function which can connect from outside. Let me try to implement that...

> > As it stands now, we have to be very careful about how we update the objects, and that is easy to get wrong because the effect is very subtle when done incorrectly.
> 
> In fact, I guess these patch calls would not result in updating the object at the server-side, that is the reason is not caught by driver validation of unexpected update check in e2e tests.

I'm not so sure. Some of the test failures in https://github.com/intel/pmem-csi/issues/924 were about unexpected object changes.

pohly (approved):

PR #934, opened 2021-04-15
test: retrieve metrics with port forwarding instead of socat

pohly:
With port forwarding through the API server we no longer need a
running container with socat inside it. For the metrics test it makes
no big difference, but the new code might be useful also for other
tests.


avalluri (approved):
Looks good to me.

PR #937, opened 2021-04-15
deploy: set fsgroup for the pod in application example(s)

### LINKED ISSUE #926: examples: run as root or change ownership
The pmem-app.yaml files run the PMEM-CSI container with the `pmem-csi` user. That makes the /data directory in the example unusable.

We either must change the ownership of it or change the user.

avalluri:
The 'pmem-csi-driver-test' container runs with 'pmem-csi' user which
does not allow to access to mounted /data volume. Change the permissions
on the /data mount by using pod security context.

FIXES #926

### DIFF
--- a/deploy/common/pmem-app-cache.yaml
+++ b/deploy/common/pmem-app-cache.yaml
@@ -12,6 +12,12 @@ spec:
       labels:
         app: my-csi-app
     spec:
+      # The 'pmem-csi-driver-test' container runs apps as user/group 1000/1000.
+      # This security context causes permissions of volume mounts
+      # to be adapted accordingly, see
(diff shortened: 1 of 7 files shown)


pohly:
I don't think this is the right solution. Suid binaries in a container smell fishy.

If `pmem-dax-check` was the application, then it would still run as root. If that is the goal, then `SecurityContext` with `RunAsUser: 0` and `RunAsGroup: 0` is a simpler and more flexible solution (decided during deployment instead of image build).

avalluri:
I changed the PR to set `securityContext.fsGroup` for the application pods as discussed in #926.

pohly (approved):

PR #938, opened 2021-04-17
Namespace creation size-related fixes

okartau:
We may adjust requested size upwards, so lets check for available
space after that.
This PR is meant to address points in #933

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,13 @@ ARG GO_VERSION="1.16.1"
 # run instead of just using some older, cached result.
 ARG CACHEBUST
 
+# We want newer ndctl that is available in buster:
+RUN echo 'deb http://ftp.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/buster-backports.list
 # In contrast to the runtime image below, here we can afford to install additional
 # tools and recommended packages. But this image gets pushed to a registry by the CI as a cache,
 # so it still makes sense to keep this layer small by removing /var/cache.
 RUN ${APT_GET} update && \
-    ${APT_GET} install -y gcc libndctl-dev make git curl iproute2 pkg-config xfsprogs e2fsprogs parted openssh-client python3 python3-venv equivs && \
+    ${APT_GET} install -y gcc libndctl-dev/buster-backports make git curl iproute2 pkg-config xfsprogs e2fsprogs parted openssh-client python3 python3-venv equivs && \
     rm -rf /var/cache/*
 RUN curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -zxf - -C / && \
     mkdir -p /usr/local/bin/ && \
@@ -43,10 +45,11 @@ COPY --from=build python3_100.0_all.deb /var/cache/python3_100.0_all.deb
 # lvm2 - volume management
 # ndctl - pulls in the necessary library, useful by itself
 # fio - only included in testing images
+RUN echo 'deb http://ftp.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/buster-backports.list
 RUN ${APT_GET} update && \
     mkdir -p /usr/local/share && \
     dpkg -i /var/cache/python3_100.0_all.deb && \
-    bash -c 'set -o pipefail; ${APT_GET} install -y --no-install-recommends file xfsprogs e2fsprogs lvm2 ndctl \
+    bash -c 'set -o pipefail; ${APT_GET} install -y --no-install-recommends file xfsprogs e2fsprogs lvm2 libndctl-dev/buster-backports ndctl/buster-backports \
        | tee --append /usr/local/share/package-install.log' && \
     rm -rf /var/cache/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ ARG CACHEBUST
 
 # We want newer ndctl that is available in buster:
 RUN echo 'deb http://ftp.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/buster-backports.list
+RUN echo 'deb-src http://ftp.debian.org/debian buster-backports main' >> /etc/apt/sources.list.d/buster-backports.list
 # In contrast to the runtime image below, here we can afford to install additional
 # tools and recommended packages. But this image gets pushed to a registry by the CI as a cache,
 # so it still makes sense to keep this layer small by removing /var/cache.
@@ -46,6 +47,7 @@ COPY --from=build python3_100.0_all.deb /var/cache/python3_100.0_all.deb
 # ndctl - pulls in the necessary library, useful by itself
 # fio - only included in testing images
 RUN echo 'deb http://ftp.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/buster-backports.list
+RUN echo 'deb-src http://ftp.debian.org/debian buster-backports main' >> /etc/apt/sources.list.d/buster-backports.list
 RUN ${APT_GET} update && \
     mkdir -p /usr/local/share && \
     dpkg -i /var/cache/python3_100.0_all.deb && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,7 +115,7 @@ RUN cd /usr/local/share/package-sources && \
         else \
             echo "    $pkg"; \
         fi; \
-    done | sort -u; \
+    done && \
     rm -rf /var/cache/*
 
 # build pmem-csi-driver
--- a/pkg/ndctl/fake/region.go
+++ b/pkg/ndctl/fake/region.go
@@ -25,6 +25,7 @@ type Region struct {
 	Enabled_            bool
 	Readonly_           bool
 	InterleaveWays_     uint64
+	RegionAlign_        uint64
 
 	Mappings_   []ndctl.Mapping
 	Namespaces_ []ndctl.Namespace
@@ -105,8 +106,11 @@ func (r *Region) SeedNamespace() ndctl.Namespace {
 	return nil
 }
 
+func (r *Region) GetAlign() uint64 {
+	return r.RegionAlign_
+}
+
 func (r *Region) CreateNamespace(opts ndctl.CreateNamespaceOpts) (ndctl.Namespace, error) {
-	defaultAlign := mib2
 	var err error
 	/* Set defaults */
 	if opts.Type == "" {
@@ -150,11 +154,7 @@ func (r *Region) CreateNamespace(opts ndctl.CreateNamespaceOpts) (ndctl.Namespac
 		if opts.Size > available {
 			return nil, fmt.Errorf("create namespace with size %v: %w", opts.Size, pmemerr.NotEnoughSpace)
 		}
-	}
-	opts.Align = defaultAlign
-
-	if opts.Size != 0 {
-		align := opts.Align
+		align := mib2
 		if opts.Size%align != 0 {
 			// Round up size to align with next block boundary.
 			opts.Size = (opts.Size/align + 1) * align
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -65,6 +65,8 @@ type Region interface {
 	AdaptAlign(align uint64) (uint64, error)
 	// FsdaxAlignment returns the default alignment for an fsdax namespace.
 	FsdaxAlignment() (uint64, error)
+	// GetAlign returns region alignment.
+	GetAlign() uint64
 }
 
 type region = C.struct_ndctl_region
@@ -143,8 +145,11 @@ func (r *region) SeedNamespace() Namespace {
 	return C.ndctl_region_get_namespace_seed(r)
 }
 
+func (r *region) GetAlign() uint64 {
+	return uint64(C.ndctl_region_get_align(r))
+}
+
 func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
-	defaultAlign := mib2
 	var err error
 	/* Set defaults */
 	if opts.Type == "" {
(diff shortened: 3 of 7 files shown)


test/start-kubernetes.sh
```diff
@@ -29,7 +29,7 @@ KVM_CPU_OPTS="${KVM_CPU_OPTS:-\
  -machine pc,accel=kvm,nvdimm=on}"
 EXTRA_QEMU_OPTS="${EXTRA_QWEMU_OPTS:-\
  -object memory-backend-file,id=mem1,share=${TEST_PMEM_SHARE},\
-mem-path=/data/nvdimm0,size=$((${TEST_PMEM_MEM_SIZE} * 1024 * 1024 + ${TEST_PMEM_LABEL_SIZE})) \
+mem-path=/data/nvdimm0,size=$((${TEST_PMEM_MEM_SIZE} * 1024 * 1024 + ${TEST_PMEM_LABEL_SIZE})),align=2M \
```
pohly, line 32: Have you tried this locally?

If you rely on the CI to try out changes like this, then progress will be very slow.

okartau: yes you are right, this was not the cause. I then switched to local dev for this, to address overall sizes and alignment issues.

okartau:
As we discussed off-line, it could help if we used more recent libndctl, which has get_region_align function starting with v68 I believe. But our containerized build in current config will use v63

okartau:
I found one way to install newer ndctl, it worked on my local setup, lets try CI now. There is also call added to get region alignment value and adjust size accordingly.

pkg/ndctl/fake/region.go
```diff
@@ -105,6 +105,10 @@ func (r *Region) SeedNamespace() ndctl.Namespace {
 	return nil
 }
 
+func (r *Region) GetAlign() uint64 {
+	return uint64(mib * 16)
```
pohly, line 109: This should better be a member variable to enable unit testing with different alignments.

Dockerfile
```diff
@@ -43,10 +45,11 @@ COPY --from=build python3_100.0_all.deb /var/cache/python3_100.0_all.deb
 # lvm2 - volume management
 # ndctl - pulls in the necessary library, useful by itself
 # fio - only included in testing images
+RUN echo 'deb http://ftp.debian.org/debian buster-backports main' > /etc/apt/sources.list.d/buster-backports.list
```
pohly, line 48: With only the binary repo enabled, does downloading the source code of that newer ndctl work?

I'll check...

pohly: It fails, but due to missing error handling the image build continues:
```
Reading package lists...
Picking 'ndctl' as source package instead of 'libdaxctl1'
E: Can not find version '71.1-1~bpo10+1' of package 'libdaxctl1'
E: Unable to find a source package for ndctl
```

I'll fix this.

pohly: See https://github.com/okartau/pmem-CSI/pull/1

pkg/ndctl/region.go
```diff
@@ -208,13 +204,36 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
+		by region align which is 16 or 96 MB so we may leave this code out
+		ways := r.InterleaveWays()
+				align = align * ways
+				if opts.Size%align != 0 {
+					// Align down to namespace-alignment boundary.
+					opts.Size /= align
+					opts.Size *= align
+					klog.V(4).Infof("%s: namespace size must align to interleave-width:%d * alignment:%d, adjust to %d",
+						regionName, ways, align, opts.Size)
+				}*/
```
pohly, line 218: Please remove the old code and instead add comments explaining how the current code works.

pkg/ndctl/region.go
```diff
@@ -208,13 +204,36 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
+		regionalign := r.GetAlign()
+		if opts.Size%regionalign != 0 {
+			// Align down to region-alignment boundary.
+			opts.Size /= regionalign
+			opts.Size *= regionalign
+			klog.V(4).Infof("%s: namespace size must align to region alignment:%d, adjust to %d",
+				regionName, regionalign, opts.Size)
```
pohly, line 225: Instead of modifying opts, can your create a local variable? Then in this info message you can print the original value and the new one.

pkg/pmem-device-manager/pmd-lvm.go
```diff
@@ -349,22 +339,15 @@ func setupNS(r ndctl.Region, percentage uint) error {
...
-		})
-		if err != nil {
-			return fmt.Errorf("failed to create PMEM namespace with size '%d' in region '%s': %v", canUse, r.DeviceName(), err)
-		}
+	klog.V(3).Infof("Create %v-bytes fsdax-namespace", canUse)
+	_, err := r.CreateNamespace(ndctl.CreateNamespaceOpts{
+		Name:  "pmem-csi",
+		Mode:  "fsdax",
+		Size:  canUse,
+		Align: align,
```
pohly: Why do we need to set an alignment here at all? Is there any scenario where the caller of `CreateNamespace` needs to set the alignment? If not, then let's remove `Align` from `CreateNamespaceOpts`.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -14,9 +14,16 @@ import (
...
+	ndctlAlign uint64 = 2 * 1024 * 1024
+	// Allocation overhead compensation per 4096 data bytes that we need to ask more to get what was requested.
+	// libndctl needs to store meta data and will use some of the allocated
+	// space for that (https://github.com/pmem/ndctl/issues/79).
+	// According to existing doc, the overhead is 64 bytes per 4096 data bytes.
+	// Trials however show that the overhead appears to be larger:
+	// 64..65 bytes (avg) if alignment=2M
+	// 66..69 bytes (avg) if alignment=1G
+	// TODO: base overhead value on real doc or code instead of trial-based value.
+	// ndctlOverheadCompensation uint64 = 66
```
pohly, line 26: Remove dead code please.

pkg/pmem-device-manager/pmd-ndctl.go
```diff
@@ -14,9 +14,16 @@ import (
 )
 
 const (
-	// 1 GB align in ndctl creation request has proven to be reliable.
-	// Newer kernels may allow smaller alignment but we do not want to introduce kernel dependency.
-	ndctlAlign uint64 = 1024 * 1024 * 1024
+	ndctlAlign uint64 = 2 * 1024 * 1024
```
pohly, line 17: Document what this alignment is (just a fallback, right?).

pkg/ndctl/region.go
```diff
@@ -208,13 +204,36 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
-		align = align * ways
-		if opts.Size%align != 0 {
-			// Round up size to align with next block boundary.
-			opts.Size = (opts.Size/align + 1) * align
-			klog.V(4).Infof("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
-				regionName, ways, align, opts.Size)
+		/* old method, using namespace align and interleave ways.
+		If we use fixed 2M namespace align, this is always overshadowed
+		by region align which is 16 or 96 MB so we may leave this code out
+		ways := r.InterleaveWays()
```
pohly, line 210: I'd prefer to not make that assumption. Better align with both to be on the safe side.

test/start-kubernetes.sh
okartau: I removed that change as it seemed not relevant after all, probably we get 2M align by default?

Dockerfile
okartau: merged

pkg/ndctl/region.go
pohly: More specifically, check the alignment requirement of the pfn (= `ndctl_pfn_get_align`). If that is not set, then we can fall back to some builtin default, like this code does:
https://github.com/pmem/ndctl/blob/ea014c0c9ec8d0ef945d072dcc52b306c7a686f9/ndctl/namespace.c#L799-L814

pohly: ndctl also checks both:
https://github.com/pmem/ndctl/blob/ea014c0c9ec8d0ef945d072dcc52b306c7a686f9/ndctl/namespace.c#L838-L843

okartau: I was thinking about that, too. Its simple to find common align value if one is 2G, but what if some day was need to switch to 1G, then what is common align between 1G and 96MB, isnt it getting quite large?

okartau: hm I meant, if one is 2MB

pohly: Although there it aborts when not aligned. We need to decrease the size.

pohly: Wait, *increase* the size.

pohly: In the GetCapacity call we round down. When creating, we round up.

pohly: If we have regionAlign and namespaceAlign, then we need to align at the smallest common multiple of both. Then the resulting size will be a multiple of both.

pohly: I most cases, the smallest common multiple will be the region alignment. We just don't make that assumption by actually doing the math.

okartau: > In the GetCapacity call we round down. When creating, we round up.

in current draft, in creation we align down. idea is, if we give up compensating idea, it does not matter much, and it is safer to align down, as aligning up has risk to exceed total size, or percentage what is allowed. Total sizes seem not to be aligned, which is bit unfriendly in that respect

pohly: > but what if some day was need to switch to 1G, then what is common align between 1G and 96MB, isnt it getting quite large

And? If the requirement is to align with both, we have no choice.

> in current draft, in creation we align down. idea is, if we give up compensating idea, it does not matter much, 

It does matter. Our promise to the user is that when they request a volume of size 1GB, we use that amount of PMEM, not less.

> aligning up has risk to exceed total size

In that case the host does not have enough capacity left for the volume and it needs to be created elsewhere. Rounding down in GetCapacity is supposed to avoid this scenario.

pohly: > hm I meant, if one is 2MB

The smallest common multiple of 2Mi and 1Gi is 1Gi.

okartau: I meant, if in theory, we need to switch to 1GB namespace align again, then for example LCM of 96 MB and 1GB will be 3072 MB. But does not matter, I added LCM code that calculates it in all cases. I hope borrowing LCM function from Go playground is allowed, with added comment of course.

pkg/pmem-device-manager/pmd-ndctl.go
okartau: this goes away, I removed passing Align from device manager to region.go, means it will be unused zero field

pkg/ndctl/region.go
pohly: > I meant, if in theory, we need to switch to 1GB namespace align again, then for example LCM of 96 MB and 1GB will be 3072 MB.

If someone configures their hardware so that these two alignment constraints apply, then yes, that's what we'll have to do.

pkg/ndctl/region.go
```diff
@@ -208,13 +204,25 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
-			opts.Size = (opts.Size/align + 1) * align
-			klog.V(4).Infof("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
-				regionName, ways, align, opts.Size)
+		namespacealign := align * r.InterleaveWays()
+		regionalign := r.GetAlign()
+		// Size has to be aligned both by namespace alignment times interleave_ways, and also by region alignment
+		lcmalign := LCM(namespacealign, regionalign)
+		klog.V(3).Infof("%s: Least Common Multiple of namespacealign:%d and regionalign:%d is %d",
+			regionName, namespacealign, regionalign, lcmalign)
+		if opts.Size%regionalign != 0 {
```
pohly, line 213: s/regionalign/lcmalign/ here.

pkg/ndctl/region.go
```diff
@@ -208,13 +204,25 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
 
 	if opts.Size != 0 {
-		ways := uint64(C.ndctl_region_get_interleave_ways(r))
-		align = align * ways
-		if opts.Size%align != 0 {
-			// Round up size to align with next block boundary.
-			opts.Size = (opts.Size/align + 1) * align
-			klog.V(4).Infof("%s: namespace size must align to interleave-width:%d * alignment:%d, force-align to %d",
-				regionName, ways, align, opts.Size)
+		namespacealign := align * r.InterleaveWays()
```
pohly, line 207: What if `Size` really is zero?

Shouldn't we then create a minimal-sized namespace, i.e. go through this alignment code and round up?

pkg/ndctl/region.go
```diff
@@ -184,16 +190,6 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
...
-		available := r.MaxAvailableExtent()
-		if available == uint64(C.ULLONG_MAX) {
-			available = r.AvailableSize()
-		}
-		if opts.Size > available {
-			return nil, fmt.Errorf("create namespace with size %v: %w", opts.Size, pmemerr.NotEnoughSpace)
-		}
-	}
-
 	align := defaultAlign
```
pohly, line 193: The `defaultAlign` variable serves no purpose and just makes the code harder to read. Please remove it.

pohly: opts.Align is unused. We can remove it and the dead code checking it:
```
diff --git a/pkg/ndctl/ndctl.go b/pkg/ndctl/ndctl.go
index 9545c6e1a..cbaac22f3 100644
--- a/pkg/ndctl/ndctl.go
+++ b/pkg/ndctl/ndctl.go
@@ -29,7 +29,6 @@ type CreateNamespaceOpts struct {
 	Name       string
 	Size       uint64
 	SectorSize uint64
-	Align      uint64
 	Type       NamespaceType
 	Mode       NamespaceMode
 	Location   MapLocation
diff --git a/pkg/ndctl/region.go b/pkg/ndctl/region.go
index abc714d36..990122527 100644
--- a/pkg/ndctl/region.go
+++ b/pkg/ndctl/region.go
@@ -150,7 +150,6 @@ func (r *region) GetAlign() uint64 {
 }
 
 func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
-	defaultAlign := mib2
 	var err error
 	/* Set defaults */
 	if opts.Type == "" {
@@ -190,19 +189,7 @@ func (r *region) CreateNamespace(opts CreateNamespaceOpts) (Namespace, error) {
 		}
 	}
 
-	align := defaultAlign
-	if opts.Align != 0 {
-		if opts.Mode == SectorMode || opts.Mode == RawMode {
-			klog.V(4).Infof("%s mode does not support setting an alignment, hence ignoring alignment", opts.Mode)
-		} else {
-			var err error
-			align, err = r.AdaptAlign(opts.Align)
-			if err != nil {
-				return nil, err
-			}
-		}
-	}
-
+	align := mib2
 	if opts.Size != 0 {
 		namespacealign := align * r.InterleaveWays()
 		regionalign := r.GetAlign()
```

okartau: I was not sure do you mean stop using it, or stop and remove it from Opts. I was thinking, perhaps we keep it so that it can possibly used again in the future? BTW we have also other fields in Opts that we don't use or always set to one default, like SectorSize. Type, Mode, Location . Although Sector mode may see it's comeback based on another discussion.

pkg/ndctl/region.go
okartau: oops, leftover in replace, good catch, thanks!

pkg/ndctl/region.go
okartau: From LVM dev-mgr, we do not call with zero arg. There has been check for too small/zero value in previous code. Now in simplifying stage, I tried to remove it, and I saw one test (namespace conversion IIRC) caused call of LVM-mode creation with zero size, which did not went OK, so I restored check there and we only call Create if size >0.
In direct case, we always use size what is passed into CreateDevice without checking, and IIRC thats always been the case and has not caused trouble. I believe libndctl then creates max-sized namespace, at least that happens in cmd-line.
yes your idea is good, lets check it and create minimal i.e. lcmalign-sized

pohly: > In direct case, we always use size what is passed into CreateDevice without checking, and IIRC thats always been the case and has not caused trouble.

"not caused trouble" is a dangerous argument - it could be that no-one has ever tested that...

So we agree, let's treat "zero size" as "smallest possible namespace".

pkg/ndctl/region.go
pohly: I understand that conceptually `pkg/ndctl` is allowed to support more than we need. But this particular code has no test coverage, therefore I prefer to remove it. If it turns out that we need it, we bring it back and then ensure that it really works.

If we keep it in place without testing, we might not remember that it is potentially broken when we start to use it.

okartau:
pushed version addressing recent comments. I had to change code on fake/ side to adapt to change of Opts, now the CreateNamespace is slightly different there, do you think we need to unify that to real one in this PR? This would need more work, for example new LCM() should be made common then, I did not start that part now.

pohly:
> now the CreateNamespace is slightly different there, do you think we need to unify that to real one in this PR? This would need more work, for example new LCM() should be made common then, I did not start that part now.

This can be done later. I had noticed that the new math functions do not really belong into the file where they were added, but decided to ignore that.

pkg/ndctl/region.go
pohly: Not quite solved the way I meant (`opts.Size` still gets modified eventually), but the info message is more informative, so let's consider this resolved.

okartau:
> Not quite solved the way I meant (opts.Size still gets modified eventually)

My thinking was: opts.Size gets checked and used later in function, so I put modified value back and avoid using those other places, keeping current change impact smaller. Other fields of Opts are modified in that function as well, making Size local variable would raise question why Size deserves local variable on changing but not others

PR #945, opened 2021-04-26
avoid pod eviction

pohly:
Eviction can be avoided by:
- setting resource requirements (done by operator, but not the YAML
  files)
- *not* setting resource limits (because our estimate might have
  been too restrictive, in which case allow the pods to continue
  is better than killing them - the operator had default limits)
- setting a priority class
- tolerating NoSchedule and NoExecute

Fixes: #261


pkg/apis/pmemcsi/v1beta1/deployment_types.go
```diff
@@ -276,39 +276,24 @@ const (
 	// recommendations reported by VirtualPodAutoscaler(LowerBound -> Requests and UpperBound -> Limits)
 
 	// DefaultControllerResourceRequestCPU default CPU resource request used for controller driver container
-	DefaultControllerResourceRequestCPU = "12m" // MilliSeconds
-
+	DefaultControllerResourceRequestCPU = "12m"
```
pohly, line 279: The "m" stands for "millicpu". I don't think we need to document that here.

pkg/apis/pmemcsi/v1beta1/deployment_types.go
```diff
@@ -276,39 +276,24 @@ const (
...
 	// DefaultControllerResourceRequestMemory default memory resource request used for controller driver container
-	DefaultControllerResourceRequestMemory = "128Mi" // MB
+	DefaultControllerResourceRequestMemory = "128Mi"
```
pohly, line 281: "Mi" doesn't need further explanation.

avalluri (approved):
Looks good to me.

Issue #943, opened 2021-04-26
operator: add test for unexpected patching of objects


pohly:
We already merged the code (PR https://github.com/intel/pmem-csi/pull/931) but it is tricky to get the code right and avoid future regressions, therefore we need a test for it.

avalluri:
#946 also covers unexpected sub-object updates by the operator using metrics.

PR #946, opened 2021-04-30
Operator metrics

avalluri:
Added support for operator metrics, by default at port 8080, which could be configured using `--metrics-addr` command-line argument.

Changed the validation logic in e2e tests to make use of these metrics to know when a deployment CR went through the reconciliation loop and, to detect if any unexpected object updates initiated by the operator.

### DIFF
--- a/deploy/operator/pmem-csi-operator.yaml
+++ b/deploy/operator/pmem-csi-operator.yaml
@@ -140,6 +140,19 @@ subjects:
   name: pmem-csi-operator
   namespace: pmem-csi
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: pmem-csi-operator-metrics
+  namespace: pmem-csi
+spec:
+  ports:
+  - port: 8080
+    targetPort: 8080
+  selector:
+    app: pmem-csi-operator
+    name: pmem-csi-operator
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -149,6 +162,7 @@ spec:
   replicas: 1
   selector:
     matchLabels:
+      app: pmem-csi-operator
       name: pmem-csi-operator
   template:
     metadata:
@@ -160,6 +174,7 @@ spec:
       containers:
       - command:
         - /usr/local/bin/pmem-csi-operator
+        - -metrics-addr=:8080
         - -v=3
         env:
         - name: WATCH_NAMESPACE
@@ -177,6 +192,9 @@ spec:
         image: intel/pmem-csi-driver:canary
         imagePullPolicy: IfNotPresent
         name: pmem-csi-operator
+        ports:
+        - containerPort: 8080
+          name: metrics
         securityContext:
           readOnlyRootFilesystem: true
         volumeMounts:
--- a/docs/install.md
+++ b/docs/install.md
@@ -1485,6 +1485,18 @@ deployment is in the `Failed` state, then one can look into the event(s) using
 > driver could run can be configured by using `nodeSelector` property of
 > [`DeploymentSpec`](#deployment-crd-api).
 
+#### Operator metrics data
+
+PMEM-CSI operator exposes below metrics data about active PmemCSIDeployment
+custom resources and it's sub-object in addition to the
+[metrics data provided by the controller-runtime](https://book-v1.book.kubebuilder.io/beyond_basics/controller_metrics.html):
+
+Name | Type | Explanation
+-----|------|------------
+`pmem_csi_deployment_reconcile` | counter | Counter that gets incremented on each time a PmemCSIDeployment CR gone through a reconcile loop, labeled with the deployment name and uid.
+`pmem_csi_deployment_sub_resource_created_at` | gauge | Timestamp at which a sub resource of the PmemCSIDeployment CR was created  by the operator. Labeled by resource details ("name, "namespace", "group", "version", "kind", "uid", "ownedBy").
+`pmem_csi_deployment_sub_resource_updated_at` | gauge | Timestamp at which a sub resource of the PmemCSIDeployment CR was updated by the operator. Labeled by resource details ("name, "namespace", "group", "version", "kind", "uid", "ownedBy").
+
 
 ## Filing issues and contributing
 
--- a/pkg/pmem-csi-operator/metrics/metrics.go
+++ b/pkg/pmem-csi-operator/metrics/metrics.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package metrics
+
+import (
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	// PmemCSIDeploymentSubsystemKey represents the key used for
+	// PMEM-CSI deployment metrics sub-system.
+	PmemCSIDeploymentSubsystemKey = "pmem_csi_deployment"
+)
+
+var (
+	// Reconcile creates new prometheus metrics counter
+	// that gets incremented on each reconcile loop
+	// of a PmemCSIDeployment CR, with information: {"name", "uid"}.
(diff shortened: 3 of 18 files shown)


docs/install.md
```diff
@@ -1485,6 +1485,18 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
 > [`DeploymentSpec`](#deployment-crd-api).
 
+#### Operator metrics data
+
+PMEM-CSI operator exposes below metrics data in addition to the
+[metrics data provided by the controller-runtime](https://book-v1.book.kubebuilder.io/beyond_basics/controller_metrics.html):
+
+Name | Type | Explanation
+-----|------|------------
+`pmem_csi_deployment_reconcile_started_at` | gauge | Timestamp at which a PmemCSIDeployment reconcile started by the operator. Labeled by the deployment name.
```
pohly, line 1495: Recording a timestamp seems to be driven by the intention to use this for E2E tests. But I cannot imaging how it could be useful for anything else, like a dashboard showing a timeline.

Wouldn't it be more useful to just count events, like `pmem_csi_deployment_reconcile_started`, `pmem_csi_deployment_sub_resource_created` and `pmem_csi_deployment_sub_resource_updated`?

Those then can be plotted over time. The E2E test then of course needs to be updated, but it can still determine when the operator did some work, so it should be doable.

docs/install.md
```diff
@@ -1485,6 +1485,18 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+`pmem_csi_deployment_reconcile_ended_at` | gauge | Timestamp at which a PmemCSIDeployment reconcile ended by the operator. Labeled by the deployment name.
+`pmem_csi_deployment_sub_resource_created_at` | gauge | Timestamp at which a sub resource of the PmemCSIDeployment CR was created  by the operator. Labled by resource details ("name, "namespace",  "group", "version", "kind").
```
pohly, line 1497: s/Labled/Labeled/

docs/install.md
```diff
@@ -1485,6 +1485,18 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+`pmem_csi_deployment_sub_resource_updated_at` | gauge | Timestamp at which a sub resource of the PmemCSIDeployment CR was updated by the operator. Labled by resource details ("name, "namespace",  "group", "version", "kind").
```
pohly, line 1498: s/Labled/Labeled/

deploy/operator/pmem-csi-operator.yaml
```diff
@@ -140,6 +140,20 @@ subjects:
...
+  namespace: pmem-csi
+spec:
+  ports:
+  - port: 8080
+    targetPort: 8080
+  selector:
+    app: pmem-csi-operator
+    name: pmem-csi-operator
+  type: NodePort
+---
```
pohly, line 156: We cannot assume that the 8080 node port is free. Why do you need a node port?

pohly (changes requested):

docs/install.md
avalluri: > Recording a timestamp seems to be driven by the intention to use this for E2E tests.

I was more referring to [operator-framework's CR metrics API](https://github.com/operator-framework/operator-lib/blob/main/handler/instrumented_enqueue_object.go) and tried to more align with that. The part of this API that was mostly driver by E2E test was `pmem_csi_deployment_sub_resource_updated_at` to address #943.

>

deploy/operator/pmem-csi-operator.yaml
avalluri: > Why do you need a node port?

Yes, it shouldn't be. I took this from the driver metrics service which is using NodePort. I will fix this.

docs/install.md

deploy/operator/pmem-csi-operator.yaml
avalluri: Done.

pohly:
Any idea where the `dialer failed: pods "17.0.4" not found` error in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-946/6/tests comes from?

avalluri:
> Any idea where the `dialer failed: pods "17.0.4" not found` error in https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/PR-946/6/tests comes from?

I guess that was with my changes to reading metrics. I tried to keep the metrics code to a single function that could be shared by both the driver and the operator tests. That code was broken with the recent change to use port-forward, where the driver metrics are read using the nodeport service, whereas the operator tests use the operator Pod.

docs/install.md
avalluri: Changed the PR such that now it records `pmem_csi_deployment_reconcile_{start,end}` counter vectors labeled with deployment name and `pmem_csi_deployment_sub_resource_{created,updated}_at`  gauge vectors with timestamps.

docs/install.md
```diff
@@ -1485,6 +1485,19 @@ deployment is in the `Failed` state, then one can look into the event(s) using
...
+#### Operator metrics data
+
+PMEM-CSI operator exposes below metrics data about active PmemCSIDeployment
+custom resources and it's sub-object in addition to the
+[metrics data provided by the controller-runtime](https://book-v1.book.kubebuilder.io/beyond_basics/controller_metrics.html):
+
+Name | Type | Explanation
+-----|------|------------
+`pmem_csi_deployment_reconcile_start` | counter | Counter that gets incremented on each time a PmemCSIDeployment CR enters a reconcile loop, labeled with the deployment name.
+`pmem_csi_deployment_reconcile_end` | counter | Counter that gets incremented on end of each reconcile loop of a PmemCSIDeployment CR, labeled with the deployment name.
```
pohly, line 1497: How useful is it to count function entry and exit?

What I had suggested on Monday was to count reconciliation, i.e. have a single `pmem_csi_deployment_reconcile` counter.

pkg/pmem-csi-operator/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func SetSubResourceUpdateMetric(obj client.Object) error {
+	if obj == nil {
+		return nil
+	}
+	return setGauge(SubResourceUpdatedAt, GetSubResourceLabels(obj))
+}
+
+func DeleteSubResourceMetric(obj client.Object) {
```
pohly, line 82: Why do we need to delete samples? 

Doesn't that introduce the risk that a metrics gatherer misses the information? Suppose a sub resource gets added and then deleted without any intermediate metrics gathering. The client will then not see the value for `created_at`.

pohly: I suppose if the number of label tuples is unbounded, metrics data could grow without any bounds. But for our purposes I find that unlikely.

test/e2e/deploy/deploy.go
```diff
@@ -95,6 +97,63 @@ func WaitForOLM(c *Cluster, namespace string) *v1.Pod {
...
+			}
+		}
+	}
+
+	return "", fmt.Errorf("no metrics port found (pod: %s)", pod.Name)
+}
+
+// GetMetricsPortForward retrieves metrics from given metricsURL
+// by Pod's container port forwarding. The pod name, namespace,
+// and port number must be embedded in metricsURL.
```
pohly, line 131: "embedded" how? It's not clear from this comment.

pohly:
Tests didn't pass: "unexpected sub-object updates by the operator".

test/e2e/operator/validate/validate.go
```diff
@@ -21,75 +21,181 @@ import (
...
-				return fmt.Errorf("timed out waiting for deployment, last error: %v", err)
+		updates, ok := mf["pmem_csi_deployment_sub_resource_updated_at"]
+		if !ok {
+			return nil
+		}
+
+		unExpectedList := []string{}
+		for _, m := range updates.GetMetric() {
+			lblMap := labelPairToMap(m.GetLabel())
+			if strings.Contains(lblMap["ownedBy"], string(deployment.UID)) && !isUpdateExpected(lblMap) {
```
pohly, line 168: Nit: why is this a substring search and not a direct comparison?

I'll merge the PR despite this, but it looks odd.

pohly (approved):

PR #948, opened 2021-05-03
test: update csi-test to 4.x

pohly:
The resource cleanup API and implementation changed, replacing
explicit register calls with wrapping of the calls which create
resources.

The new "target path removed" test depends on another remote operation
for checking paths.


pohly:
This is WIP because the new test is not in a regular release yet.

pohly:
See https://github.com/kubernetes-csi/csi-test/pull/338/

avalluri (approved):

PR #949, opened 2021-05-04
WIP: Namespace creation: use memory for map location

okartau:
This should reduce the overhead we lose on device side.


pohly:
We cannot take away RAM from the node for this, it will cause pod scheduling issues when the node unexpectedly has less RAM available for pods.

PR #955, opened 2021-05-19
remove testing node ports

pohly:
We either already stopped using them (metrics service) or can connect directly to pods now (socat).

The advantage is that deployments are more likely to succeed because these node ports do not need be available (again).


avalluri (approved):
Looks good to me

PR #956, opened 2021-05-20
nodeserver: Improve idempotency in ephemeral volume creation

okartau:
This is WIP as it needs a change on testing code side which fixes repeating of idempotency testing.
The current changes are result of improvements I made during debugging an issue
that occured thanks to kubelet repeating PublishVolume requests, which
seems happens due to some kubelet internal issue.

----
The check for existing file system before mkfs was made on
one call path only, but ephemeral took another path when creating
an inline volume. Due to 2 additional side effects: incomplete
idempotency testing, and issue in kubelet that caused repeated
publishVolume requests, it was exposed that nodeserver attempts
to mkfs a device with existing file system, which fails,
and there is no useful error information.
The use of dax=always allows comparing of mount flags to work
correctly, as otherwise there would be mismatch:
"no dax found in dax=always"
According to docs, -o dax is legacy stiyle which could become
deprecated in the future.

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -174,7 +174,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			return nil, err
 		}
 		srcPath = device.Path
-		mountFlags = append(mountFlags, "dax")
+		mountFlags = append(mountFlags, "dax=always")
 	} else {
 		// Validate parameters.
 		v, err := parameters.Parse(parameters.PersistentVolumeOrigin, req.GetVolumeContext())
@@ -526,7 +526,7 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	mountOptions = append(mountOptions, "dax")
+	mountOptions = append(mountOptions, "dax=always")
 
 	if err = ns.mount(device.Path, stagingtargetPath, mountOptions, false /* raw block */); err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
@@ -630,6 +630,19 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 		fsType = defaultFilesystem
 	}
 
+	// Check does devicepath already contain a filesystem?
+	existingFsType, err := determineFilesystemType(device.Path)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	if existingFsType != "" {
+		// Is existing filesystem type same as requested?
+		if existingFsType == fsType {
+			klog.V(4).Infof("Skip mkfs as %v file system already exists on %v", existingFsType, device.Path)
+			return nil
+		}
+		return status.Error(codes.AlreadyExists, "File system with different type exists")
+	}
 	cmd := ""
 	var args []string
 	// hard-code block size to 4k to avoid smaller values and trouble to dax mount option
@@ -646,15 +659,7 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 	}
 	output, err := pmemexec.RunCommand(cmd, args...)
 	if err != nil {
-		// If the filesystem is already mounted, then
-		// formatting it fails.  In that case we don't need to
-		// format and can ignore that error.  We could check
-		// for that ourselves in advance, but that would be
-		// extra code.
-		if strings.Contains(output, "contains a mounted filesystem") {
-			return nil
-		}
-		return fmt.Errorf("mkfs failed: %s", output)
+		return fmt.Errorf("mkfs failed: output:[%s] err:[%v]", output, err)
 	}
 
 	return nil
@@ -664,7 +669,7 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType 
 func (ns *nodeServer) mount(sourcePath, targetPath string, mountOptions []string, rawBlock bool) error {
 	notMnt, err := ns.mounter.IsLikelyNotMountPoint(targetPath)
 	if err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("failed to determain if '%s' is a valid mount point: %s", targetPath, err.Error())
+		return fmt.Errorf("failed to determine if '%s' is a valid mount point: %s", targetPath, err.Error())
 	}
 	if !notMnt {
 		return nil
@@ -798,7 +803,9 @@ func findMountFlags(flags []string, findIn []string) bool {
 		}
 		found := false
 		for _, fIn := range findIn {
-			if f == fIn {
+			if f == "dax=always" && fIn == "dax" ||
+				f == "dax" && fIn == "dax=always" ||
+				f == fIn {
 				found = true
 				break
 			}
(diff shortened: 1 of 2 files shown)


pohly:
> This is WIP as it needs a change on testing code side which fixes repeating of idempotency testing.

You can simply copy my commit into this PR.

okartau:
The change seems to work as planned as I see repeated:
```
nodeserver.go:641 Skip mkfs as ext4 file system already exists on /dev/ndbus0region0fsdax/fake-e-318fbab89545328222eec8ca23590e972be8d2fcf27908ac3316cdbe
nodeserver.go:233 NodePublishVolume[fake-ephemeral-volume-id]: Existing mountFlags:'[rw seclabel relatime dax=always]', fsTyp: 'ext4'
nodeserver.go:235 NodePublishVolume[fake-ephemeral-volume-id]: parameters match existing, return OK
```
Most recent run however failed, but seems for unrelated reason, I will re-trigger.

P.S. what is your opinion about code that tried to match text output. I commented it out with TODO marking now, is it ok to delete? I believe the check before mkfs coves this scenario already, and string match is not reliable.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -526,7 +526,7 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	mountOptions = append(mountOptions, "dax")
+	mountOptions = append(mountOptions, "dax=always")
```
pohly, line 529: Does that increase our minimum supported kernel version?

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -630,6 +630,20 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType
...
+	existingFsType, err := determineFilesystemType(device.Path)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+	if existingFsType != "" {
+		// Is existing filesystem type same as requested?
+		if existingFsType == fsType {
+			klog.V(4).Infof("Skip mkfs as %v file system already exists on %v", existingFsType, device.Path)
+			return nil
+		} else {
```
pohly, line 643: Go idiom is to not use else.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -651,10 +665,13 @@ func (ns *nodeServer) provisionDevice(device *pmdmanager.PmemDeviceInfo, fsType
...
-			return nil
-		}
-		return fmt.Errorf("mkfs failed: %s", output)
+		// TODO: REMOVE LATER
+		// comparing output does not work as wording is different, will never match.
+		// The added check above for existing file system works better.
+		//if strings.Contains(output, "contains a mounted filesystem") {
+		//	return nil
+		//}
+		return fmt.Errorf("mkfs failed: output:[%s] err:[%v]", output, err)
```
pohly, line 674: Remove the commented code.

pkg/pmem-csi-driver/nodeserver.go
okartau: deleted unnecessary else, comes in next push

pkg/pmem-csi-driver/nodeserver.go
okartau: That's a valid concern. If we do not want to change mount option, we have to loosen our options compare function to accept 'dax' matching to 'dax=always'. I was thinking about  that, too. It could be done, for example either writing special handling for dax=always, or trying to be more generic and strip always part with equal sign so that we treat option itself as significant part and value as stripped part

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -798,6 +803,10 @@ func findMountFlags(flags []string, findIn []string) bool {
 		}
 		found := false
 		for _, fIn := range findIn {
+			if f == "dax" && fIn == "dax=always" {
```
okartau, line 806: In current version, instead of adding dax=always, I reverted to adding plain dax, and this check. This is more conservative, however will fail if dax gets deprecated, but that is trouble for the more distant future

okartau:
updated commit msg to reflect dax=always change correctly

pkg/pmem-csi-driver/nodeserver.go
pohly: > This is more conservative, however will fail if dax gets deprecated, but that is trouble for the more distant future

Where exactly is it deprecated? In the "mount" command or in the Linux kernel? We control the former, so as soon as we update the mount util and it breaks, we can change our code without affecting users.

okartau: Possible future deprecation is mentioned in [kernel doc](https://www.kernel.org/doc/Documentation/filesystems/dax.txt) which means the plan is probably about kernel. mount command has then two possibilities, either remove -o dax support or make it permanent alias to -o dax=always

pohly: My understanding is the opposite: it describes ""-o dax", which is a mount command line option.

pohly: The kernel doesn't use "dax" or "dax=always", so this has to be the command line tool.

okartau: sure, we use mount command API, so we depend what is supported by it. Just that the page I mentioned, explains same concept using kernel flags, and I understand they mention deprecating certain use of flags. So it is probably combined change of mount and kernel which will change it. In all ways, we have to react if something stop working. Our current proposal here is backwards-looking, keeping the syntax which does work with older systems.

pohly: > Our current proposal here is backwards-looking, keeping the syntax which does work with older systems.

That's my point: how we call the mount command depends on the content of our container, not the system we run on. We can use "dax=always" already now, as our "mount" command understands it.

Where we need to be careful is in what we get back from the kernel. That may show "dax" (old kernel) or "dax=always" (newer kernel), so we have to accept both.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -798,6 +803,11 @@ func findMountFlags(flags []string, findIn []string) bool {
 		}
 		found := false
 		for _, fIn := range findIn {
+			if f == "dax=always" && fIn == "dax" {
```
pohly, line 806: There's no need to have two if checks:
```
if f == "dax=always" && fIn == "dax" ||
  f == "dax" && fIn == "dax=always" ||
  f == fIn {
   found = true
   break
}
```

Note that I added the case for "we use dax" and the kernel reports "dax=always". I don't know whether we need it, but it feels safer to have it.

pohly (changes requested):

okartau:
The latest [run failed](https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-956/7/testReport/junit/fedora-1_19.operator/API/Testing___1_19___updating_all_in_default_deployment_while_running/) with unrelated error, "operator.API.updating all in default deployment while running", any ideas?
I had rebased hoping recent commit helps with  those, but seems not

pohly:
> The latest [run failed](https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/view/change-requests/job/PR-956/7/testReport/junit/fedora-1_19.operator/API/Testing___1_19___updating_all_in_default_deployment_while_running/) with unrelated error, "operator.API.updating all in default deployment while running", any ideas?

@avalluri: ^^^

> I had rebased hoping recent commit helps with those, but seems not

Testing happens against a merge with the target branch. It is not necessary to rebase to pick up changes in that branch.

pohly (approved):

PR #968, opened 2021-05-26
deploy: add 'app' label to operator deployment matchLabels

avalluri:
Recent change (d3ce705cdaf66e9b17d4166e8da9ef7f8ebbe311) in devel added
a new label to selector.matchLabels list, that breaks to upgrade the
v0.9.1 operator to devel due to mismatch in label selector.

This change backports that new missing label.

### DIFF
--- a/deploy/bindata_generated.go
+++ b/deploy/bindata_generated.go
@@ -103,7 +103,7 @@ func deployKubernetes117DirectPmemCsiYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "deploy/kubernetes-1.17/direct/pmem-csi.yaml", size: 13025, mode: os.FileMode(420), modTime: time.Unix(1619608498, 0)}
+	info := bindataFileInfo{name: "deploy/kubernetes-1.17/direct/pmem-csi.yaml", size: 13025, mode: os.FileMode(436), modTime: time.Unix(1622056533, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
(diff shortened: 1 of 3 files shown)


deploy/bindata_generated.go
```diff
@@ -103,7 +103,7 @@ func deployKubernetes117DirectPmemCsiYaml() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "deploy/kubernetes-1.17/direct/pmem-csi.yaml", size: 13025, mode: os.FileMode(420), modTime: time.Unix(1619608498, 0)}
+	info := bindataFileInfo{name: "deploy/kubernetes-1.17/direct/pmem-csi.yaml", size: 13025, mode: os.FileMode(436), modTime: time.Unix(1622056533, 0)}
```
pohly, line 106: Only the time stamps changed. You can remove the entire bindata_generated.go change from the PR.

pohly (commented):
How does that affect real-world upgrades of the operator?

Why was I not able to reproduce the failure locally?

pohly:
> How does that affect real-world upgrades of the operator?

More specifically, if someone has installed 0.9.1 with olm, can they upgrade to 0.9.2 with the new label?

This might be the wrong fix: instead adding the label on the "release-0.9" branch, we might have to remove in on "devel".

avalluri:
> > How does that affect real-world upgrades of the operator?
> 
> More specifically, if someone has installed 0.9.1 with olm, can they upgrade to 0.9.2 with the new label?

No, The upgrade using OLM will fail with the same error.

> 
> This might be the wrong fix: instead adding the label on the "release-0.9" branch, we might have to remove it on "devel".

Okay, I will submit that patch then.

PR #969, opened 2021-05-26
Jenkinsfile: fix logs file creation, do it always

okartau:
Due to misused shell redirection, kubelet log contained the
output from last worker only, and no cluster information.
Compress longer job logs, save kubelet logs in all cases,
means we do not need exception intercepting block any more.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -490,15 +490,13 @@ void TestInVM(worker, distro, distroVersion, kubernetesVersion, skipIfPR) {
                            testrun=\$(echo '${distro}-${distroVersion}-${kubernetesVersion}' | sed -e s/--*/-/g | tr . _ ) && \
                            make test_e2e TEST_E2E_REPORT_DIR=${WORKSPACE}/build/reports.tmp/\$testrun \
                                          TEST_E2E_SKIP=\$(if [ \"${env.CHANGE_ID}\" ] && [ \"${env.CHANGE_ID}\" != null ]; then echo \\\\[Slow\\\\]@${skipIfPR}; fi) \
-                           ' |tee joblog-${BUILD_TAG}-test-${kubernetesVersion}.log |egrep 'Passed|FAIL:|^ERROR' 2>&1 \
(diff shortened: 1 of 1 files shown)


okartau:
In  this change few points will affect working with logs:
- kubelet logs are saved in all cases (was: only failing jobs). Note, kubelet logs are catenated into single file in current change (I tried to extract last char from $cmd to be part of filename but did not succeed easily, so thats the current tradeoff).
- job and kubelet logs get compressed by gzip. This means huge space savings for jenkins server, a log file compresses more than 10x

Does anybody object compressing of log files? Does it affect your working pattern negatively?
I have recently settled to style (as getting a job log in browser window is very slow and unreliable) where I copy link in browser and wget that file for local checking. In my flow, I will need to add gunzip step. If one has habit to open log in browser directly, then the current change will be a blocker and we need to give up the idea of compressing.

pohly:
I view log files in Firefox.

Can't we get them served with a content type that indicates the compression? Then the browser can decompress on-the-fly and we save both space on the server and bandwidth.

okartau:
> Can't we get them served with a content type that indicates the compression?

I dont find that feature in quick search. There is however another jenkins plugins that we don't use, "compress artifacts". This will do the compressing part transparently, thus solving the space issue. But I think the file is presented to download step as regular-sized file then, so it does not save anything in download case. But use of this plugin would be improvement over current state, without creating changes to working patterns, so I think the right way is to revert compression commands in this PR, and install that compression plugin.

pohly:
> I think the right way is to revert compression commands in this PR, and install that compression plugin.

Agreed.

pohly:
Has the plugin been installed?

okartau:
> Has the plugin been installed?

I tried after seeing this msg. Installed plugin, noticed it also needs enabling in config, so I tried that, but got weird-looking error msg about some color, seems unrelated to what I am doing. The I recalled, we have that issue that config save may invalidate our builder setup. Not sure yet about that, will investigate the config state.

pohly (approved):

PR #970, opened 2021-05-27
deploy/operator: remove newly added app label from selector list

avalluri:
Adding this label breaks operator upgrades using OLM, as the previous
(v0.9.x) versions do not have this label that results in mismatched
label selector.

### DIFF
--- a/deploy/kustomize/operator/operator.yaml
+++ b/deploy/kustomize/operator/operator.yaml
@@ -119,7 +119,6 @@ metadata:
 spec:
   selector:
     name: pmem-csi-operator
-    app: pmem-csi-operator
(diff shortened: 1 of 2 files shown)


pohly:
How important was it to add this additional label?

It's unfortunate that we can't change it, but this looks like the right fix to me.

avalluri:
> How important was it to add this additional label?
> 
Not so important in my opinion. When I was writing metrics service, I found that the deployment templated spec is using both the app & name labels. So tempted to match the label selection to them. I was not expecting that to lead to the upgrade issues.

pohly (approved):

PR #975, opened 2021-06-01
Jenkinsfile: check for skipping 1.18 before spawning agent

okartau:
This causes creating 2 workers instead of 3 in PR CI jobs.


okartau:
hundreds of failures, connectivity issues, will re-trigger

Issue #971, opened 2021-05-28
test flake: invalid memory address in metrics test


pohly:
https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/devel/367/tests

```
/mnt/workspace/pmem-csi_devel/test/e2e/metrics/metrics.go:50
Test Panicked
/go/src/runtime/panic.go:212
Panic: runtime error: invalid memory address or nil pointer dereference
Full stack:
github.com/intel/pmem-csi/test/e2e/metrics.glob..func1.2.1()
	/mnt/workspace/pmem-csi_devel/test/e2e/metrics/metrics.go:81 +0x7f0
github.com/onsi/gomega.InterceptGomegaFailures(0xc004d0e090, 0xc00373bc10, 0x0, 0xc002d66000)
	/mnt/workspace/gopath/pkg/mod/github.com/onsi/gomega@v1.10.5/gomega_dsl.go:107 +0xbb
github.com/intel/pmem-csi/test/e2e/metrics.glob..func1.2.2(0x0, 0x0)
	/mnt/workspace/pmem-csi_devel/test/e2e/metrics/metrics.go:106 +0x2a
reflect.Value.call(0x1af1fa0, 0xc00373bc00, 0x13, 0x1de44d6, 0x4, 0xc0011c0f10, 0x0, 0x0, 0x1af1fa0, 0x1, ...)
	/go/src/reflect/value.go:476 +0x8e7
reflect.Value.Call(0x1af1fa0, 0xc00373bc00, 0x13, 0xc0011c0f10, 0x0, 0x0, 0x11a90db8bc47, 0xc00005e6d8, 0xc00005c000)
	/go/src/reflect/value.go:337 +0xb9
github.com/onsi/gomega/internal/asyncassertion.(*AsyncAssertion).pollActual(0xc0030b4980, 0xc001a48000, 0x2e20fe0, 0x78c782, 0x2e52b10)
	/mnt/workspace/gopath/pkg/mod/github.com/onsi/gomega@v1.10.5/internal/asyncassertion/async_assertion.go:78 +0xee
github.com/onsi/gomega/internal/asyncassertion.(*AsyncAssertion).match(0xc0030b4980, 0x21064b0, 0x2e52b10, 0x3b9aca01, 0x0, 0x0, 0x0, 0x989680)
[...]

PR #984, opened 2021-06-07
test: nil pointer in metrics test, II

pohly:
Apparently resp (?!) can be nil even when there was no error code.

Panic: runtime error: invalid memory address or nil pointer dereference
Full stack:
github.com/intel/pmem-csi/test/e2e/metrics.glob..func1.2.1()
	/mnt/workspace/pmem-csi_PR-977/test/e2e/metrics/metrics.go:81 +0x7f0
github.com/onsi/gomega.InterceptGomegaFailures(0xc0052d4ab0, 0xc0014b2660, 0x0, 0xc002627800)
	/mnt/workspace/gopath/pkg/mod/github.com/onsi/gomega@v1.10.5/gomega_dsl.go:107
        +0xbb

metrics.go:81 now had:

     Expect(resp.Body).NotTo(BeNil(), "have response body")

Fixes: https://github.com/intel/pmem-csi/issues/971

### DIFF
--- a/test/e2e/metrics/metrics.go
+++ b/test/e2e/metrics/metrics.go
@@ -78,7 +78,13 @@ var _ = deploy.Describe("direct-testing", "direct-testing-metrics", "", func(d *
 								pod.Namespace, pod.Name, port.ContainerPort)
 							resp, err := client.Get(url)
 							framework.ExpectNoError(err, "GET failed")
-							Expect(resp.Body).NotTo(BeNil(), "have response body")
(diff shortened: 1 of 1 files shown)


pohly:
This goes against what `go doc net/http.Client.Get` says:
```
    When err is nil, resp always contains a non-nil resp.Body. Caller should
    close resp.Body when done reading from it.
```

avalluri (approved):

PR #994, opened 2021-06-11
test: force image pull

pohly:
After "make push" we want to be sure that a running cluster actually
uses the new images. Our default image pull policy is "IfNotPresent",
so it won't pull each time a Pod starts. We have to use some other
mechanism.

The new test/pull-images.sh does that by definining a DaemonSet that
uses both kinds of PMEM-CSI images and has image pull policy
"Always". By changing the annotation of the pods, we force a pod
restart on each node each time the script gets invoked.

### DIFF
--- a/test/pull-images.sh
+++ b/test/pull-images.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+set -o errexit
+
+TEST_DIRECTORY=${TEST_DIRECTORY:-$(dirname $(readlink -f $0))}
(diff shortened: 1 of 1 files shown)


pohly:
@okartau Nothing to test here, we can merge once you have reviewed this.

okartau:
This is meant to be invoked manually?

pohly:
Yes.

okartau (approved):

Issue #935, opened 2021-04-15
OpenShift: permissions


pohly:
It looks like privileged containers need special permissions on OpenShift. Here's an example: https://github.com/democratic-csi/charts/commit/bfdf4b174b76c191f8057fba40394122ac3ed72a

krisiasty:
Right, I found I had to assign "privileged" security context constraint (SCC) to the *-webhooks and *-controller service accounts in the pmem-csi namespace before creating PmemCSIDeployment. You can do this manually like in the example below, but it would be much better if that could be handled automatically by the operator:
```
# oc adm policy add-scc-to-user privileged -z pmem-csi-intel-com-controller -n pmem-csi
clusterrole.rbac.authorization.k8s.io/system:openshift:scc:privileged added: "pmem-csi-intel-com-controller"

# oc  adm policy add-scc-to-user privileged -z pmem-csi-intel-com-webhooks -n pmem-csi
clusterrole.rbac.authorization.k8s.io/system:openshift:scc:privileged added: "pmem-csi-intel-com-webhooks"
```

pohly:
Why does the webhooks service account need this? Its container doesn't need to run as privileged container.

In other words, what was failing when you didn't add that?

krisiasty:
The webhooks service account is used by the controller pod with the following spec (fragment):
```
  securityContext:
    runAsNonRoot: true
    runAsUser: 1000
  serviceAccount: pmem-csi-intel-com-webhooks
  serviceAccountName: pmem-csi-intel-com-webhooks
```

I think the problem here is the line `runAsUser: 1000`. The UID 1000 is out of allowed UID range in the OpenShift:
```
# oc describe project pmem-csi
Name:			pmem-csi
Created:		2 weeks ago
Labels:			<none>
Annotations:		kubectl.kubernetes.io/last-applied-configuration={"apiVersion":"v1","kind":"Namespace","metadata":{"annotations":{},"name":"pmem-csi"}}

			openshift.io/sa.scc.mcs=s0:c24,c19
			openshift.io/sa.scc.supplemental-groups=1000590000/10000
			openshift.io/sa.scc.uid-range=1000590000/10000
```
[...]

PR #996, opened 2021-06-11
OpenShift permissions

pohly:
This consists of two changes:
- running the controller without a fixed UID to allow OpenShift to assign a UID dynamically (see commit for a detailed discussion)
- adding one additional role binding for the node driver

Fixes: https://github.com/intel/pmem-csi/issues/935

### DIFF
--- a/deploy/kustomize/driver/pmem-csi.yaml
+++ b/deploy/kustomize/driver/pmem-csi.yaml
@@ -147,10 +147,6 @@ spec:
         app.kubernetes.io/instance: pmem-csi.intel.com
         pmem-csi.intel.com/webhook: ignore
     spec:
-      securityContext:
-        runAsNonRoot: true
-        # UID 1000 is defined in Dockerfile
-        runAsUser: 1000
       serviceAccountName: pmem-csi-intel-com-webhooks
       # Allow this pod to run on all nodes and
       # prevent eviction (https://github.com/kubernetes-csi/csi-driver-host-path/issues/47#issuecomment-538469081).
--- a/deploy/kustomize/driver/pmem-csi.yaml
+++ b/deploy/kustomize/driver/pmem-csi.yaml
@@ -4,6 +4,26 @@ metadata:
   name: pmem-csi-intel-com-controller
   namespace: default
 ---
+# This role binding corresponds to what the following command creates:
+# oc adm policy add-scc-to-user privileged -z pmem-csi-intel-com-controller
+#
+# One difference is that we don't use the name system:openshift:scc:privileged
+# for the object. That object is managed by oc and the admin, whereas this
+# one here is under control of PMEM-CSI (YAML or operator).
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: pmem-csi-intel-com-node-openshift-cfg
+  namespace: default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:openshift:scc:privileged
+subjects:
+- kind: ServiceAccount
+  name: pmem-csi-intel-com-controller
+  namespace: default
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/deploy/kustomize/driver/pmem-csi.yaml
+++ b/deploy/kustomize/driver/pmem-csi.yaml
@@ -7,16 +7,12 @@ metadata:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  labels:
-    pmem-csi.intel.com/deployment: lvm-testing
   name: pmem-csi-intel-com-webhooks
   namespace: default
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  labels:
-    pmem-csi.intel.com/deployment: lvm-testing
   name: pmem-csi-intel-com-webhooks-cfg
   namespace: default
 rules:
@@ -32,8 +28,6 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  labels:
-    pmem-csi.intel.com/deployment: lvm-testing
   name: pmem-csi-intel-com-webhooks-runner
 rules:
 - apiGroups:
@@ -79,8 +73,6 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  labels:
-    pmem-csi.intel.com/deployment: lvm-testing
   name: pmem-csi-intel-com-webhooks-role-cfg
   namespace: default
 roleRef:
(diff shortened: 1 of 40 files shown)


deploy/common/kata-app.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Pod
```
avalluri, line 1: @pohly Is this change related to this commit?

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -147,10 +147,6 @@ spec:
         app.kubernetes.io/instance: pmem-csi.intel.com
         pmem-csi.intel.com/webhook: ignore
     spec:
-      securityContext:
-        runAsNonRoot: true
```
avalluri, line 151: Does running the controller as non-root(though not choosing the UID) is still not relevant?

deploy/olm-catalog/0.9.0/pmem-csi-operator_v1_serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
avalluri, line 1: I think this is not intentional to add files under 'deploy/olm-catalog', is it?

deploy/kustomize/driver/pmem-csi.yaml
```diff
@@ -4,6 +4,26 @@ metadata:
   name: pmem-csi-intel-com-controller
   namespace: default
 ---
+# This role binding corresponds to what the following command creates:
```
avalluri, line 7: Isn't it good to keep these kinds of changes in a kustomize layer specific to openshift?

avalluri (commented):
> On non-OpenShift systems the additional RoleBinding is ignored,

But it might raise unnecessary confusion for the cluster-admin.

deploy/common/kata-app.yaml
pohly: No, will remove.

deploy/kustomize/driver/pmem-csi.yaml
pohly: Have you seen the commit message? There I am trying to discuss that aspect.

In a nutshell, actual users will still run this as non-root.

deploy/olm-catalog/0.9.0/pmem-csi-operator_v1_serviceaccount.yaml
pohly: Not intentional. `.gitignore` should have been updating when adding that directory to the built files.

deploy/kustomize/driver/pmem-csi.yaml
pohly: That would imply that we maintain different deployments and deployment instructions for upstream Kubernetes and OpenShift, with the resulting user errors when they pick the wrong one.

I'm trying to avoid that.

pohly:
> > On non-OpenShift systems the additional RoleBinding is ignored,
> But it might raise unnecessary confusion for the cluster-admin.

Only if they run on upstream Kubernetes *and* inspect role bindings. I think this is very unlikely to be a problem.

deploy/olm-catalog/0.9.0/pmem-csi-operator_v1_serviceaccount.yaml
pohly: Ah, it was the other way around: I still had the directory, but 435f821906c055b8dfd6a918ba36dc3e2137dcea removed the exclude rule.

deploy/kustomize/driver/pmem-csi.yaml
pohly: docs/design.md also gets updated to explain this.

deploy/kustomize/testing/controller-coverage-patch.yaml
```diff
@@ -14,23 +14,3 @@
     hostPath:
       path: /var/lib/pmem-csi-coverage
       type: DirectoryOrCreate
-# driver in controller mode runs as non-root user
-# which cannot have enough permissions to access
-# coverage-dir(/var/lib/pmem-csi-coverage)
-# Change the ownership of this folder to "pmem-csi:pmem-csi"
-# so that driver could write coverage reports.
-- op: add
-  path: /spec/template/spec/initContainers
```
avalluri, line 23: Seems like this change(removing init container) is not related to this commit?

deploy/kustomize/driver/pmem-csi.yaml
avalluri: My point is can't we keep `runAsNotRoot: true` by not setting the UID?

deploy/kustomize/testing/controller-coverage-patch.yaml
pohly: It is. Because we now run as root (at least where we test), we don't need this workaround for non-root anymore.

deploy/kustomize/driver/pmem-csi.yaml
pohly: But on upstream Kubernetes we do run as root. We can't keep `runAsNotRoot: true` unless we set the fixed UID, can we?

pohly:
@avalluri did I address your comments?

pohly:
Also note the new final commit `deploy: test with "testing" deployments instead of "coverage"`. That was what I found when trying out this PR on an OpenShift 4.6 cluster. The Pods didn't come up because image pulling in that cluster is impossible (access to Internet not configured), but at least the Pods all got created.

avalluri (approved):

pohly:
I have rebased and tested the rebased deployment on an OpenShift 4.8 (in development!) cluster that was created with libvirt such that the worker node has 896.00 MiB of PMEM.

```
test/setup-ca-kubernetes.sh
sed -e 's;intel/pmem-csi-driver:canary;pohly/pmem-csi-driver:canary;' deploy/kubernetes-1.21/pmem-csi-direct.yaml | kubectl apply -f -
kubectl label node/localhost storage=pmem
```

The "localhost" name is a quirk of the cluster - somehow the worker node lost its real name.

All pods come up and PMEM-CSI is able to find the PMEM:
```
I0623 15:07:43.865223       1 main.go:73] "PMEM-CSI started." version="v0.9.0-197-g157b2c818"
I0623 15:07:43.865516       1 mount_linux.go:163] Detected OS without systemd
I0623 15:07:43.867050       1 mount_linux.go:163] Detected OS without systemd
I0623 15:07:43.867720       1 server.go:54] "GRPC-server: Listening for connections" endpoint="unix:///csi/csi.sock"
I0623 15:07:43.868426       1 pmem-csi-driver.go:305] "PMEM-CSI ready." capacity="896Mi maximum volume size, 896Mi available, 896Mi managed, 896Mi total"
```

There was an issue with creating a 100Mi namespace in direct mode, but LVM mode worked. That's unrelated to this PR, so I'll merge it.

PR #997, opened 2021-06-12
operator: set the generated bundle channel name to 'stable'

### LINKED ISSUE #960: Operator Hub: next stability level
We are listed with an "alpha" channel and without the "Full Lifecycle" capability. Depending on the definition of the later, we should support it.

We also should replace "alpha". What's the next level? "beta"?

avalluri:
FIXES #960

### DIFF
--- a/operator/operator.make
+++ b/operator/operator.make
@@ -46,6 +46,12 @@ MANIFESTS_DIR=$(REPO_ROOT)/deploy/kustomize/olm-catalog
 CATALOG_DIR=$(REPO_ROOT)/deploy/olm-catalog
 BUNDLE_DIR=$(REPO_ROOT)/deploy/olm-bundle/$(MAJOR_MINOR_PATCH_VERSION)
 CRD_DIR=$(REPO_ROOT)/deploy/crd
+# CHANNELS: A comma-separated list of channels the generated
+# bundle belongs to. This supposed to be all the channel names we support.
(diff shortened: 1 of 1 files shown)


operator/operator.make
```diff
@@ -46,6 +46,11 @@ MANIFESTS_DIR=$(REPO_ROOT)/deploy/kustomize/olm-catalog
 CATALOG_DIR=$(REPO_ROOT)/deploy/olm-catalog
 BUNDLE_DIR=$(REPO_ROOT)/deploy/olm-bundle/$(MAJOR_MINOR_PATCH_VERSION)
 CRD_DIR=$(REPO_ROOT)/deploy/crd
+# CHANNELS: A comma-separated list of channels the generated
+# bundle belongs to.
+CHANNELS?=stable
```
pohly, line 51: Should this include `alpha`?

Please include instructions when to use more than one channel or link to further information.

avalluri: > Should this include `alpha`?

Maybe yes. I myself not sure from where the OLM gets the channels in the bundle format, because these are defined in annotations inside the version-specific folders. There is no common file to define all available channels, where it used to be in package format. Anyways I will add the 'alpha' to the channel list.

pohly: Please add it and I'll merge.

avalluri: Done.

pohly (approved):

PR #999, opened 2021-06-14
build: replace go-bindata with Go 1.16 embed

pohly:
This makes our build simpler because we don't need an extra make rule
and don't need to commit a generated file into our source repo (less
merge conflicts!).

Fixes: https://github.com/intel/pmem-csi/issues/998

PR #1000, opened 2021-06-14
Probes

pohly:
Liveness probes serve as a fallback mechanisms for cases where the
container gets stuck but fails to terminate. We piggy-back onto the
metrics endpoint with HTTP probes because that endpoint is readily
available and, in the case of the PMEM-CSI node driver, already does
some work that mirrors what the driver needs to do during
operation (interact with underlying storage to retrieve capacity).
    
A startup probe is used because the node driver might start up more
slowly when setting up LVM. We can also add that for direct mode for
the sake of simplicity.
    
The liveness probes could be useful, although in practice the node
driver should already terminate if it runs into fatal issues, so we
probably won't need to rely on the probe. To avoid potential false
positives (killing the container although it only had a temporary
outage), the threshold is 5.
    
Readiness probes make less sense. The node driver does not do any load
balancing. It might make sense if we had more than one instance and
let the scheduler access only ready instances, but right now we only
have one instance.


pohly:
@avalluri this builds on top of PR #999  to avoid merge conflicts.

avalluri:
Looks good to me. But there are some test failures.

pohly:
> Looks good to me. But there are some test failures.

I debugged that this morning and pushed a new commit which helps get containers into the ready state faster.

pohly:
@avalluri okay now? The revised liveness settings seem to have fixed the test failures.

avalluri:
> @avalluri okay now? The revised liveness settings seem to have fixed the test failures.

Yes. looks good.

PR #1003, opened 2021-06-15
docs links check

pohly:
We want to check links to detect stale external links and anchors that we didn't get right or that changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ Manifest
 /_build
 /deploy/olm-bundle
 /deploy/kustomize/olm-catalog/crd
+
+*.pyc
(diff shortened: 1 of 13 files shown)


okartau:
one commit msg has typo:
`skips URLs that only consist of an anchorm`

pohly:
> one commit msg has typo:
> skips URLs that only consist of an anchorm

Fixed. I put that in to keep reviewers on their toes and to check whether they read the commit messages :grin:

pohly:
@okartau is it okay now?

okartau:
yes. I think too it is good idea to get rid of TOC

Issue #1004, opened 2021-06-17
test flake: switch device mode - pods (not?) restarted


pohly:
It's not clear from the assertion whether it expects a restart or no restart:
```
Pod restart 'direct-to-lvm-use-volume-in-different-mode-node-5p65f'
Expected
    <bool>: false
to be true
/mnt/workspace/pmem-csi_devel/test/e2e/operator/deployment_api.go:1173
```

An `if (...) framework.Failf()` is clearer.

But the actual issue of course is that this test flake now occurs. It started with the liveness probe PR (https://github.com/intel/pmem-csi/pull/1000) and occurred again on devel after merging that PR (https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/blue/organizations/jenkins/pmem-csi/detail/devel/388/pipeline).

pohly:
> An if (...) framework.Failf() is clearer.

It can't be used here because we check the result of an `Eventually`. "Restarted" is still confusing because what it really waits for here is the deletion of the pod.

pohly:
Our DaemonSet uses the defaults of `.spec.updateStrategy.type: RollingUpdate` with `.spec.updateStrategy.rollingUpdate.maxUnavailable: 1`. This means it will restart one pod at a time, which makes the total duration proportional to the number of nodes. We need to consider this when specifying the timeout (= have one timeout for all pods, proportional to the number of pods) or make sure that all pods get restarted at the same time (= `maxUnavailable: <number of nodes>`.

As it should make our testing run faster, I prefer the latter approach.

PR #1005, opened 2021-06-17
operator rollout strategy

pohly:
This speeds up testing and avoids a test flake by restarting driver pods in parallel.

Fixes: https://github.com/intel/pmem-csi/issues/1004


pohly:
One can see that this works both because the test runs faster (108s instead of 195s) and because the logging shows that all pods get the TERM signal at roughly the same time:
```
direct-to-lvm-use-volume-in-same-mode-node-hlclf/pmem-driver@pmem..ker3: I0617 10:01:00.176427       1 pmem-csi-driver.go:343] Caught signal terminated, terminating.
direct-to-lvm-use-volume-in-same-mode-node-cfbq6/pmem-driver@pmem..ker1: I0617 10:01:00.251572       1 pmem-csi-driver.go:343] Caught signal terminated, terminating.
direct-to-lvm-use-volume-in-same-mode-node-f6j4n/pmem-driver@pmem..ker2: I0617 10:01:00.419268       1 pmem-csi-driver.go:343] Caught signal terminated, terminating.
```

avalluri (approved):

PR #1006, opened 2021-06-17
volume handling cleanup and fixes

pohly:
NodeUnstageVolume did not return the correct error code anymore when the volume didn't exist. This was found while working on a new test where volumes get deleted before running the normal cleanup code.

The actual goal was to have a test for space fragmentation and checking that the maximum volume size is handled correctly. While working on that, I ended up addressing various todos.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/aws/aws-sdk-go v1.35.37 // indirect
-	github.com/container-storage-interface/spec v1.3.0
+	github.com/container-storage-interface/spec v1.5.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v0.4.0
 	github.com/google/go-cmp v0.5.5
@@ -24,6 +24,7 @@ require (
 	golang.org/x/sys v0.0.0-20210603081109-ebe580a85c40
 	google.golang.org/genproto v0.0.0-20210207032614-bba0dbe2a9ea // indirect
 	google.golang.org/grpc v1.35.0
+	google.golang.org/protobuf v1.26.0
 	gopkg.in/freddierice/go-losetup.v1 v1.0.0-20170407175016-fc9adea44124
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.21.1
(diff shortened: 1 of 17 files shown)


okartau:
looks like delicate balance in alignment land is shaking again:
```
/mnt/workspace/pmem-csi_PR-1006/test/e2e/storage/sanity.go:773
Jun 17 20:53:27.453: additional volume size on node pmem-csi-govm-worker1
Expected
    <int64>: 37748736
to equal
    <int64>: 34603008
```

probably related to changes on LVM side which print this?
`pmd-lvm.go:152] CreateDevice increased size from 33Mi to 36Mi to satisfy LVM alignment of 4Mi`

pohly:
> looks like delicate balance in alignment land is shaking again

The test was broken and did not handle volumes that were larger than requested. I think I have addressed all test failures, can you have another look?

See https://github.com/intel/pmem-csi/compare/3b36a3f1ab8909a5862734fe567a67e04caa27bb..1a9af657de3391fbaf0f63442fdcee73b93249fb for the latest changes.

okartau:
align and size fields have mostly/all? been uint64, now there are numerous int64 values introduced. This is planned and not possibly creating any trouble? size and align values hardly have any meaning on negative side, but can cause some unwanted effect, can  they...

pohly:
> align and size fields have mostly/all? been uint64, now there are numerous int64 values introduced.

int64 is what `resource.Quantity` uses. I think there is a preference to use signed types in Go even when the expected value range is just for positive numbers. For example, `os.FileInfo.GetSize` returns `int64`, not `uint64`. The advantage is that arithmetic is less likely to wrap around. A negative value as result of substraction might not make sense, but at least it is obviously wrong instead of turning into a large positive value. I was tempted to change pkg/ndctl, but that would have been a too large change.

> This is planned and not possibly creating any trouble?

The additional type casts are annoying, but I can't imagine why they should cause trouble. Only valid sizes get cast, and those should all fit into int64.

pohly:
> size and align values hardly have any meaning on negative side, but can cause some unwanted effect, can they...

Like what?

Modulo and division are the same for int64 as long as the value is the same. But that's moot because those calculations still use uint64.

okartau (approved):

PR #1011, opened 2021-06-22
pmem-csi-driver: structured, contextual logging

pohly:
All log messages from PMEM-CSI itself are produced with log message
plus key/value pairs. Loggers accumulate a prefix and additional
key/value pairs as they are passed down the call chain.

All gRPC calls have the method name as prefix and incrementing request
counter, to distinguish between different requests for the same
operation.

Fixes: #825


avalluri (approved):
Looks good to me.

PR #1015, opened 2021-06-23
pmem-csi-driver: revert to mounting with "-o dax"

pohly:
"dax=always" gets passed verbatim to the kernel by the mount
command. Older kernels like 5.4 then fail with "wrong fs type, bad
option, bad superblock on /dev/pmem1.1, missing codepage or helper
program, or other error."

Although "-o dax" is said to be a legacy option that may get removed,
in practice the risk of that happening is low, so we can revert to
using it without any fallbacks. It's functionally the same.


pohly:
This was tested on the failing system by Julio.

Issue #1029, opened 2021-07-17
v1.0.0: olm upgrade test failed


pohly:
After tagging v1.0.0, the "olm upgrade" test caused following olm tests to fail because it removed the "deploy/olm-bundle/1.0.0" directory that is needed for other tests. From https://cloudnative-k8sci.southcentralus.cloudapp.azure.com/job/pmem-csi/job/v1.0.0/1/artifact/joblog-jenkins-pmem-csi-v1.0.0-1-test-1.19.log:
```
pmem-csi-operator-55584cb8b9-p4r6w/pmem-csi-operator@pmem..ker1: I0717 08:11:51.070825       1 reflector.go:255] Listing and watching *v1.RoleBinding from sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:241
[1mSTEP[0m: Operator is ready!
I0717 08:11:51.107474     724 exec.go:34] "Starting command" command="/bin/rm" args=[rm -rf /mnt/workspace/pmem-csi_v1.0.0/deploy/olm-bundle/1.0.0]
I0717 08:11:51.108784     724 exec.go:52] "Command terminated" command="/bin/rm" stdout-len=0 combined-len=0 error=<nil>
Jul 17 08:11:51.108: INFO: upgrade: Operator changed to: 0.9
Jul 17 08:11:51.108: INFO: Ensuring that the CRs got reconciled
...
I0717 08:13:46.273436     724 exec.go:34] "EnsureDeployment: Starting command" command="test/start-operator.sh" args=[test/start-operator.sh -olm]
I0717 08:13:46.285487     724 exec.go:72] "EnsureDeployment/stderr: ++ ls -1t /mnt/workspace/pmem-csi_v1.0.0/deploy/olm-bundle" command="test/start-operator.sh"
I0717 08:13:46.285530     724 exec.go:72] "EnsureDeployment/stderr: ++ head -1" command="test/start-operator.sh"
I0717 08:13:46.287185     724 exec.go:72] "EnsureDeployment/stderr: + BUNDLE_VERSION=" command="test/start-operator.sh"
I0717 08:13:46.287216     724 exec.go:72] "EnsureDeployment/stderr: + echo 'Starting the operator using OLM....'" command="test/start-operator.sh"
I0717 08:13:46.287233     724 exec.go:72] "EnsureDeployment/stdout: Starting the operator using OLM...." command="test/start-operator.sh"
[...]

PR #1034, opened 2021-08-02
test: bump up fictional future version number for OLM upgrade test

pohly:
Since releasing 1.0.0, the "current = 1.0" version number was not high
enough anymore with the effect that the OLM upgrade test didn't
actually upgrade and after a test run removed the files under deploy
which were still needed.


pohly:
This is a stop-gap fix for https://github.com/intel/pmem-csi/issues/1029. Testing "devel" and thus updating "master" is currently broken because of that issue.

A more thorough solution which introduces upgrade testing from 0.9 can follow later.

pohly:
Full branch testing was successful, therefore I am merging this.

PR #1035, opened 2021-08-03
test: skew testing against 1.0, consolidate testing

pohly:
Skew testing for the operator used special code because the bundle
format was new and not supported by previous releases at the time of
writing the test. Now we can test against 1.0 and use the normal skew
testing suite also for OLM.

Deployments through the operator were tested with a subset of the full
test suites. As operator and OLM are now more important, it seems
prudent to test those deployments more throroughly. If that turns out
to increase the overall test time too much, then we might remove some
of the other deployment cases, like lvm-production and
direct-production.

Multiple different framework.Framework instances were create before
some of the tests, which created unnecessary work and namespaces. Now
they share the same instance.


pohly:
@avalluri: overall test times are IMHO acceptable now (7 hours for PR testing, 10:30 for branch) and we have better coverage.

Shall we merge?

test/e2e/deploy/deploy.go
```diff
@@ -915,6 +932,23 @@ func findDriver(c *Cluster) (*Deployment, error) {
...
+	for i := 0; ; i++ {
+		deployment, err := findOperatorOnce(c)
+		if err == nil {
+			return deployment, nil
+		}
+		if i > 6 {
+			return nil, err
+		}
+		framework.Logf("finding operator failed during attempt #%d, will retry: %v", err)
+		time.Sleep(10 * time.Second)
```
avalluri, line 947: Sounds like 10 seconds retry timeout might result in a delay. I think we could retry after 1 or 2 seconds.

avalluri (approved):
Looks good to me.

test/e2e/deploy/deploy.go
pohly: Changed to one second.

PR #1041, opened 2021-08-31
fix various linter warnings

pohly:
Mostly they are around inefficient initialization, missing error checking, deprecated APIs, and
some unreachable code.

In several cases, error handling is intentionally omitted, either because the
error cannot occur or cannot be handled. This can be expressed by assigning the
error to _.

### DIFF
--- a/cmd/pmem-csi-driver/main.go
+++ b/cmd/pmem-csi-driver/main.go
@@ -10,7 +10,7 @@ package main
 import (
 	"os"
 
-	"github.com/intel/pmem-csi/pkg/pmem-csi-driver"
+	pmemcsidriver "github.com/intel/pmem-csi/pkg/pmem-csi-driver"
 )
 
 func main() {
--- a/cmd/pmem-csi-driver/main_test.go
+++ b/cmd/pmem-csi-driver/main_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	"github.com/intel/pmem-csi/pkg/coverage"
-	"github.com/intel/pmem-csi/pkg/pmem-csi-driver"
+	pmemcsidriver "github.com/intel/pmem-csi/pkg/pmem-csi-driver"
 )
 
 func TestMain(t *testing.T) {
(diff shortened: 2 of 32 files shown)


pohly:
I've fixed everything reported by `golangci-lint run` in its default configuration except for:
```
pkg/ndctl/region.go:229:5: SA4023: this comparison is never true (staticcheck)
	if ns == nil {
	   ^
pkg/ndctl/region.go:228:8: SA4023(related information): the lhs of the comparison is the 1st return value of this function call (staticcheck)
	ns := r.SeedNamespace()
	      ^
pkg/ndctl/region.go:145:18: SA4023(related information): (*github.com/intel/pmem-csi/pkg/ndctl._Ctype_struct_ndctl_region).SeedNamespace never returns a nil interface value (staticcheck)
func (r *region) SeedNamespace() Namespace {
                 ^
```

This static check is false. `ndctl_region_get_namespace_seed` can return NULL and thus SeedNamespace can return nil.

This warning can be suppressed with:
```
func (r *region) SeedNamespace() Namespace {
	seed := C.ndctl_region_get_namespace_seed(r)
	if seed == nil {
		return nil
	}
	return seed
}
```

But that feels wrong.

```
pkg/volumepathhandler/volume_path_handler.go:119:2: ineffectual assignment to ctx (ineffassign)
	ctx, logger := pmemlog.WithName(ctx, "mapBindMountDevice")
	^
pkg/volumepathhandler/volume_path_handler.go:238:2: ineffectual assignment to ctx (ineffassign)
	ctx, logger := pmemlog.WithName(ctx, "RemoveMapPath")
	^
pkg/volumepathhandler/volume_path_handler.go:295:2: ineffectual assignment to ctx (ineffassign)
	ctx, logger := pmemlog.WithName(ctx, "GetDeviceBindMountRefs")
	^
```

Here the assignment to `ctx` can indeed be omitted with `_, logger := ...` because `ctx` is not used. But I find that dangerous: if the code after that later gets changed so that `ctx` *is* used, it'll use the one without the modified logger. I'd rather rely on the Go compiler to detect and remove the dead assignment.

avalluri (approved):

Issue #1046, opened 2021-11-03
liveness probe failure during long-running volume operations


pohly:
Deleting a 200GB volume failed while the PMEM-CSI driver was busy working on the DeleteVolume call. Presumably it was scrubbing the volume at the time (logs not detailed enough to be sure):

```
I1103 16:19:59.883947       1 tracing.go:25] "Probe: Processing gRPC call" request-counter=10 full-method="/csi.v1.Identity/Probe"
I1103 16:19:59.884414       1 tracing.go:25] "GetPluginInfo: Processing gRPC call" request-counter=11 full-method="/csi.v1.Identity/GetPluginInfo"
I1103 16:19:59.884762       1 tracing.go:25] "GetPluginCapabilities: Processing gRPC call" request-counter=12 full-method="/csi.v1.Identity/GetPluginCapabilities"
I1103 16:19:59.885242       1 tracing.go:25] "ControllerGetCapabilities: Processing gRPC call" request-counter=13 full-method="/csi.v1.Controller/ControllerGetCapabilities"
I1103 16:19:59.885709       1 tracing.go:25] "NodeGetInfo: Processing gRPC call" request-counter=14 full-method="/csi.v1.Node/NodeGetInfo"
I1103 16:20:06.393008       1 tracing.go:25] "GetCapacity: Processing gRPC call" request-counter=15 full-method="/csi.v1.Controller/GetCapacity"
I1103 16:20:06.493085       1 tracing.go:25] "DeleteVolume: Processing gRPC call" request-counter=16 full-method="/csi.v1.Controller/DeleteVolume"
I1103 16:20:46.036305       1 tracing.go:25] "NodeGetCapabilities: Processing gRPC call" request-counter=17 full-method="/csi.v1.Node/NodeGetCapabilities"
I1103 16:20:50.785601       1 tracing.go:25] "NodeGetCapabilities: Processing gRPC call" request-counter=18 full-method="/csi.v1.Node/NodeGetCapabilities"
I1103 16:21:06.393911       1 tracing.go:25] "GetCapacity: Processing gRPC call" request-counter=19 full-method="/csi.v1.Controller/GetCapacity"
I1103 16:21:08.310609       1 pmem-csi-driver.go:358] "Caught signal, terminating." signal="terminated"
```

The reason for the signal is a failed liveness probe from kubelet:
```
[...]

PR #1047, opened 2021-11-04
pmem-csi: avoid false negative liveness probe

pohly:
During a long-running DeleteVolume for a 200GB volume the LVM lock is held and
the liveness probe for metrics data timed out after a minute, killing the
container, because metrics retrieval must take the LVM lock.

As operation duration is non-deterministic, the solution is to probe only
metrics data which can be served immediately when the process is up and
running. This is less precise (= doesn't really check any functionality), but
that's an inherent problem of liveness probes: they simply cannot cover the
full functionality.

Fixes: #1046

### DIFF
--- a/deploy/kubernetes-1.19/direct/pmem-csi.yaml
+++ b/deploy/kubernetes-1.19/direct/pmem-csi.yaml
@@ -382,7 +382,7 @@ spec:
         livenessProbe:
           failureThreshold: 6
           httpGet:
-            path: /metrics
+            path: /metrics/simple
             port: metrics
             scheme: HTTP
           periodSeconds: 10
@@ -401,7 +401,7 @@ spec:
         startupProbe:
           failureThreshold: 60
           httpGet:
-            path: /metrics
+            path: /metrics/simple
             port: metrics
             scheme: HTTP
           periodSeconds: 1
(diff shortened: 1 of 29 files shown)


pohly:
@avalluri does this look alright?

pohly:
I tried out the fix on the cluster with the 200GB volume that was pending deletion. PMEM-CSI reports:
```
I1105 15:21:21.563828       1 tracing.go:25] "DeleteVolume: Processing gRPC call" request-counter=10 full-method="/csi.v1.Controller/DeleteVolume"
```

And external-provisioner:
```
I1105 15:21:21.563579       1 controller.go:1472] delete "pvc-db1b6e84-29cf-444a-a553-d52d547ac8cf": started
I1105 15:24:50.285521       1 controller.go:1487] delete "pvc-db1b6e84-29cf-444a-a553-d52d547ac8cf": volume deleted
I1105 15:24:50.292446       1 controller.go:1537] delete "pvc-db1b6e84-29cf-444a-a553-d52d547ac8cf": persistentvolume deleted
I1105 15:24:50.292463       1 controller.go:1542] delete "pvc-db1b6e84-29cf-444a-a553-d52d547ac8cf": succeeded
```

Data point: deleting a 200GB volume takes ~3:30 min.

avalluri (approved):
Looks good to me.

PR #1049, opened 2021-11-11
Kubernetes 1.22

pohly:


pohly:
@avalluri: does this PR look okay?

Shall we backport to release-1.0? I think that would be valuable.

PR #1053, opened 2021-11-24
avoid Delete call in controller

pohly:
Main reason for the PR is an observation that some tests because the controller container restarted directly after doing some useless Delete (see commit message).

Also contains some other, minor enhancements.


avalluri (approved):
Looks good to me.

pohly:
I did a full branch test already by pushing to the main repo.

Let's wait for the PR test anyway, to see whether we have remaining test flakes that could be fixed by this PR.

Issue #1058, opened 2022-01-02
OpenShift: setting up XFS volumes fails during xfs_io


pohly:
@kirin1994 ran into an issue when testing on OpenShift: volumes with XFS get created, but then the xfs_io command which sets the 2m extsize for huge pages hangs.

Invoking that binary in the PMEM-CSI DaemonSet container via `kubectl exec` without any parameters also hangs. Invoking it under strace shows that it gets stuck very early, probably even before it reaches `main()`. This looks like an issue with the kernel or container runtime.

pohly:
There is a potential workaround: the work done my `xfs_io -c "extsize 2m" <mount point>` can also be done directly by the PMEM-CSI driver through the same ioctl calls that xfs_io uses. Using those from Go is possible through cgo.

pohly:
Corresponding upstream issue: https://bugzilla.redhat.com/show_bug.cgi?id=2034194

pohly:
Reopening - we probably need to backport and do a v1.0.2.

pohly:
This is contained in 1.0.2.

PR #1059, opened 2022-01-02
XFS: fix creating volumes on OpenShift

pohly:
On OpenShift, the xfs_io command that was used while creating XFS volumes hangs
while getting loaded (kernel or container runtime bug?). This problem can be
avoided by making the same ioctl calls as in that binary directly from
the PMEM-CSI driver.

Fixes: https://github.com/intel/pmem-csi/issues/1058

### DIFF
--- a/pkg/pmem-csi-driver/nodeserver.go
+++ b/pkg/pmem-csi-driver/nodeserver.go
@@ -30,6 +30,7 @@ import (
 	"github.com/intel/pmem-csi/pkg/pmem-csi-driver/parameters"
 	pmdmanager "github.com/intel/pmem-csi/pkg/pmem-device-manager"
 	"github.com/intel/pmem-csi/pkg/volumepathhandler"
+	"github.com/intel/pmem-csi/pkg/xfs"
 )
 
 const (
@@ -285,12 +286,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	if ephemeral && fsType == "xfs" {
-		// FS was created only in ephemeral case.
-		// Only if created filesytem and it was XFS:
-		// Tune XFS file system to serve huge pages:
-		// Set file system extent size to 2 MiB sized and aligned block allocations.
-		_, err := pmemexec.RunCommand(ctx, "xfs_io", "-c", "extsize 2m", hostMount)
-		if err != nil {
+		if err := xfs.ConfigureFS(hostMount); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -289,8 +290,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		// Only if created filesytem and it was XFS:
 		// Tune XFS file system to serve huge pages:
 		// Set file system extent size to 2 MiB sized and aligned block allocations.
-		_, err := pmemexec.RunCommand(ctx, "xfs_io", "-c", "extsize 2m", hostMount)
-		if err != nil {
+		if err := xfs.SetXFSExtsize(hostMount); err != nil {
```
okartau, line 293: Here, it seems awkward that comment says "set to 2M", the subroutine is called "set size", but where this size gets set? One would expect that caller sets size via parameter. Sure that you already considered that and reached conclusion we now only need one size so we hard code it in function, and can change it later if another sizes may become into picture. But to make it look closer to what it is doing, the function could be named SetXFSExtsize2m or something.

pohly: What about `ConfigureFS`? The caller doesn't need to care what the function does, only that it has to be called after mkfs.xfs.

pkg/pmem-csi-driver/nodeserver.go
```diff
@@ -587,11 +583,8 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if existingFsType == "" && requestedFsType == "xfs" {
```
pohly, line 590: I think `existingFsType == ""` was too strict here. Suppose mkfs.xfs succeeded but then the function gets interrupted before `xfs_io`. During the next attempt it would find the existing XFS but then skip `xfs_io` although it hasn't been called.

`xfs_io` and now `xfs.ConfigureFS` are idempotent. It's therefore safe to do this step multiple times.

pkg/pmem-csi-driver/nodeserver.go
pohly: @okartau: changed, please take another look and also see my other comment about the if check.

okartau: I agree, doing it using ConfigureFS is good solution

pkg/pmem-csi-driver/nodeserver.go
okartau: yes, this makes sense.

PR #1072, opened 2022-04-19
build: Kubernetes 1.24

pohly:
The biggest change is that the corresponding klog now supports contextual
logging, so quite a bit of our custom code can be replaced.


pohly:
Closing in favor of the more complete PR #1073

PR #1078, opened 2022-07-12
doc: use design-proposals-archive path for scheduer_extender link

okartau:


okartau:
something wrong with checks here, CI job was triggered (and failed, although not justified as this is doc change), but status is not propagating here, lint job pass status visible here only.

pohly:
There's another problem with "make vhtml" (newer markdown package not compatible). PR #1079 is a more complete solution.

PR #1096, opened 2023-05-15
deps: github.com/docker/distribution v2.8.2

pohly:
Because of CVE-2023-2253...


pohly:
PMEM-CSI is not affected, but we have to fix CVE alerts.

